PR #1, opened 2017-02-01
Css and package refactor

evansiroky:
The package.json was refactored to avoid having competing react packages.  If there are 2 different versions of react trying to render something, that will cause an error.


demory:
Manually merged this to the dev branch.

PR #9, opened 2017-03-08
React native improvements

evansiroky:
* use global fetch if available
* check for existence of window global


evansiroky:
@landonreed I updated a few things in the itinerary display to make them pass the web accessibility lint tests.

lib/components/narrative/access-leg.js
```diff
@@ -12,48 +12,58 @@ export default class AccessLeg extends Component {
...
     if (index === this.props.activeStep) {
       this.props.setActiveStep(null)
     } else {
       this.props.setActiveStep(index, step)
     }
   }
+
   render () {
     const { active, activeStep, index, leg } = this.props
+    const legClickHandler = (e) => this._onLegClick(e, leg, index)
```
trevorgerhardt: Instantiate this with the class so the function doesn't get recreated on each render.

evansiroky: I'm a little confused on how to do that since it's in a loop where leg and index are changed.  Wouldn't a function need to get created each time for this to work?

trevorgerhardt: `leg` and `index` are retrieved from the props.

```js
  legClickHandler = (e) => this._onLegClick(e, this.props.leg, this.props.index)
```

Should work the same and not be re-created on each render. It's not going to be a serious performance hit here, but it's good practice.

evansiroky: Ah, I see it now, I was looking at another piece of code further down when I made that comment.

lib/components/narrative/access-leg.js
```diff
@@ -12,48 +12,58 @@ export default class AccessLeg extends Component {
...
     return (
       <div
         key={index}
         className={`leg${active ? ' active' : ''}`}>
-        <div
+        <a
```
landonreed: `a`s don't work here (click doesn't trigger div expansion).  Also, previously a click anywhere in the dev triggered expansion, now you have to click exactly on the link (except that it doesn't work at the moment)
image

lib/components/narrative/narrative-itinerary.js
```diff
@@ -37,15 +37,16 @@ export default class NarrativeItinerary extends Component {
     } = this.props
     return (
       <div className={`option itinerary${active ? ' active' : ''}`}>
-        <div
+        <a
```
landonreed: `a` focus looks really bad here: 

image

evansiroky: With button:

screen shot 2017-03-10 at 1 15 21 pm

lib/components/narrative/access-leg.js
evansiroky: With button:

screen shot 2017-03-10 at 1 15 53 pm

evansiroky:
Looks like we get much better results with using `button`.  A drawback of us following the [no-static-element-interactions](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-static-element-interactions.md) rule is that custom css is needed to make it look decent now.  We could turn off this particular rule in our mastarm config to get around it.

landonreed:
I think it's worth following the rule, even if it means some custom css.  I still think this needs work from a design perspective (the blue focus doesn't work quite right), but we'll merge for now.  @evansiroky could you open an issue to investigate this further?  We may want to draw on some external resources.  One example I found (Google maps) uses a button for the up/down caret, and seems to link the rest of the clickable area to the button:

image

PR #14, opened 2017-03-23
Switching locations

evansiroky:
This adds an action that will switch the from and to locations and then dispatch a trip plan action.

This PR builds on top of #13


example.js
```diff
@@ -7,7 +7,7 @@ import thunk from 'redux-thunk'
 import createLogger from 'redux-logger'
 
 // import Bootstrap Grid components for layout
-import { Navbar, Grid, Row, Col } from 'react-bootstrap'
+import { Navbar, Grid, Row, Col, Button } from 'react-bootstrap'
```
landonreed: `Button` import is not used.

lib/components/form/switchButton.js
```diff
@@ -0,0 +1,33 @@
+import React, { Component, PropTypes } from 'react'
```
landonreed: We seem to be naming files in hear with dashes `fil-name.js`, rather than camelCase.  Any reason it was switched up here?

evansiroky:
I added the changes requested.  Also, I don't particularly like where the switch button show up in the example.  I merely added it so I could test out the functionality.  It seems like every trip planner implements the switch button in a slightly different visual way.  In my other project I'm mainly interested in using the new action that I made.

PR #18, opened 2017-03-24
Bundle all css together

evansiroky:
I moved all of the css from example.css into the lib/index.css file.  This could probably use some more refactoring, but it’ll get other projects working in the meantime.


landonreed:
@evansiroky sorry, I just meant that we should add in the narrative button css.  I think that other css in `example.css` should remain outside of the core library (and just exist for the example).  @demory, I think that was your intention here, right?

PR #19, opened 2017-03-27
Add itinerary carousel

landonreed:
Add swipeable itinerary carousel for use in mobile views.

### DIFF
--- a/lib/components/narrative/itinerary-carousel.js
+++ b/lib/components/narrative/itinerary-carousel.js
@@ -20,20 +20,25 @@ class ItineraryCarousel extends Component {
     setActiveLeg: PropTypes.func,
     setActiveStep: PropTypes.func
   }
+
   _onItineraryClick = () => {
     this.setState({expanded: !this.state.expanded})
   }
-  _leftClick = () => {
-    const newIndex = this.props.activeItinerary === 0 ? this.props.itineraries.length - 1 : this.props.activeItinerary - 1
-    this.props.setActiveItinerary(newIndex)
+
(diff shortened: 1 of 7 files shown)


lib/components/narrative/itinerary-carousel.js
```diff
@@ -0,0 +1,110 @@
...
+  static propTypes = {
+    itineraries: PropTypes.array,
+    pending: PropTypes.bool,
+    activeItinerary: PropTypes.number,
+    hideHeader: PropTypes.bool,
+    setActiveItinerary: PropTypes.func,
+    setActiveLeg: PropTypes.func,
+    setActiveStep: PropTypes.func
+  }
+  _onItineraryClick = () => {
```
evansiroky: Can you put a blank line between all these methods?  It's a little hard to read.

lib/components/narrative/itinerary-carousel.js
```diff
@@ -0,0 +1,110 @@
...
+    this.setState({expanded: !this.state.expanded})
+  }
+  _leftClick = () => {
+    const newIndex = this.props.activeItinerary === 0 ? this.props.itineraries.length - 1 : this.props.activeItinerary - 1
```
evansiroky: If you used destructuring here, it could produce more concise code.  Furthermore, you could return the result of the conditional operation without having to create a new variable.  Same in the _rightClick method.

evansiroky:
whoops, didn't see that the target branch was master.

landonreed:
That's definitely my bad.  Sorry about that.

PR #23, opened 2017-03-28
Add react router

landonreed:
Add initial routing for application and make map entirely controlled component.

Pending questions:

- pushState vs hashChange
- exact configuration of routes

### DIFF
--- a/lib/components/map/base-map.js
+++ b/lib/components/map/base-map.js
@@ -4,92 +4,56 @@ import { connect } from 'react-redux'
 import { Map } from 'react-leaflet'
 
 import { setLocation, updateMapState } from '../../actions/map'
-import { constructLocation, getItineraryBounds } from '../../util/map'
+import { constructLocation } from '../../util/map'
 import { getActiveItinerary, getActiveSearch } from '../../util/state'
 
 class BaseMap extends Component {
+  state = {}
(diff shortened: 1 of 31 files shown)


lib/components/map/base-map.js
```diff
@@ -125,7 +76,8 @@ const mapStateToProps = (state, ownProps) => {
 
 const mapDispatchToProps = (dispatch, ownProps) => {
   return {
-    setLocation: (type, location) => { dispatch(setLocation({ type, location })) }
+    setLocation: (type, location) => { dispatch(setLocation({ type, location })) },
+    updateMapState: (props) => { dispatch(updateMapState(props)) }
```
trevorgerhardt: Btw, I would highly recommend matching argument patterns here if you can. For the obvious reason (matching) and also so that you can write a simpler `mapDispatchToProps`. If the argument patterns match then you could write it like:

```js
const mapDispatchToProps = {setLocation, updateMapState}
```

landonreed: Thanks for the note @trevorgerhardt.  I didn't realize it could be reduced to that.

landonreed:
Codebase has already integrated react-router hash routing. It may need additional work, but this PR is now out of date.

PR #24, opened 2017-04-04
Add config options for form control

evansiroky:
* allow debouncing of trip plan requests
* add option to not auto-plan trips

### DIFF
--- a/lib/actions/form.js
+++ b/lib/actions/form.js
@@ -1,3 +1,4 @@
+import debounce from 'lodash.debounce'
 import { createAction } from 'redux-actions'
 
 import { planTrip } from './api'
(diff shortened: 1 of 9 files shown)


lib/actions/form.js
```diff
@@ -37,12 +38,18 @@ export function setTime (payload) {
   }
 }
 
+let debouncedPlanTrip  // store as global, so it can be reused.  Maybe there's a better way?
```
evansiroky: @trevorgerhardt let me now if there is a better way to do this.  Since a new function is returned on each formChanged event I put the debouncedPlanTrip variable in this scope so it would be available in subsequent calls.

trevorgerhardt: If `debouncePlanTimeMs` can change then you would need to recreate the debounced function when it does. If it can't change then it's application configuration and shouldn't be retrieved within the function and can be created once with `dispatch` passed in as a parameter:

```js
const debouncedPlanTrip = debounce(({dispatch}) => dispatch(planTrip()), DEBOUNCE_PLAN_TIME_MS)
```

evansiroky: the debounce time is a variable that comes from the store, so it could change.  I updated my code so that it'll update the debounce function if that happens.

evansiroky:
I added some comments for you, @landonreed.  lmk if it makes sense.

PR #27, opened 2017-04-17
Mobile support

demory:
Various enhancements to support implementation of TriMet mobile-first UI. (2nd pass at this PR; places `autoPlan` in `config` section of store per discussion w/ @evansiroky @landonreed)


lib/components/form/mode-selector.js
```diff
@@ -9,25 +9,44 @@ class ModeSelector extends Component {
...
+      case 'TRANSIT,BICYCLE': return 'Bike to Transit'
+      case 'WALK': return 'Walk Only'
+      case 'BICYCLE': return 'Bike Only'
+    }
+    return mode
+  }
+
   render () {
     const { config, mode } = this.props
+    const label = this.props.label || 'Mode:'
```
landonreed: I'm not sure if we've started in otp-rr yet, but for default props maybe we should start using `static defaultProps = {label: 'Mode'}` https://babeljs.io/blog/2015/06/07/react-on-es6-plus.

PR #32, opened 2017-05-10
Settings screen

demory:


lib/components/form/modes-panel.js
```diff
@@ -0,0 +1,71 @@
...
+        {modeGroups.map((group, k) => {
+          return (
+            <div className='mode-group-row' key={k}>
+              <div className='group-name'>{group.name}</div>
+              {group.modes.map(mode => {
+                const modeSelected = queryModes.indexOf(mode) !== -1
+                return (
+                  <div key={mode} className='mode-icon-highlight'>
+                    <button
+                      onClick={() => {
```
landonreed: We should move this inline function into a class function.

lib/reducers/create-otp-reducer.js
```diff
@@ -17,7 +17,7 @@ const defaultQuery = {
   departArrive: 'NOW',
   date: getCurrentDate(),
   time: getCurrentTime(),
-  mode: 'TRANSIT,WALK'
+  mode: 'TRAM,BUS,GONDOLA,WALK'
```
landonreed: What happens when a mode supplied here doesn't exist in modes available in config?  It looks like `GONDOLA` cannot be removed from the list.  Perhaps the `modes-panel` should filter out any modes that are not available to prevent a mode from sneaking in and messing up plan request results.

lib/components/form/settings-bar.js
```diff
@@ -0,0 +1,79 @@
...
+    modeGroups.forEach(g => { totalModeCount += g.modes.length })
+
+    const selectedModeCount = this.props.queryModes.length
+
+    return (
+      <div className='settings-bar'>
+        <div className='button-container'>
+          <Button
+            className='settings-button'
+            onClick={() => { this.props.setShowExtendedSettings(!this.props.showExtendedSettings) }}
```
landonreed: Move inline func to class func.

lib/components/form/modes-panel.js
```diff
@@ -0,0 +1,71 @@
...
+const mapStateToProps = (state, ownProps) => {
+  return {
+    modeGroups: state.otp.config.modeGroups,
+    queryModes: state.otp.currentQuery.mode.split(',')
+  }
+}
+
+const mapDispatchToProps = (dispatch, ownProps) => {
+  return {
+    setMode: (mode) => { dispatch(setMode({ mode })) }
```
landonreed: I think we should try to simplify actions so that the arguments match.  That way, refactoring becomes easier down the road (we're not scratching our heads about an undefined argument) and the syntax becomes simpler.  For example, if the args match, this becomes `const mapDispatchToProps = {setMode}`

lib/reducers/create-otp-reducer.js
landonreed: Moving this to an issue for later consideration.

PR #35, opened 2017-06-19
Add ability to post-process results

evansiroky:
Extend the create-otp-reducer function to allow it to accept a function that will post-process the results and save to the reducer state.


landonreed:
This PR added a feature needed by the https://github.com/conveyal/ModeifyReactNative project, which still uses the `result-post-processor` branch of this repo. There have likely been many changes to the otp-react-redux project since this PR was created, so it may not be worth the effort to reconcile them until ModeifyReactNative needs significant maintenance. tldr; we're closing this PR for now (but NOT deleting the `result-post-processor` branch).

PR #36, opened 2017-07-05
Profile routing

demory:


lib/actions/api.js
```diff
@@ -45,21 +47,88 @@ export function planTrip (customOtpQueryBuilder) {
...
 }
 
-// function setURLSearch (params) {
-//   window.location.hash = `#plan?${params.split('plan?')[1]}`
-// }
+// Profile Query
+
+export const receivedProfileError = createAction('PROFILE_ERROR')
+export const receivedProfileResponse = createAction('PROFILE_RESPONSE')
+export const requestProfileResponse = createAction('PROFILE_REQUEST')
```
evansiroky: I sort of wonder if it may be a bit redundant to have these actions for profile requests/responses.  Perhaps we can just re-use the existing request/response actions but still have different logic for making a profile vs itinerary request.

lib/actions/config.js
```diff
@@ -1,3 +1,6 @@
 import { createAction } from 'redux-actions'
 
 export const setAutoPlan = createAction('SET_AUTOPLAN')
+
+// TODO: this should eventually be handled via mapState
+export const setMapCenter = createAction('SET_MAP_CENTER')
```
evansiroky: Is this used anywhere else in the app?  If it's just a trimet-mod-otp thing, it should probably live in that project.

lib/actions/map.js
```diff
@@ -26,6 +27,43 @@ export function clearLocation (payload) {
 
 export function setLocation (payload) {
   return function (dispatch, getState) {
```
evansiroky: This could be refactored to use woonerf's promise and multi middleware.

lib/components/form/date-time-selector.js
```diff
@@ -13,39 +14,107 @@ class DateTimeSelector extends Component {
...
+      return (
+        <Form>
+          <FormGroup style={{marginBottom: '15px'}} className='date-time-selector'>
+            <Row>
+              <Col xs={12}>
+                <FormControl
+                  componentClass='select'
+                  style={{width: '100%'}}
+                  onChange={(evt) => {
+                    this.props.setDate(moment().weekday(evt.target.value).format('YYYY-MM-DD'))
```
evansiroky: This handler should be a class function

lib/components/form/date-time-selector.js
```diff
@@ -13,39 +14,107 @@ class DateTimeSelector extends Component {
...
+              </Col>
+            </Row>
+            <Row style={{ marginTop: 20 }}>
+              <Col xs={5}>
+                <FormControl
+                  className='time-selector'
+                  type='time'
+                  value={startTime}
+                  style={{width: '100%'}}
+                  onChange={evt => { this.props.setStartTime(evt.target.value) }}
```
evansiroky: This handler should be a class function

lib/components/form/date-time-selector.js
```diff
@@ -13,39 +14,107 @@ class DateTimeSelector extends Component {
...
+                />
+              </Col>
+              <Col xs={2}>TO</Col>
+              <Col xs={5}>
+                <FormControl
+                  className='time-selector'
+                  type='time'
+                  value={endTime}
+                  style={{width: '100%'}}
+                  onChange={evt => { this.props.setEndTime(evt.target.value) }}
```
evansiroky: This handler should be a class function

lib/components/form/date-time-selector.js
```diff
@@ -13,39 +14,104 @@ class DateTimeSelector extends Component {
...
+            <Row>
+              <Col xs={12}>
+                <FormControl
+                  componentClass='select'
+                  style={{width: '100%'}}
+                  onChange={(evt) => {
+                    setQueryParam({
+                      date: moment().weekday(evt.target.value).format('YYYY-MM-DD')
+                    })
+                  }}
```
evansiroky: This handler function should be a class method

lib/components/form/date-time-selector.js
```diff
@@ -13,39 +14,104 @@ class DateTimeSelector extends Component {
...
+              </Col>
+            </Row>
+            <Row style={{ marginTop: 20 }}>
+              <Col xs={5}>
+                <FormControl
+                  className='time-selector'
+                  type='time'
+                  value={startTime}
+                  style={{width: '100%'}}
+                  onChange={evt => { setQueryParam({startTime: evt.target.value}) }}
```
evansiroky: This handler function should be a class method

lib/components/form/date-time-selector.js
```diff
@@ -13,39 +14,104 @@ class DateTimeSelector extends Component {
...
+                />
+              </Col>
+              <Col xs={2}>TO</Col>
+              <Col xs={5}>
+                <FormControl
+                  className='time-selector'
+                  type='time'
+                  value={endTime}
+                  style={{width: '100%'}}
+                  onChange={evt => { setQueryParam({endTime: evt.target.value}) }}
```
evansiroky: This handler function should be a class method

lib/components/form/dropdown-selector.js
```diff
@@ -0,0 +1,53 @@
...
+        <FormGroup style={{marginBottom: '15px'}} className='date-time-selector'>
+          <Row>
+            <Col xs={6} className='setting-label'>{label}</Col>
+            <Col xs={6}>
+              <FormControl
+                componentClass='select'
+                value={value}
+                onChange={(evt) => {
+                  setQueryParam({ [name]: evt.target.value })
+                }}
```
evansiroky: This handler function should be a class method

lib/components/form/location-field.js
```diff
@@ -1,75 +1,278 @@
...
+    /** the text input element **/
+    const textControl = <FormControl ref='formControl'
+      className={formControlClassname}
+      type='text'
+      value={this.state.value}
+      placeholder={this.props.label || this.props.type}
+      onChange={(e) => {
+        this.setState({ value: e.target.value })
+        this._geocode(e.target.value)
+      }}
```
evansiroky: This handler function should be a class method

lib/components/form/location-field.js
```diff
@@ -1,75 +1,278 @@
...
+      onClick={() => {
+        if (typeof this.props.onClick === 'function') this.props.onClick()
+        this.setState({ menuVisible: true })
+      }}
```
evansiroky: This handler function should be a class method

lib/components/form/location-field.js
```diff
@@ -1,75 +1,278 @@
...
+    />
+
+    /** the clear ('X') button add-on */
+    const clearButton = <InputGroup.Addon onClick={() => {
+      this.props.clearLocation(this.props.type)
+      this.setState({
+        value: '',
+        geocodedFeatures: []
+      })
+    }}>
```
evansiroky: This handler function should be a class method

lib/components/form/location-field.js
```diff
@@ -1,75 +1,278 @@
...
+            <InputGroup>
+              {/* location field icon -- also serves as dropdown anchor */}
+              <DropdownButton
+                componentClass={InputGroup.Button}
+                open={this.state.menuVisible}
+                onToggle={(v, e) => {
+                  // if clicked on input form control, keep dropdown open; otherwise, toggle
+                  const targetIsInput = e.target.className.indexOf(formControlClassname) !== -1
+                  this.setState({ menuVisible: targetIsInput ? true : !this.state.menuVisible })
+                }}
```
evansiroky: This handler function should be a class method

lib/actions/form.js
```diff
@@ -1,62 +1,41 @@
...
 
-export function setMode (payload) {
+export function setQueryParam (payload) {
   return function (dispatch, getState) {
-    dispatch(settingMode(payload))
-    dispatch(formChanged())
-  }
-}
-
-export function setDepart (payload) {
```
evansiroky: Removing all these is a breaking change.  See my additional discussion on [the commit where this was initially refactored](https://github.com/opentripplanner/otp-react-redux/commit/58530afb5df5c4d69bb4a4aa3906ed84a867b89b#commitcomment-22954224).

lib/common/query-params.js
```diff
@@ -0,0 +1,248 @@
+import { getCurrentDate, getCurrentTime } from '../util/time'
```
evansiroky: For better discovery, I think this file should exist within the util directory.

lib/actions/api.js
```diff
@@ -45,22 +49,75 @@ export function planTrip (customOtpQueryBuilder) {
...
+
+function constructPlanQuery (api, query) {
+  const planEndpoint = `${api.host}${api.port ? ':' + api.port : ''}${api.path}/plan`
+
+  let params = {}
+  queryParams.filter(qp => qp.planTypes.indexOf('ITINERARY') !== -1 && qp.name in query).forEach(qp => {
```
evansiroky: I think it would be good to have this filtering out of values also print warnings when an unrecognized key is sent.

lib/actions/api.js
```diff
@@ -45,22 +49,75 @@ export function planTrip (customOtpQueryBuilder) {
...
+    params = Object.assign(params, qp.itineraryRewrite
+      ? qp.itineraryRewrite(query[qp.name])
+      : { [qp.name]: query[qp.name] })
```
evansiroky: I think it would be good to run some kind of validation on each param that is sent to make sure that it is the right type.

lib/common/query-params.js
```diff
@@ -0,0 +1,248 @@
...
+ * options: an array of text/value pairs used with a dropdown selector
+ */
+
+const queryParams = [
+  { /* from - the trip origin. stored internally as a location (lat/lon/name) object  */
+    name: 'from',
+    planTypes: [ 'ITINERARY', 'PROFILE' ],
+    default: null,
+    itineraryRewrite: value => ({ fromPlace: `${value.lat},${value.lon}` }),
+    profileRewrite: value => ({ from: { lat: value.lat, lon: value.lon } })
```
evansiroky: Both of these will throw an error if a user sends `null`, `undefined` or a number as the value.  It would be good to have some kind of validation for what shape the expected value should be.

lib/components/narrative/itinerary-carousel.js
```diff
@@ -99,10 +109,23 @@ class ItineraryCarousel extends Component {
...
+  let itineraries = null
+  let profileOptions = null
+  let showProfileSummary = false
+  if (activeSearch && activeSearch.planResponse && activeSearch.planResponse.plan) {
+    itineraries = activeSearch.planResponse.plan.itineraries
+  } else if (activeSearch && activeSearch.profileResponse && activeSearch.profileResponse.otp) {
+    profileOptions = activeSearch.profileResponse.otp.profile
+    itineraries = profileOptionsToItineraries(profileOptions)
+    showProfileSummary = true
+  }
```
evansiroky: It is really needed to separate the itinerary and profile responses into different objects?

lib/components/narrative/narrative-profile-summary.js
```diff
@@ -0,0 +1,78 @@
...
+              padding: '3px',
+              color: 'white' }}
+            >
+              <div style={{ height: '20px', fill: 'white', marginTop: '6px' }}>{getModeIcon(option.icon, this.props.customIcons)}</div>
+              <div style={{ fontSize: '10px', textAlign: 'center', marginTop: '8px', textTransform: 'uppercase', letterSpacing: '2px' }}>{option.title}</div>
+              <div style={{ textAlign: 'center', marginTop: '2px', height: '30px' }}>
+                {option.time > 0
+                  ? <span><span style={{ fontSize: 24, fontWeight: '500' }}>{Math.round(option.time / 60)}</span> min</span>
+                  : <span style={{ fontSize: '11px' }}>(Not Found)</span>
+                }
```
evansiroky: Hmm, this is a little different than the app where it will hide options that don't get returned.  Not sure which would be better to display.

lib/index.js
```diff
@@ -62,10 +69,14 @@ export {
...
   NarrativeItinerary,
+  NarrativeProfileOptions,
 
   // actions
+  findNearbyStops,
+  getCurrentPosition,
   setAutoPlan,
-  setShowExtendedSettings,
+  setLocationToCurrent,
+  setMapCenter,
```
evansiroky: interesting that set `setQueryParam` isn't exported.  Would be interested to hear your thoughts on why.

lib/reducers/create-otp-reducer.js
```diff
@@ -1,25 +1,20 @@
...
-  date: getCurrentDate(),
-  time: getCurrentTime(),
-  mode: 'TRAM,BUS,GONDOLA,WALK'
-}
-// getDefaultQuery()
+// construct the initial/default query
+const defaultQuery = { type: 'ITINERARY' }
+queryParams.filter(qp => 'default' in qp).forEach(qp => {
+  defaultQuery[qp.name] = qp.default
+})
```
evansiroky: much cleaner.  I like it.

lib/reducers/create-otp-reducer.js
```diff
@@ -51,11 +51,21 @@ function createOtpReducer (config, initialQuery) {
             query: clone(state.currentQuery),
             pending: true,
             planResponse: null,
+            type: 'ITINERARY',
```
evansiroky: The request type could be included in the action and then the PROFILE_REQUEST wouldn't be needed.

lib/reducers/create-otp-reducer.js
```diff
@@ -129,22 +147,35 @@ function createOtpReducer (config, initialQuery) {
...
-      case 'SET_DEPART':
-        return update(state, { currentQuery: { departArrive: { $set: action.payload.departArrive } } })
-      case 'SET_DATE':
-        return update(state, { currentQuery: { date: { $set: action.payload.date } } })
-      case 'SET_TIME':
-        return update(state, { currentQuery: { time: { $set: action.payload.time } } })
+
+      case 'SET_QUERY_PARAM':
+        console.log('merging QPs', action.payload)
+        return update(state, { currentQuery: { $merge: action.payload } })
```
evansiroky: It could also be possible for a user to send an action with this name.  Perhaps we should namespace all of these actions so that there is less of a chance for collisions with redux reducers implemented in upstream apps.

evansiroky: Also, the upstream app could send over their own payload which may or may not conform to what is expected by the rest of OTP-RR.  Another opportunity for some validation.

lib/actions/config.js
evansiroky: nvm, I see where it's used.

lib/util/distance.js
```diff
@@ -1,7 +1,7 @@
-export function distanceStringImperial (meters) {
+export function distanceStringImperial (meters, abbreviate) {
```
evansiroky: Consider replacing this file with [humanize-distance](https://www.npmjs.com/package/humanize-distance).

PR #38, opened 2017-09-13
Fix content m2

landonreed:
This PR fixes miscellaneous issues with UI organization, text content, and formatting.

Created on behalf of @demory.

### DIFF
--- a/lib/util/profile.js
+++ b/lib/util/profile.js
@@ -16,13 +16,13 @@ export function filterProfileOptions (options) {
 
 /** profileOptionsToItineraries **/
 
-export function profileOptionsToItineraries (options) {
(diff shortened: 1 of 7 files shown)


lib/util/profile.js
```diff
@@ -73,7 +73,7 @@ function optionToItinerary (option) {
         duration: walkOnTime,
         transitLeg: false,
         from: {
-          name: 'Origin'
+          name: locationString(query && query.from.name, 'Destination')
```
landonreed: @demory I would suggest that we put `query.from.name` and `query.to.name` into variables -- for clarity in the code and to make the whole thing less verbose.  These values are used a number of times in this function.

PR #43, opened 2018-02-28
Milestone 4 enhancements

demory:


lib/actions/api.js
```diff
@@ -130,6 +165,216 @@ export function bikeRentalQuery () {
...
+      dispatch(responseAction(payload))
+    }
+
+    if (typeof postprocess === 'function') {
+      postprocess(payload, dispatch, getState)
+    }
+  }
+}
+
+// TODO: convert the rest of these to use the above createQueryAction helper
```
evansiroky: We may even want to try to convert these to use [woonerf fetch](https://github.com/conveyal/woonerf#fetch).  Maybe once we get around to refactoring we can revisit and standardize all of our api calls.

lib/components/form/location-field.js
```diff
@@ -76,6 +81,23 @@ class LocationField extends Component {
     })
   }
 
+  _geocodeSearch (text) {
```
evansiroky: We've got a lot of autocomplete components floating around.  Maybe we need to abstract out some logic of storing and displaying autocomplete results so that we have less duplicate code to maintain.

lib/components/map/base-map.js
```diff
@@ -52,46 +88,78 @@ class BaseMap extends Component {
...
+      map.leafletElement.fitBounds([
+        [newProps.query.from.lat, newProps.query.from.lon],
+        [newProps.query.to.lat, newProps.query.to.lon]
+      ], { padding })
+
+    // If only from or to is set, pan to that
+    } else if (newProps.query.from) {
+      map.leafletElement.panTo([newProps.query.from.lat, newProps.query.from.lon])
+    } else if (newProps.query.to) {
+      map.leafletElement.panTo([newProps.query.to.lat, newProps.query.to.lon])
```
evansiroky: We should be avoiding direct calls to the leaflet map whenever possible and instead send props to the Leaflet map

landonreed: To be addressed in #44.

lib/components/form/location-field.js
landonreed: @evansiroky, what autocomplete components are you referring to besides `location-field.js`?

lib/actions/api.js
landonreed: I have replaced the remaining API calls with `createQueryAction`

Let's revisit woonerf#fetch in #45.

landonreed:
I've addressed @evansiroky's comments, so I am going to go ahead and merge.

PR #50, opened 2019-05-09
Merge to dev

evansiroky:
It's been a while, but it's time to merge everything into `dev` and eventually `master` in order to start making regular releases.


landonreed:
Works for me. Let's get back on track with releases.

PR #51, opened 2019-05-10
Elevation enhancements

landonreed:
Enhances the elevation profile chart  according to the changes outlined in conveyal/trimet-mod-otp#171. Here's how the new chart appears:

image


lib/components/map/leg-diagram.js
```diff
@@ -130,84 +155,153 @@ class LegDiagram extends Component {
...
+            y={y - 10}
+            fontSize={11}
+            textAnchor='middle'>
+            {this._formatElevation(elevConverted)}
+          </text>
+        )
+      }
     }
-
+    // console.log({ traversed, width, xAxisCompression })
```
evansiroky: Still needed?

landonreed: Nah. Will remove.

lib/components/map/base-map.js
```diff
@@ -248,8 +248,8 @@ class BaseMap extends Component {
 
   render () {
     const { config, children, diagramLeg, elevationPoint, popupLocation } = this.props
-    const { baseLayers } = this.props.config.map
-
+    const { baseLayers } = config.map
+    const showElevationProfile = Boolean(config.elevationProfile)
```
evansiroky: Can you make a PR to https://github.com/conveyal/trimet-mod-otp/ with this added config setting enabled?

landonreed: https://github.com/conveyal/trimet-mod-otp/pull/186

landonreed:
Just pushed another commit that accounts for the move of a bunch of files from the TriMet repo to this one.

PR #57, opened 2019-06-10
Fix example

landonreed:
Fix library example usage and update config


example-config.yml
```diff
@@ -1,15 +1,17 @@
 api:
-  host: http://trimet.dev.conveyal.com
+  host: https://trimet-otp.ibi-transit.com
```
evansiroky: Let's make this point to localhost instead since this is merely an example and we don't want to tie use to any quasi-production server.

PR #59, opened 2019-06-18
Micromobility

evansiroky:
Adds in the ability to plan trips with Micromobility (eScooters).

## API changes

- New vehicle rental API endpoint action.
- automatic throttling of API queries. This is needed to avoid each vehicle rental overlay calling the vehicle rental endpoint when only one call is needed.

## UI changes

- Refactor of mode-button to be more readable and to use new icon lookup function
- Refactor of settings-selector-panel to be able to have eScooter rentals and personal eScooter mode options. Also has ability to select from multiple companies.
- Addition of Micromobility icon

## Map changes

- Consolidation of bike rental, car rental and eScooter rentals into single vehicle_rental overlay component
- Moving of car rental company css to trimet-mod project
- Zoom-dependent styling of map markers based off of custom config

## Narrative

- Refactor to rename some functions to more meaningful names and use of refactored icon lookup function
- Styling for micromobility mode
- Refactor of icon lookup function so that company and mode icons are all in same lookup file.

## Query Params

- Addition of `maxEScooterDistance` and `eScooter Power` selectors for various micromobility configuration.


evansiroky:
This PR is dependent on a new release of transitive.js. See https://github.com/conveyal/transitive.js/pull/47.

lib/actions/api.js
```diff
@@ -30,25 +30,22 @@ export function routingQuery () {
...
-      .then(json => {
-        dispatch(routingResponse({ response: json, searchId }))
-      })
-      .catch(error => {
-        dispatch(routingError({ error, searchId }))
-      })
+    try {
+      const realtimeResponse = await fetch(constructRoutingQuery(otpState))
+      const realtimeJson = await getJsonAndCheckResponse(realtimeResponse)
+      await dispatch(routingResponse({ response: realtimeJson, searchId }))
```
landonreed: Are `getJsonAndCheckResponse` and `routingResponse` really async functions? Do we actually need to await them? I'm also curious if there was a reason this change was made. There is no comment associating the change, so it makes me think this is just coding style/for the heck of it?

lib/actions/api.js
```diff
@@ -30,25 +30,22 @@ export function routingQuery () {
...
-        dispatch(nonRealtimeRoutingResponse({ response: json, searchId }))
-      })
-      .catch(error => {
-        console.error(error)
-        // do nothing
-      })
+    try {
+      const staticResponse = await fetch(constructRoutingQuery(otpState, true))
+      const staticJson = await getJsonAndCheckResponse(staticResponse)
+      await dispatch(nonRealtimeRoutingResponse({ response: staticJson, searchId }))
```
landonreed: Same as above comment on async/await.

lib/actions/api.js
```diff
@@ -188,6 +185,15 @@ export function carRentalQuery (params) {
   return createQueryAction('car_rental', carRentalResponse, carRentalError)
 }
 
+// Vehicle rental (e.g. Lime eScooter) locations lookup query
+
+export const vehicleRentalResponse = createAction('VEHICLE_RENTAL_RESPONSE')
+export const vehicleRentalError = createAction('VEHICLE_RENTAL_ERROR')
```
landonreed: "Vehicle rental" feels poorly named. We have very mode-specific names for every other rental type (e.g., bike, car). I'm guessing this name is aimed at a future refactor where everything is a vehicle so the generic name makes sense. If this is the case, it deserves a comment somewhere or we should honestly just name the thing scooter rental -- otherwise it could get confusing.

lib/actions/api.js
```diff
@@ -530,8 +536,26 @@ export function findStopsWithinBBox (params) {
...
+const throttledUrls = {}
+
+function now () {
+  return (new Date()).getTime()
+}
+
+const TEN_SECONDS = 10000
+
+// automatically clear throttled urls older than 10 seconds
+setInterval(() => {
```
landonreed: I like including `window.` on globals like `setInterval` -- I think it makes it easier to search for these types of functions in the code. Also, I don't think the comment fully describes what the function is doing and this kind of standalone, repeating execution deserves a good description.

lib/actions/api.js
```diff
@@ -561,6 +585,15 @@ function createQueryAction (endpoint, responseAction, errorAction, options) {
       const api = otpState.config.api
       url = `${api.host}${api.port ? ':' + api.port : ''}${api.path}/${endpoint}`
     }
+
+    // don't make a request to a URL that has already seen the same request
+    // within the last 10 seconds
+    if (throttledUrls[url] && throttledUrls[url] > now() - TEN_SECONDS) {
+      // URL already had a request within last 10 seconds, exit
```
landonreed: Should this log a warning? My impression is yes. Otherwise, future developers will probably have no idea what is going on when they can't figure out why a request is short circuiting.

lib/components/form/mode-button.js
```diff
@@ -1,6 +1,6 @@
-import React, {PropTypes, Component} from 'react'
+import React, {PropTypes, Component, PureComponent} from 'react'
```
landonreed: Any reason PlusTransit is a PureComponent? This feels like a bit of an arbitrary decision.

lib/components/form/settings-selector-panel.js
```diff
@@ -41,75 +57,196 @@ class SettingsSelectorPanel extends Component {
...
+    setQueryParam({ mode: [...nonOwnModes, newMode].join(',') })
+  }
+
+  _setOwnBike = () => this._replaceOwnMode('BICYCLE', 'BICYCLE')
+
+  _setRentedBike = () => this._replaceOwnMode('BICYCLE_RENT', 'BICYCLE')
+
+  _setOwnVehicle = () => this._replaceOwnMode('MICROMOBILITY', 'MICROMOBILITY')
+
+  _setRentedVehicle = () => {
```
landonreed: Again, the use of the generic vehicle language feels like a bit of a mismatch.

lib/components/form/settings-selector-panel.js
```diff
@@ -137,6 +274,7 @@ class SettingsSelectorPanel extends Component {
                 <ModeButton
                   enabled
                   active={modeHasTransit && this._modeIsActive(mode)}
+                  company={mode.company}
```
landonreed: The company prop in ModeButton doesn't appear to be used. Should this be removed? (`this.props.mode.company` is used in ModeButton.)

lib/components/map/base-map.js
```diff
@@ -123,6 +124,10 @@ class BaseMap extends Component {
     // }
   }
 
+  _onVeiwportChanged = e => {
```
landonreed: misspelling.

lib/components/map/base-map.js
```diff
@@ -321,6 +333,7 @@ class BaseMap extends Component {
         zoom={config.map.initZoom || 13}
         onOverlayAdd={this._onOverlayAdd}
         onOverlayRemove={this._onOverlayRemove}
+        onViewportChanged={this._onVeiwportChanged}
```
landonreed: misspelling (see above)

lib/components/narrative/line-itin/itin-summary.js
```diff
@@ -37,7 +37,11 @@ export default class ItinerarySummary extends Component {
           <div className='header'>{formatDuration(itinerary.duration)}</div>
 
           {/* Duration as time range */}
-          <div className='detail'>{formatTime(itinerary.startTime, timeOptions)} - {formatTime(itinerary.endTime, timeOptions)}</div>
+          <div className='detail'>
+            {formatTime(itinerary.startTime, timeOptions)}
+            -
+            {formatTime(itinerary.endTime, timeOptions)}
```
landonreed: Are the spaces surrounding the `-` preserved here properly?

lib/components/narrative/line-itin/itinerary.css
```diff
@@ -162,6 +162,18 @@
   );
 }
 
+.otp .line-itin .leg-line-vehicle {
```
landonreed: Again, 'vehicle' terminology feels confusing.

lib/util/itinerary.js
```diff
@@ -125,36 +174,30 @@ export function getStepStreetName (step) {
...
-    modeStr = `${modeStr}_${mode.company.toUpperCase()}`
-  }
-
-  // Check if there is a custom icon for this mode
-  if (customIcons && modeStr in customIcons) {
-    return customIcons[modeStr]
+export function getIcon (iconId, customIcons) {
+  // Check if there is a custom icon
+  if (customIcons && iconId in customIcons) {
+    return customIcons[iconId]
```
landonreed: Just curious: what happened to the company-specific stuff?

lib/actions/api.js
evansiroky: This change was more to make it easier to debug. It's easier to trace back errors with async/await as opposed to giant promise chains that catc errors that bubble up from dispatches unrelated to fetching. I can remove this if you like.

And yes, all of these things can be awaited. See [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch) and [redux-thunk dispatch chaining](https://github.com/reduxjs/redux-thunk#composition).

lib/actions/api.js
evansiroky: Yeah, it's intended to be for a future refactor. I'll add some comments. Also, if you have any thoughts on combining the APIs in a future refactor, that is appreciated.

lib/actions/api.js
evansiroky: Will add

lib/actions/api.js
evansiroky: Will add

lib/components/form/mode-button.js
evansiroky: PureComponent increases performance by having a default `shouldComponentUpdate` method defined. This does a shallow comparison of the props sent to the component and will cut back on the number of renders the component does.

lib/components/form/settings-selector-panel.js
evansiroky: Will add comments

lib/components/form/settings-selector-panel.js
evansiroky: Good catch, will remove.

lib/components/map/base-map.js
evansiroky: Will fix

lib/components/narrative/line-itin/itin-summary.js
evansiroky: Good point, the spaces would not be preserved, will fix.

lib/util/itinerary.js
evansiroky: There is now just one icon lookup, so an iconId can be either a company or a mode.

evansiroky:
I addressed most of the comments. I did my best to rename stuff with vehicle to micromobility to hopefully make things more understandable.

Also, I added one more thing which is to make the exclusive modes (Walk only, Bike only) be configurable and only added as needed.

lib/util/itinerary.js
landonreed: Can you add a comment to this method to describe that it will return an icon for either a company or a mode?

evansiroky:
Added more comments, fixed merge conflicts and fixed tests. Reassigning to you, @landonreed, to get your official PR approval.

landonreed:
Sorry, looks like my merge to dev caused a few more conflicts (as expected). Do you want me to tackle or do you want to wrap them up? Once we fix them, I'll take a final look.

evansiroky:
I'll work on the merge conflicts today.

PR #60, opened 2019-06-18
Get linting working on all files and fix all resulting errors

evansiroky:
Linting wasn't working on all files. Now it is. And all files pass the linting test now.


lib/components/map/transitive-styles.js
```diff
@@ -1,12 +1,8 @@
...
-const bikeSvg = '<?xml version="1.0" encoding="utf-8"?><svg width="2304" height="1792" viewBox="0 0 2304 1792" xmlns="http://www.w3.org/2000/svg"><path d="M762 1152h-314q-40 0-57.5-35t6.5-67l188-251q-65-31-137-31-132 0-226 94t-94 226 94 226 226 94q115 0 203-72.5t111-183.5zm-186-128h186q-18-85-75-148zm480 0l288-384h-480l-99 132q105 103 126 252h165zm1120 64q0-132-94-226t-226-94q-60 0-121 24l174 260q15 23 10 49t-27 40q-15 11-36 11-35 0-53-29l-174-260q-93 95-93 225 0 132 94 226t226 94 226-94 94-226zm128 0q0 185-131.5 316.5t-316.5 131.5-316.5-131.5-131.5-316.5q0-97 39.5-183.5t109.5-149.5l-65-98-353 469q-18 26-51 26h-197q-23 164-149 274t-294 110q-185 0-316.5-131.5t-131.5-316.5 131.5-316.5 316.5-131.5q114 0 215 55l137-183h-224q-26 0-45-19t-19-45 19-45 45-19h384v128h435l-85-128h-222q-26 0-45-19t-19-45 19-45 45-19h256q33 0 53 28l267 400q91-44 192-44 185 0 316.5 131.5t131.5 316.5z" fill="#fff"/></svg>'
-
```
landonreed: Is there a chance these could be needed in the future?

evansiroky: The from/to svg's are those that are used in CarFreeAtoZ, so I don't think those are in use anymore. There is also a different icon for bikes with an svg in the icons of the project. I don't think they're needed.

evansiroky:
Just got linting working again with mastarm 5

PR #62, opened 2019-06-18
Multi agency

landonreed:
### URL Routing
- adds connected-react-router to do all of the URL routing and preserve stop/route viewer states

### Multi-agency
- adds configurable agency branding URLs for itinerary narrative and route viewer

### Elevation graph
- adds some minor fixes for rendering elevation graph

### Store recent places/searches/favorites
- uses localStorage to store recent places, searches and settings, plus favorite places and transit stops


lib/reducers/create-otp-reducer.js
```diff
@@ -1,34 +1,52 @@
...
-const defaultQuery = { routingType: 'ITINERARY' }
-queryParams.filter(qp => 'default' in qp).forEach(qp => {
-  defaultQuery[qp.name] = qp.default
-})
-
+// Load user override settings from local storage.
+// TODO: Make this work with settings fetched from user profile API service.
+const userOverrides = getItem('defaultQuery')
+const defaults = Object.assign(getDefaultQuery(), userOverrides)
+const autoRefresh = getItem('autoRefresh', true)
```
evansiroky: It'd be nice to have some comments for what these are all used for. Like for this one, it'd be nice if there was a comment that it's auto refreshing stoptime updates. Maybe this variable could be renamed to be more specific since the vehicle rental layers also auto-refresh.

lib/actions/form.js
```diff
@@ -133,7 +101,7 @@ export function formChanged (oldQuery, newQuery) {
     }
 
     // Check whether a trip should be auto-replanned
-    const {autoPlan, debouncePlanTimeMs} = otpState.config
+    const { autoPlan, debouncePlanTimeMs } = otpState.config
```
evansiroky: oh man, I prefer the other way. I think since Trevor gave us full control over mastarm we oughta update the eslint config to have one bracket spacing rule.

lib/components/app/responsive-webapp.js
```diff
@@ -31,34 +29,69 @@ class ResponsiveWebapp extends Component {
...
+              break
+          }
+          if (paramIsDifferent) {
+            hasChangedParam = true
+          }
+        })
+      if (hasChangedParam) {
+        console.log('handle search change', prevProps.location.search, location.search)
+        console.log(query, planParams)
+        // this.props.setQueryParam(planParams)
```
evansiroky: What does `hasChangedParam` do? Is it used anymore?

lib/components/form/user-settings.js
```diff
@@ -0,0 +1,279 @@
...
+
+  render () {
+    const { user } = this.props
+    const { favoriteStops, trackRecent, recentPlaces, recentSearches } = user
+    // Clone locations in order to prevent blank locations from seeping into the
+    // app state/store.
+    const locations = this._getLocations(user)
+    const order = ['home', 'work', 'suggested', 'stop', 'recent']
+    const sortedLocations = locations.sort((a, b) => {
+      const aIndex = order.indexOf(a.type)
```
evansiroky: this could be shortened to `order.indexOf(a.type) - order.indexOf(b.type)`.

lib/components/form/user-settings.js
```diff
@@ -0,0 +1,279 @@
...
+                : <small>No recent searches</small>
+              }
+            </ul>
+          </div>
+        }
+        <hr />
+        <div>
+          <div className='section-header'>My preferences</div>
+          <small>Remember recent searches/places?</small>
+          <Button
```
evansiroky: This could use some spacing between the <small> element.

lib/actions/form.js
landonreed: Yea, I go back and forth on this rule, but have just been doing whatever the linter tells me to do lately, haha.

lib/components/app/responsive-webapp.js
landonreed: Ah, forgot to finish this up. It's supposed to basically handle when the back button is pressed, detecting the URL search change and resending the trip plan request.

lib/components/map/route-viewer-overlay.js
```diff
@@ -13,27 +17,11 @@ class RouteViewerOverlay extends MapLayer {
...
-        routeData &&
-        routeData.patterns &&
-        Object.values(routeData.patterns).reduce(
-          (acc, ptn) => acc && typeof ptn.geometry !== 'undefined',
-          true
-        )
-      )
-    }
-
+    console.log(nextProps);
```
evansiroky: Please remove

lib/components/viewers/route-viewer.js
```diff
@@ -10,29 +10,47 @@ import { setMainPanelContent, setViewedRoute } from '../../actions/ui'
...
   }
 
   render () {
-    const { routes, hideBackButton, languageConfig } = this.props
-
+    const { operators, routes, hideBackButton, languageConfig } = this.props
+    const sortedRoutes = routes ? Object.values(routes).sort(routeComparator) : []
+    const agencySortedRoutes = operators.length > 0
+      ? sortedRoutes.sort((a, b) => {
+        const aOperator = operatorIndexForRoute(operators, a)
```
evansiroky: If `operatorIndexForRoute` always returns an integer, this can be simplified to:

```javascript
return operatorIndexForRoute(operators, a) - operatorIndexForRoute(operators, b)
```

lib/components/viewers/route-viewer.js
```diff
@@ -51,47 +69,91 @@ class RouteViewer extends Component {
...
+              )
+            })
+          }
         </div>
       </div>
     )
   }
 }
 
 class RouteRow extends Component {
```
evansiroky: With the React Developer tools, it can be seen that clicking on a new route results in a very sluggish css transition. This is happening largerly because each RouteRow is freshly rendered whenever another RouteRow changes. All of these renders are not needed since clicking on one route shouldn't require a rerender of another RouteRow 20 spots above. A simple fix for this is to change the RouteRow to extend PureComponent and make sure the input props change only when absolutely needed.

See this article about [how PureComponents can help with performance](https://blog.logrocket.com/make-react-fast-again-part-2-why-did-you-update-dd1faf79399f/).

lib/components/viewers/stop-viewer.js
```diff
@@ -99,17 +147,42 @@ class StopViewer extends Component {
...
+              <div>
+                <b>Stop ID</b>: {stopId}
+                <button
+                  className='link-button pull-right'
+                  style={{ fontSize: 'small' }}
+                  onClick={this._refreshStopTimes}>
+                  <Icon
+                    className={spin ? 'fa-spin' : ''}
+                    type='refresh' />{' '}
+                  {moment(stopData.stopTimesLastUpdated).format('h:mm:ss a')}
```
evansiroky: There is a config.yml entry for `datetime.timeFormat`. Perhaps we should use that here?

lib/components/viewers/stop-viewer.js
```diff
@@ -158,20 +244,25 @@ class RouteRow extends Component {
...
+    const { route } = pattern
 
     // sort stop times by next departure
     let sortedStopTimes = null
     if (stopTimes) {
       sortedStopTimes = stopTimes.sort((a, b) => {
-        const aTime = a.serviceDay + a.realtimeDeparture
-        const bTime = b.serviceDay + b.realtimeDeparture
+        const aTime = a.serviceDay + a.realtimeArrival
+        const bTime = b.serviceDay + b.realtimeArrival
```
evansiroky: Why the change to arrival? Also, previous comments about implementing sort by subtracting numbers apply here.

lib/components/viewers/stop-viewer.js
```diff
@@ -240,35 +342,61 @@ class RouteRow extends Component {
...
+  const userTimeZone = moment.tz.guess()
+  const inHomeTimezone = homeTimezone && homeTimezone === userTimeZone
   // Determine whether to show departure as countdown (e.g. "5 min") or as HH:MM time
-  const showCountdown = inHomeTimezone && !differentDay &&
-    stopTime.realtimeDeparture - currentTime < 3600 &&
-    stopTime.realtimeDeparture > currentTime
-
+  const secondsUntilArrival = stopTime.realtimeArrival - currentHomeTime
+  const arrivesInFuture = stopTime.realtimeArrival > currentHomeTime
+  const showCountdown = !differentDay &&
```
evansiroky: Won't this not show arrivals that occur after midnight?

evansiroky: Looks like it'll just show the exact time and not a countdown.

lib/components/viewers/stop-viewer.js
```diff
@@ -240,35 +342,61 @@ class RouteRow extends Component {
...
+const ONE_HOUR_IN_SECONDS = 3600
+
+function getHomeTime (homeTimezone) {
+  const now = moment()
+  return now.tz(homeTimezone).diff(now.clone().startOf('day'), 'seconds')
+}
+
 // helper method to generate stop time w/ status icon
-function getFormattedStopTime (stopTime, homeTimezone) {
+function getFormattedStopTime (stopTime, homeTimezone, arrivingText = 'Due') {
```
evansiroky: Any complex function delaying with displaying time and timezones ought to have tests with all kinds of edge cases. Let's add an issue for this.

lib/components/viewers/stop-viewer.js
```diff
@@ -240,35 +342,61 @@ class RouteRow extends Component {
...
-        {differentDay && <div style={{ marginBottom: -4 }}>{serviceDay.format('dddd')}</div>}
-        <div>{showCountdown
-          ? formatDuration(stopTime.realtimeDeparture - currentTime)
-          : formatStopTime(stopTime.realtimeDeparture)
-        }</div>
+        {differentDay &&
+          <div style={{ marginBottom: -4 }}>{serviceDay.format('dddd')}</div>
+        }
+        <div>
+          {showCountdown
```
evansiroky: Seems incorrect to not show a countdown for arrivals after midnight.

lib/actions/api.js
```diff
@@ -563,7 +649,7 @@ function createQueryAction (endpoint, responseAction, errorAction, options) {
     }
     let payload
     try {
-      const response = await fetch(url)
+      const response = await fetch(url, options.fetchOptions)
```
evansiroky: These options will mess a bit with the fetch throttling I implemented in the micromobility branch.

lib/util/storage.js
```diff
@@ -0,0 +1,34 @@
+// Prefix to use with local storage keys.
+const STORAGE_PREFIX = 'otp'
```
evansiroky: This is all very specific for use with localStorage. Would it be advantageous to rewrite this so that we can use a generic persistence worker that can automatically figure out how to read and save data?

lib/util/time.js
```diff
@@ -43,8 +44,10 @@ export function formatTime (ms, options) {
  * @param {number} seconds time since midnight in seconds
  * @returns {string} formatted text representation
  */
-export function formatStopTime (seconds) {
-  return moment().startOf('day').seconds(seconds).format('h:mm a')
+export function formatStopTime (seconds, timezone = null) {
+  const m = timezone ? moment().tz(timezone) : moment()
+  const format = timezone ? 'h:mm a z' : 'h:mm a'
```
evansiroky: Another possible place to use time display string defined in config.

lib/actions/form.js
evansiroky: What linting rules are you using? I don't see [space-in-brackets](https://eslint.org/docs/rules/space-in-brackets) defined in [mastarm's eslintrc.json](https://github.com/conveyal/mastarm/blob/dev/lib/eslintrc.json) file.

lib/util/storage.js
landonreed: I think that's the plan in the long run. If you have any specific thoughts on a refactor for right now, I'm open to them.

evansiroky: I was thinking that there could be a class with generic methods that could be implemented to initialize a new session and then get and save data. The config could specify the type of persistence manager and any need keys, etc.

lib/actions/form.js
landonreed: I think I'm just using [standard](https://standardjs.com/).

lib/components/app/responsive-webapp.js
landonreed: OK, the back button behavior has been fully implemented.

lib/components/viewers/stop-viewer.js
landonreed: Actually, this appears to be a mistake. I thought I remember seeing something in [the ticket](https://github.com/ibi-group/trimet-mod-otp/issues/184) about changing from departures to arrivals, but I must be mistaken. I will revert back to departures.

lib/components/viewers/stop-viewer.js
landonreed: #63.

lib/components/viewers/stop-viewer.js
landonreed: Yes, that's correct. This was already the behavior. I'm not entirely sure why it was designed this way. For now, I'll just add a FIXME.

lib/components/viewers/stop-viewer.js
landonreed: See [previous comment](https://github.com/opentripplanner/otp-react-redux/pull/62/files#r297342085).

lib/util/storage.js
landonreed: @evansiroky, what do you think about just adding a `TODO` here about this? My concern is that to truly make something like you're describing it might be worthwhile to have some other persistence strategy to build against. I think given that our immediate task is to store data in local storage, we should aim to build that and encapsulate it as much as possible for the next project, when we expect to swap in a different persistence strategy. There are factors with future projects that we don't have complete information about, so I'm worried we would waste our time writing too much code for hypothetical systems.

landonreed:
@evansiroky, I think I've addressed everything except for the fetch options, route viewer optimizations, and merge conflicts. I'll wait to hear from you on the route viewer stuff and will fix the conflicts/fetch options, once we first merge micromobility into dev.

__tests__/actions/__snapshots__/form.js.snap
```diff
@@ -0,0 +1,39 @@
...
+    [Function],
+  ],
+  Array [
+    [Function],
+  ],
+  Array [
+    [Function],
+  ],
+  Array [
+    [Function],
```
evansiroky: This snapshot isn't providing much useful information, the test that generates it could probably use a refactor.

__tests__/actions/api.js
```diff
@@ -2,7 +2,12 @@
 
 import nock from 'nock'
 
-import {routingQuery} from '../../lib/actions/api'
+import * as api from '../../lib/actions/api'
+
+// Use mocked randId function and pass in searchId for routingQuery calls so that
+// snapshots are deterministic (i.e., the random IDs don't change).
+let idCounter = 1234
+const randId = () => `abcd${idCounter++}`
```
evansiroky: I'm not a huge fan of this since it's injecting a test function into the code. Perhaps we should just consider introducing a uuid dependency so that we can do something like this: https://github.com/conveyal/analysis-ui/blob/dev/__mocks__/uuid.js

__tests__/actions/form.js
```diff
@@ -20,18 +31,15 @@ describe('actions > api > form', () => {
...
         searches: []
       }
     }
 
-    const formChangedAction = formChanged()
-
     const mockDispatch = jest.fn()
     for (var i = 0; i < 10; i++) {
+      // Alternate back and forth between two queries on each iteration.
+      const formChangedAction = formChanged(queries[i % 2], queries[(i + 1) % 2])
```
evansiroky: This test can definitely use a refactor. It seems like no debouncing is happening anymore due to what the snapshot looks like. It may make sense to comment this out again and reassess later.

lib/actions/api.js
```diff
@@ -17,32 +19,113 @@ export const nonRealtimeRoutingResponse = createAction('NON_REALTIME_ROUTING_RES
...
     // fetch a realtime route
-    fetch(constructRoutingQuery(otpState))
+    const query = constructRoutingQuery(otpState)
+    fetch(query)
       .then(getJsonAndCheckResponse)
       .then(json => {
         dispatch(routingResponse({ response: json, searchId }))
+        // If tracking is enabled, store locations and search after successful
+        // search is completed.
+        // TODO recent searches
```
evansiroky: What does this TODO mean?

__tests__/actions/form.js
landonreed: Hmm. Ok -- you think we should rename the file again or just comment out the code?

lib/actions/api.js
```diff
@@ -17,32 +19,113 @@ export const nonRealtimeRoutingResponse = createAction('NON_REALTIME_ROUTING_RES
...
+        }
       })
       .catch(error => {
         dispatch(routingError({ error, searchId }))
       })
-
+    // Update OTP URL params if a new search. In other words, if we're
+    // performing a search based on query params taken from the URL after a back
+    // button press, we don't need to update the OTP URL.
+    // TODO: Should we be replacing the URL params here?
```
evansiroky: Seems like it'd be a good idea to replace the url params whenever a state change occurs that is reproducible based on the given url params.

lib/actions/api.js
landonreed: It's old/irrelevant now. It was a reminder to remember recent searches. I'll remove.

lib/actions/api.js
landonreed: I need to elaborate on this TODO. I was running into issues where if the app was re-running an old search found in the URL params and it pushed a new path with the query params it would funk with the history.  The TODO should say `TODO: For old searches that we are re-running, should we be **replacing** the URL params here (instead of **pushing** a new path to history like what currently happens in updateOtpUrlParams)? That way we could ensure that the path absolutely accurately reflects the app state.`

lib/actions/api.js
```diff
@@ -574,14 +694,63 @@ function createQueryAction (endpoint, responseAction, errorAction, options) {
...
+    Object.keys(urlParams)
+      .filter(key => key.indexOf('_') !== -1)
+      .forEach(key => { params[key] = urlParams[key] })
+    params.ui_activeSearch = searchId
+    // Assumes this is a new search and the active itinerary should be reset.
+    params.ui_activeItinerary = 0
+    // Merge in the provided OTP params and update the URL
+    dispatch(setUrlSearch(Object.assign(params, otpParams)))
+  }
+}
```
evansiroky: This file is getting long. Let's break it up in the next round of feature development.

lib/actions/ui.js
```diff
@@ -1,21 +1,89 @@
 import { createAction } from 'redux-actions'
+import { push } from 'connected-react-router'
+
+export function routeTo (url, replaceSearch) {
```
evansiroky: Needs some jsdoc

lib/actions/ui.js
```diff
@@ -1,21 +1,89 @@
...
+      path = `${path}${search}`
+    }
+    dispatch(push(path))
+  }
+}
 
 export const setMobileScreen = createAction('SET_MOBILE_SCREEN')
 
-export const setMainPanelContent = createAction('SET_MAIN_PANEL_CONTENT')
+export function setMainPanelContent (payload) {
```
evansiroky: jsdoc

lib/components/app/responsive-webapp.js
```diff
@@ -102,17 +104,96 @@ class ResponsiveWebapp extends Component {
...
+  }
+
+  _onBackButtonPress = (e) => {
+    // Get new search ID from URL after back button pressed.
+    // console.log('back button pressed', e)
+    const urlParams = getUrlParams()
+    const newSearchId = urlParams.ui_activeSearch
+    const newItinIndex = +urlParams.ui_activeItinerary || 0
+    const newSearch = this.props.searches[newSearchId]
+    if (newSearch) {
```
evansiroky: Would it make more sense to rename these as `previous...` instead of `new...`?

lib/components/app/responsive-webapp.js
```diff
@@ -122,71 +203,91 @@ class ResponsiveWebapp extends Component {
...
+  setActiveItinerary,
+  setActiveSearch,
+  setLocationToCurrent,
   setMainPanelContent,
+  setMapCenter,
+  setMapZoom,
   setQueryParam,
-  parseUrlQueryString
+  setViewedRoute,
+  setViewedStop
```
evansiroky: These actions are increasing quite a bit in number. Would it make more sense to combine some of the component handlers as actions and just import those large actions?

lib/components/app/responsive-webapp.js
landonreed: I see what you mean there, but on second thought, this function is also called when the 'forward' button is pressed, so previous might not always be an appropriate term.

landonreed: I went ahead and just changed to previous.

landonreed:
OK @evansiroky, I think I resolved all of your comments. Great suggestions!

lib/util/storage.js
evansiroky: Let's create an issue for OTP-RR to be able to save user settings to something other than localstorage and call it good for now.

lib/components/viewers/route-viewer.js
evansiroky: I'll approve this PR and make a separate PR to dev in the interest of moving things along.

__tests__/actions/form.js
evansiroky: Let's comment it all out for now. I initially put this test case in here to verify the functionality when developing the CFAZ react native app because it always auto-plans trips. But that was a while ago and things have drifted in the codebase, so I'm fine with removing the test case.

lib/actions/api.js
evansiroky: Ah, I see. Let's await user feedback and stuff and keep that expanded TODO comment.

evansiroky:
Good and thorough first take at user settings.

lib/components/app/responsive-webapp.js
```diff
@@ -31,31 +26,42 @@ class ResponsiveWebapp extends Component {
...
+          this.props.setMapZoom({ zoom: this.props.initZoomOnLocate })
         }
       }
     }
-
+    // If the path changes (e.g., via a back button press) check whether the
+    // main content needs to switch between, for example, a viewer and a search.
+    if (!isEqual(location.pathname, prevProps.location.pathname)) {
+      // console.log('url changed to', location.pathname)
+      this.props.matchContentToUrl(location)
```
evansiroky: This is causing a 2nd dispatch of the `findRoute` action when selecting a new route in the route viewer.

evansiroky: Workaround idea here: https://github.com/opentripplanner/otp-react-redux/issues/64

Issue #64, opened 2019-06-27
Premature optimization idea: cache route viewer data


evansiroky:
In the current design of the route viewer as of 26b9a8a0d431fc7bd2f6b7097a8657dde82a43b6, the route polyline data is fetched anew each time a user clicks on a new route. In order to cut back on the number of API fetches for users that click back and forth between routes, we could store the response for certain routes in a in-browser cache and return that data to reduce the need for API requests.

evansiroky:
In addition to caching this data we could also refactor the fetching of pattern geometry data so that only one action is sent after receiving all responses as opposed to having one action per response as exists here.

https://github.com/opentripplanner/otp-react-redux/blob/0e48ad4a82829ffe46e3b9e58fead605f0800466/lib/actions/api.js#L597-L602

PR #65, opened 2019-06-27
Route row pure component refactor

evansiroky:
Refactors the RouteRow component and the props sent to it within `components/viewers/route-viewer.js` to be a PureComponent that only re-renders when needed.

This PR depends on #62.


landonreed:
@evansiroky, when using the React module in Chrome devtools, I'm still seeing every route row re-render when I click a new route. Is there something I'm missing here?

evansiroky:
> @evansiroky, when using the React module in Chrome devtools, I'm still seeing every route row re-render when I click a new route. Is there something I'm missing here?

Yes, that devtools display is odd. However, if you add in a `console.log('render')` into the render function of RouteRow and then switch the extended class back and forth between Component and PureComponent you'll notice significantly less logged statements when the Component is a PureComponent.

landonreed:
@evansiroky, I just merged the latest `dev` branch in. I'm not really experiencing what you've described. There are still a ton of renders whether or not I'm using PureComponent. Let's maybe do a screenshare to see if we can come up with a solution.

lib/components/viewers/route-viewer.js
```diff
@@ -76,13 +84,17 @@ class RouteViewer extends Component {
           {agencySortedRoutes
             .map(route => {
               // Find operator based on agency_id (extracted from OTP route ID).
-              const operator = operatorForRoute(operators, route) || {}
```
evansiroky: @landonreed, this is where the problem was. On each re-render of the RouteViewer, the operatorForRoute will return a new `{}`. Each of these new `{}` objects does not strictly equal the other `{}` instances due to [the way JavaScript calculates object equality](http://adripofjavascript.com/blog/drips/object-equality-in-javascript.html), therefore the props are different which causes the rerenders.

landonreed: Gotcha, thanks for fixing. Before I approve this though, I think we should address the fact that a re-render occurs every time a new pattern geometry response is processed and added to the route in the reducer. Whether this fix happens by deconstructing the route into its parts and passing them piecemeal to the `RouteRow` component or through some change in the reducer, I don't think this PR will have achieved its objective until those extra renders are fixed.

evansiroky: Yes, that would further optimize things, but that is a separate optimization entirely. This refactor merely touches one file. The primary objective of ensuring that each route row doesn't re-render when a change occurs to an unrelated route row has already been accomplished within this PR.

If I were to attempt to implement the pattern geometry optimization, I'd edit the api actions so that a new action wouldn't get dispatched until all of the geometries for the pattern were fetched.

landonreed: Fair enough. Can you create a ticket describing that approach? After that's done feel free to merge.

evansiroky: Added a comment to #64.

PR #67, opened 2019-06-29
Don't throttle stop time update requests

evansiroky:
My latest changes to throttling were causing some stoptime update requests to be throttled when they probably shouldn't have been.


evansiroky:
Added a comment.

PR #69, opened 2019-07-02
Make user settings configurable, fix print layout

landonreed:
This PR:
- Makes the stored places/recent searches box configurable
- Fixes ibi-group/trimet-mod-otp#79

### DIFF
--- a/lib/components/map/endpoints-overlay.js
+++ b/lib/components/map/endpoints-overlay.js
@@ -3,7 +3,7 @@ import { connect } from 'react-redux'
 
 import Endpoint from './endpoint'
 import { clearLocation, forgetPlace, rememberPlace, setLocation } from '../../actions/map'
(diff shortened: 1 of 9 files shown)


lib/components/map/endpoints-overlay.js
```diff
@@ -27,9 +36,11 @@ const mapStateToProps = (state, ownProps) => {
   // current query is no search is available.
   const activeSearch = getActiveSearch(state.otp)
   const query = activeSearch ? activeSearch.query : state.otp.currentQuery
+  const showUserSettings = state.otp.config.persistence && state.otp.config.persistence.enabled
```
evansiroky: This same check is showing up in a bunch of places. It'd probably be a good idea to make a util function or selector for it.

evansiroky:
This needs a complimentary PR in https://github.com/ibi-group/trimet-mod-otp/.

PR #74, opened 2019-07-08
Fix reverse locations and clear location actions

landonreed:
Previous changes caused reverse locations to not fire a search when the locations were switched.
Similar changes caused the clear location action to keep the current search visible after a location
was cleared. This commit fixes both issues.

fix ibi-group/trimet-mod-otp#215

### DIFF
--- a/lib/actions/map.js
+++ b/lib/actions/map.js
@@ -1,5 +1,7 @@
 import { createAction } from 'redux-actions'
 
+import { routingQuery } from './api'
+import { clearActiveSearch } from './form'
 import { reverse } from '../util/geocoder'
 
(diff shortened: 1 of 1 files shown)


lib/actions/map.js
```diff
@@ -13,18 +15,23 @@ import { reverse } from '../util/geocoder'
  *   }
  */
 
-export const clearingLocation = createAction('CLEAR_LOCATION')
-export const settingLocation = createAction('SET_LOCATION')
-export const switchingLocations = createAction('SWITCH_LOCATIONS')
+// Private actions
```
evansiroky: what are private actions?

evansiroky: ah I see. nvm.

PR #75, opened 2019-07-08
Revert GraphQL

landonreed:
This PR reverts the GraphQL queries employed to fetch routes and stops. After these queries were implemented to assist with the enhancements related to multi-agency implementations, it was later realized that the alternative transit index used by TriMet (backed by gtfsdb) would not support these queries.

### DIFF
--- a/lib/components/map/route-viewer-overlay.js
+++ b/lib/components/map/route-viewer-overlay.js
@@ -2,9 +2,7 @@ import React from 'react'
 import { connect } from 'react-redux'
 import { FeatureGroup, MapLayer, Polyline } from 'react-leaflet'
 
-function geomToArray (point) {
-  return [point.lat, point.lon]
-}
+import polyline from '@mapbox/polyline'
 
 class RouteViewerOverlay extends MapLayer {
   static propTypes = {}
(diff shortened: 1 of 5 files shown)


lib/components/map/route-viewer-overlay.js
```diff
@@ -15,10 +13,27 @@ class RouteViewerOverlay extends MapLayer {
   componentWillUnmount () {}
 
   componentWillReceiveProps (nextProps) {
+    // helper fn to check if geometry has been populated for all patterns in route
+    const isGeomComplete = routeData => {
+      return (
+        routeData &&
+        routeData.patterns &&
+        Object.values(routeData.patterns).reduce(
```
evansiroky: You could use [Array.every](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every) for this.

evansiroky:
Is there a way we can maintain both of these side-by-side? ie, one for gtfsdb and another for OTP TI? Seems like both share common elements and we may end up needed to employ either in the future.

landonreed:
@evansiroky, I think the simplest way to maintain these side-by-side is to just not use GraphQL for now. We'll need to coordinate with TriMet on what makes sense moving forward. I think mixing GraphQL with Rest queries would complicate things unnecessarily in the reducer and some of the component logic. We might be able to string things together with payload rewriters, but I don't want to make things too complicated/difficult to maintain for now.

PR #79, opened 2019-07-10
Fix place names for eScooter rentals

landonreed:
Previously place names for vehicle (scooter) rentals defaulted to the place name, which often tend
to be UUIDs that have no relevance to the ID found on the scooter/vehicle. This changes the place
name to be $COMPANY $VEHICLE_TYPE and uses the same shared component for bike, car, and
micromobility rental.

fix ibi-group/trimet-mod-otp#213

### DIFF
--- a/lib/components/narrative/line-itin/place-row.js
+++ b/lib/components/narrative/line-itin/place-row.js
@@ -1,8 +1,14 @@
 import React, { Component, PureComponent } from 'react'
+import { connect } from 'react-redux'
 
 import LocationIcon from '../../icons/location-icon'
 import ViewStopButton from '../../viewers/view-stop-button'
-import { getPlaceName, isTransit } from '../../../util/itinerary'
+import {
+  getCompanyForNetwork,
+  getModeStringForCompany,
+  getPlaceName,
+  isTransit
+} from '../../../util/itinerary'
 import { formatTime } from '../../../util/time'
 
 import TransitLegBody from './transit-leg-body'
@@ -11,7 +17,7 @@ import AccessLegBody from './access-leg-body'
 // TODO: make this a prop
 const defaultRouteColor = '#008'
 
-export default class PlaceRow extends Component {
+class PlaceRow extends Component {
   _createLegLine (leg) {
     switch (leg.mode) {
       case 'WALK': return <div className='leg-line leg-line-walk' />
@@ -31,7 +37,7 @@ export default class PlaceRow extends Component {
 
   /* eslint-disable complexity */
   render () {
-    const { customIcons, leg, legIndex, place, time, timeOptions, followsTransit, previousLeg } = this.props
+    const { config, customIcons, leg, legIndex, place, time, timeOptions, followsTransit, previousLeg } = this.props
     const stackIcon = (name, color, size) => <i className={`fa fa-${name} fa-stack-1x`} style={{ color, fontSize: size + 'px' }} />
 
     let icon
@@ -59,6 +65,10 @@ export default class PlaceRow extends Component {
     }
 
     const interline = leg && leg.interlineWithPreviousLeg
+    // TODO: This changeVehicles condition might should be showing the full stop
+    // place name because it could be helpful to give the user the stop viewer to
+    // click on in order to see what time the next vehicle will be arriving to the
+    // shared stop.
     const changeVehicles = previousLeg && previousLeg.to.stopId === leg.from.stopId && isTransit(previousLeg.mode) && isTransit(leg.mode)
     const special = interline || changeVehicles
     return (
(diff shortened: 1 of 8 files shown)


lib/components/narrative/line-itin/place-row.js
```diff
@@ -59,6 +65,10 @@ export default class PlaceRow extends Component {
     }
 
     const interline = leg && leg.interlineWithPreviousLeg
+    // TODO: This changeVehicles condition might should be showing the full stop
+    // place name because it could be helpful to give the user the stop viewer to
+    // click on in order to see what time the next vehicle will be arriving to the
```
evansiroky: This comment could use a few grammatical fixes.

lib/components/narrative/line-itin/place-row.js
```diff
@@ -156,14 +170,23 @@ class RentedVehicleLeg extends PureComponent {
       )
     }
 
-    if (leg.rentedVehicleData) {
+    if (leg.rentedVehicle || leg.rentedBike || leg.rentedCar) {
+      // console.log(leg.from.networks, configCompanies)
```
evansiroky: please remove

lib/components/narrative/line-itin/place-row.js
```diff
@@ -156,14 +170,23 @@ class RentedVehicleLeg extends PureComponent {
...
+      const companies = leg.from.networks.map(n => getCompanyForNetwork(n, configCompanies))
+      const companyLabel = companies.map(co => co.label).join('/')
+      const modeString = getModeStringForCompany(companies[0])
```
evansiroky: Since a company can offer multiple modes, this may not be future-proof. Can't we just use the mode of the leg to get the label?

lib/util/itinerary.js
```diff
@@ -390,21 +390,55 @@ export function getLegIcon (leg, customIcons) {
...
-  return place.address ? place.address.split(',')[0] : place.name
+  if (place.address) return place.address.split(',')[0]
+  if (place.networks && place.vertexType === 'VEHICLERENTAL') {
+    // For vehicle rental pick up, do not use the place name. Rather, use
+    // company name + vehicle type (e.g., SPIN eScooter). Place name is often just
+    // a UUID that has no relevance to the actual vehicle. For bikeshare, however,
+    // there are often hubs or bikes that have relevant names to the user.
+    const company = getCompanyForNetwork(place.networks[0], companies)
+    if (company) {
+      return `${company.label} ${getModeStringForCompany(company)}`
```
evansiroky: Since a company can offer multiple modes, this may not be future-proof. Can't we just use the mode of the leg to get the label?

lib/util/itinerary.js
```diff
@@ -390,21 +390,55 @@ export function getLegIcon (leg, customIcons) {
...
-export function getPlaceName (place) {
+export function getCompanyForNetwork (networkString, companies = []) {
+  const company = companies.find(co => co.id === networkString)
+  if (!company) {
+    console.warn(`No company found in config.yml that matches rented vehicle network: ${networkString}`, companies)
+  }
+  return company
+}
+
+export function getModeStringForCompany (company) {
```
evansiroky: Since a company can offer multiple modes, this may not be future-proof. Can't we just use the mode of the leg to get the label?

lib/components/narrative/line-itin/place-row.js
```diff
@@ -92,22 +102,11 @@ export default class PlaceRow extends Component {
...
           )}
 
-          {/* Place subheading: rented bike pickup */}
-          {leg && leg.rentedBike && (
-            <div className='place-subheader'>
-              Pick up shared bike
-            </div>
+          {/* Place subheading: rented vehicle (e.g., scooter, bike, car) pickup */}
+          {leg && (leg.rentedVehicle || leg.rentedBike || leg.rentedCar) && (
+            <RentedVehicleLeg config={config} leg={leg} />
```
evansiroky: I like this consolidation

lib/components/narrative/line-itin/place-row.js
landonreed: I'm just going to go ahead and implement this `TODO`.

lib/util/itinerary.js
landonreed: @evansiroky, after trying this out, I don't think this proposed approach really works that well. Basically, as I mentioned on MS Teams, we would need to check the mode on the next leg in order to determine what vehicle/place the user is walking towards, which would require a larger refactor and feels a little scattered. Perhaps, instead of using the company though, we could check against the vertex type, which will tell us whether it is a car, bike, or "vehicle". How does this sound?

PR #80, opened 2019-07-18
Properly render eScooter legs occurring after forced walking

evansiroky:
The latest code changes for eScooter rental didn't properly handle transitions from forced walking back to renting a vehicle. There was always an assumption that the from leg had vehicle rental info when in reality it does not.

### DIFF
--- a/lib/components/narrative/line-itin/place-row.js
+++ b/lib/components/narrative/line-itin/place-row.js
@@ -167,10 +167,10 @@ class RentedVehicleLeg extends PureComponent {
       )
     }
 
-    let pickUpString = 'Pick up'
+    let rentalDescription = 'Pick up'
     if (leg.rentedBike) {
       // TODO: Special case for TriMet may need to be refactored.
-      pickUpString += ` shared bike`
+      rentalDescription += ` shared bike`
     } else {
       // Add company and vehicle labels.
       let vehicleName = ''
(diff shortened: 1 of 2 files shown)


lib/components/narrative/line-itin/place-row.js
```diff
@@ -156,42 +156,58 @@ class RentedVehicleLeg extends PureComponent {
...
     }
-    if (leg.rentedVehicle || leg.rentedBike || leg.rentedCar) {
-      let pickUpString = 'Pick up'
-      if (leg.rentedBike) {
-        // TODO: Special case for TriMet may need to be refactored.
-        pickUpString += ` shared bike`
-      } else {
-        // Add company and vehicle labels.
+
+    let pickUpString = 'Pick up'
```
landonreed: Since this variable sometimes does not contain `Pick up`, maybe we should rename it to `rentalString` or something.

lib/components/narrative/line-itin/place-row.js
```diff
@@ -156,42 +156,58 @@ class RentedVehicleLeg extends PureComponent {
...
+      let vehicleName = ''
+      // TODO allow more flexibility in customizing these mode strings
+      let modeString = leg.rentedVehicle
+        ? 'eScooter'
+        : leg.rentedBike
+          ? 'bike'
+          : 'car'
+
+      // The networks attribute of the from data will only appear at the very
+      // beggining of the rental. It is possible that there will be some forced
```
landonreed: beggining typo

PR #81, opened 2019-07-18
Handle any route w webapp + support otp.js legacy start route

landonreed:
Previously visiting any random route (e.g., https://trimet-mod-dev.ibi-transit.com/#/foo) causes a blank page to be rendered. This fixes that.

It also adds support for the legacy `start` route used in otp.js to specify lat/lon/zoom/router.

### DIFF
--- a/lib/actions/ui.js
+++ b/lib/actions/ui.js
@@ -74,7 +74,6 @@ export function matchContentToUrl (location) {
           // pathname to specify lat, lon, etc.)
           [,, lat, lon, zoom, routerId] = idToParams(location.pathname, '/')
         }
-        console.log('Setting start position/zoom/router', lat, lon, zoom, routerId)
         // Update map location/zoom and optionally override router ID.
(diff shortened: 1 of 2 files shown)


lib/actions/ui.js
```diff
@@ -65,9 +65,16 @@ export function matchContentToUrl (location) {
...
       case '@':
         // Parse comma separated params (ensuring numbers are parsed correctly).
-        const [lat, lon, zoom, routerId] = id.split(',').map(s => isNaN(s) ? s : +s)
+        let [lat, lon, zoom, routerId] = id ? idToParams(id) : []
+        if (!lat || !lon) {
+          // Attempt to parse path. (Legacy UI otp.js used slashes in the
+          // pathname to specify lat, lon, etc.)
+          [,, lat, lon, zoom, routerId] = idToParams(location.pathname, '/')
+        }
+        console.log('Setting start position/zoom/router', lat, lon, zoom, routerId)
```
evansiroky: Please remove

PR #85, opened 2019-07-23
Replace hash-obj w object-hash

### LINKED ISSUE #84: hash-obj is not compatible with IE 11
I finally tracked down the root cause of the browser compatibility issues (or at least one of them). `hash-obj` is not transpiled for IE, so we should not be using it in otp-rr. There may have also been some issues with incorrect polyfills being included, but that should be resolved with the latest version of trimet-mod-otp. Thoughts on an alternative for this package, @evansiroky?

landonreed:
Fix #84 by replacing hash-obj with object-hash (which is transpiled).

This also updates the example app and makes it IE compatible, which now makes it useful for quickly deploying to test whether the library itself is transpiled/polyfilled correctly (deployed at https://trimet-mod-staging.ibi-transit.com right now), rather than having to bundle through the TriMet app.

Note: this is also waiting on changes from ~conveyal/mastarm#273~ conveyal/mastarm#277


package.json
```diff
@@ -3,15 +3,16 @@
...
     "cover": "mastarm test --coverage",
     "jest": "mastarm test",
     "lint": "mastarm lint lib --quiet",
     "lint-docs": "documentation lint lib/**/*.js",
     "prepublish": "mastarm prepublish --config configurations/prepublish",
     "prestart": "yarn",
     "test": "yarn run lint && yarn run lint-docs && yarn run jest",
-    "start": "mastarm build --serve example.js"
+    "start": "mastarm build --serve example.js",
+    "zip": "zip -r dist.zip index.html example.css dist"
```
evansiroky: I'm pretty sure the zip command won't work on windows. What is the purpose of this script?

landonreed: We can remove this. I just wanted to preserve how to deploy the example. Thoughts on an alternative approach?

.gitignore
```diff
@@ -14,6 +14,7 @@ coverage
 
 # built example app
 dist
+dist.zip
```
evansiroky: This can probably be removed too.

evansiroky:
These changes now make otp-rr work on IE 11. This PR is dependent on https://github.com/conveyal/mastarm/pull/279 and references a commit in mastarm in order for things to work temporarily. Unfortunately, this solution requires the import of `es6-math` in the root js file to polyfill the needed math functions.

evansiroky:
Assigning back to you, @landonreed. Let me know if you think this'll work for now.

landonreed:
What about building the library `dist` for publication on npm? Does that contain the math polyfill or would we also need to import `es6-math` into trimet-mod-otp?

evansiroky:
We'd also need to add es6-math into trimet-mod-otp. The import order of things would matter, so it's best to require the polyfill from the root file.

landonreed:
Gotcha. Ideally mastarm would just work out of the box, but for now this should be good enough. Let's just keep a ticket open in mastarm referencing this solution until we come up with something better.

PR #86, opened 2019-07-25
Geocode refactor

evansiroky:
* Consolidates all geocoding methods into one util file
* Adds the ability to do geocoding with the ArcGIS geocoder


lib/util/geocoder.js
```diff
@@ -1,9 +1,126 @@
...
+    const { 'point.lat': lat, 'point.lon': lon } = response.isomorphicMapzenSearchQuery
+    return {
+      lat,
+      lon,
+      name: response[0].address
+    }
+  }
+}
+
+export default memoize(geocoderConfig => {
```
landonreed: I'm not crazy about having this export not named `getGeocoder`. If I'm searching for a specific function name within the codebase, I would expect to be able to search for and find it (e.g., I often search `functionName (` (with the whitespace in there) to quickly locate the function definition.

lib/util/geocoder.js
```diff
@@ -1,9 +1,126 @@
...
+      return rewriters[key](response)
+    }
+    return response
+  }
 }
 
-export function reverse (point, gcConfig) {
-  return getGeocoder(gcConfig).reverse(point, gcConfig)
+/**
+ * Create customized geocoder funcitons given a certain geocoding API, the
```
landonreed: typo `funcitons`

lib/reducers/create-otp-reducer.js
```diff
@@ -49,6 +51,38 @@ const MAX_RECENT_STORAGE = 5
...
+  'geocoder.type'
+]
+
+function validateInitalState (initialState) {
+  const {config} = initialState
+
+  const errors = []
+
+  // iterate through required keys to ensure integrity of config
+  // TODO: replace with JSON Schema validation, probably in a script that lives outside of the app
```
landonreed: Maybe this TODO should be elevated to the function def? Also, please give this function js doc.

lib/util/geocoder.js
```diff
@@ -1,9 +1,126 @@
...
+ * config for the geocoder and response rewrite functions specific to this
+ * application. Any geocoder api that is added is expected to have an API that
+ * behaves very closely to https://github.com/conveyal/isomorphic-mapzen-search
+ */
+function makeGeocoder (geocoderApi, geocoderConfig, responseRewriters) {
+  const {apiKey, baseUrl, boundary, focusPoint} = geocoderConfig
+  return {
+    autocomplete: query => {
```
landonreed: Could you give each of these functions (autocomplete, getLocationFromGeocodedFeature) some basic comments on their behavior?

lib/reducers/create-otp-reducer.js
```diff
@@ -49,6 +51,38 @@ const MAX_RECENT_STORAGE = 5
 
 // TODO: fire planTrip action if default query is complete/error-free
 
+const requiredConfigKeys = [
```
landonreed: Can you give this variable a comment?

example-config.yml
```diff
@@ -38,17 +38,23 @@ map:
...
     lon: -122.6821804
-  # Optional custom Pelias instance (defaults to Mapzen Search)
+  # required geocoder type. Can be either PELIAS or ARCGIS
+  # WARNING: using the ARCGIS geocoder with a persistence strategy of
+  #   `localStorage` will result in an error being thrown because ARCGIS
+  #   requires the use of a paid account to store geocode results.
+  #   See https://developers.arcgis.com/rest/geocode/api-reference/geocoding-free-vs-paid.htm
+  type: PELIAS
+  # Optional custom api endpoint
   # baseUrl: https://geocoder.example.com/pelias/v1
```
landonreed: Could you add a commented out example config block for the `ARCGIS` geocoder?

__tests__/util/geocoder.js
```diff
@@ -0,0 +1,106 @@
...
+}
+
+describe('geocoder', () => {
+  const geocoders = [
+    {
+      apiKey: 'dummy-mapzen-key',
+      baseUrl: 'https://ws-st.trimet.org/pelias/v1',
+      type: 'PELIAS'
+    }, {
+      type: 'ARCGIS'
```
landonreed: Why is no baseUrl defined for ARCGIS? Isn't this required by the `makeGeocoder` func?

landonreed: Oh, never mind. I guess the default base URL is set in the arcgis geocoder library.

lib/util/geocoder.js
evansiroky: Will fix

lib/reducers/create-otp-reducer.js
```diff
@@ -49,6 +51,46 @@ const MAX_RECENT_STORAGE = 5
 
 // TODO: fire planTrip action if default query is complete/error-free
 
+// a list of required keys that must be present in the initial config of the
+// initial state
+const requiredConfigKeys = [
+  'geocoder.type'
```
landonreed: Sorry I didn't think of this before, but should `geocoder.type` actually be a required key? What if someone wants to set up the app without a geocoder defined? I don't think we should throw an error that prevents the site from loading in this case.

evansiroky: Hmmm, if it's possible to not have a geocoder, some other things should probably change to make things work properly.

evansiroky:
@landonreed, I'm reassigning to you for review since I've made some significant refactors and added support for properly running the app without a geocoder.

PR #89, opened 2019-07-31
Remove star, fix tz issue, and fix index.html

landonreed:
A couple of fixes:
- remove stop viewer star button when user settings is disabled
- fix undefined issue when homeTimezone config not set
- add comment in `index.html` about which js to use

### DIFF
--- a/index.html
+++ b/index.html
@@ -9,6 +9,11 @@
   </head>
   <body>
     <div id="root"></div>
+    <!--
+      Only one of the following js files is needed. For development
+      use example.js. For production use dist/index.js.
+    -->
     <script src="dist/index.js"></script>
+    <script src="/example.js"></script>
   </body>
 </html>
--- a/index.html
+++ b/index.html
@@ -13,7 +13,7 @@
       Only one of the following js files is needed. For development
       use example.js. For production use dist/index.js.
     -->
-    <script src="dist/index.js"></script>
+    <!-- <script src="dist/index.js"></script> -->
     <script src="/example.js"></script>
   </body>
 </html>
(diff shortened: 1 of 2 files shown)


lib/components/viewers/stop-viewer.js
```diff
@@ -166,17 +168,20 @@ class StopViewer extends Component {
...
-              style={{
-                color: this._isFavorite() ? 'yellow' : 'black',
-                padding: 0,
-                marginLeft: '5px'
-              }}
-              bsStyle='link'>
-              <Icon type={this._isFavorite() ? 'star' : 'star-o'} />
-            </Button>
+            {showUserSettings
+              ? <Button
```
evansiroky: You can just do `showUserSettings && <Button...` instead of using a ternary.

index.html
```diff
@@ -9,6 +9,11 @@
   </head>
   <body>
     <div id="root"></div>
+    <!--
+      Only one of the following js files is needed. For development
+      use example.js. For production use dist/index.js.
+    -->
     <script src="dist/index.js"></script>
+    <script src="/example.js"></script>
```
evansiroky: Why are both being required here? Shouldn't one be commented out?

lib/components/viewers/stop-viewer.js
landonreed: Yea, I'm aware -- I was just kind of feeling the ternary here. Does this stand out as a major style concern to you? I use both, but I did like something I read recently that called out using ternaries in favor of `&&` because it was more readable. Also, this: https://github.com/airbnb/javascript/issues/520#issuecomment-167003046

>>> The [`condition && <Button...`] example looks cleanest, but it suffers from a footgun: if condition is zero, it will be toString'ed and rendered.

evansiroky: Ah, that reminds me of https://stackoverflow.com/a/43375944/269834. Maybe if we care a lot about this, we can add a linting rule or something.

PR #91, opened 2019-08-01
Stop viewer fixes

landonreed:
- fix npe if no stop times exist for pattern (test case http://localhost:8080/#/stop/TriMet:13170)
- move noisy warning in vehicle rental overlay to componentDidMount

### DIFF
--- a/lib/components/viewers/stop-viewer.js
+++ b/lib/components/viewers/stop-viewer.js
@@ -13,7 +13,7 @@ import { setMainPanelContent, toggleAutoRefresh } from '../../actions/ui'
 import { findStop, findStopTimesForStop } from '../../actions/api'
 import { forgetStop, rememberStop, setLocation } from '../../actions/map'
 import { routeComparator } from '../../util/itinerary'
-import { getShowUserSettings } from '../../util/state'
+import { getShowUserSettings, getStopViewerConfig } from '../../util/state'
 import { formatDuration, formatStopTime, getTimeFormat } from '../../util/time'
 
 class StopViewer extends Component {
@@ -116,6 +116,7 @@ class StopViewer extends Component {
       showUserSettings,
       stopData,
       stopViewerArriving,
+      stopViewerConfig,
       timeFormat
     } = this.props
     const { spin } = this.state
@@ -237,6 +238,7 @@ class StopViewer extends Component {
                         pattern={patternTimes.pattern}
                         route={patternTimes.route}
                         stopTimes={patternTimes.times}
+                        stopViewerConfig={stopViewerConfig}
                         key={patternTimes.id}
                         stopViewerArriving={stopViewerArriving}
                         homeTimezone={homeTimezone}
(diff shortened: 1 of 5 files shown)


lib/components/viewers/stop-viewer.js
```diff
@@ -285,20 +285,26 @@ class PatternRow extends Component {
       timeFormat
     } = this.props
     // sort stop times by next departure
-    let sortedStopTimes = null
-    if (stopTimes) {
+    let sortedStopTimes = []
+    const hasStopTimes = stopTimes && stopTimes.length > 0
+    if (hasStopTimes) {
       sortedStopTimes = stopTimes.sort((a, b) => {
```
evansiroky: The [Array.sort](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) method sorts an array in-place, so this reassignment is not needed.

lib/components/viewers/stop-viewer.js
```diff
@@ -285,20 +285,26 @@ class PatternRow extends Component {
...
       })
       // Cap the number of times shown for any Route at 5. TODO: make configurable
       if (sortedStopTimes.length > 0) sortedStopTimes = sortedStopTimes.slice(0, 5)
-      // Do not show any patterns with no departures happening soon.
-      const timeIsOverThreshold = sortedStopTimes[0].realtimeDeparture - getHomeTime(homeTimezone) > ONE_HOUR_IN_SECONDS * 3
-      if (sortedStopTimes[0] && timeIsOverThreshold) {
-        return null
-      }
+      // Do not show any patterns with no departures happening soon or departures
+      // that have already occurred (there appear to be cases where tomorrow's
```
evansiroky: Should the comment say "yesterday's" instead of "tomorrow's"? Also, this seems like a bug that should be reported to the main OTP repo.

lib/components/viewers/stop-viewer.js
```diff
@@ -285,20 +285,26 @@ class PatternRow extends Component {
...
+      // times will show up when they should not).
+      const secondsUntilFirstTime = sortedStopTimes[0].realtimeDeparture - getHomeTime(homeTimezone)
+      const firstTimeIsOverThreshold = secondsUntilFirstTime > ONE_HOUR_IN_SECONDS * 3
+      const firstTimeIsInThePast = secondsUntilFirstTime < 0
```
evansiroky: These two thresholds would be good things to make configurable, but for now it's fine to leave these values hardcoded.

lib/components/viewers/stop-viewer.js
evansiroky: Well, I guess this reassignement sort of is needed in order to not reassign a const. Still seems kind of like a bad practice to be manipulating a variable that is a property incoming from a parent component.

lib/components/viewers/stop-viewer.js
landonreed: Let me look into this more... yesterday (Thursday) it was showing times for Friday, so I'm not quite sure what was going on with http://localhost:8080//#/stop/TriMet:13170

lib/components/viewers/stop-viewer.js
landonreed: What if we just concat the array before sorting?

lib/components/viewers/stop-viewer.js
landonreed: OK, this query is basically what I was running into yesterday: https://trimet-otp.ibi-transit.com/otp/routers/default/index/stops/TriMet:13170/stoptimes?timeRange=345600&numberOfDepartures=5&startTime=1564686760

This is asking for stop times at 12:12 PM PT on Thursday (with a time range of four days, which is what the param is configured for in the current code). So it makes sense that we would be getting times for Friday. The departure time values in the response are reported in seconds since midnight, so it would make sense that I was seeing these Friday times show up in the UI.

So this doesn't appear to be a bug in OTP, but we should consider whether we want to filter out these times in the request to OTP or here in the UI.

lib/components/viewers/stop-viewer.js
```diff
@@ -282,23 +284,28 @@ class PatternRow extends Component {
...
-      // Do not show any patterns with no departures happening soon.
-      const timeIsOverThreshold = sortedStopTimes[0].realtimeDeparture - getHomeTime(homeTimezone) > ONE_HOUR_IN_SECONDS * 3
-      if (sortedStopTimes[0] && timeIsOverThreshold) {
-        return null
-      }
+    let sortedStopTimes = []
+    const hasStopTimes = stopTimes && stopTimes.length > 0
+    if (hasStopTimes) {
+      sortedStopTimes = stopTimes
+        .concat()
```
evansiroky: Better, but there's more that could be done. Although this does now avoid modifying the data from the parent component, these values are still being created anew in each render in both this PatternRow component and also the StopViewer render function. Instead of recalculating and resorting on each render, we could create a selector that takes in the stop data and produces the exact display values we need. That way all these sorting, slicing and other operations would happen only once. However, that is all a premature optimization on arrays of a length of typically 5 elements, so I guess this will work for now.

lib/components/viewers/stop-viewer.js
landonreed: I believe prior communication with TriMet revealed that when users are viewing a stop on Friday Monday times should show up for routes that do not have service over the weekend. So we should just leave this as is and remove this filter that I added.

PR #93, opened 2019-08-06
Properly render stop times with certain TriMet index data, fix some timezone problems

evansiroky:
- In certain responses from the TriMet transit index (e.g., https://betaplanner.trimet.org/ws/ti/v0/index/stops/TriMet:715/routes), certain routes don't show up in the list of routes of routes even though they do show up in the list of stop times. In these cases, the stop time data is not added and as a result, the component doesn't run into a NPE when rendering the data.
- This PR also adds a bunch of groundwork that'll make testing react component rendering easier. A fair amount of code was copied from datatools-ui.
- Test cases were added for rendering the StopViewer
- The reducer was refactored slightly in order to make testing with connected components possible.
- A test was added to make sure the reducer can generate the proper initial state
- The home timezone config value is now used more prevalently and correctly. The default query date and time now use this config value and some calculations of stop times also use it as well.
- A refactor was done to minimize the usage of magic time formatting strings in order to bring more clarity as to what time formatting regime is being used since two are used concurrently (one for display purposes and another for OTP query purposes).


lib/components/viewers/stop-viewer.js
```diff
@@ -136,6 +136,12 @@ class StopViewer extends Component {
         const id = `${routeId}-${headsign}`
         if (!(id in stopTimesByPattern)) {
           const route = stopData.routes.find(r => r.id === routeId)
+          // in some cases, the TriMet transit index will not return all routes
+          // that serve a stop. Perhaps it doesn't return some routes if the
+          // route only performs a drop-off at the stop... not quite sure. So a
+          // check is needed to make sure we don't add data for routes not found
+          // from the routes query.
+          if (!route) return
```
landonreed: Let's log a warning here. Otherwise, it might confuse someone expecting to see stop_times for a given route/stop.

evansiroky:
@landonreed, these timezone issues were a real pain to test perfectly. I kept noticing one more thing to do here and there, but I believe I'm finally at a comfortable stopping point. I'm sure there are more edge cases where the calculated time is wrong, but making even more tests with different timezone configurations and edge cases should be easier to implement given these changes. In the process, I also resolved some problems that occurred when a user's timezone differs from the home timezone. And finally, I refactored the time formatting to remove all magic formatting strings in order to make it more clear what time format regime is being used.

PR #94, opened 2019-08-12
Fix date string display in service alert

landonreed:
Use the correct moment method to format date string.

Fix ibi-group/trimet-mod-otp#136.

### DIFF
--- a/lib/components/narrative/line-itin/transit-leg-body.js
+++ b/lib/components/narrative/line-itin/transit-leg-body.js
@@ -203,18 +203,14 @@ class AlertsBody extends Component {
           .sort((a, b) => b.effectiveStartDate - a.effectiveStartDate)
           .map((alert, i) => {
             const effectiveStartDate = moment(alert.effectiveStartDate)
-            let effectiveDateString = 'Effective as of '
             const daysAway = moment().diff(effectiveStartDate, 'days')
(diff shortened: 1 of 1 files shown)


lib/components/narrative/line-itin/transit-leg-body.js
```diff
@@ -203,18 +203,14 @@ class AlertsBody extends Component {
...
-            }
-            effectiveDateString += effectiveStartDate
-              .calendar(
-                null,
-                { sameElse: longDateFormat }
-              )
-              .split(' at')[0]
+            // Add time if alert is effective within one day. Otherwise, use
+            // calendar long date format (e.g., July 31, 2019).
+            const dateTimeFormatter = Math.abs(daysAway) <= 1
```
evansiroky: This variable isn't a formatter, rather it's just a string. I think it'd make sense to rename this variable to indicate that. Everything else looks good.

msrichardson:
To clarify, if service alert effective date is anything other than yesterday/today/tomorrow, please use the actual date (ex: Effective as of August 6, 2019).  If yesterday/today/tomorrow, please use that language (ex: Effective as of 2:33 pm today).   [We don't anticipate ever having a 'tomorrow', but wanted to include that possibility for other entities.]

PR #95, opened 2019-08-13
Log statement to mysteriously fix IE location select onclick

landonreed:
After some amount of debugging I discovered that for some reason on IE, the only way to select a
geocoded result with a click in the location field was to click on the upper part of the MenuItem
(above the text). But by adding a log statement (first I tried to add the statement in the
locationSelected function) anywhere in the app, it seemed to make the entire MenuItem clickable. I
have no idea why this black magic seems to work consistently and had trouble finding info on the
Internet related to this issue.

fix ibi-group/trimet-mod-otp#237

Note: this also contains a tweak that refs ibi-group/trimet-mod-otp#136

### DIFF
--- a/lib/actions/map.js
+++ b/lib/actions/map.js
@@ -38,21 +38,15 @@ export function clearLocation (payload) {
 export function setLocation (payload) {
   return function (dispatch, getState) {
     const otpState = getState().otp
-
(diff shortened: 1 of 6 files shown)


landonreed:
Closing in favor of #96. Thank god this didn't make it into the master branch haha.

PR #96, opened 2019-08-13
Be able to handle clicking a geocoded location in IE11

evansiroky:
A counter-proposal to #95. It appears that there is some funkiness going on when clicking a geocode result in IE 11. If the top part of a result is clicked, then it works, but click on the text and you're out of luck. That made me wonder if it's some weird styling issue, so I removed all the fancy styling and merely display the text alone if the user is using IE and it appears to work now.

Fixes ibi-group/trimet-mod-otp#237


landonreed:
Touché.

PR #98, opened 2019-08-13
Fix service day for arrival times after midnight

landonreed:
Fixes ibi-group/trimet-mod-otp#235.

### DIFF
--- a/lib/components/viewers/stop-viewer.js
+++ b/lib/components/viewers/stop-viewer.js
@@ -398,6 +398,7 @@ class PatternRow extends Component {
 }
 
 const ONE_HOUR_IN_SECONDS = 3600
+const ONE_DAY_IN_SECONDS = 86400
 
 /**
  * @param  {string} homeTimezone Timezone string
@@ -423,19 +424,23 @@ function getFormattedStopTime (stopTime, homeTimezone, soonText = 'Due', timeFor
 
   const now = moment().tz(homeTimezone)
   const serviceDay = moment(stopTime.serviceDay * 1000).tz(homeTimezone)
-  const currentHomeSecondsAfterMidnight = getHomeSecondsAfterMidnight(homeTimezone)
-  const differentDay = now.date() !== serviceDay.date()
-
+  const arrivesAfterMidnight = stopTime.realtimeDeparture >= ONE_DAY_IN_SECONDS
+  // Calculate current home seconds after midnight. If arrival occurs after
+  // midnight, adjust the value by one day in seconds.
+  const currentHomeSecondsAfterMidnight = getHomeSecondsAfterMidnight(homeTimezone) + (arrivesAfterMidnight ? ONE_DAY_IN_SECONDS : 0)
+  // Determine if arrival occurs on different day, making sure to account for an
+  // extra day added to the service day if it arrives after midnight.
+  const arrivalDay = arrivesAfterMidnight
+    ? serviceDay.add(1, 'day')
+    : serviceDay
+  const differentDay = now.dayOfYear() !== arrivalDay.dayOfYear()
   // Determine whether to show departure as countdown (e.g. "5 min") or as HH:mm time
   const secondsUntilDeparture = stopTime.realtimeDeparture - currentHomeSecondsAfterMidnight
-  const departsInFuture = stopTime.realtimeDeparture > currentHomeSecondsAfterMidnight
-  // Show the exact time if the departure occurs after midnight and if the
-  // departure happens within an hour.
-  // FIXME: It's unclear why this was designed to show exact time after midnight.
-  //  We should determine whether this is the desired behavior.
-  const showCountdown = !differentDay &&
-    secondsUntilDeparture < ONE_HOUR_IN_SECONDS &&
-    departsInFuture
+  // Determine if vehicle arrives after midnight in order to advance the day of
+  // the week when showing arrival time/day.
+  const departsInFuture = secondsUntilDeparture > 0
+  // Show the exact time if the departure if the departure happens within an hour.
+  const showCountdown = secondsUntilDeparture < ONE_HOUR_IN_SECONDS && departsInFuture
 
   // Use "soon text" (e.g., Due) if vehicle is approaching.
   const countdownString = secondsUntilDeparture < 60
@@ -447,6 +452,11 @@ function getFormattedStopTime (stopTime, homeTimezone, soonText = 'Due', timeFor
     // in New York, but viewing a trip planner for service based in Los Angeles).
     inHomeTimezone ? timeFormat : `${timeFormat} z`
   )
+  // We only want to show the day of the week if the arrival is on a
+  // different day and we're not showing the countdown string. This avoids
+  // cases such as when it's Wednesday at 11:55pm and an arrival occurs at
+  // Thursday 12:19am. We don't want the time to read: 'Thursday, 24 minutes'.
+  const showDayOfWeek = differentDay && !showCountdown
   return (
     <div>
       <div style={{ float: 'left' }}>
@@ -459,9 +469,9 @@ function getFormattedStopTime (stopTime, homeTimezone, soonText = 'Due', timeFor
             style={{ color: '#888', fontSize: '0.8em', marginRight: 2 }} />
         }
       </div>
-      <div style={{ marginLeft: 20, fontSize: differentDay ? 12 : 14 }}>
-        {differentDay &&
-          <div style={{ marginBottom: -4 }}>{serviceDay.format('dddd')}</div>
+      <div style={{ marginLeft: 20, fontSize: showDayOfWeek ? 12 : 14 }}>
+        {showDayOfWeek &&
+          <div style={{ marginBottom: -4 }}>{arrivalDay.format('dddd')}</div>
         }
         <div>
           {showCountdown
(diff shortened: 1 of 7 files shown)


lib/components/viewers/stop-viewer.js
```diff
@@ -423,19 +424,23 @@ function getFormattedStopTime (stopTime, homeTimezone, soonText = 'Due', timeFor
 
   const now = moment().tz(homeTimezone)
   const serviceDay = moment(stopTime.serviceDay * 1000).tz(homeTimezone)
-  const currentHomeSecondsAfterMidnight = getHomeSecondsAfterMidnight(homeTimezone)
-  const differentDay = now.date() !== serviceDay.date()
-
+  const arrivesAfterMidnight = stopTime.realtimeDeparture >= ONE_DAY_IN_SECONDS
+  // Calculate current home seconds after midnight. If arrival occurs after
+  // midnight, adjust the value by one day in seconds.
+  const currentHomeSecondsAfterMidnight = getHomeSecondsAfterMidnight(homeTimezone) + (arrivesAfterMidnight ? ONE_DAY_IN_SECONDS : 0)
```
evansiroky: This won't work for stop times that occur more than 1 day after the service date. To resolve this you could do something like this:

```javascript
const currentHomeSecondsAfterMidnight = getHomeSecondsAfterMidnight(homeTimezone) + (stopTime.realtimeDeparture - (stopTime.realtimeDeparture % ONE_DAY_IN_SECONDS))
```

lib/components/viewers/stop-viewer.js
```diff
@@ -423,19 +424,23 @@ function getFormattedStopTime (stopTime, homeTimezone, soonText = 'Due', timeFor
...
+  // Determine if arrival occurs on different day, making sure to account for an
+  // extra day added to the service day if it arrives after midnight.
+  const arrivalDay = arrivesAfterMidnight
+    ? serviceDay.add(1, 'day')
```
evansiroky: Won't work for stop times that occur more than 1 day after the service date. Recommended change:

```javascript
serviceDay.add(
  (stopTime.realtimeDeparture - (stopTime.realtimeDeparture % ONE_DAY_IN_SECONDS)) / ONE_DAY_IN_SECONDS, 
  'day'
)

File: lib/components/viewers/stop-viewer.js [hunk 3]
```diff
@@ -423,19 +424,23 @@ function getFormattedStopTime (stopTime, homeTimezone, soonText = 'Due', timeFor
...
-  // departure happens within an hour.
-  // FIXME: It's unclear why this was designed to show exact time after midnight.
-  //  We should determine whether this is the desired behavior.
-  const showCountdown = !differentDay &&
-    secondsUntilDeparture < ONE_HOUR_IN_SECONDS &&
-    departsInFuture
+  // Determine if vehicle arrives after midnight in order to advance the day of
+  // the week when showing arrival time/day.
+  const departsInFuture = secondsUntilDeparture > 0
+  // Show the exact time if the departure if the departure happens within an hour.
```
evansiroky: typo: time _of_ the departure

lib/components/viewers/stop-viewer.js
```diff
@@ -459,9 +469,9 @@ function getFormattedStopTime (stopTime, homeTimezone, soonText = 'Due', timeFor
             style={{ color: '#888', fontSize: '0.8em', marginRight: 2 }} />
         }
       </div>
-      <div style={{ marginLeft: 20, fontSize: differentDay ? 12 : 14 }}>
-        {differentDay &&
-          <div style={{ marginBottom: -4 }}>{serviceDay.format('dddd')}</div>
+      <div style={{ marginLeft: 20, fontSize: showDayOfWeek ? 12 : 14 }}>
+        {showDayOfWeek &&
+          <div style={{ marginBottom: -4 }}>{arrivalDay.format('dddd')}</div>
```
evansiroky: Please make this format string configurable.

landonreed: I'd like to avoid making things configurable for the sake of being configurable. Is there some other format that we expect users of otp-rr might want here? I suppose something like `MM/dd` might be a desirable option, but I'm not totally convinced this is necessary.

lib/components/viewers/stop-viewer.js
landonreed: How could a stop time occur more than 1 day after the service date? Are you suggesting that a stop time could be `49:00:00` (i.e., 1am two days after the service date)? Is that even permitted in GTFS? If so, this is such an edge case that I really don't think it's worth handling. Do we even know how OTP would handle such a stop time?

lib/components/viewers/stop-viewer.js
landonreed: Again, I'm not understanding when a stop time would ever be more than 1 day after the service date https://github.com/opentripplanner/otp-react-redux/pull/98#discussion_r313905657

lib/components/viewers/stop-viewer.js
landonreed: OK, I guess I'll eat my words: https://groups.google.com/d/msg/transit-developers/ZkfnuNv1gho/ic6ObK5PBQAJ

>I haven’t analyzed existing feeds to look for patterns in when overnight service breaks happen, but as someone who has worked on several open source GTFS consuming apps I can say we’ve always considered the fact that some trains run for several days, and would at least accept and attempt to properly handle times up to 48:00. - Andrew Byrd

lib/components/viewers/stop-viewer.js
landonreed: Actually, I think this is incorrect (both mine and yours). I've got a fix coming in though based on your other comment.

lib/components/viewers/stop-viewer.js
```diff
@@ -423,19 +424,30 @@ function getFormattedStopTime (stopTime, homeTimezone, soonText = 'Due', timeFor
 
   const now = moment().tz(homeTimezone)
   const serviceDay = moment(stopTime.serviceDay * 1000).tz(homeTimezone)
+  // Determine if arrival occurs on different day, making sure to account for
+  // any extra days added to the service day if it arrives after midnight. Note:
+  // this can handle the rare (and non-existent?) case where an arrival occurs
+  // 48:00 hours (or more) from the start of the service day.
+  const departureTimeRemainder = stopTime.realtimeDeparture % ONE_DAY_IN_SECONDS
+  const daysAfterServiceDay = (stopTime.realtimeDeparture - departureTimeRemainder) / ONE_DAY_IN_SECONDS
+  const arrivalDay = serviceDay.add(daysAfterServiceDay, 'day')
```
evansiroky: To be consistent, this should be renamed `departureDay` since the data is derived from the departure time.

lib/components/viewers/stop-viewer.js
evansiroky: All of the other calls in otp-rr to `moment.format` don't use hardcoded strings. When these strings are introduced it can easily lead ot inconsistencies with date formatting, so I think it would be good to have consistent formats that can be referred to so it is easy to be reason about why a time is being formatted this way.

lib/components/viewers/stop-viewer.js
```diff
@@ -423,19 +424,30 @@ function getFormattedStopTime (stopTime, homeTimezone, soonText = 'Due', timeFor
...
-    departsInFuture
+  const differentDay = now.dayOfYear() !== arrivalDay.dayOfYear()
+  // Determine whether to show departure as countdown (e.g. "5 min") or as HH:mm
+  // time. If arrival occurs on different (AKA future) day, use unaltered stop
+  // time departure. If the arrival occurs on the same day, use remainder
+  // departure time so that we're comparing times based on actual seconds since
+  // midnight (without interference from times that occur greater than 24 hours
+  // after start of service).
+  const secondsUntilDeparture = differentDay
+    ? stopTime.realtimeDeparture - currentHomeSecondsAfterMidnight
```
evansiroky: I don't think this will work properly for the edge case of a departure time occurring more than 48 hours from the beginning service day. This would be a great opportunity to make a test case for this specific edge case.

landonreed: Can you clarify why you don't think this will work? Let's say it is Wednesday at 5pm (`61200 `) and we have a stop time with a Wednesday service day and a departure time at `172860` (12:01am Friday morning). `secondsUntilDeparture` will be `172860 - 61200 = 111660` or ~ 31 hours. From what I can tell this is the answer we're looking for.

As far as a test case for this -- I don't necessarily agree that it is worth our time to write a test case for an extreme edge case like this.

lib/components/viewers/stop-viewer.js
landonreed: I'm also going to refactor the `differentDay` boolean to be `vehicleDepartsToday`, which I think is a little more clear and make it less likely to be confused with `serviceDay` and `departureDay`.

landonreed:
That is good to know that such a feed exists, but I still don't necessarily think its existence warrants a test case at this point.

lib/components/viewers/stop-viewer.js
evansiroky: Ok, try Thursday at 5pm (`61200` for today, but actually `147600` from the service day before). This will still output `111660` or ~ 31 hours when it should be just `25260` or ~ 7 hours.

Also, in my example feed below, I think this shows that departure times occurring after 48:00:00 are not extreme edge cases. It's quite common actually in Portland where there is a train route that take more than 2 days to get there from Chicago.

PR #99, opened 2019-08-14
Do not send Pelias query sources by default

evansiroky:
Refactors the geocoder a bit in order to make sure Pelias queries don't send the `sources` parameter unless explicitly defined in either the config or individual query.

Fixes https://github.com/ibi-group/trimet-mod-otp/issues/239


landonreed:
@evansiroky, I think you need to update a snapshot.

evansiroky:
Oops

PR #106, opened 2019-09-06
Vehicle wording fixes

evansiroky:
* Changes the wording in map popups to match what is requested in https://github.com/ibi-group/trimet-mod-otp/issues/241
* Will now correctly show available docks and bikes at any zoom level for biketown hubs
* Renames all "eScooter" references to "E-scooter" as requested in https://github.com/ibi-group/trimet-mod-otp/issues/246


evansiroky:
The reason I didn't add to the language files is laziness, but also we haven't yet committed to making sure i18n works well, so maybe after this, we should start doing that?

PR #108, opened 2019-09-10
Properly render service alert body and fix retina tile display

### LINKED ISSUE #107: Support service alert body content
Presently, if a GTFS-rt alert has a body value, it is not rendered in the UI. TriMet alerts do not have bodies, so the original UI was designed without a way to render this. Many other agencies *do* have alert bodies, so we need a way to display these.

landonreed:
Fixes #107 and #109.


lib/components/map/base-layers.js
```diff
@@ -8,16 +9,23 @@ class BaseLayers extends Component {
...
-          <BaseLayer
-            name={l.name}
-            checked={i === 0}
-            key={i}>
-            <TileLayer
-              {...l}
-              detectRetina />
-          </BaseLayer>
-        ))}
+        {baseLayers && baseLayers.map((l, i) => {
```
evansiroky: Please rename `l` to `layers` since we also have this `L` variable. Maybe even rename `L` to `Leaflet`.

evansiroky: Actually, is this file even needed anymore? It doesn't appear to be used anywhere. If not we should delete it.

PR #111, opened 2019-09-18
Fix a bug that occurred when rendering geometry with 0 length

evansiroky:
This PR:

- fixes a bug that occurs when zooming in to a leg that doesn't have geometry info
- allows the configuration of the max zoom level of the map

Fixes https://github.com/ibi-group/trimet-mod-otp/issues/250

### DIFF
--- a/lib/util/itinerary.js
+++ b/lib/util/itinerary.js
@@ -232,7 +232,7 @@ export function getItineraryBounds (itinerary) {
  * Return a leaflet LatLngBounds object that encloses the given leg's geometry.
  */
 export function getLegBounds (leg) {
-  let coords = polyline
+  const coords = polyline
     .toGeoJSON(leg.legGeometry.points)
     .coordinates.map(c => [c[1], c[0]])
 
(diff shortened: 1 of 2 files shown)


lib/util/itinerary.js
```diff
@@ -228,11 +228,24 @@ export function getItineraryBounds (itinerary) {
...
+
+  // in certain cases, there might be zero-length coordinates in the leg
+  // geometry. In these cases, build us an array of coordinates using the from
+  // and to data of the leg.
+  if (coords.length === 0) {
+    coords = [
+      [leg.from.lat, leg.from.lon],
+      [leg.to.lat, leg.to.lon]
+    ]
+  }
```
landonreed: This is a bit of a nit picky style thing, but I kind of like the idea of declaring `coords` with `const` and if the length is zero, pushing the leg endpoints into the array:

```
coords.push([leg.from.lat, leg.from.lon], [leg.to.lat, leg.to.lon])
```

evansiroky: huh, I never knew you could push one _or more_ elements using the `push` method.

PR #113, opened 2019-10-07
Bump mastarm to 5.2.0 for a11y linting

landonreed:
There are a number of a11y issues that keep this library out of compliance with web accessibility standards. The latest version of mastarm now provided a11y linting, so we should fix up any errors it finds.

Errors: https://travis-ci.org/opentripplanner/otp-react-redux/builds/594745948#L219-L262


evansiroky:
Unassigning myself until we decide to work on this.

Issue #122, opened 2020-01-28
Unexpected sorting for routes in route viewer


landonreed:
The sorting used for routes displayed in the route viewer does not follow any intelligible pattern. This is due to the fact that when OTP does not find `route_sort_order` for routes, it defaults to `-999`.
https://github.com/OneBusAway/onebusaway-gtfs-modules/commit/8c54d57d14fb4c5421b63afcbb187ea7cc4276a6

![image](https://user-images.githubusercontent.com/2370911/73285673-614c3080-41c4-11ea-8f5e-086831852132.png)

PR #123, opened 2020-01-28
Fix route sort, color issues

landonreed:
This PR fixes the route sort order issue in #122.

It also addresses some issues with route text color found with SEPTA. When viewing the route viewer ([link](https://beta-plan.septa.org/#/route)), the route text color is not available to render the route label because it is not included in the OTP short response for route. See image with issue:

image

This fix uses a utility function (which should move to otp-ui soon) to identify the right contrasting color for a given bg color.


lib/components/viewers/route-viewer.js
```diff
@@ -110,8 +127,14 @@ class RouteRow extends PureComponent {
...
     const {defaultRouteColor, defaultRouteTextColor, longNameSplitter} = operator || {}
-    const color = `#${defaultRouteTextColor || route.textColor || '000000'}`
     const backgroundColor = `#${defaultRouteColor || route.color || 'ffffff'}`
+    // NOTE: text color is not a part of short response route object, so there
+    // is no way to determine from OTP what the text color should be if the
+    // background color is, say, black. Instead, determine the appropriate
+    // contrast color and use that if no text color is available.
+    const contrastColor = getContrastYIQ(backgroundColor)
+    const color = `#${defaultRouteTextColor || route.textColor || contrastColor}`
+    console.log(route, color, backgroundColor, contrastColor)
```
evansiroky: Please remove console.log

lib/util/itinerary.js
```diff
@@ -247,13 +247,21 @@ export function getLegBounds (leg) {
...
+  // values. NOTE: OTP/OBA defaults sort order values to -999, so instead of
+  // checking that the sort order is not null, we just check that the values are
+  // not equal. Technically, there could be cases where route_sort_order values
+  // are equivalent, but ideally they should be unique.
+  if (a.sortOrder !== b.sortOrder) {
     aComp = a.sortOrder
     bComp = b.sortOrder
   } else if (!isNaN(parseInt(a.shortName)) && !isNaN(parseInt(b.shortName))) {
+    // Otherwise, if both short names can be parsed as integers, use these
+    // numbers for sorting.
```
evansiroky: There is a chance that two routes could have the exact same number value for shortName. In that case, I think it would make sense to defer to sorting based off of shortName/longName.

binh-dam-ibigroup: The viewer doesn't distinguish by agency, does it?

evansiroky: @binh-dam-ibigroup, it used to, but it's been commented out:

https://github.com/opentripplanner/otp-react-redux/blob/121066124e7804575e8dc3a73ad7209f5a755453/lib/components/viewers/route-viewer.js#L135-L140

landonreed: @binh-dam-ibigroup, it does handle agencies:
https://github.com/opentripplanner/otp-react-redux/blob/12d90349be3c7f68e041f422e1092ea79a32d77d/lib/components/viewers/route-viewer.js#L61-L65

PR #124, opened 2020-01-28
Make route comparator tolerate all kinds of data

evansiroky:
- Adds a few functions to determine validity of various route values.
- Refactors route sort order function to sort using various different data formats for routes
- Adds a test suite to check for all kinds of possible route data being sorted properly

### DIFF
--- a/__tests__/util/__mocks__/itinerary.json
+++ b/__tests__/util/__mocks__/itinerary.json
@@ -75,5 +75,17 @@
     "mode": "SUBWAY",
     "shortName": "Yellow",
     "sortOrder": 2
+  },
+  "route14": {
+    "longName": "Xpress route C",
+    "mode": "BUS",
+    "shortName": "30C",
+    "sortOrder": 2
+  },
+  "route15": {
+    "longName": "Express route X",
+    "mode": "BUS",
+    "shortName": "30X",
+    "sortOrder": 2
   }
 }
--- a/__tests__/util/__mocks__/itinerary.json
+++ b/__tests__/util/__mocks__/itinerary.json
@@ -1,0 +1,79 @@
+{
+  "route1": {
+    "longName": "Across town",
+    "mode": "BUS",
+    "shortName": "10",
+    "sortOrder": 10
+  },
+  "route2": {
+    "longName": "Around town",
+    "mode": "BUS",
+    "shortName": "20",
+    "sortOrder": 2
+  },
+  "route3": {
+    "longName": "Around another town",
+    "mode": "BUS",
+    "shortName": "3",
+    "sortOrder": -999
+  },
+  "route4": {
+    "longName": "Loop route",
+    "mode": "BUS",
+    "shortName": "2",
+    "sortOrder": -999
+  },
+  "route5": {
+    "longName": "A-line",
+    "mode": "BUS",
+    "shortName": "A",
+    "sortOrder": -999
+  },
+  "route6": {
+    "longName": "B-line",
+    "mode": "BUS",
+    "shortName": "B",
+    "sortOrder": -999
+  },
+  "route7": {
+    "longName": "A meandering route",
+    "mode": "BUS",
+    "sortOrder": -999
+  },
+  "route8": {
+    "longName": "Zig-zagging route",
+    "mode": "BUS",
+    "shortName": "30",
+    "sortOrder": 2
+  },
+  "route9": {
+    "longName": "Express route",
+    "mode": "BUS",
+    "shortName": "30",
+    "sortOrder": 2
+  },
+  "route10": {
+    "longName": "Variation of express route",
+    "mode": "BUS",
+    "shortName": "30",
+    "sortOrder": 2
+  },
+  "route11": {
+    "longName": "Local route",
+    "mode": "BUS",
+    "shortName": "6",
+    "sortOrder": 2
+  },
+  "route12": {
+    "longName": "Intercity Train",
+    "mode": "RAIL",
+    "shortName": "IC",
+    "sortOrder": 2
+  },
+  "route13": {
+    "longName": "Yellow line Subway",
+    "mode": "SUBWAY",
+    "shortName": "Yellow",
+    "sortOrder": 2
+  }
+}
--- a/__tests__/util/__snapshots__/itinerary.js.snap
+++ b/__tests__/util/__snapshots__/itinerary.js.snap
@@ -1,22 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`util > itinerary routeComparator should prioritize routes with shortNames over those with just longNames 1`] = `
-Array [
-  Object {
-    "longName": "B-line",
-    "mode": "BUS",
-    "shortName": "B",
-    "sortOrder": -999,
-  },
-  Object {
-    "longName": "A meandering route",
-    "mode": "BUS",
-    "sortOrder": -999,
-  },
-]
-`;
-
-exports[`util > itinerary routeComparator should prioritize routes with valid integer shortNames 1`] = `
+exports[`util > itinerary routeComparator should prioritize routes with integer shortNames over alphabetic shortNames 1`] = `
 Array [
   Object {
     "longName": "A-line",
@@ -28,6 +12,22 @@ Array [
     "longName": "Loop route",
     "mode": "BUS",
     "shortName": "2",
+    "sortOrder": -999,
+  },
+]
+`;
+
+exports[`util > itinerary routeComparator should prioritize routes with shortNames over those with just longNames 1`] = `
+Array [
+  Object {
+    "longName": "B-line",
+    "mode": "BUS",
+    "shortName": "B",
+    "sortOrder": -999,
+  },
+  Object {
+    "longName": "A meandering route",
+    "mode": "BUS",
     "sortOrder": -999,
   },
 ]
@@ -62,23 +62,6 @@ Array [
     "longName": "Intercity Train",
     "mode": "RAIL",
     "shortName": "IC",
-    "sortOrder": 2,
-  },
-]
-`;
-
-exports[`util > itinerary routeComparator should sort routes based off of alphabetic shortName with routes with same sortOrder 1`] = `
-Array [
-  Object {
-    "longName": "Around town",
-    "mode": "BUS",
-    "shortName": "20",
-    "sortOrder": 2,
-  },
-  Object {
-    "longName": "Zig-zagging route",
-    "mode": "BUS",
-    "shortName": "30",
     "sortOrder": 2,
   },
 ]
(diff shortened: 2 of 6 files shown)


lib/util/itinerary.js
```diff
@@ -245,26 +245,98 @@ export function getLegBounds (leg) {
...
+ * multi-criteria sort is needed to account for various differences. The sort
+ * occurs as follows:
+ * 1. Attempt to sort based off of routeSortOrder. Routes that do not have a
+ *    valid routeSortOrder will be placed beneath those that do.
+ * 2. In case of the same routeSortOrder criteria, attempt to sort based off of
+ *    routeShortNames that are parseable as integers. routeShortNames that
+ *    cannot be parsed as integers will be placed beneath those that are valid.
+ * 3. In case of the same routeShortName criteria, attempt to sort based off of
+ *    the string routeShortName or routeLongName if the routeShortName is
+ *    undefined.
```
landonreed: I'm confused... this JSdoc suggests that it checks the route sort order and then uses route short names to further sort (and so on for short name/long name). Where in the code is that actually handled? It appears that once a valid routeSortOrder comparison is made, the function simply returns.

evansiroky: Bah, you're right.

landonreed: Dang, you really had me scratching my head!

evansiroky:
All right, this one should actually work.

__tests__/util/itinerary.js
```diff
@@ -1,7 +1,111 @@
...
+      shortName: '30',
+      sortOrder: 2
+    }
+    const route10 = {
+      longName: 'Variation of express route',
+      shortName: '30',
+      sortOrder: 2
+    }
+
+    function sortRouteArray (...routes) {
```
landonreed: Nit pick: function name `sortRoutes` seems more appropriate (or `getRoutesSorted`?).

lib/util/itinerary.js
```diff
@@ -245,31 +245,133 @@ export function getLegBounds (leg) {
...
+ *
+ * @param  {Object} a Route A
+ * @param  {Object} b Route B
+ * @return {number}   A numeric sort comparison value
+ */
+export const routeComparator = makeMultiCriteriaSort(
+  makeNumericValueComparator(obj => getRouteSortOrderValue(obj.sortOrder)),
+  makeNumericValueComparator(obj => parseInt(obj.shortName)),
+  makeObjectStringValueComparator('shortName'),
+  makeObjectStringValueComparator('longName')
```
landonreed: I'm wondering if the `makeObjectStringValueComparator` should just be `makeStringValueComparator` and accept an object getter function like `makeNumericValueComparator`. Seems useful to have parallel structures for these make comparator functions (both for legibility and maintainability).

lib/util/itinerary.js
```diff
@@ -245,31 +245,133 @@ export function getLegBounds (leg) {
...
+ * 1. Attempt to sort based off of sortOrder. Routes that do not have a valid
+ *    sortOrder will be placed beneath those that do.
+ * 2. In case of the same sortOrder criteria, attempt to sort based off of
+ *    shortNames that are parseable as integers. shortNames that cannot be
+ *    parsed as integers will be placed beneath those that are valid.
+ * 3. In case of the same integer shortName criteria, attempt to sort based
+ *    off of the string shortName. Routes without shortNames will be placed
+ *    beneath those with shortNames.
+ * 4. In case of the same string shortName, attempt to sort based off of
+ *    the string longName.
```
landonreed: The way this list of sorting criteria is worded is somewhat confusing to me. Instead of nestling the different cases/conditions in each item in the comment, I think it would be more helpful to state that when a criteria is found to be equivalent for a pair of routes, the next criteria is used. Then just list the criteria being used. This is somewhat of a restatement of the `makeMultiCriteriaSort` jsdoc, but I think it bears repeating. How I might rewrite it:

```
Compares routes for the purposes of sorting and displaying in a user interface. Due to GTFS feeds having varying levels of data quality, a multi-criteria sort is needed to account for various differences. The criteria included here are each applied to the routes in the order listed. If a given sort criterion yields equivalence (e.g., two routes have the short name "20"), the comparator falls back onto the next sort criterion (e.g., long name). The sort occurs as follows:
 1. sortOrder. Routes that do not have a valid sortOrder will be placed beneath those that do.
 2. shortName as integer. shortNames that cannot be parsed as integers will be placed beneath those that are valid.
 3. shortName as string. Routes without shortNames will be placed beneath those with shortNames.
 4. longName as string.
```

lib/util/itinerary.js
```diff
@@ -245,31 +245,133 @@ export function getLegBounds (leg) {
...
+    } else {
+      aVal = a
+      bVal = b
+    }
+    // if both values aren't valid numbers, use the next sort criteria
+    if (isNaN(aVal) && isNaN(bVal)) return 0
+    // b is a valid number, b gets priority
+    if (isNaN(aVal)) return 1
+    // a is a valid number, a gets priority
+    if (isNaN(bVal)) return -1
```
landonreed: I'm a little worried about embedding this priority for numbers over strings in this comparator. If we decide that strings should come before numbers, we would have to refactor this function pretty heavily. There are a number of cases where we may actually want to give strings priority over ints (think about how TriMet [lists their routes](https://newplanner.trimet.org/map/#/route)). Perhaps this function should also except a boolean arg `stringsFirst` to indicate whether strings should be given higher priority in the sort (and these lines could use that arg)?

lib/util/itinerary.js
```diff
@@ -245,31 +245,133 @@ export function getLegBounds (leg) {
...
+function makeObjectStringValueComparator (key) {
+  return (a, b) => {
+    const aVal = a[key]
+    const bVal = b[key]
+    // both a and b are uncomparable strings, return equivalent value
+    if (!aVal && !bVal) return 0
+    // a is not a comparable string, b gets priority
+    if (!aVal) return 1
+    // b is not a comparable string, a gets priority
+    if (!bVal) return -1
```
landonreed: Again, wondering about this embedded logic similar to https://github.com/opentripplanner/otp-react-redux/pull/124/files#r372464064

lib/util/itinerary.js
evansiroky: I went ahead and added an option to forgo sorting by integer shortName. I think it'd be best to keep that because I can see it being quite common to have routes with "numeric" shortNames (ex: "3", "10", "20") and if we didn't sort by integer shortName, we'd get non-numerically sorted values (ex: "10", "20", "3").

evansiroky:
See additional changes and this comment: https://github.com/opentripplanner/otp-react-redux/pull/124#discussion_r372803892

lib/util/itinerary.js
landonreed: Sorry, I think maybe my initial point wasn't super clear. I do think we should sort shortNames as integers; however, I think that in some cases (maybe all) it might be preferable to have non-integer shortNames (and perhaps longNames) show up before integer shortNames in a list. E.g., Agency X may have numerically-named bus routes and color-named train lines. Because the train lines are more heavy utilized (and likely a shorter list) they might should show up first for better discoverability. So the order we're looking for would be:
```
BLUE
GREEN
RED
1
2
10
19
22
25
100
etc.
```

binh-dam-ibigroup: I think so too. I'm also in favor of having one snapshot that has all the cases, so it shows what a normal route list could look like.

__tests__/util/itinerary.js
```diff
@@ -1,7 +1,122 @@
-import {isTransit} from '../../lib/util/itinerary'
+import {isTransit, makeRouteComparator} from '../../lib/util/itinerary'
 
 describe('util > itinerary', () => {
   it('isTransit should work', () => {
     expect(isTransit('CAR')).toBeFalsy()
   })
+
+  describe('routeComparator', () => {
+    const route1 = {
```
binh-dam-ibigroup: Does it make sense to move these declarations to another file?

lib/util/itinerary.js
evansiroky: @landonreed and @binh-dam-ibigroup, I just pushed a new commit that introduces two more comparator functions which I think will address these concerns. I've added a route type comparator that sorts based on route type. This will prioritize fixed guideway transit and ferries over buses. Secondly, I added a comparator that prioritizes routes with shortNames that begin with an alphabetic character.

__tests__/util/__mocks__/itinerary.json
```diff
@@ -0,0 +1,79 @@
...
+    "shortName": "IC",
+    "sortOrder": 2
+  },
+  "route13": {
+    "longName": "Yellow line Subway",
+    "mode": "SUBWAY",
+    "shortName": "Yellow",
+    "sortOrder": 2
+  }
+}
```
landonreed: I'm not sure if much is gained from moving this into a separate json file, especially since it gets destructured in the test file. Probably not worth another refactor though.

__tests__/util/__snapshots__/itinerary.js.snap
```diff
@@ -0,0 +1,235 @@
...
+    "longName": "A-line",
+    "mode": "BUS",
+    "shortName": "A",
+    "sortOrder": -999,
+  },
+  Object {
+    "longName": "Loop route",
+    "mode": "BUS",
+    "shortName": "2",
+    "sortOrder": -999,
```
landonreed: I think this test might need to be renamed?

__tests__/util/__snapshots__/itinerary.js.snap
```diff
@@ -0,0 +1,235 @@
...
+    "longName": "Around town",
+    "mode": "BUS",
+    "shortName": "20",
+    "sortOrder": 2,
+  },
+  Object {
+    "longName": "Zig-zagging route",
+    "mode": "BUS",
+    "shortName": "30",
+    "sortOrder": 2,
```
landonreed: Test name mismatch?

__tests__/util/__mocks__/itinerary.json
landonreed: It feels a bit like maybe some of the tests should just take JSON objects directly as arguments to the `sortRoutes` func. Everything feels a bit scattered by having to check which routes are which back in this file. It seems like a few of the tests have become incorrectly named as a result of this disjointedness.

__tests__/util/__mocks__/itinerary.json
```diff
@@ -0,0 +1,79 @@
...
+  "route8": {
+    "longName": "Zig-zagging route",
+    "mode": "BUS",
+    "shortName": "30",
+    "sortOrder": 2
+  },
+  "route9": {
+    "longName": "Express route",
+    "mode": "BUS",
+    "shortName": "30",
```
binh-dam-ibigroup: Name this one with shortName "30X" to see how it will get sorted!

evansiroky:
Just pushed a commit that renames a test and removes one that's no longer needed.

evansiroky:
...and I added a test case to prove that alphanumeric short names seem to get sorted properly.

PR #125, opened 2020-02-13
From-To-Location-Picker

binh-dam-ibigroup:
This PR starts integrating the from/to location pickers from @opentripplanner/otp-ui into otp-react-redux.


lib/components/map/set-from-to.js
```diff
@@ -18,19 +18,7 @@ export default class SetFromToButtons extends Component {
...
-        >
-          <LocationIcon type='from' /> From here
-        </Button>
-        <Button bsSize='xsmall' className='set-to-button'
-          onClick={this._setToClicked}
-        >
-          <LocationIcon type='to' /> To here
-        </Button>
-      </div>
+      <FromToLocationPicker onFromClick={this._setFromClicked} onToClick={this._setToClicked} />
```
evansiroky: Is there a way to completely replace the `SetFromToButtons` component with the `FromToLocationPicker`? The `FromToLocationPicker` is able to call a `setLocation` method with the same arguments that are in here.

binh-dam-ibigroup: I will decline for now. We will need our own Popup class for use inside overlays and inside `BaseMap` before we can inline `SetFromToButtons`.  Right now, `SetFromToButtons` takes in a map so it can close the popup when you click on the links.

PR #129, opened 2020-02-14
eslint packages for showing style errors inside IDEs

binh-dam-ibigroup:
This PR adds babel/eslint plugins to package.json so that code editors can show style errors on files you are working on. Right now, it is so that the current code passes the yarn lint stage.

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,13 +79,11 @@
   "devDependencies": {
     "@babel/cli": "^7.4.4",
     "@babel/core": "^7.4.5",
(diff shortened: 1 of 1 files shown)


landonreed:
Closing for now. The approach we typically use is to reference mastarm's eslint config file in the IDE, rather than define a project-specific file. Mastarm's linting config (or at least the config defined in the mastarm version for our projects) may be a bit out of date though, so it might be good to update that across our projects.

PR #131, opened 2020-02-18
Fix 'disappearing' routes on click

### LINKED ISSUE #130: Route type comparator does not consider GTFS route types in long route response
OTP has two different response types in the transit index for routes: long and short.

The short response (which the route type comparator in otp-rr) currently handles contains a `Route@mode` field, which maps to the OTP transit mode enum values (e.g., `BUS` and `RAIL`). However, in the long response from OTP, the `mode` field is dropped in favor of `route_type`, which uses the [int-based route types from GTFS](https://developers.google.com/transit/gtfs/reference/#routestxt). So, when a route is clicked in the otp-rr route viewer, the route object is replaced with the long response, and it drops the `mode` field. Originally, I thought the route was just disappearing from the list, but in fact it was dropping to the bottom of the list (in accordance with the route type comparator logic -- default to `9999`).

https://github.com/opentripplanner/otp-react-redux/blob/a16abb4a8d99b16a24802673fe0542318e2fc837/lib/util/itinerary.js#L278-L289

landonreed:
Fixes #130.

### DIFF
--- a/lib/util/itinerary.js
+++ b/lib/util/itinerary.js
@@ -262,9 +262,9 @@ function getSortValues (getterFn, a, b) {
   return { aVal, bVal }
 }
 
-/**
- * Lookup for the sort values associated with various OTP modes.
- */
+// Lookup for the sort values associated with various OTP modes.
+// Note: JSDoc format not used to avoid bug in documentationjs.
+// https://github.com/documentationjs/documentation/issues/372
 const modeComparatorValue = {
   SUBWAY: 1,
   TRAM: 2,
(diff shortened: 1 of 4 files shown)


lib/util/itinerary.js
```diff
@@ -262,29 +262,52 @@ function getSortValues (getterFn, a, b) {
...
-      return 7
-    case 'BUS':
-      return 8
-    default:
-      return 9999
+  // For some strange reason, the short route response in OTP returns the
+  // string-based modes, but the long route response returns the
+  // integer route type. This attempts to account for both of those cases.
+  if (typeof route.mode !== 'undefined') {
+    return modeComparatorValue[route.mode]
```
binh-dam-ibigroup: Warning, this block can return `undefined` can can break other pieces of code that rely on = the `9999` return value...

PR #134, opened 2020-02-19
Fix blank screen error on set locations

### LINKED ISSUE #133: Setting from / to location via location search on mobile causes a white screen
Note: this issue is only detectable on a mobile device (using Chrome dev tools to simulate mobile on a desktop is not enough to trigger).

When a user sets the from and to locations via a geocoded result in the location search component, there is a fitBounds invocation in `BaseMap` that causes an error to be thrown resulting in a blank white screen.
https://github.com/opentripplanner/otp-react-redux/blob/7c175ee524ef2366d41478c48a970b3960a7b10c/lib/components/map/base-map.js#L128-L133
The error logged in the console suggests an invalid latlng object, but adding a timeout or removing the fitBounds invocation on mobile both appear to be viable solutions. Attempting to set a timeout (with a value of 250 seconds), works, but the map doesn't actually appear to fit to the bounds of the from/to locations. So a better approach may be to just disable the fitBounds call on mobile devices.

landonreed:
fix #133

### DIFF
--- a/lib/components/map/base-map.js
+++ b/lib/components/map/base-map.js
@@ -11,6 +11,7 @@ import LocationIcon from '../icons/location-icon'
 import { constructLocation } from '../../util/map'
 import { getActiveItinerary, getActiveSearch } from '../../util/state'
 import { getItineraryBounds, getLegBounds, legLocationAtDistance } from '../../util/itinerary'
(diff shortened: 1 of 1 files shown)


landonreed:
I'm in agreement here as well. Thanks, all.

PR #137, opened 2020-03-02
otp-ui location field and geocoder refactor

evansiroky:
This PR replaces the location field and geocoding functionality using code sourced from otp-ui.

Also, this code fixes an annoying bug that would overlay the UserSettings component on the welcome screen when user data persistence was active.


landonreed:
@evansiroky, I believe the user settings overlay was a feature ;)

binh-dam-ibigroup:
IMO the UI tweak should be addressed as a separate issue.

evansiroky:
See commit pushed to resolve styling issues. See https://github.com/opentripplanner/otp-ui/pull/89 for fixes to blurring of location-field.

PR #139, opened 2020-03-11
DateTimeSelector from OTP-UI, Remove Profile Routing

binh-dam-ibigroup:
### Checklist

- [x] Appropriate branch selected _(all PRs must first be merged to `dev` before they can be merged to `master`)_
- [x] Any modified or new methods or classes have helpful JSDoc and code is thoroughly commented
- [na] The description lists all applicable issues this PR seeks to resolve
- [na] The description lists any configuration setting(s) that differ from the default settings
- [x] All tests and CI builds passing

### Description

This PR replaces the DateTimeSelector component with [the one from OTP-UI](https://github.com/opentripplanner/otp-ui/blob/master/packages/trip-form/src/DateTimeSelector/index.js). For the DateTimeSelector component, the bootstrap and OTP-rr CSS attributes are reproduced inside a styled-component implementation. 

The PR also removes code dealing with profile routing, most of which was previously commented out.


lib/components/form/date-time-modal.js
```diff
@@ -2,74 +2,126 @@
...
         <div className='main-panel'>
-          {rtDefaults.find(d => d.key === routingType).component}
+          <StyledDateTimeSelector
+            className='date-time-selector'
+            date={date}
+            dateFormatLegacy={dateFormatLegacy}
+            departArrive={departArrive}
+            onQueryParamChange={setQueryParam}
+            time={time}
+            timeFormatLegacy={timeFormatLegacy}
```
landonreed: Could you explain why we're using props called `*Legacy`? This kind of splitting of code paths within the components doesn't seem necessary -- we should not be considering the otp-react-redux library "legacy" software. I don't think I had a chance to review opentripplanner/otp-ui#32, otherwise, I wouldn't have supported the splitting of the handlers like this.

lib/index.js
```diff
@@ -63,7 +62,6 @@ export {
   // form components
   DateTimeModal,
   DateTimePreview,
-  DateTimeSelector,
```
landonreed: Hmm, do we actually want to remove this component here? Or do we want to export the styled date time selector in its place? Just trying to think through how this library and its exports might continue to be used.

binh-dam-ibigroup: Good question. @evansiroky what is your take?

lib/components/form/date-time-modal.js
binh-dam-ibigroup: I agree on adding more documentation to `DateTimeSelector` in OTP-ui. There is a tiny bit [here](https://github.com/opentripplanner/otp-ui/blob/master/packages/trip-form/src/DateTimeSelector/index.js#L213).

binh-dam-ibigroup: See my adjusted the comments that are relevant to OTP-RR.

binh-dam-ibigroup: I added additional docs on the OTP-UI side, see https://github.com/opentripplanner/otp-ui/pull/98

lib/index.js
evansiroky: Seems like it should be in the index file. But, our only real use of otp-react-redux is via trimet-mod-otp so I'm not sure.

binh-dam-ibigroup: We have been deleting the refactored components from index.js. See also here: https://github.com/opentripplanner/otp-react-redux/pull/147/files

PR #141, opened 2020-03-19
OTP-UI trip settings

binh-dam-ibigroup:
This PR brings components of OTP-UI trip-form package into OTP-react-redux.
The SettingsSelectorPanel and DateTimeSelector components have been replaced with their counterparts in OTP-UI. Form components that became unused have been removed.

The PR leaves only one thing out pending [otp-ui PR 92](https://github.com/opentripplanner/otp-ui/pull/92). When you select an access mode (e.g. Transit + Bike), the title does not get bold (it was previously).


binh-dam-ibigroup:
When did `mastarm` become so uptight on a11y requirements??

lib/components/form/user-trip-settings.js
```diff
@@ -0,0 +1,78 @@
...
+
+import Icon from '../narrative/icon'
+import {
+  clearDefaultSettings,
+  resetForm,
+  storeDefaultSettings
+} from '../../actions/form'
+import { getTripOptionsFromQuery, isNotDefaultQuery } from '../../util/query'
+
+class UserTripSettings extends Component {
```
landonreed: Could you add javadoc to this component?

binh-dam-ibigroup: I added the jsdoc for the component.

PR #142, opened 2020-03-23
Trip Details, Trip Viewer Overlay, and Printable Itinerary.

binh-dam-ibigroup:
This PR replaces `TripDetails` in lib/components/narrative, `TripViewerOverlay` in lib/components/map, and `PrintableItinerary` in `lib/components/narrative/printable`  with their counterparts from OTP-ui.

Visual changes in `TripDetails` are shown in the screenshot below.
image

Visual changes in `PrintableItinerary` are shown in the screenshot below.
image


binh-dam-ibigroup:
I am also adding `PrintableItinerary` to this PR because that component (actually, `PrintLayout`) uses `TripDetails`.

lib/components/app/print-layout.js
```diff
@@ -53,8 +54,24 @@ class PrintLayout extends Component {
     root.removeAttribute('class')
   }
 
+  /**
+   * Use one of the customIcons if provided,
+   * (similar to @opentraiplanner/trip-form/ModeIcon)
```
landonreed: typo

lib/components/app/print-layout.js
```diff
@@ -82,13 +99,15 @@ class PrintLayout extends Component {
...
         {/* The main itinerary body */}
-        {itinerary
-          ? <PrintableItinerary
-            configCompanies={configCompanies}
-            customIcons={customIcons}
-            itinerary={itinerary}
-            timeFormat={timeFormat} />
-          : null
+        {itinerary &&
+          <div>
```
landonreed: Maybe we should use a [fragment](https://reactjs.org/docs/fragments.html) here instead of a div.
```
<>
           <PrintableItinerary
              config={config}
              itinerary={itinerary}
              LegIcon={this.customLegIcon(customIcons)}
            />
            <TripDetails itinerary={itinerary} />
</>
```

lib/components/map/connected-trip-viewer-overlay.js
```diff
@@ -0,0 +1,19 @@
...
+  const viewedTrip = state.otp.ui.viewedTrip
+  return {
+    tripData: viewedTrip
+      ? state.otp.transitIndex.trips[viewedTrip.tripId]
+      : null
+  }
+}
+
+const mapDispatchToProps = {
+}
```
landonreed: Stylistically, putting `{ }` on the same line seems better to me.

lib/components/narrative/default/default-itinerary.js
```diff
@@ -3,7 +3,7 @@ import React from 'react'
 import NarrativeItinerary from '../narrative-itinerary'
 import ItinerarySummary from './itinerary-summary'
 import ItineraryDetails from './itinerary-details'
-import TripDetails from '../trip-details'
+import TripDetails from '../connected-trip-details'
```
landonreed: Are we following this convention with the new otp-ui components everywhere? cc @evansiroky. Just want to make sure we're consistent with this.

lib/components/narrative/connected-trip-details.js
```diff
@@ -0,0 +1,25 @@
+import { connect } from 'react-redux'
+import styled from 'styled-components'
+import TripDetailsBase from '@opentripplanner/trip-details'
```
landonreed: Wondering if there's a better name for this import. `UnstyledTripDetails`? I'm fine with what you have, but it does seem like there might be a better name out there.

lib/components/app/print-layout.js
binh-dam-ibigroup: Good suggestion.

lib/components/narrative/default/default-itinerary.js
binh-dam-ibigroup: @evansiroky and I have discussed that, I'm following his lead on that.

binh-dam-ibigroup:
There are no changes to the `config` object as far as I know.

lib/components/app/print-layout.js
```diff
@@ -53,8 +54,24 @@ class PrintLayout extends Component {
     root.removeAttribute('class')
   }
 
+  /**
+   * Use one of the customIcons if provided,
+   * (similar to @opentriplanner/trip-form/ModeIcon)
+   * otherwise, fall back on TriMetLegIcon.
+   * TODO: Combine all custom icon rendering in one place.
+   */
+  customLegIcon = icons => {
```
evansiroky: I was initially going to say that we should go ahead and immediately start using the TriMetLegIcon, but ideally we'd make this TriMetLegIcon be available via a slot in the overall application because we aren't always going to want to use the TriMetLegIcon in all of our deployments.

lib/components/narrative/connected-trip-details.js
evansiroky: I'm not super opinionated about this. When I do this I usually do this the other way around a la:

```js
import TripDetails from  '@opentripplanner/trip-details'

const StyledTripDetails = styled(TripDetails)`
...
```

evansiroky: Would be nice to agree on some kind of convention.

lib/components/narrative/default/default-itinerary.js
evansiroky: Guess it's settled then.

binh-dam-ibigroup:
Unfortunately the size of the question mark icon is hard-coded inside OTP-UI...do we want a ticket for this? Also, the bus and walk icons have these proportions natively in OTP-UI icons, so we would have to either resize the icons in OTP-UI or pollute the code with size adjustments... what do you say?

binh-dam-ibigroup:
When https://github.com/opentripplanner/otp-ui/pull/104 and https://github.com/opentripplanner/otp-ui/pull/105 get released, you will see the icon size adjustments in OTP-RR. There are no other code work on the OTP-RR repo for integrating trip-details and printable itinerary.

binh-dam-ibigroup:
Thanks for the approval. I need to merge the latest changes from dev, then I'll merge this into dev.

PR #151, opened 2020-04-07
OTP-UI humanize-distance

binh-dam-ibigroup:
This PR replaces the distance functions with the ones from the humanize-distance package from OTP-UI.


binh-dam-ibigroup:
Good idea on the configuration to switch to metric units.

binh-dam-ibigroup:
We will deal with configuration to show narratives in metric in https://github.com/opentripplanner/otp-ui/issues/116 and https://github.com/opentripplanner/otp-react-redux/issues/158.

PR #152, opened 2020-04-08
Add a component for the elevation point marker

evansiroky:
As a part of the map refactor we left out the elevation point marker display on the map. This PR adds it back in a new map component. This PR depends on https://github.com/opentripplanner/otp-ui/pull/107 in order to work properly.


lib/components/map/elevation-point-marker.js
```diff
@@ -0,0 +1,45 @@
...
+import React, { Component } from 'react'
+import { connect } from 'react-redux'
+import { CircleMarker } from 'react-leaflet'
+
+class ElevationPointMarker extends Component {
+  render () {
+    const { diagramLeg, elevationPoint, showElevationProfile } = this.props
+
+    let elevationPointMarker = null
+    if (showElevationProfile && diagramLeg && elevationPoint) {
```
landonreed, line 11: I think this used to have a comment:

```
// Compute the elevation point marker, if activeLeg and elevation profile is enabled.
```

lib/components/map/elevation-point-marker.js
```diff
@@ -0,0 +1,45 @@
+import coreUtils from '@opentripplanner/core-utils'
+import React, { Component } from 'react'
+import { connect } from 'react-redux'
+import { CircleMarker } from 'react-leaflet'
+
+class ElevationPointMarker extends Component {
```
landonreed, line 6: Let's give the class some jsdoc about how this works/what it's showing/how to turn it on.

binh-dam-ibigroup:
@landonreed, @evansiroky is right about opentripplanner/otp-ui#107. To make this work, I am reverting the core-utils call to the old itinerary function with the corrected turlAlong syntax.

PR #156, opened 2020-04-22
Add User Sign-In

binh-dam-ibigroup:
This PR adds a sign-in component, implemented as a native react-bootstrap `<NavItem>` or `<NavDropdown>`, as shown in the screenshots below.

Unlike what the branch name may imply, there is no sign up functionality in this PR.

The following features are added:
* To enable Auth0 authentication, add the following entries to`config.yml`:
```yml
persistence:
  enabled: true
  strategy: <anything except localStorage>
  auth0:
    audience: <auth0 audience>
    domain: <auth0 authentication domain>
    clientId: <auth0 client id>
```
* When logged in, the UI shows the user's auth0 avatar, account links (not implemented), and sign out link.
* The user's current trip search is preserved when logging in or out.

## Screenshots

Mobile:
image

Desktop:
image


binh-dam-ibigroup:
The PR is out there, but it shouldn't be merged until the OTP-UI port is complete IMO.

lib/util/auth.js
```diff
@@ -0,0 +1,105 @@
...
+ * - persistence.strategy is defined and is not 'localStorage',
+ * - persistence.auth0 is defined.
+ * @param otpConfig The OTP configuration params from config.yml.
+ * @returns The Auth0 configuration, or null if the conditions are not met.
+ */
+export function getAuth0Config (otpConfig) {
+  const { persistence } = otpConfig
+  if (persistence) {
+    const { enabled = false, strategy = null, auth0 = null } = persistence
+    return (enabled && strategy && strategy !== 'localStorage') ? auth0 : null
```
evansiroky, line 48: This all seems very specific to Auth0, so I think we should make the strategy required to be "Auth0" here.

lib/components/mobile/navigation-bar.js
```diff
@@ -1,12 +1,15 @@
-import React, { Component } from 'react'
 import PropTypes from 'prop-types'
-import { connect } from 'react-redux'
+import React, { Component } from 'react'
 import { Navbar } from 'react-bootstrap'
 import FontAwesome from 'react-fontawesome'
+import { connect } from 'react-redux'
 
-import AppMenu from '../app/app-menu'
 import { setMobileScreen } from '../../actions/ui'
```
evansiroky, line 8: Unnecessary empty line

lib/components/user/nav-login-button.js
```diff
@@ -0,0 +1,117 @@
...
+        }
+        .otp.mobile .navbar li.dropdown.open {
+          background-color: #080808;
+        }
+        .otp.mobile .navbar .container-fluid > li.dropdown.open > a {
+          color: #fff;
+        }
+      `}
+      </style>
+    )
```
landonreed, line 81: For now, can you just make the link color match the hamburger link? Color: white, onhover color: `#ddd`. Not sure why this is grey right now (in mobile and desktop).

lib/util/auth.js
binh-dam-ibigroup: I don't think the strategy itself should be set to Auth0. In the configuration, that is set to `otp_middleware` (you might have middleware with different authentication methods). The auth0-specific check is for `persistence.auth0` (in the deconstructed var declarations from `presistence`).
```yml
persistence:
  enabled: true
  strategy: otp_middleware
  auth0:
    audience: https://otp-middleware
    domain:
    clientId:
```

lib/components/user/nav-login-button.js
binh-dam-ibigroup: Bootstrap makes drop-down links in dark navigation bars grey by default.

lib/util/auth.js
evansiroky: I like that you've clarified the way the config should be setup. However, in the code there is still a check that the strategy isn't `localStorage`. I'd prefer this to directly check if the strategy is `otp-middleware`.

lib/components/user/nav-login-button.js
```diff
@@ -0,0 +1,124 @@
...
+    })),
+    onSignInClick: PropTypes.func.isRequired,
+    onSignOutClick: PropTypes.func.isRequired,
+    profile: PropTypes.shape({
+      email: PropTypes.string.isRequired,
+      name: PropTypes.string.isRequired,
+      nickname: PropTypes.string,
+      picture: PropTypes.string
+    })
+  }
```
landonreed, line 31: @evansiroky, what is your view on using proptypes here? Personally, I favor flow.js over proptypes definitions -- it's much more concise and we actually get the advantage of linting the types (maybe that happens with proptypes, too?). I understand that proptypes was useful for the otp-ui project, but I'm concerned about the typing being actually useful during development.

lib/components/user/nav-login-button.js
```diff
@@ -0,0 +1,124 @@
...
+
+  static defaultProps = {
+    links: null,
+    profile: null
+  }
```
landonreed, line 36: Are these default props really necessary?

lib/components/user/nav-login-button.js
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+      * Override bootstrap's defaults, especially for nav links.
+      * HACK: Because we don't want to use the rendered <NavDropDown> inside a <Nav> while in mobile mode
+      * (otherwise, the menu items get displayed full screen width, which is undesirable),
+      * the CSS rules below let us fit the sign-in component to the top-right corner
+      * and add some of the default visuals provided by react-bootstrap's <Nav> while in mobile mode.
+      * We change the link color to white for consistency with the AppMenu component.
+      */
+    const styleOverrides = (
```
landonreed, line 63: @binh-dam-ibigroup, I would prefer for this css to be in a separate css file, similar to how we handle the other pieces of this application. Any reason why it's inline with the js?

lib/components/user/nav-login-button.js
```diff
@@ -0,0 +1,124 @@
...
+          {links && links.map((item, index) => (
+            <MenuItem href={item.url} key={index} target={item.target}>{item.text}</MenuItem>
+          ))}
+          <MenuItem divider />
+          <MenuItem onSelect={onSignOutClick}>Sign out</MenuItem>
+        </NavDropdown>
+      )
+    }
+
+    return (
```
landonreed, line 117: Could you give some basic comments next to the above conditional and this return statement about the presumed login states and what we're returning? I.e.:

```
User is logged in, return dropdown with links to account information and log out.
```

and 

```
User not logged in, showing sign in link.
```

lib/util/auth.js
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * Handle logging out with Auth0 and return to the provided path.
+ */
+export function logout (path = '') {
+  const returnTo = encodeURIComponent(`${getAuthRedirectUri()}${path}`)
+  window.location.href = `https://${process.env.AUTH0_DOMAIN}/v2/logout?returnTo=${returnTo}&client_id=${process.env.AUTH0_CLIENT_ID}`
+}
+
+export const getAuthRedirectUri = () => {
```
landonreed, line 26: Is this even needed? I remember running into some issues with admin dashboard and coming up with this function, but I'm not sure if it is relevant here?

lib/util/auth.js
landonreed: Yes, I agree. Checking that something is not `localhost` is just asking for bugs down the road.

lib/components/user/nav-login-button-auth0.js
```diff
@@ -0,0 +1,23 @@
...
+ */
+export default function NavLoginButtonAuth0 (props) {
+  const { isAuthenticated, login, logout, user } = useAuth()
+
+  const handleLogin = () => login({ appState: {returnTo: window.location.href} })
+  const handleLogout = () => logout({ returnTo: window.location.href })
+
+  return (
+    <NavLoginButton
+      {...props}
```
landonreed, line 17: Let's not spread these props.

lib/components/user/nav-login-button.js
binh-dam-ibigroup: It's PropTypes documentation, a habit from OTP-UI.

lib/util/auth.js
binh-dam-ibigroup: I am not sure, I took those bits from otp-admin-ui!

lib/components/user/nav-login-button-auth0.js
binh-dam-ibigroup: We still need to forward `id`, `links`, `profile`, `className` and `style` to that component.

lib/util/auth.js
binh-dam-ibigroup: Ok, so `getAuthRedirectUri()` is needed for Auth0. The `logout()` method above it is not, I'm removing it. Thanks for pointing this out.

lib/components/user/nav-login-button-auth0.js
landonreed: Yea, I understand. I'm just commenting on the syntax. Instead of using the spread operator (`...props`), I'm asking if you can specify them individually in NavLoginButton's props.

lib/util/auth.js
landonreed: That's fine, but I don't think`getAuthRedirectUri` is actually needed here either. It's only used one time and I'm not sure if the `return 'http://localhost:9966'` is reachable. Could you try removing and replace the one invocation of this with `${window.location.protocol}//${window.location.host}`?

lib/components/user/nav-login-button.js
evansiroky: I also prefer static-typing, but think that until we make a commitment to doing that everywhere in otp-rr moving forward, allowing propTypes is better than nothing.

lib/components/user/nav-login-button.js
landonreed: If you remove these defaultProps, nothing changes though, right? It's essentially (not exactly) the same as setting defaultProps to `undefined`.

lib/components/user/nav-login-button-auth0.js
```diff
@@ -6,18 +6,28 @@ import NavLoginButton from './nav-login-button'
...
+const NavLoginButtonAuth0 = ({
+  className,
+  id,
+  links,
+  style
+}) => {
   const { isAuthenticated, login, logout, user } = useAuth()
 
   const handleLogin = () => login({ appState: {returnTo: window.location.href} })
   const handleLogout = () => logout({ returnTo: window.location.href })
```
landonreed, line 18: I really don't think we want to run the risk of upsetting Auth0 by potentially returning the user to a bad URL (this could happen if we use href and somehow down the road are not using hash routing or if the href has a path set for some other reason).

PR #150, opened 2020-04-03
OTP-UI Icons

binh-dam-ibigroup:
This PR removes icons from lib/components/icons and uses those from the OTP-UI icons package instead. The PR also removes unused narrative profile components.

EDITS: Please see the Remarks section below for edits and testing notes.

### Remarks

1. [Outdated as of 9219719bac7f92f1d7ead440e50f9e218947b3d5] Full support of custom icons in OTP-ui is pending opentripplanner/otp-ui#119, however the refactor is mostly compatible with the currently released OTP-ui library.
2. [Outdated as of 9219719bac7f92f1d7ead440e50f9e218947b3d5] In  https://github.com/opentripplanner/otp-react-redux/blob/02fef2765dd684acadfbdcf0d1676853a0cdd4db/example.js#L30, there is a `MyModeIcon` variable passed to various `ModeIcon` props, however `ModeIcon` props will take effect pending a change in `SettingsSelectorPanel` in opentripplanner/otp-ui#119.
3. [Outdated as of 6c10ea651c951d0497289ac8111d7adf867dedc1] As a result, the default icon for "Take Transit" in OTP-RR is changed, and there are subtle visual changes to the itinerary narratives as shown in the screenshots below.
(Note that the bus and walk icon sizes are pending OTP release per https://github.com/opentripplanner/otp-ui/pull/105 to address [this comment](https://github.com/opentripplanner/otp-react-redux/pull/142#pullrequestreview-383985093)) Beware, visual changes taking place! (see screenshots below)
 
image

image

screenshot if using LineItinerary
image

4. As of 6c10ea651c951d0497289ac8111d7adf867dedc1, visual styles for trip form elements (not the map or nav bar) match newplanner.trimet.org, and the old styling is abandoned.

PR #157, opened 2020-04-24
OTP-UI core-utils

binh-dam-ibigroup:
This PR migrates code to use the OTP-UI core-utils package.
It also removes legacy user profile code.
(The PR does not include OTP-ui's icons.)

In this PR, we generalize this notation for most OTP-UI packages:
```js
import coreUtils from '@opentripplanner/core-utils'
```
which removes the path to the lib folder in each OTP-UI package.


__tests__/util/itinerary.js
```diff
@@ -1,4 +1,6 @@
-import {isTransit, routeComparator} from '../../lib/util/itinerary'
+import coreUtils from '@opentripplanner/core-utils'
+
+import { routeComparator } from '../../lib/util/itinerary'
```
landonreed, line 3: Didn't we add route comparator to otp-ui? 🤔

binh-dam-ibigroup: We did. It is pending release.

__tests__/util/itinerary.js
```diff
@@ -1,4 +1,6 @@
-import {isTransit, routeComparator} from '../../lib/util/itinerary'
+import coreUtils from '@opentripplanner/core-utils'
```
landonreed, line 1: Re: import style, it'd be good to have an answer on whether the style influences build size. If so, let's use the more precise import style (that refs the `lib/` directory).

lib/actions/form.js
```diff
@@ -21,6 +13,16 @@ import {
...
 
+const {
+  getDefaultQuery,
+  getTripOptionsFromQuery,
+  getUrlParams,
+  planParamsToQuery
+} = coreUtils.query
+const { getItem, randId } = coreUtils.storage
+const { OTP_API_TIME_FORMAT } = coreUtils.time
+const { isMobile } = coreUtils.ui
```
landonreed, line 24: This whole block seems a lot better suited to destructuring in the import statement.

lib/components/form/default-search-form.js
```diff
@@ -1,19 +1,25 @@
...
 import TabbedFormPanel from './tabbed-form-panel'
+import SwitchButton from './switch-button'
 import defaultIcons from '../icons'
 
+// Use the icons props to override icons for certain modes in the mode selector panel.
+// If no icon is provided for a specific mode, the default OTP-ui icon will be used.
+// const customIcons = {
+//  TRANSIT: <TriMetModeIcon label='rail' />
+// }
+
```
landonreed, line 14: Isn't this possible with `config.js` in trimet-mod-otp as well? Maybe add a TODO statement here about consolidating that repo/code into otp-rrr

binh-dam-ibigroup: Correct, this is a TODO item that is related to #150.

lib/components/narrative/default/access-leg.js
```diff
@@ -1,12 +1,10 @@
-import React, {Component} from 'react'
+import coreUtils from '@opentripplanner/core-utils'
 import PropTypes from 'prop-types'
+import React, {Component} from 'react'
 
 import Icon from '../icon'
-
 import LegDiagramPreview from '../leg-diagram-preview'
 import { distanceString } from '../../../util/distance'
```
landonreed, line 7: Let's use `@opentripplanner/humanizeDistance`. I think it's ok if it's in this same PR (unless you have it in some other PR already). Not 100% sure why that's a separate package... I think that just leads to confusion/oversights like this ^^.

lib/components/narrative/narrative-routing-results.js
```diff
@@ -32,12 +31,8 @@ class NarrativeRoutingResults extends Component {
     if (mainPanelContent) return null
 
     return (
-      routingType === 'ITINERARY'
-        ? <TabbedItineraries itineraryClass={itineraryClass} itineraryFooter={itineraryFooter} itineraries={itineraries} customIcons={customIcons} />
-        : <NarrativeProfileOptions
-          itineraryClass={itineraryClass}
-          customIcons={customIcons}
-        />
+      routingType === 'ITINERARY' &&
```
landonreed, line 34: I'm not sure this check for routingType is necessary. Although... maybe we should keep it in place because we may have a new routing type (`MATRIX`) in short order...

lib/util/index.js
```diff
@@ -1,22 +1,12 @@
 import * as distance from './distance'
```
landonreed, line 1: Let's say farewell to distance.

lib/util/index.js
```diff
@@ -1,22 +1,12 @@
...
 import * as itinerary from './itinerary'
```
landonreed, line 2: Does itinerary need to exist anymore? What is unique in that file from otp-ui?

lib/util/index.js
```diff
@@ -1,22 +1,12 @@
...
-import * as map from './map'
-import * as profile from './profile'
-import * as query from './query'
-import * as reverse from './reverse'
 import * as state from './state'
-import * as time from './time'
 import * as ui from './ui'
```
landonreed, line 4: I think the stuff currently in `ui` can move to `state`.

lib/util/index.js
```diff
@@ -1,22 +1,12 @@
...
 const OtpUtils = {
   distance,
   itinerary,
-  map,
-  profile,
-  query,
-  reverse,
   state,
-  time,
   ui
```
landonreed, line 10: Maybe we should export coreUtils in addition to whatever utils still live in otp-rr. I.e.,

```
const OtpUtils = {
  ...coreUtils,
  state
}
```

lib/util/index.js
binh-dam-ibigroup: distance is handled by #151.

lib/util/index.js
binh-dam-ibigroup: There are some functions in itinerary that didn't originally get transferred over to OTP-ui.

lib/util/ui.js
```diff
@@ -1,59 +1,11 @@
+import coreUtils from '@opentripplanner/core-utils'
+
 import { MainPanelContent } from '../actions/ui'
-import { summarizeQuery } from './query'
 import { getActiveSearch } from './state'
 
 // Set default title to the original document title (on load) set in index.html
 const DEFAULT_TITLE = document.title
```
landonreed, line 7: As mentioned elsewhere, I think this stuff can be consolidated into `state`. I think it's a more appropriate location and simplifies things.

lib/util/storage.js
```diff
@@ -1,42 +0,0 @@
-// Prefix to use with local storage keys.
```
landonreed, line 1: Where did this file go?

binh-dam-ibigroup: To OTP-ui...

lib/components/form/default-search-form.js
binh-dam-ibigroup: I'm removing this comment for this PR.

lib/util/index.js
binh-dam-ibigroup: I'm not sure... Should people import from OTP-ui directly? @evansiroky what do you think?

lib/actions/form.js
binh-dam-ibigroup: Inlining the ones not from query for now.

lib/util/index.js
landonreed: Yea, you're probably right. We can keep as is for now.

landonreed:
@binh-dam-ibigroup, could you fix the conflicts. Then we can go ahead with a merge.

binh-dam-ibigroup:
Conflicts resolved.

PR #163, opened 2020-05-06
Create/Edit Account

binh-dam-ibigroup:
## New account creation and edition functionality
- Panes regarding different steps/aspects of user accounts: notifications, consent, adding favorite locations.
- Basic account creation workflow that transitions through the panes.
- Basic account editing layout that displays all the panes.
- Basic 'Signed-In' screens when transitioning between Auth0 login and the account page.
- Error message popup when errors are encountered.

Screnshots
See the software live at https://fdot-otp.ibi-transit.com/
image

## Dependencies
- #156
- https://github.com/ibi-group/otp-middleware/pull/15

## Behind the scenes

When signing in:
1.  User starts with an itinerary search at `/<search_params>` (e.g. `localhost:9966/#/<search_params>`) and signs in/signs up via Auth0.
2.  Browser is redirected to whitelist URL `/signedin` by Auth0.
(no action taken at this stage – our redirect callback in `auth.js` is about to get called by Auth0.)
4.  Browser is redirected to `/account/welcome/<search_params>` by our redirect callback
(`<search_params>` was “saved” prior to login)
(don’t show previous URL in navigation history)
5.  Try to fetch user data to determine whether this is a new user:
*  If new user, then show the account setup screens (don’t change URL).
  When the new user is done with that screen, navigate to `/<search_params>`.
*  If existing user, then just redirect to `/<search_params` without showing `/account/welcome` in the navigation history.

When a logged-in user clicks My Account from the profile drop down:
1. Browser is directed to `/account/<search_params>`
2. User makes changes.
3. Browser is directed to `/<search_params`. The browsing history is preserved.


lib/components/user/auth0-token-wrapper.js
```diff
@@ -0,0 +1,65 @@
...
+
+  _fetchToken = () => {
+    /**
+     * An Auth0 access token, required to call the middleware API,
+     * is typically not available right away when this function is called.
+     * So, we set up a timer to poll for an accessToken from the props every second.
+     * Once the token is available, update state so token can be forwarded to children.
+     * While we wait for the token, display a wait screen.
+     */
+    this._tokenTimer = setInterval(async () => {
```
evansiroky, line 32: This doesn't need to be an async function. There is no await inside it.

lib/components/user/auth0-token-wrapper.js
```diff
@@ -0,0 +1,65 @@
...
+  render () {
+    const { auth, children } = this.props
+    const token = auth.accessToken || this.state.accessToken
+
+    if (!token) {
+      // Display a wait screen while waiting for token.
+      return <AwaitingScreen />
+    }
+
+    // console.log(token)
```
evansiroky, line 55: Please remove

lib/components/user/nav-login-button.js
```diff
@@ -99,7 +102,12 @@ export default class NavLoginButton extends Component {
           <MenuItem header>{displayedName}</MenuItem>
 
           {links && links.map((item, index) => (
-            <MenuItem href={item.url} key={index} target={item.target}>{item.text}</MenuItem>
+            <MenuItem
+              onSelect={() => this.props.routeTo(item.url)}
```
evansiroky, line 106: Make this onSelect a class method so that a new function isn't created every render.

lib/components/user/user-account-screen.js
```diff
@@ -0,0 +1,560 @@
...
+    )
+
+    const panesForNewAccount = {
+      terms: { title: 'Create a new account', pane: termsOfUsePane, nextId: 'notifications', disableNext: !hasConsentedToTerms },
+      notifications: { title: 'Notification preferences', pane: notificationPrefsPane, nextId: notificationChannel === 'sms' ? 'verifyPhone' : 'places', prevId: 'terms' },
+      verifyPhone: { title: 'Verify your phone', pane: phoneVerificationPane, nextId: 'places', prevId: 'notifications', disableNext: true },
+      places: { title: 'Add your locations', pane: placesPane, nextId: 'finish', prevId: 'notifications' },
+      finish: { title: 'Account setup complete!', pane: finishPane, prevId: 'places' }
+    }
+    const realActivePane = panesForNewAccount[activePaneId]
```
evansiroky, line 469: Just make this `activePane`.

lib/components/user/user-account-screen.js
```diff
@@ -0,0 +1,560 @@
...
+    const panesForExistingAccount = [
+      {title: 'Terms', pane: termsOfUsePane},
+      {title: 'Notifications', pane: notificationPrefsPane},
+      {title: 'My locations', pane: placesPane}
+    ]
+
+    return (
+      <form>
+        {isNewAccount
+          ? (
```
evansiroky, line 480: To improve readability, I recommend breaking up each suboption into its own render function. Ex:

```js
<form>
  {isNewAccount ? this._renderNewAccount() : this._renderExistingAccount()}
</form>
```

lib/components/user/user-account-screen.js
```diff
@@ -0,0 +1,560 @@
...
+
+    const {
+      hasConsentedToTerms,
+      notificationChannel,
+      phoneNumber,
+      savedLocations,
+      storeTripHistory
+    } = userData
+
+    const termsOfUsePane = (
```
evansiroky, line 331: Each of these panes should probably be its own component in a separate file.

lib/util/auth.js
```diff
@@ -77,13 +70,27 @@ export function getAuth0Callbacks (reduxStore) {
...
-      if (appState && appState.returnTo) {
+      if (appState && appState.urlHash) {
         /**
-         * Immediately after successful Auth0 login, the user will be redirected to the MOD UI URL with no search.
-         * We want to redirect to the URL just before the Auth0 operation (stored in appState.returnTo),
+         * Immediately after successful login, Auth0 will redirect to a URL (in whitelist) with no search parameters.
+         * If the user has set up their MOD UI account, we want to redirect to
+         * the state just before the Auth0 operation (stored in appState.urlHash),
          * and prevent the URL with no search from appearing in the user's browsing history.
          */
```
landonreed, line 79: For multi-line comments that are not jsdoc, I prefer not to use this style. Instead, please use the double slash `//`

lib/util/auth.js
```diff
@@ -77,13 +70,27 @@ export function getAuth0Callbacks (reduxStore) {
...
+      } else if (appState && appState.returnTo) {
+        // TODO: Handle other after-login situations.
+        /* Careful!
+         * - When redirecting from a login-protected page while logged out,
+         *     then returnTo is set by Auth0 to this object format:
+         *     {
+         *       pathname: "/"
+         *       query: { ... }
+         *     }
+         */
```
landonreed, line 93: double slash (`//`) comments instead, please.

lib/components/user/user-account-screen.js
```diff
@@ -0,0 +1,595 @@
...
+    }
+
+    // Build an 'effective' locations list for display.
+    const effectiveLocations = [
+      homeLocation,
+      workLocation,
+      ...savedLocations.filter(loc => loc.type !== 'home' && loc.type !== 'work')
+    ]
+
+    const placesPane = (
```
evansiroky, line 435: Needs a refactor to allow geocoding of locations. Should be in a separate component/file like the other panes.

lib/components/user/user-account-screen.js
```diff
@@ -0,0 +1,595 @@
...
+                      style={{float: 'right'}}
+                    >
+                      Save Preferences
+                    </Button>
+                  </nav>
+                </FormGroup>
+              </div>
+            )
+          }
+        </form>
```
evansiroky, line 576: To improve readability, I recommend breaking up each suboption into its own render function. Ex:

```js
<form>
  {isNewAccount ? this._renderNewAccount() : this._renderExistingAccount()}
</form>
```

lib/util/middleware.js
```diff
@@ -0,0 +1,64 @@
...
+      'x-api-key': apiKey
+    },
+    ...options
+  })
+  if ((res.status && res.status >= 400) || (res.code && res.code >= 400)) {
+    const result = await res.json()
+    let message = `Error ${method}-ing user: ${result.message}`
+    if (result.detail) message += `  (${result.detail})`
+
+    console.log(message)
```
evansiroky, line 21: Please remove console.log

lib/util/middleware.js
```diff
@@ -0,0 +1,64 @@
...
+      message
+    }
+  }
+  return {
+    status: 'success',
+    data: await res.json()
+  }
+}
+
+export async function fetchUser (middlewareConfig, idField, id, token) {
```
evansiroky, line 33: Needs a comment that this method will fetch a user by their Auth0Id and not the Mongo Id.

evansiroky: Might want to consider renaming to `fetchUserByAuth0Id`.

lib/components/user/user-account-screen.js
binh-dam-ibigroup: Component has been extracted. Per internal IBI discussion, we will handle geocoding in a separate PR.

lib/util/middleware.js
binh-dam-ibigroup: The methods in here are agnostic to Auth0 (even though its caller specifies `idField` to be the Auth0 id).

lib/components/user/account-setup-finish-pane.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+
+const AccountSetupFinishPane = () => (
```
evansiroky, line 3: This doesn't even need to be a function if the same thing is always returned.

lib/components/user/after-signin-screen.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react'
+import FontAwesome from 'react-fontawesome'
+
+/**
+ * Screen that is flashed just after user sign in.
+ * TODO: Improve this screen.
+ */
+const AfterSignIn = () => (
```
evansiroky, line 8: Convert from function to constant.

lib/components/user/awaiting-screen.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+
+/**
+ * Screen that is flashed while retrieving user data.
+ * TODO: Improve this screen.
+ */
+const AwaitingScreen = () => <div>Processing...</div>
```
evansiroky, line 7: Convert react from function to constant.

lib/components/user/nav-login-button.js
```diff
@@ -35,6 +38,10 @@ export default class NavLoginButton extends Component {
     profile: null
   }
 
+  _handleLinkClick = item => {
```
evansiroky, line 41: Ah, didn't realize this was a handler for an item in a loop. This refactor still creates a function for every item on each render. The way to avoid this for items in a loop is to memorize this handler like this: https://github.com/ibi-group/datatools-ui/blob/48ec0da2e303a9b20a81f3fb757a61d3eef9fa32/lib/manager/components/CreateFeedSource.js#L86

lib/components/user/phone-verification-pane.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import { Alert, FormControl, FormGroup } from 'react-bootstrap'
+
+/**
+ * User phone verification pane.
+ * TODO: to be completed.
+ */
+const PhoneVerificationPane = () => (
```
evansiroky, line 8: Convert react from function to constant.

lib/components/user/user-account-screen.js
```diff
@@ -0,0 +1,437 @@
...
+      activePaneId,
+      userData
+    } = this.state
+
+    const {
+      hasConsentedToTerms,
+      notificationChannel = 'email'
+    } = userData
+
+    const panes = this._makePanes()
```
evansiroky, line 265: This will go through the process of generating the react for every pane of every render. See comments in how to change `_makePanes()`.

lib/components/user/user-account-screen.js
```diff
@@ -0,0 +1,437 @@
...
+  _handleToPrevPane = panes => {
+    return () => {
+      const { activePaneId } = this.state
+      this.setState({
+        activePaneId: panes[activePaneId].prevId
+      })
+    }
+  }
+
+  _makePanes = () => {
```
evansiroky, line 200: This generates way too much react that isn't needed on every render. It should be refactored to be something like: 

```js
renderActivePane = () => {
  switch(this.state.activePaneId) {
    case 'terms': return this.renderTerms()
    case 'notifications': return this.renderNotifications()
    ...
  }
}
```

lib/components/user/user-account-screen.js
```diff
@@ -0,0 +1,437 @@
...
+  _handleExitAndSave = async () => {
+    await this._updateUserPrefs()
+    this._handleExit()
+  }
+
+  _handleLocationsChange = newLocations => {
+    this._updateUserState({ savedLocations: newLocations })
+  }
+
+  _handleNotificationChannelChange = e => {
```
evansiroky, line 164: All these special handlers that only call updateUserState could be contained within the individual components for each pane.

lib/components/user/account-setup-finish-pane.js
binh-dam-ibigroup: I think they should remain components. @landonreed?

lib/components/user/nav-login-button.js
binh-dam-ibigroup: Thanks for the suggestion!

lib/components/user/user-account-screen.js
binh-dam-ibigroup: Good idea

lib/components/user/notification-prefs-pane.js
```diff
@@ -0,0 +1,102 @@
...
+  },
+  {
+    type: 'none',
+    text: 'Don\'t notify me'
+  }
+]
+
+/**
+ * User notification preferences pane.
+ */
```
landonreed, line 22: ~Can you add a TODO here for handling the SMS verification stuff? We should do that in another PR.~ OK, I guess there is a TODO for that already.

Also, we need a TODO about field validation.

lib/components/user/terms-of-use-pane.js
```diff
@@ -0,0 +1,67 @@
...
+      <div>
+        <ControlLabel>You must agree to the terms of service to continue.</ControlLabel>
+
+        <FormGroup>
+          <Checkbox
+            checked={checkTerms}
+            disabled={disableCheckTerms}
+            onChange={disableCheckTerms ? null : this._handleCheckTermsChange}
+          >
+            I have read and consent to the <a href='/#/terms-of-service'>Terms of Service</a> for using the Trip Planner.
```
landonreed, line 49: please add TODO about adding this link.

lib/components/user/terms-of-use-pane.js
```diff
@@ -0,0 +1,67 @@
...
+          </Checkbox>
+        </FormGroup>
+
+        <FormGroup>
+          <Checkbox
+            checked={checkHistory}
+            onChange={this._handleCheckHistoryChange}
+          >
+            Optional: I consent to the Trip Planner storing my historical planned trips in order to
+            improve transit services in my area. <a href='/#/terms-of-storage'>More info...</a>
```
landonreed, line 59: please add TODO about adding this link.

lib/components/user/user-account-screen.js
```diff
@@ -0,0 +1,422 @@
...
+    recentLocations: [],
+    savedLocations: [],
+    storeTripHistory: false // User must opt in.
+  }
+}
+
+/**
+ * This screen handles creating/updating OTP user account settings.
+ */
+class UserAccountScreen extends Component {
```
landonreed, line 37: If possible, please split into two components as mentioned above. This file is very complex/long. Will review further after change is made.

lib/components/user/verify-email-screen.js
```diff
@@ -0,0 +1,34 @@
...
+const _handleClick = () => window.location.reload()
+
+/**
+ * This component contains the prompt for the user to verify their email address.
+ * It also contains a button that lets the user finish account setup.
+ *
+ * (One way to make sure the parent page fetches the latest email verification status
+ * is to simply reload the page.)
+ */
+const verifyEmailScreen = (
```
landonreed, line 13: Did you ever look into the possibility of Auth0 redirecting the user to the OTP app after they click the email verification link? Also, this needs styling work (probably makes sense to just give all of these full screen views a classname that applies styling or otherwise use Bootstrap containers or something).

lib/util/auth.js
```diff
@@ -77,13 +70,24 @@ export function getAuth0Callbacks (reduxStore) {
...
-        window.location.replace(appState.returnTo)
+      if (appState && appState.urlHash) {
+        // Immediately after successful login, Auth0 will redirect to a URL (in whitelist) with no search parameters.
+        // If the user has set up their MOD UI account, we want to redirect to
+        // the state just before the Auth0 operation (stored in appState.urlHash),
+        // and prevent the URL with no search from appearing in the user's browsing history.
+        const newUserPath = appState.urlHash.replace('#/', '#/account/welcome')
+        const newUserUrl = `${getAuthRedirectUri()}${newUserPath}`
+
+        window.location.replace(newUserUrl)
```
landonreed, line 81: I'm guessing this needs to change based on changes in add user sign in PR.

lib/util/auth.js
```diff
@@ -77,13 +70,24 @@ export function getAuth0Callbacks (reduxStore) {
...
+      } else if (appState && appState.returnTo) {
+        // TODO: Handle other after-login situations.
+        // Careful!
+        // - When redirecting from a login-protected page while logged out,
+        //     then returnTo is set by Auth0 to this object format:
+        //     {
+        //       pathname: "/"
+        //       query: { ... }
+        //     }
```
landonreed, line 90: What other scenarios do we want to handle?

lib/util/middleware.js
landonreed: @binh-dam-ibigroup, it's very unlikely that this code will ever be refactored to not use Auth0. I appreciate your efforts to isolate Auth0 and make certain auth things generic, but we need the code in its present state to be readable when we revisit and actually reflect the components that its built with. In my view, part of that would be to make naming changes like @evansiroky suggested above.

lib/util/middleware.js
```diff
@@ -0,0 +1,64 @@
...
+  const { apiBaseUrl, apiKey } = middlewareConfig
+  const requestUrl = `${apiBaseUrl}${API_USER_PATH}/find/${idField}/${id}`
+
+  return secureFetch(requestUrl, token, apiKey)
+}
+
+export async function addUser (middlewareConfig, token, data) {
```
landonreed, line 40: It really seems like we would want these to be actions that update the redux store when they receive the response.

binh-dam-ibigroup: Yes... coming in the next PR. The next PR will make the current user a redux state!

lib/components/user/awaiting-screen.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+
+/**
+ * Screen that is flashed while retrieving user data.
+ * TODO: Improve this screen.
+ */
+const awaitingScreen = <div>Processing...</div>
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Need a separate discussion on that.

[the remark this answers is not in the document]
binh-dam-ibigroup: The `<AfterSignIn>` page is loaded while this block of code is executed, before the account (or the next destination) is extracted and passed to the router. This step is unavoidable. There is an opportunity to improve the appearance for the wait screens, and I will maintain that should be a separate discussion.

lib/components/user/nav-login-button.js
```diff
@@ -35,6 +39,10 @@ export default class NavLoginButton extends Component {
     profile: null
   }
 
+  _handleLinkClick = memoize(
+    item => () => this.props.routeTo(item.url)
+  )
+
```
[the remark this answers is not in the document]
binh-dam-ibigroup: @landonreed @evansiroky please agree on the approach.

lib/components/user/verify-email-screen.js
binh-dam-ibigroup: Redirecting from Auth0's link should be handled separately.

lib/components/user/after-signin-screen.js
```diff
@@ -0,0 +1,20 @@
...
+      <br />
+      <FontAwesome
+        name='hourglass-half'
+        size='4x'
+      />
+    </h1>
+  </div>
+)
+
+export default afterSignInScreen
```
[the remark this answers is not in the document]
binh-dam-ibigroup: See comment above.

lib/components/user/favorite-locations-pane.js
```diff
@@ -0,0 +1,175 @@
...
+    color: #fff;
+  }
+  &:focus {
+    background-color: unset;
+    color: unset;
+    ::placeholder {
+      color: unset;
+    }
+  }
+`
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Inline CSS if short and sweet?

lib/components/user/auth0-token-wrapper.js
```diff
@@ -0,0 +1,68 @@
+import PropTypes from 'prop-types'
+import { Component } from 'react'
+
+import { renderChildrenWithProps } from '../../util/ui'
+
+import awaitingScreen from './awaiting-screen'
+
+/**
+ * This screen handles creating/updating OTP user accoun settings.
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Good one. Unfortunately Auth0Wrapper needs to remain a function component per `useAuth` hook rule.

lib/components/user/nav-login-button.js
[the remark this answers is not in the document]
landonreed: Pretty much everywhere in our code we have made a habit of not inlining handlers. And typically this has not been via memoizing functions. I really prefer that we just stick to this habit for now. I still say we just create a very simple component (can be functional) that just takes an item as a prop and handles the click.

lib/components/user/user-account-screen.js
```diff
@@ -0,0 +1,243 @@
...
+      // Display a hint while fetching user data (from componentDidMount).
+      return <AwaitingScreen />
+    }
+
+    return (
+      <div className='otp'>
+        {/* TODO: Do mobile view. */}
+        <AppNav />
+        <form className='container'>
+          {(isNewAccount
```
evansiroky, line 203: This is a lot of nesting. I'd prefer for it to be simplified by having something like:

```js
let formContents
if (!isNewAccount) {
  formContents = (
    <ExistingAccountDisplay
      onCancel={this._handleExit}
      onComplete={this._handleExitAndSave}
      panes={this._panes}
    />
  )
} ...

return (
...
  <form className='container'>
    {formContents}
  </form>
...
)
```

lib/components/user/nav-login-button.js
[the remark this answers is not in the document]
binh-dam-ibigroup: @landonreed This will probably not work if I have to add a `<Nav>` element inside the mobile view: https://github.com/opentripplanner/otp-react-redux/blob/70a9d96b31722b836d3ec099609d100c245a1439/lib/components/mobile/navigation-bar.js#L58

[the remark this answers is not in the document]
binh-dam-ibigroup: I found a better way by adding an onClick attribute to each link object.

binh-dam-ibigroup:
> Approved, but on the condition that `MenuItem` is refactored into a simple component that takes the URL prop to fix the inlining of handler (per discussion with @binh-dam-ibigroup).

I just pushed the corresponding change.

PR #164, opened 2020-05-12
refactor(icons): move example custom icons into own file

landonreed:
Move things around a little to clean up the example/make it easier to override classic icons.

### DIFF
--- a/example.js
+++ b/example.js
@@ -27,7 +27,7 @@ import otpConfig from './config.yml'
 
 // Set useCustomIcons to true to override classic icons with the exports from
 // custom-icons.js
-const useCustomIcons = true
+const useCustomIcons = false
 
 // Define icon sets for modes.
 let MyLegIcon = ClassicLegIcon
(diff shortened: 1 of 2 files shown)


example.js
```diff
@@ -33,13 +22,22 @@ import {
   AppMenu,
   createOtpReducer
 } from './lib'
-
 // load the OTP configuration
 import otpConfig from './config.yml'
 
+// Set useCustomIcons to true to override classic icons with the exports from
+// custom-icons.js
+const useCustomIcons = true
```
binh-dam-ibigroup, line 30: This should be false by default.

example.js
```diff
@@ -50,44 +48,6 @@ const MyModeIcon = ClassicModeIcon
 // const MyLegIcon = () => <Ferry />
 // const MyModeIcon = () => <AerialTram />
```
binh-dam-ibigroup, line 49: I am getting rid of this example in #150 in favor of the one using CustomIcons.

PR #169, opened 2020-06-01
Get user by auth0 token

binh-dam-ibigroup:
This PR accommodates the change introduced by https://github.com/ibi-group/otp-middleware/pull/22.

### DIFF
--- a/lib/util/middleware.js
+++ b/lib/util/middleware.js
@@ -1,6 +1,7 @@
 if (typeof (fetch) === 'undefined') require('isomorphic-fetch')
 
 const API_USER_PATH = '/api/secure/user'
+const API_AUTH0USER_PATH = '/api/secure/auth0user'
 
 export async function secureFetch (url, accessToken, apiKey, method = 'get', options = {}) {
   const res = await fetch(url, {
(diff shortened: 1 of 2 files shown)


lib/util/middleware.js
```diff
@@ -30,9 +31,9 @@ export async function secureFetch (url, accessToken, apiKey, method = 'get', opt
   }
 }
 
-export async function fetchUser (middlewareConfig, token, id) {
+export async function fetchUser (middlewareConfig, token) {
   const { apiBaseUrl, apiKey } = middlewareConfig
-  const requestUrl = `${apiBaseUrl}${API_USER_PATH}/find/auth0UserId/${id}`
+  const requestUrl = `${apiBaseUrl}${API_AUTH0USER_PATH}`
```
landonreed, line 36: If this is the only use of this constant, it probably isn't worth putting in a variable that's at the top of the file.

binh-dam-ibigroup:
Please wait until https://github.com/ibi-group/otp-middleware/pull/22 and https://github.com/ibi-group/otp-middleware/pull/15 are merged into dev before merging this PR.

binh-dam-ibigroup:
Adding BLOCKED label until https://github.com/ibi-group/otp-middleware/pull/15 is merged.

lib/util/middleware.js
landonreed: @binh-dam-ibigroup, the point of having the `API_USER_PATH` constant is so that we can reuse it in functions in this file. So, there's really no need to have all of these interim variables. I think the clearest/most maintainable way to write this is just:

```javascript
return secureFetch(`${apiBaseUrl}${API_USER_PATH}/fromtoken`, token, apiKey)
```

or if that string is too long, just put it in `requestUrl`

binh-dam-ibigroup: Good idea @landonreed, it is now:
```
  const requestUrl = `${apiBaseUrl}${API_USER_PATH}/fromtoken`
  return secureFetch(requestUrl, ...
```

PR #167, opened 2020-05-26
Log trip requests

binh-dam-ibigroup:
This PR adds the following HTTP headers to trip requests forwarded to {OTP/the Middleware}:
- An `Authorization` header with the Auth0 access token, for logged-in users,
- An `x-api-key` header with the API key to OTP/middleware API interface.

To enable that, the PR introduces a new reducer for middleware users, new application states: `state.user.loggedInUser`, `state.user.accessToken` (fix #173 ), and related refactorings.

### Other changes
- BREAKING: The Auth0 `audience` parameter has been removed from `config.yml`.
- `Auth0Wrapper` and related components are removed.
- A new `withLoggedInUserSupport` higher-order component is introduced that ensures wrapped components such as the user account page have access to `state.user.loggedInUser`.
- Routing management after sign-in has been moved away from `UserAccountScreen` and consolidated in the `AfterSignInScreen`. The URL prior to login is saved as a state.

### Dependencies

- Minimum middleware: https://github.com/ibi-group/otp-middleware/pull/15
- For storing trip requests: https://github.com/ibi-group/otp-middleware/pull/17

PR #175, opened 2020-06-15
improvement(example): Move Auth0Provider plumbing out of example.js a…

binh-dam-ibigroup:
This pull request addresses this comment: https://github.com/opentripplanner/otp-react-redux/pull/167#discussion_r439492550 and moves the plumbing to set up `Auth0Provider` to the `ConnectedRouter` component.

### DIFF
--- a/example.js
+++ b/example.js
@@ -16,8 +16,8 @@ import { Grid, Row, Col } from 'react-bootstrap'
 
 // import OTP-RR components
 import {
-  AppNav,
   DefaultMainPanel,
+  DesktopNav,
   Map,
   MobileMain,
   ResponsiveWebapp,
@@ -81,7 +81,7 @@ class OtpRRExample extends Component {
     /** desktop view **/
     const desktopView = (
       <div className='otp'>
-        <AppNav otpConfig={otpConfig} title='OpenTripPlanner' />
+        <DesktopNav />
         <Grid>
           <Row className='main-row'>
             <Col sm={6} md={4} className='sidebar'>
--- a/example.js
+++ b/example.js
@@ -11,21 +11,15 @@ import { Provider } from 'react-redux'
 import thunk from 'redux-thunk'
 import createLogger from 'redux-logger'
 
-// Auth0
-import { Auth0Provider } from 'use-auth0-hooks'
-import { accountLinks, getAuth0Callbacks, getAuth0Config } from './lib/util/auth'
-import { AUTH0_AUDIENCE, AUTH0_SCOPE, URL_ROOT } from './lib/util/constants'
-
 // import Bootstrap Grid components for layout
-import { Nav, Navbar, Grid, Row, Col } from 'react-bootstrap'
+import { Grid, Row, Col } from 'react-bootstrap'
 
 // import OTP-RR components
 import {
-  AppMenu,
+  AppNav,
   DefaultMainPanel,
   Map,
   MobileMain,
-  NavLoginButtonAuth0,
   ResponsiveWebapp,
   createOtpReducer,
   createUserReducer
@@ -81,37 +75,13 @@ const store = createStore(
   compose(applyMiddleware(...middleware))
 )
 
-// Auth0 config and callbacks.
-const auth0Config = getAuth0Config(otpConfig.persistence)
-const auth0Callbacks = getAuth0Callbacks(store)
-
 // define a simple responsive UI using Bootstrap and OTP-RR
 class OtpRRExample extends Component {
   render () {
     /** desktop view **/
     const desktopView = (
       <div className='otp'>
-        <Navbar fluid inverse>
-          <Navbar.Header>
-            <Navbar.Brand>
-              <div style={{ float: 'left', color: 'white', fontSize: 28 }}>
-                <AppMenu />
-              </div>
-              <div className='navbar-title' style={{ marginLeft: 50 }}>OpenTripPlanner</div>
-            </Navbar.Brand>
-          </Navbar.Header>
-
-          {auth0Config && (
-            <Navbar.Collapse>
-              <Nav pullRight>
-                <NavLoginButtonAuth0
-                  id='login-control'
-                  links={accountLinks}
-                />
-              </Nav>
-            </Navbar.Collapse>
-          )}
-        </Navbar>
+        <AppNav otpConfig={otpConfig} title='OpenTripPlanner' />
         <Grid>
           <Row className='main-row'>
             <Col sm={6} md={4} className='sidebar'>
@@ -146,34 +116,20 @@ class OtpRRExample extends Component {
   }
 }
 
-const innerProvider = (
-  <Provider store={store}>
-    { /**
+// render the app
+render(
+  (
+    <Provider store={store}>
+      { /**
      * If not using router history, simply include OtpRRExample here:
      * e.g.
      * <OtpRRExample />
      */
-    }
-    <OtpRRExample />
-  </Provider>
-)
-
-// render the app
-render(auth0Config
-  ? (
-    <Auth0Provider
-      audience={AUTH0_AUDIENCE}
-      scope={AUTH0_SCOPE}
-      domain={auth0Config.domain}
-      clientId={auth0Config.clientId}
-      redirectUri={URL_ROOT}
-      {...auth0Callbacks}
-    >
-      {innerProvider}
-    </Auth0Provider>
+      }
+      <OtpRRExample />
+    </Provider>
   )
-  : innerProvider
-,
+  ,
 
-document.getElementById('root')
+  document.getElementById('root')
 )
--- a/lib/actions/auth.js
+++ b/lib/actions/auth.js
@@ -30,7 +30,7 @@ export function showLoginError (err) {
 /**
  * This function is called by the Auth0Provider component, with the described parameter(s),
  * after the user signs in.
- * @param {Error} err
+ * @param {Object} appState The state that was stored when calling useAuth().login().
  */
 export function processSignIn (appState) {
   return function (dispatch, getState) {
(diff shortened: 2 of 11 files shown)


example.js
```diff
@@ -81,37 +75,13 @@ const store = createStore(
...
-              <Nav pullRight>
-                <NavLoginButtonAuth0
-                  id='login-control'
-                  links={accountLinks}
-                />
-              </Nav>
-            </Navbar.Collapse>
-          )}
-        </Navbar>
+        <AppNav otpConfig={otpConfig} title='OpenTripPlanner' />
```
landonreed, line 84: This appears to be a connected component, so there isn't really any need to pass in the otpConfig or the title as props.

lib/util/auth.js
```diff
@@ -42,9 +42,9 @@ export function getAuth0Config (persistence) {
  * Gets the callback methods for Auth0.
  * Note: Methods inside are originally copied from https://github.com/sandrinodimattia/use-auth0-hooks#readme
  * and some methods from that example may still be untouched.
- * @param reduxStore The redux store to carry out the navigation action.
+ * @param dispatch The dispatch method to carry out the actions.
  */
-export function getAuth0Callbacks (reduxStore) {
+export function getAuth0Callbacks (dispatch) {
```
landonreed, line 47: These callbacks are effectively actions, right? I feel like there are some gymnastics happening to try to make everything a util function instead of an action. I don't really like the idea of passing `dispatch` from the component as a param here. I think this is a spot where things could be simplified by making these actions and connecting them in `mapDispatchToProps`. Conventions exist for a reason: we do things in a common predictable way so that the app can be navigated easily by the rest of the team. Also, I don't think we need to combine them all into a `getAuth0Callbacks` function.

lib/components/app/responsive-webapp.js
```diff
@@ -180,8 +183,12 @@ class RouterWrapper extends Component {
   }
 
   render () {
-    const { routerConfig } = this.props
-    return (
+    const { dispatch, persistence, routerConfig } = this.props
+    // Initialize Auth0Provider in this component so it is available everywhere.
+    const auth0Config = getAuth0Config(persistence)
+    const auth0Callbacks = getAuth0Callbacks(dispatch)
```
landonreed, line 189: As mentioned in another comment, these callbacks should be connected to the component as actions.

lib/components/app/responsive-webapp.js
```diff
@@ -241,7 +248,27 @@ class RouterWrapper extends Component {
...
+    return (
+      auth0Config
+        ? (
+          <Auth0Provider
+            audience={AUTH0_AUDIENCE}
+            scope={AUTH0_SCOPE}
+            domain={auth0Config.domain}
+            clientId={auth0Config.clientId}
+            redirectUri={URL_ROOT}
+            {...auth0Callbacks}
```
landonreed, line 261: Please avoid spreading these props. There's only three of them and our goal should be readable, maintainable code.

lib/components/app/app-nav.js
```diff
@@ -4,26 +4,33 @@ import { connect } from 'react-redux'
 
 import NavLoginButtonAuth0 from '../user/nav-login-button-auth0.js'
 import { accountLinks, getAuth0Config } from '../../util/auth'
-
 import AppMenu from './app-menu'
 
-const AppNav = ({ otpConfig }) => {
+const AppNav = ({ otpConfig, title }) => {
```
landonreed, line 9: This is only intended for Desktop use, right? Please add jsdoc to that effect and/or rename to DesktopNavbar.

lib/components/app/app-nav.js
```diff
@@ -4,26 +4,33 @@ import { connect } from 'react-redux'
...
-            // This style is applied here because it is only intended for
-            // desktop view.
-            style={{ marginLeft: 50 }}
-          />
+
+          {branding && (
+            <div
+              className={`icon-${branding}`}
+              // This style is applied here because it is only intended for
+              // desktop view.
```
landonreed, line 26: Not sure this comment is necessary since this file is only intended for Desktop use as far as I can tell.

lib/components/app/app-nav.js
```diff
@@ -4,26 +4,33 @@ import { connect } from 'react-redux'
...
+              style={{ marginLeft: 50 }}
+            />
+          )}
+
+          {title && (
+            <div className='navbar-title' style={{ marginLeft: 50 }}>OpenTripPlanner</div>
```
landonreed, line 32: `title` prop is not actually used? Move "OpenTripPlanner" to default prop perhaps?

lib/components/app/responsive-webapp.js
```diff
@@ -241,7 +248,27 @@ class RouterWrapper extends Component {
...
+        : router
+    )
   }
 }
-const mapStateToWrapperProps = (state, ownProps) => ({ routerConfig: state.otp.config.reactRouter })
+const mapStateToWrapperProps = (state, ownProps) => ({
+  persistence: state.otp.config.persistence,
+  routerConfig: state.otp.config.reactRouter
+})
 export default connect(mapStateToWrapperProps)(RouterWrapper)
```
landonreed, line 274: The RouterWrapper name should be changed to something like `AuthRouterWrapper`. Open to other suggestions. Also, it could probably use jsdoc (can be kept simple) at this point.

lib/components/app/app-nav.js
binh-dam-ibigroup: At some point we'll have to figure out combining this and the mobile navigation bar.

lib/components/app/app-nav.js
binh-dam-ibigroup: Oops!

lib/components/app/responsive-webapp.js
binh-dam-ibigroup: `RouterWrapperWithAuth0`

lib/components/app/app-nav.js
binh-dam-ibigroup: I choose `DesktopNavBar`.

lib/components/app/responsive-webapp.js
binh-dam-ibigroup: Good point. Very good point.

lib/util/auth.js
binh-dam-ibigroup: Sample code legacy...

lib/components/app/responsive-webapp.js
binh-dam-ibigroup: This and turning the auth0 callbacks into actions are addressed in https://github.com/opentripplanner/otp-react-redux/pull/175/commits/0cf1737698da95a83d65cc9a40a696afa9f5db48.

lib/components/app/desktop-nav.js
```diff
@@ -1,29 +1,40 @@
...
 import { accountLinks, getAuth0Config } from '../../util/auth'
-
 import AppMenu from './app-menu'
 
-const AppNav = ({ otpConfig }) => {
+/**
+ * The desktop navigation bar, featuring a sign-in button/menu.
+ * TODO: merge with the mobile navigation bar.
+ */
+const DesktopNav = ({ otpConfig, title }) => {
```
landonreed, line 14: I'm still a little confused as to why there is a `title` prop. Shouldn't this be taking the title from `otpConfig`?

lib/actions/auth.js
```diff
@@ -0,0 +1,56 @@
...
+  return function (dispatch, getState) {
+    // TODO: improve this.
+    if (err) dispatch(push('/oops'))
+  }
+}
+
+/**
+ * This function is called by the Auth0Provider component, with the described parameter(s),
+ * after the user signs in.
+ * @param {Error} err
```
landonreed, line 33: copy pasta param?

lib/actions/auth.js
```diff
@@ -0,0 +1,56 @@
...
+export function processSignIn (appState) {
+  return function (dispatch, getState) {
+    if (appState && appState.urlHash) {
+      // At this stage after login, Auth0 has already redirected to /signedin (Auth0-whitelisted)
+      // which shows the AfterLoginScreen.
+      //
+      // Here, we save the URL hash prior to login (contains a combination of itinerary search, stop/trip view, etc.),
+      // so that the AfterLoginScreen can redirect back there when logged-in user info is fetched.
+      // (For routing, it is easier to deal with the path without the hash sign.)
+      const urlHashWithoutHash = (appState.urlHash.split('#')[1] || '/')
```
landonreed, line 44: Is it possible for the split to yield an array of length = 1? I'm wondering if it's risky to try to access the second element (for fear of array out of index error). Is a better string operation to use here `appState.urlHash.replace('#', '')`?

lib/components/app/desktop-nav.js
binh-dam-ibigroup: I don't know! `OpenTripPlanner` was hardcoded as title before, and `trimet-mod-otp` implementations don't use a title at all, so I was trying to accommodate both.

landonreed: OK, so previously `example.js` just had a very barebones navbar and was just serving as an example for how to piece together otp-react-redux components within a web frame. However, now that we have otp-ui, I think it's OK for `example.js` to change from that original purpose somewhat. That's all just backstory to give you a little more context. For this component: just remove the title prop, get the title from otpConfig, and set the default value to OpenTripPlanner. There's no reason to have a `title` prop, when it's not being used by anything -- that's just dead code that someone down the road will think is important for some unknown reason.

lib/actions/auth.js
evansiroky: I agree it'd be good to avoid array index out of bounds exceptions, although, I think a brute force replacement of the hash character isn't nuanced enough.

lib/actions/auth.js
binh-dam-ibigroup: Pasta is sticky...

lib/actions/auth.js
binh-dam-ibigroup: Using substring instead of array.split in bb9f9e8926a9e638be4f1a9a33340e6c48792067.

lib/components/app/desktop-nav.js
binh-dam-ibigroup: Using title from otpConfig in bb9f9e8926a9e638be4f1a9a33340e6c48792067.

PR #179, opened 2020-06-22
fix(actions/api): Send routing requests to middleware

binh-dam-ibigroup:
This PR is a temporary remedy until https://github.com/ibi-group/otp-middleware/issues/26 is addressed in the middleware.

To use, set the middleware configuration as normal,
but change your OTP API host in `config.yml` to be a regular OTP server (not the middleware):
```yml
# Default OTP API
api:
  host: https://regular-otp-server.example.com
  path: /otp/routers/default
```

### DIFF
--- a/__tests__/components/viewers/__snapshots__/stop-viewer.js.snap
+++ b/__tests__/components/viewers/__snapshots__/stop-viewer.js.snap
@@ -376,7 +376,7 @@ exports[`components > viewers > stop viewer should render countdown times after 
                                     From Location Icon
                                   </title>
                                   <path
-                                    d="M256 8C119.033 8 8 119.033 8 256s111.033 248 248 248 248-111.033 248-248S392.967 8 256 8zm80 248c0 44.112-35.888 80-80 80s-80-35.888-80-80 35.888-80 80-80 80 35.888 80 80z"
+                                    d="M256 56c110.532 0 200 89.451 200 200 0 110.532-89.451 200-200 200-110.532 0-200-89.451-200-200 0-110.532 89.451-200 200-200m0-48C119.033 8 8 119.033 8 256s111.033 248 248 248 248-111.033 248-248S392.967 8 256 8zm0 168c-44.183 0-80 35.817-80 80s35.817 80 80 80 80-35.817 80-80-35.817-80-80-80z"
                                     fill="currentColor"
                                     key="k0"
                                   />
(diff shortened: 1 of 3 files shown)


__tests__/components/viewers/__snapshots__/stop-viewer.js.snap
```diff
@@ -4746,7 +4746,7 @@ exports[`components > viewers > stop viewer should render with TriMet transit in
                         onClick={[Function]}
                       >
                         <button
-                          className="styled__Button-p56b41-2 XBKVp"
+                          className="styled__Button-p56b41-2 bjZNPA"
```
evansiroky, line 4749: oh my, we need to figure out how to remedy these changing snapshots ASAP. There has to be some kind of jest plugin for this.

lib/actions/api.js
```diff
@@ -214,6 +220,26 @@ function constructRoutingQuery (otpState, ignoreRealtimeUpdates) {
   // Check for routingType-specific API config; if none, use default API
   const rt = config.routingTypes && config.routingTypes.find(rt => rt.key === routingType)
   const api = (rt && rt.api) || config.api
+
+  // FIXME: Remove this block below once there is a fix
+  // for https://github.com/ibi-group/otp-middleware/issues/26.
```
evansiroky, line 225: I'm going to add the blocked label until that issue is resolved then. We shouldn't create PRs with FIXMEs and instead should fix the underlying issues.

binh-dam-ibigroup: I have reverted the snapshot. It's a good ideas to avoid changes if the file is irrelevant.

__tests__/components/viewers/__snapshots__/stop-viewer.js.snap
binh-dam-ibigroup: I have reverted the snapshot. It's a good ideas to avoid changes if the file is irrelevant.

lib/actions/api.js
binh-dam-ibigroup: As discussed, ibigroup/otp-middleware#26 will be worked on. Discuss if we need to show the working stop/trip viewer along with logging trip requests meanwhile.

binh-dam-ibigroup:
Closing in favor of https://github.com/ibi-group/otp-middleware/pull/31.

Issue #181, opened 2020-06-26
ArcGis geocoder does not handle getting locations from 'search' query correctly.


landonreed:
The ArcGis geocoder adapter does not correctly handle getting the location from a search query. The response for this query type is simply a GeoJSON-style feature with geometry and properties (not the magicKey/text result that comes back from autocomplete). Currently `ArcGISGeocoder#getLocationFromGeocodedFeature` only handles the autocomplete style result:
https://github.com/opentripplanner/otp-react-redux/blob/e04aac7de96a0eedaf09403affc00b73f2df896f/lib/util/geocoder.js#L129-L142

PR #182, opened 2020-06-26
Geocoding fixes

### LINKED ISSUE #180: Handle ungeocoded from and to places when parsing query params
Some transit agencies have simple forms on their website that connect with Google Transit/Maps by sending the place names, time, and date. otp-react-redux should accommodate the same kind of thing, by simply attempting to geocode the incoming place names when parsing the query params.

landonreed:
This PR fixes #180 and #181. It is intended for `master` because it contains maintenance changes for an existing implementation, but we should also ensure that the changes make their way into the otp-ui side of things.

To test the geocoding of ungeocoded place names:
1. use the SEPTA config
2. `yarn start`
3. Visit http://localhost:9966/#/?fromPlace=wells%20fargo%20center&toPlace=strawberry%20mansion

### DIFF
--- a/lib/util/query.js
+++ b/lib/util/query.js
@@ -151,7 +151,7 @@ export function getDefaultQuery (config) {
 
 /**
  * Geocode utility for returning the first result for the provided place name text.
- * @param  {String} text - text to search
+ * @param  {string} text - text to search
  * @param  {Object} geocoderConfig
  * @return {Location}
  */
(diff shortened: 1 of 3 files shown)


lib/util/query.js
```diff
@@ -148,22 +149,52 @@ export function getDefaultQuery (config) {
   return defaultQuery
 }
 
+/**
+ * Geocode utility for returning the first result for the provided place name text.
+ * @param  {String} text - text to search
+ * @param  {Object} geocoderConfig
+ * @return {Location}
+ */
+async function getFirstGeocodeResult (text, geocoderConfig) {
```
binh-dam-ibigroup, line 158: Is the underlying geocoder API async?

lib/util/query.js
```diff
@@ -148,22 +149,52 @@ export function getDefaultQuery (config) {
...
         query.from = parseLocationString(params.fromPlace)
+        // If a valid location was not found, but the place name text exists,
+        // attempt to geocode the name.
+        if (!query.from && params.fromPlace) {
+          query.from = await getFirstGeocodeResult(params.fromPlace, config.geocoder)
+        }
         break
       case 'toPlace':
         query.to = parseLocationString(params.toPlace)
+        // If a valid location was not found, but the place name text exists,
```
binh-dam-ibigroup, line 193: Add a TODO comment to note potential for combining the code of these two cases.

lib/util/query.js
landonreed: The geocoder API makes asynchronous fetches, so if we want to be able to use the result inline (instead of in a then clause), we need to make this function async.

lib/util/query.js
landonreed: Fixed.

PR #183, opened 2020-06-29
Add SettingsPreview and TripDetails messages

landonreed:
Adds new messages pattern for SettingsPreview and TripDetails components.

Non-breaking config changes:

```yaml
languages:
  settingsPreview:
    label: Trip options
  tripDetails:
    # see TripDetails#defaultProps
```

### DIFF
--- a/lib/components/form/settings-preview.js
+++ b/lib/components/form/settings-preview.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types'
 import { Button } from 'react-bootstrap'
 import { connect } from 'react-redux'
 
+import { mergeMessages } from '../../util/messages'
 import { isNotDefaultQuery } from '../../util/query'
 
 class SettingsPreview extends Component {
@@ -22,11 +23,16 @@ class SettingsPreview extends Component {
   }
 
   static defaultProps = {
-    editButtonText: <i className='fa fa-pencil' />
+    editButtonText: <i className='fa fa-pencil' />,
+    messages: {
+      // Label accepted as two-element array or single (not array) string literal.
+      label: ['Transit Options', '& Preferences']
+    }
   }
 
   render () {
-    const { config, query, caret, editButtonText } = this.props
+    const { caret, config, query, editButtonText } = this.props
+    const messages = mergeMessages(SettingsPreview.defaultProps, this.props)
     // Show dot indicator if the current query differs from the default query.
     let showDot = isNotDefaultQuery(query, config)
     const button = (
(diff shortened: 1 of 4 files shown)


lib/components/narrative/trip-details.js
```diff
@@ -47,30 +68,39 @@ class TripDetails extends Component {
       format: timeFormat,
       offset: getTimeZoneOffset(itinerary)
     }
-
+    console.log(messages.at, messages)
```
evansiroky, line 71: remove please

lib/components/form/settings-preview.js
```diff
@@ -40,7 +46,22 @@ class SettingsPreview extends Component {
 
     return (
       <div className='settings-preview' onClick={this.props.onClick}>
-        <div className='summary'>Transit Options<br />&amp; Preferences</div>
+        <div className='summary'>
+          {// If settings preview is multiple lines, split
+            Array.isArray(messages.label)
+              ? <span>
+                {messages.label[0]}
```
evansiroky, line 53: Since this is an array, maybe it should be a be rendered using a map function.

lib/components/form/settings-preview.js
```diff
@@ -22,11 +23,16 @@ class SettingsPreview extends Component {
   }
 
   static defaultProps = {
-    editButtonText: <i className='fa fa-pencil' />
+    editButtonText: <i className='fa fa-pencil' />,
+    messages: {
+      // Label accepted as two-element array or single (not array) string literal.
```
binh-dam-ibigroup, line 28: What happens if someone passes `messages` with attributes that don't include `label`?

landonreed: `mergeMessages` will ensure that the default label key gets preserved.

binh-dam-ibigroup: Ok the answer was below in `mergeMessages`.

binh-dam-ibigroup, line 28: I think this comment needs to say that each array entry is rendered as a separate line for the button caption.

binh-dam-ibigroup: Or, you could use this: https://stackoverflow.com/a/55466235

lib/components/form/settings-preview.js
```diff
@@ -40,7 +46,22 @@ class SettingsPreview extends Component {
...
+                {messages.label[0]}
+                <br />
+                {messages.label[1]}
+              </span>
+              : <span
+                // Increase line height if number of lines is 1 for vertical
+                // centering.
+                style={{lineHeight: 2.6}}>
+                {messages.label}
+              </span>
```
binh-dam-ibigroup, line 62: Could this work? https://stackoverflow.com/a/55466235

landonreed: Good suggestion. It might be tricky to get the styling right on that though (the line height = 2.6 is important for centering vertically and might be hard to detect without the array).

PR #185, opened 2020-07-01
fix(default-map): Bring back vehicle rental overlay visibility support

binh-dam-ibigroup:
This PR brings back support for toggling vehicle rental overlay visibility when selecting the corresponding mode in the mode selector. This PR addresses this comment: https://github.com/ibi-group/trimet-mod-otp/pull/265#issuecomment-652550928

To test, use a configuration with vehicle rentals, and observe that the vehicle rental overlays are being toggled when selecting modes in the mode selector.

### DIFF
--- a/lib/components/map/default-map.js
+++ b/lib/components/map/default-map.js
@@ -109,8 +109,6 @@ class DefaultMap extends Component {
   }
 
   componentDidUpdate (prevProps) {
-    // TODO: reinstate this call below.
-    // this._updateBounds(prevProps, this.props)
     // Check if any overlays should be toggled due to mode change
     this._handleQueryChange(prevProps.query, this.props.query)
   }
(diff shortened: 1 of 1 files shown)


lib/components/map/default-map.js
```diff
@@ -56,6 +108,13 @@ class DefaultMap extends Component {
     setLocation(payload)
   }
 
+  componentDidUpdate (prevProps) {
+    // TODO: reinstate this call below.
+    // this._updateBounds(prevProps, this.props)
```
evansiroky, line 113: The bounds updating logic has been moved to the `BoundsUpdatingOverlay` component. I think this is already accounted for.

binh-dam-ibigroup: Great. I have updated the comment.

PR #188, opened 2020-07-06
fix(StopViewerOverlay): Fix prop name for overlay.

binh-dam-ibigroup:
This PR fixes the missing teal marker normally displayed on the map when the stop viewer is shown.


evansiroky:
Should we merge this to the alpha branch first and then dev?

PR #194, opened 2020-07-08
Resolve dev->master conflicts

binh-dam-ibigroup:
This PR resolves conflicts between the dev and master branches. Also, it uses the most recent the OTP-ui package versions.


binh-dam-ibigroup:
@evansiroky and @landonreed I think the snapshot failure come from a change in the `from` icon in the `FromToLocationPicker` introduced with https://github.com/opentripplanner/otp-ui/issues/136. We should update the snapshots to reflect that.

evansiroky:
In the future, we should be more cognizant of the commit messages we use.

This is not a feature in my opinion: `feat: Resolve merge conflicts master <- dev`

This will show up in the release notes as being a new feature, when it is really not. We should use the `feat` and `fix` commit prefix very sparingly. We shouldn't even use `fix` unless we are fixing a bug in the current master branch. Often times we'll use `fix` for "fixes" we do during PR review on stuff that is actually new features or general refactoring.

binh-dam-ibigroup:
> In the future, we should be more cognizant of the commit messages we use.
> 
> This is not a feature in my opinion: `feat: Resolve merge conflicts master <- dev`
> 
> This will show up in the release notes as being a new feature, when it is really not. We should use the `feat` and `fix` commit prefix very sparingly. We shouldn't even use `fix` unless we are fixing a bug in the current master branch. Often times we'll use `fix` for "fixes" we do during PR review on stuff that is actually new features or general refactoring.

Oh... interesting... sorry I thought I didn't specify a category so it might have defaulted to `feat`?

landonreed:
@evansiroky, agreed. This is definitely not a feature. It's probably more appropriately categorized as refactor or chore.

PR #201, opened 2020-07-17
Correctly sort routes in route viewer by agency

evansiroky:
<strike>This PR looks at the route id and parses out the agency ID and then attempts to match to a transit operator id in order to properly sort routes by multiple agencies. Also, the operator image is now shown in the route viewer from this fix.

To try this out, look at the route list with a deployment with multiple agencies.</strike>

This PR will now correctly sort routes in both the route viewer and stop viewer using the core-utils v3 logic and config if provided. As such, this updates all otp-ui packages to the latest version. Part of that update is correcting how the stops-overlay works so that it works with the zoom-based-overlay. One more thing that this PR does is now show a mode icon for each route in the route viewer and also an agency icon (if one is properly defined in the transit operators config).


lib/components/viewers/route-viewer.js
```diff
@@ -159,11 +164,13 @@ class RouteRow extends PureComponent {
           onClick={this._onClick}
         >
           <div style={{display: 'inline-block'}}>
-            {// TODO: re-implement multi-agency logos for route viewer.
-              // Currently, the agency object is not nested within the get all
-              // routes endpoint and causing this to only display transitOperators for
-              // the selected route.
-              // operator && <img src={operator.logo} style={{marginRight: '5px'}} height={25} />
+            {operator &&
```
binh-dam-ibigroup, line 167: On the configs that don't define a logo, this results in a "broken image link" placeholder to be shown, so the condition for showing `` should rather be `operator && operator.logo`.

lib/components/viewers/route-viewer.js
```diff
@@ -93,14 +99,13 @@ class RouteViewer extends Component {
           {agencySortedRoutes
             .map(route => {
               // Find operator based on agency_id (extracted from OTP route ID).
-              // TODO: re-implement multi-agency logos for route viewer.
-              // const operator = operatorForRoute(transitOperators, route) || {}
+              const operator = operatorForRoute(transitOperators, route) || {}
```
landonreed, line 102: If the operator is missing, the `` tag is shown below, but it is a broken image. I think this default to empty object should be removed.

lib/components/viewers/route-viewer.js
```diff
@@ -62,7 +67,8 @@ class RouteViewer extends Component {
       : []
     const agencySortedRoutes = transitOperators.length > 0
       ? sortedRoutes.sort((a, b) => {
-        return operatorIndexForRoute(transitOperators, a) - operatorIndexForRoute(transitOperators, b)
+        return routeOperatorComparatorValue(transitOperators, a) -
+          routeOperatorComparatorValue(transitOperators, b)
```
landonreed, line 71: I wonder if we should just be embedding this logic within otp-ui? Perhaps not, but I think the question is worth asking.

lib/components/viewers/route-viewer.js
landonreed, line 167: Perhaps this check for operator existing should extend to the logo field being defined as well.

landonreed: If no logo is found, should some other indicator for agency be used (e.g., name)?

lib/components/viewers/route-viewer.js
evansiroky: This seems specific to the route-viewer, so I'd say if the route-viewer becomes an otp-ui component, then it'd make sense to put this in otp-ui core-utils.

evansiroky:
This PR is now dependent on https://github.com/opentripplanner/otp-ui/pull/188.

evansiroky:
This PR is now dependent on https://github.com/opentripplanner/otp-ui/pull/189.

evansiroky:
This PR is ready to go, except for being dependent on https://github.com/opentripplanner/otp-ui/pull/221.

binh-dam-ibigroup (approved):
Looks good. There are some code and layout items that could be refined further, but I'm in favor to pass this as is.

lib/util/viewer.js
```diff
@@ -162,3 +162,26 @@ export function getStopTimesByPattern (stopData) {
   }
   return stopTimesByPattern
 }
+
+/**
+ * Gets the mode string from either an OTP Route or RouteShort model. The OTP
+ * Route model returns the mode as an integer type whereas the RouteShort model
+ * returns the mode string.
+ */
+export function getModeFromRoute (route) {
```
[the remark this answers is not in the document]
evansiroky: I thought that too, but nothing else in otp-ui needs this at the moment, so I just put it here to not create even more otp-ui PRs.

lib/components/viewers/route-viewer.js
```diff
@@ -149,46 +178,55 @@ class RouteRow extends PureComponent {
...
-          <div style={{display: 'inline-block', marginTop: '2px'}}>
-            <Label
-              style={{
-                backgroundColor: backgroundColor === '#ffffff' ? 'rgba(0,0,0,0)' : backgroundColor,
-                fontSize: 'medium',
-                fontWeight: 400,
-                color
-              }}>
+          </RouteRowElement>
+          <ModeIconElement>
```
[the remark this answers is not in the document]
evansiroky: This was to ensure that rotues with a background color on their label had sufficient space. I went ahead and changed the styling to reduce that spacing for labels with a white background color in 6cacf9132bad719af6842ed15f6e677765b90596.

evansiroky:
Ready for review!

lib/components/app/batch-routing-panel.js
```diff
@@ -33,12 +33,13 @@ class BatchRoutingPanel extends Component {
...
       itineraryFooter,
       LegIcon,
+      ModeIcon,
       mobile,
       showUserSettings
     } = this.props
     const actionText = mobile ? 'tap' : 'click'
     return (
-      <ViewerContainer>
+      <ViewerContainer ModeIcon={ModeIcon}>
```
landonreed, line 42: Maybe out of scope for this PR, but is there a better way for us to pass around Leg/ModeIcon? Would it be useful to place them in the store on app startup (passed along with the OTP config object perhaps). I just find it a bit cumbersome to pass them to these connected components in this way.

lib/components/viewers/stop-viewer.js
```diff
@@ -250,12 +250,21 @@ class StopViewer extends Component {
...
+    // construct a lookup table mapping pattern (e.g. 'ROUTE_ID-HEADSIGN') to
+    // an array of stoptimes
     const stopTimesByPattern = getStopTimesByPattern(stopData)
+    const routeComparator = coreUtils.route.makeRouteComparator(
+      transitOperators
+    )
+    const patternComparator = (patternA, patternB) => routeComparator(
+      patternA.route,
+      patternB.route
+    )
```
landonreed, line 267: This doesn't seem to have a fallback sort if the routes are the same. Should we have it fallback on pattern headsign or something along those lines?

lib/components/viewers/route-viewer.js
```diff
@@ -113,6 +109,53 @@ class RouteViewer extends Component {
...
+  width: 100%;
+`
+
+const RouteRowElement = styled.div`
+  display: inline-block;
+`
+
+const OperatorImg = styled.img`
+  height: 25px;
+  margin-right: 8px;
```
landonreed, line 128: The operator image appears to be offset vertically slightly from the other row elements:

image

landonreed (changes requested):
Just a few minor comments.

lib/components/app/batch-routing-panel.js
evansiroky: See #274 which proposes to address this comment.

lib/components/viewers/stop-viewer.js
evansiroky: We can do that. I'd like to reuse a sorting helper that isn't exported from otp-ui. Can you please approve https://github.com/opentripplanner/otp-ui/pull/226 to make this possible?

landonreed: Approved.

lib/components/viewers/route-viewer.js
evansiroky: Fixed in 576730fe82ce0ae5ffac6e30c6f011b71b93a888

lib/components/viewers/stop-viewer.js
evansiroky: Fixed in d9410f90947481eaebe0a3c0d43c9ce0cff4af62 and e110f9aa9774b83ed8649bb7aeb76f533625a5ac

evansiroky:
Ready for review again.

example-config.yml
```diff
@@ -130,6 +130,38 @@ itinerary:
...
 
+# Provide a list of transit operators to enhance the user experience for multi-
+# agency implementations. The `order` value of each agencies will dictate the
+# order that routes appear in the route and stop viewers. Also, optionally
+# provide operator logos to appear in itineraries). Note: these logos will
+# override the values found in agency.txt#agency_branding_url. Agencies are
+# matched according to both the feedId and agencyId. This works best with the
+# IBI fork of OTP which returns the necessary information in the route index
+# API. Check the build logs of OTP to see what feedIds and agencyIds were found
+# as these IDs can sometimes be auto-generated.
```
landonreed, line 141: What about feeds that have a single agency and no `agency_id`?

landonreed (approved):
Looks good. Just a final question on how feeds with no agency id are handled.

example-config.yml
evansiroky: Additional comments added in 98ef742fc280ac2ea58a2faae8fc09220dbe7c11

PR #222, opened 2020-09-01
Add ability to show mapbox wordmark via config

evansiroky:
This PR adds the ability to add a simple HTML element for displaying the [Mapbox wordmark attribution](https://docs.mapbox.com/help/how-mapbox-works/attribution/#other-mapping-frameworks) on the map.

### DIFF
--- a/lib/components/map/default-map.js
+++ b/lib/components/map/default-map.js
@@ -141,6 +141,13 @@ class DefaultMap extends Component {
 
     return (
       <MapContainer>
+        {/* Add the mapbox wordmark if set in config. The implementing
+          application must include CSS that properly displays the wordmark. See
+          mapbox website for example CSS:
+          https://docs.mapbox.com/help/how-mapbox-works/attribution/#other-mapping-frameworks */}
+        {mapConfig.addMapboxWordmark &&
+          <a href='http://mapbox.com/about/maps' class='mapbox-wordmark' target='_blank'>Mapbox</a>
+        }
         <BaseMap
           baseLayers={mapConfig.baseLayers}
           center={center}


lib/components/map/default-map.js
```diff
@@ -141,6 +141,13 @@ class DefaultMap extends Component {
 
     return (
       <MapContainer>
+        {/* Add the mapbox wordmark if set in config. The implementing
+          application must include CSS that properly displays the wordmark. See
+          mapbox website for example CSS:
+          https://docs.mapbox.com/help/how-mapbox-works/attribution/#other-mapping-frameworks */}
+        {mapConfig.addMapboxWordmark &&
+          <a href='http://mapbox.com/about/maps' class='mapbox-wordmark' target='_blank'>Mapbox</a>
+        }
```
landonreed, line 150: I don't love the idea of adding an additional config property like this. It's just additional (unnecessary) configuration. What if we just check the active layer URL and add the wordmark if it comes from `mapbox.com`.

evansiroky:
Unassigning myself since it appears no action is required of me at this time.

lib/components/map/default-map.js
```diff
@@ -141,6 +141,13 @@ class DefaultMap extends Component {
 
     return (
       <MapContainer>
+        {/* Add the mapbox wordmark if set in config. The implementing
+          application must include CSS that properly displays the wordmark. See
+          mapbox website for example CSS:
+          https://docs.mapbox.com/help/how-mapbox-works/attribution/#other-mapping-frameworks */}
+        {mapConfig.addMapboxWordmark &&
+          <a href='http://mapbox.com/about/maps' class='mapbox-wordmark' target='_blank'>Mapbox</a>
```
binh-dam-ibigroup, line 149: It should be possible to use React-Leaflet 's built-in [attribution control](https://react-leaflet.js.org/docs/en/components#attributioncontrol) or Leaflet's equivalent. Doing so would avoid all of the CSS styling proposed in https://github.com/ibi-group/configurations/pull/82. This may require a change in OTP-UI instead of OTP-RR. Thoughts?

evansiroky:
The CSS is required to properly style the Mapbox logo. Also, the built-in attribution control is quite limited in what it can do and it doesn't seem to have an easy way to add HTML/CSS to it. I think it's probably easier to add the HTML element this way. However, I do think that @binh-dam-ibigroup's suggestion to move this to otp-ui makes sense and also that @landonreed's [idea of automagically inserting the Mapbox wordmark](https://github.com/opentripplanner/otp-react-redux/pull/222#discussion_r481132132) instead of relying on config is a good idea, so I'm closing this in favor https://github.com/opentripplanner/otp-ui/pull/211.

PR #224, opened 2020-09-04
Phone Verification

binh-dam-ibigroup:
This PR adds the phone number verification for new and existing MOD UI users when they enter a new phone number for trip notifications. The is a little bit different than in original mockups but for now it is a single flow that works for both new and existing users.

~UPDATE: commit https://github.com/ibi-group/otp-middleware/pull/83/commits/5621a131b26f784d618e0a8cb228ba3e3700c787 or equivalent is required.~

UPDATE: commit https://github.com/ibi-group/otp-middleware/pull/83/commits/a9386a6e2f46bc9529e14d51ff9e51106ec87ba6 or equivalent is required.

UPDATE: there ~is a~ are new configuration parameters for:
- the phone number regex.
- phone region code
- phone number placeholder.

This PR depends on ~https://github.com/ibi-group/otp-middleware/pull/33~ https://github.com/ibi-group/otp-middleware/pull/83 (remember to add the Twilio SMS service config variable to the middleware) ~and implements a workaround for [this issue](https://github.com/ibi-group/otp-middleware/pull/33#discussion_r483187826)~.


lib/actions/user.js
```diff
@@ -1,3 +1,4 @@
+import clone from 'lodash/cloneDeep'
```
evansiroky, line 1: The `clone` package is installed in the otp-react-redux package.json, so that package should probably be used.

lib/actions/user.js
```diff
@@ -181,3 +188,91 @@ export function deleteUserMonitoredTrip (id) {
...
+      const { accessToken } = user
+
+      // FIXME: (required by middleware) Only temporarily update the user's phone number
+      //   in the database with the pending one,
+      //   and cache the previous value and verification state so they can be reverted.
+      //   The full user record will be updated upon the user clicking Finish/Save preferences.
+      // TODO: Figure out what should happen if the user refreshes the browser at this stage.
+
+      // Make a clone of the original userData object.
+      const userData = clone(originalUserData)
```
evansiroky, line 210: Might be more readable to rename `userData` to `newUserData`.

lib/util/middleware.js
```diff
@@ -139,3 +140,20 @@ export async function deleteTrip (middlewareConfig, token, id) {
     }
   }
 }
+
+export async function getPhoneVerificationCode (middlewareConfig, token, userId) {
```
evansiroky, line 144: These 2 util functions don't clearly convey meaning to me. I suggest renaming them to `requestPhoneVerificationCode` and `confirmPhoneVerificationCode` respectively.

lib/components/user/notification-prefs-pane.js
```diff
@@ -85,13 +226,59 @@ class NotificationPrefsPane extends Component {
...
-              <FormControl onChange={this._handlePhoneNumberChange} type='tel' value={phoneNumber} />
-            </FormGroup>
+            <div>
+              <FormGroup validationState={phoneFieldValidationState}>
+                <ControlLabel>Enter your phone number for SMS notifications:</ControlLabel>
+                {/* TODO: Add field validation. */}
+
+                <FormControl
+                  onChange={this._handlePhoneNumberChange}
+                  placeholder='+12223334567'
```
evansiroky, line 236: This is actually a valid phone number, so we shouldn't use it. Instead use something with a 555 area code and 555 prefix. [Read here for more info](https://en.wikipedia.org/wiki/555_(telephone_number)).

lib/components/user/notification-prefs-pane.js
```diff
@@ -41,16 +75,123 @@ class NotificationPrefsPane extends Component {
...
+        phoneFieldValidationState = 'error'
+      }
+    } else if (isPhoneNumberSameAsInitial && isPhoneNumberVerified) {
+      phoneStatusGlyph = 'ok'
+      phoneStatusText = 'Verified'
+      phoneFieldValidationState = 'success'
+    } else {
+      phoneStatusGlyph = 'remove'
+      phoneStatusText = 'Verification required'
+      phoneFieldValidationState = 'error'
```
evansiroky, line 191: Perhaps this should be a `warning` validation state instead of an error since the user hasn't done anything wrong, just needs to confirm the number.

evansiroky (changes requested):
See comments.

lib/components/user/notification-prefs-pane.js
binh-dam-ibigroup: Good to know! Fixed in https://github.com/opentripplanner/otp-react-redux/pull/224/commits/626fdb7a50e90c7a827c98e3d9a00d29f2025365.

lib/util/middleware.js
binh-dam-ibigroup: I have renamed these functions in https://github.com/opentripplanner/otp-react-redux/pull/224/commits/626fdb7a50e90c7a827c98e3d9a00d29f2025365.

lib/components/user/notification-prefs-pane.js
binh-dam-ibigroup: Fixed in https://github.com/opentripplanner/otp-react-redux/pull/224/commits/626fdb7a50e90c7a827c98e3d9a00d29f2025365

lib/actions/user.js
binh-dam-ibigroup: Fixed in https://github.com/opentripplanner/otp-react-redux/pull/224/commits/626fdb7a50e90c7a827c98e3d9a00d29f2025365

evansiroky (approved):

lib/actions/user.js
```diff
@@ -181,3 +188,91 @@ export function deleteUserMonitoredTrip (id) {
...
+ * Validate the phone number verification code for the logged-in user.
+ */
+export function verifyPhoneNumber (originalUserData, code) {
+  return async function (dispatch, getState) {
+    const { otp, user } = getState()
+    const { otp_middleware: otpMiddleware = null } = otp.config.persistence
+
+    if (otpMiddleware) {
+      const { accessToken } = user
+      const sendResult = await validatePhoneVerificationCode(otpMiddleware, accessToken, originalUserData.id, code)
```
landonreed, line 250: This util function is used once. I don't really see the value in having the fetch code isolated, especially with a method that has four args.

lib/actions/user.js
```diff
@@ -181,3 +188,91 @@ export function deleteUserMonitoredTrip (id) {
...
+    const { otp_middleware: otpMiddleware = null } = otp.config.persistence
+
+    if (otpMiddleware) {
+      const { accessToken } = user
+
+      // FIXME: (required by middleware) Only temporarily update the user's phone number
+      //   in the database with the pending one,
+      //   and cache the previous value and verification state so they can be reverted.
+      //   The full user record will be updated upon the user clicking Finish/Save preferences.
+      // TODO: Figure out what should happen if the user refreshes the browser at this stage.
```
landonreed, line 207: If there needs to be a change to the middleware design, let's just discuss/handle that now instead of embedding things to fix in this PR.

lib/actions/user.js
```diff
@@ -181,3 +188,91 @@ export function deleteUserMonitoredTrip (id) {
...
+      const newUserData = clone(originalUserData)
+      const previousPhoneNumber = newUserData.phoneNumber
+      const previousIsPhoneNumberVerified = newUserData.isPhoneNumberVerified
+
+      newUserData.phoneNumber = newPhoneNumber
+      newUserData.isPhoneNumberVerified = false
+
+      const userUpdateResult = await updateUser(otpMiddleware, accessToken, newUserData)
+
+      if (userUpdateResult.status === 'success' && userUpdateResult.data) {
```
landonreed, line 219: This method body contains too much logic to not have any comments describing what's going on here.

lib/actions/user.js
```diff
@@ -181,3 +188,91 @@ export function deleteUserMonitoredTrip (id) {
...
+export function verifyPhoneNumber (originalUserData, code) {
+  return async function (dispatch, getState) {
+    const { otp, user } = getState()
+    const { otp_middleware: otpMiddleware = null } = otp.config.persistence
+
+    if (otpMiddleware) {
+      const { accessToken } = user
+      const sendResult = await validatePhoneVerificationCode(otpMiddleware, accessToken, originalUserData.id, code)
+      if (sendResult.status === 'success' && sendResult.data) {
+        if (sendResult.data.status === 'approved') {
```
landonreed, line 252: Please add comments about what this status indicates.

lib/actions/user.js
```diff
@@ -181,3 +188,91 @@ export function deleteUserMonitoredTrip (id) {
...
+          const newUserData = clone(originalUserData)
+
+          // Update phone number and verification in database record.
+          // FIXME: The call to updateUer below assumes the middleware requires saving the user's phone number prior to verification.
+          newUserData.isPhoneNumberVerified = true
+          newUserData.notificationChannel = 'sms'
+          const userUpdateResult = await updateUser(otpMiddleware, accessToken, newUserData)
+
+          if (userUpdateResult.status === 'success' && userUpdateResult.data) {
+            alert('Your phone is now verified and set to receive trip notifications.')
```
landonreed, line 263: We need to stop embedding alerts in the code...

landonreed (changes requested):
A few comments about how things are laid out. There are a lot of alerts embedded here, but that shouldn't really be a pattern that we start using heavily. Also, it seems like there is a lot of maneuvering to account for how things are set up in otp-middleware. If we need to change server-side things, let's just handle that now.

lib/actions/user.js
```diff
@@ -181,3 +188,91 @@ export function deleteUserMonitoredTrip (id) {
...
+
+/**
+ * Requests a verification code via SMS for the logged-in user.
+ */
+export function requestPhoneVerificationSms (originalUserData, newPhoneNumber) {
+  return async function (dispatch, getState) {
+    const { otp, user } = getState()
+    const { otp_middleware: otpMiddleware = null } = otp.config.persistence
+
+    if (otpMiddleware) {
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Good suggestion for throwing an error if `otpMiddleware` is null.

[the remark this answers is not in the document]
binh-dam-ibigroup: I have been intending to refactor the code involving the extraction and use of the `otpMiddleware` variable, it would be best to do that in a separate PR.

lib/actions/user.js
```diff
@@ -113,12 +118,14 @@ export function createOrUpdateUser (userData) {
 
       // TODO: improve the UI feedback messages for this.
       if (result.status === 'success' && result.data) {
-        alert('Your preferences have been saved.')
+        if (!silentOnSuccess) {
+          alert('Your preferences have been saved.')
+        }
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Moving to a more general discussion in https://github.com/opentripplanner/otp-react-redux/issues/245

lib/actions/user.js
binh-dam-ibigroup: I'll remove this one, as there is already a check mark feedback shown when the number is verified.

lib/actions/user.js
```diff
@@ -181,3 +188,91 @@ export function deleteUserMonitoredTrip (id) {
...
+      // FIXME: (required by middleware) Only temporarily update the user's phone number
+      //   in the database with the pending one,
+      //   and cache the previous value and verification state so they can be reverted.
+      //   The full user record will be updated upon the user clicking Finish/Save preferences.
+      // TODO: Figure out what should happen if the user refreshes the browser at this stage.
+
+      // Make a clone of the original userData object.
+      const newUserData = clone(originalUserData)
+      const previousPhoneNumber = newUserData.phoneNumber
+      const previousIsPhoneNumberVerified = newUserData.isPhoneNumberVerified
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/224/commits/45a2057bd625d2d35e15f452d5c73b35514bba3d (along with other places that needed comments).

lib/actions/user.js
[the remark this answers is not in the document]
binh-dam-ibigroup: Addressed separately in in https://github.com/opentripplanner/otp-react-redux/pull/246.

lib/actions/user.js
binh-dam-ibigroup: Will address once https://github.com/opentripplanner/otp-react-redux/pull/246 is merged.

lib/actions/user.js
binh-dam-ibigroup: I agree, as noted in the PR description.

landonreed: @binh-dam-ibigroup, can you take the lead on scheduling a call to resolve the backend design so that we can merge/fix this?

lib/components/user/notification-prefs-pane.js
```diff
@@ -85,13 +223,60 @@ class NotificationPrefsPane extends Component {
...
+                {/* Show glyphs underneath the input control with <HelpBlock>
+                   (there are some alignment issues with <FormControl.Feedback /> in mobile view). */}
+                <HelpBlock>
+                  {phoneStatusGlyph && <Glyphicon glyph={phoneStatusGlyph} />} {phoneStatusText}
+                </HelpBlock>
+              </FormGroup>
+
+              {shouldVerifyPhone && (
+                <ButtonStrip>
+                  <Button bsStyle='primary' onClick={this._handleRequestPhoneVerificationCode}>Verify my phone</Button>
```
landonreed, line 247: This should say "Send verification text".

lib/components/user/notification-prefs-pane.js
```diff
@@ -41,16 +76,119 @@ class NotificationPrefsPane extends Component {
...
+    } else if (isPhoneNumberSameAsInitial && isPhoneNumberVerified) {
+      phoneStatusGlyph = 'ok'
+      phoneStatusText = 'Verified'
+      phoneFieldValidationState = 'success'
+    } else {
+      phoneStatusGlyph = 'warning-sign'
+      phoneStatusText = 'Verification required'
+      phoneFieldValidationState = 'warning'
+      shouldVerifyPhone = true
+    }
```
landonreed, line 190: These states aren't working perfectly for me. E.g., I can click "Revert" twice in a row for some reason?

lib/components/user/notification-prefs-pane.js
```diff
@@ -85,13 +223,60 @@ class NotificationPrefsPane extends Component {
             </FormGroup>
```
landonreed, line 223: In the ToggleButton group above, it might be preferable just to handle setting the notification channel (i.e., making an updateUser request to the backend) on click.

lib/components/user/notification-prefs-pane.js
```diff
@@ -85,13 +223,60 @@ class NotificationPrefsPane extends Component {
...
+                <ControlLabel>Enter your phone number for SMS notifications:</ControlLabel>
+                {/* TODO: Add field validation. */}
+
+                <FormControl
+                  onChange={this._handlePhoneNumberChange}
+                  // Fake phone number, see https://en.wikipedia.org/wiki/555_(telephone_number)
+                  placeholder='+15555550123'
+                  type='tel'
+                  value={phoneNumber}
+                />
```
landonreed, line 237: A style note: this form element should probably have a max width instead of extending the width of the screen.

lib/components/user/notification-prefs-pane.js
```diff
@@ -85,13 +223,60 @@ class NotificationPrefsPane extends Component {
...
+              <FormControl onChange={this._handlePhoneValidationCodeChange} value={phoneValidationCode} />
+            </FormGroup>
+            <p>
+              <Button bsStyle='link' onClick={this._handleRequestPhoneVerificationCode}>Request a new code</Button>
+            </p>
+          </Modal.Body>
+
+          <Modal.Footer>
+            <Button onClick={this._handlePhoneValidationCancel}>Cancel</Button>
+            <Button bsStyle='primary' onClick={this._handleSendPhoneVerificationCode}>Verify my phone</Button>
```
landonreed, line 277: Cancel, by convention, is usually on the right (feel free to fact check me).

lib/components/user/notification-prefs-pane.js
```diff
@@ -85,13 +223,60 @@ class NotificationPrefsPane extends Component {
...
+            <p>
+              Please check the SMS messaging app on your mobile phone
+              for a text message with a verification code, and enter the code below.
+            </p>
+            <FormGroup>
+              <ControlLabel>Verification code:</ControlLabel>
+              <FormControl onChange={this._handlePhoneValidationCodeChange} value={phoneValidationCode} />
+            </FormGroup>
+            <p>
+              <Button bsStyle='link' onClick={this._handleRequestPhoneVerificationCode}>Request a new code</Button>
```
landonreed, line 271: Left padding makes this button offset in a weird way.

lib/components/user/notification-prefs-pane.js
```diff
@@ -85,13 +223,60 @@ class NotificationPrefsPane extends Component {
...
+                   (there are some alignment issues with <FormControl.Feedback /> in mobile view). */}
+                <HelpBlock>
+                  {phoneStatusGlyph && <Glyphicon glyph={phoneStatusGlyph} />} {phoneStatusText}
+                </HelpBlock>
+              </FormGroup>
+
+              {shouldVerifyPhone && (
+                <ButtonStrip>
+                  <Button bsStyle='primary' onClick={this._handleRequestPhoneVerificationCode}>Verify my phone</Button>
+                  {initialPhoneNumber.length !== 0 && <Button onClick={this._handleRevertPhoneNumber}>Revert number</Button>}
```
evansiroky, line 248: Revert shouldn't be an option here. There should be no reason to revert a change that hasn't been confirmed.

lib/components/user/user-account-screen.js
```diff
@@ -59,6 +106,81 @@ class UserAccountScreen extends Component {
     this._handleExit()
   }
 
+  /**
+   * Reverts the user phone number and verification state in the database.
+   */
+  _handleRevertPhoneNumber = async () => {
```
evansiroky, line 112: Please remove all revert code. I believe it is not needed for a change that needs confirmation.

evansiroky (changes requested):
Please remove all code for reverting a new phone number that hasn't been confirmed. I don't believe this an appropriate workflow.

Please fix and merge #246 into this PR.

lib/components/user/user-account-screen.js
binh-dam-ibigroup: Updated

lib/components/user/notification-prefs-pane.js
binh-dam-ibigroup: Fixed

lib/components/user/notification-prefs-pane.js
binh-dam-ibigroup: Fixed.

lib/components/user/notification-prefs-pane.js
binh-dam-ibigroup: Removed

lib/components/user/notification-prefs-pane.js
binh-dam-ibigroup: Should the `Save Preferences` button be removed altogether if we are going to update the backend on the spot?

binh-dam-ibigroup:
I'm going to submit this for review. I will get to @evansiroky's remark about merging #246 into this PR, the good thing is the changes to that are mostly independent from the UI elements.

lib/actions/user.js
```diff
@@ -165,21 +199,66 @@ export function createOrUpdateUserMonitoredTrip (tripData, isNew, silentOnSucces
...
+    const sendResult = await secureFetch(requestUrl, accessToken, apiKey, 'POST')
+
+    // If the check is successful, status in the returned data will be "approved".
+    if (sendResult.status === 'success' && sendResult.data) {
+      if (sendResult.data.status === 'approved') {
+        // Refetch user and update application state with new phone number and verification status.
+        // (This also refetches the user's monitored trip, and that's ok.)
+        await dispatch(fetchOrInitializeUser({ accessToken }))
+      } else {
+        alert('You entered in incorrect validation code. Please try again.')
```
evansiroky, line 258: Instead of saying "incorrect" maybe "invalid" would be a better catch-all for cases where the code is correct, but expired.

evansiroky (changes requested):
Has merge conflicts. Otherwise, seems good.

lib/actions/user.js
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/224/commits/708f5b7b8eb3f91aec836eaf0391b1c442a20bbb

lib/components/user/notification-prefs-pane.js
```diff
@@ -218,68 +155,244 @@ class NotificationPrefsPane extends Component {
...
+    const codeErrorState = getErrorState(this.props, 'validationCode')
+
+    return (
+      <>
+        {isEditing
+          ? (
+            <FormGroup validationState={phoneErrorState}>
+              <ControlLabel>Enter your phone number for SMS notifications:</ControlLabel>
+              <ControlStrip>
+                <Field as={InlineTextInput}
```
landonreed, line 331: As mentioned above, we should just use https://catamphetamine.gitlab.io/react-phone-number-input/

lib/components/user/notification-prefs-pane.js
```diff
@@ -218,68 +155,244 @@ class NotificationPrefsPane extends Component {
...
+                  name='newPhoneNumber'
+                  onKeyDown={this._handlePhoneNumberKeyDown}
+                  placeholder={phoneFormatOptions.inputPlaceholder}
+                  type='tel'
+                // onBlur, onChange, and value are passed automatically by Formik
+                />
+                <Button
+                  bsStyle='primary'
+                  disabled={isBlank(newPhoneNumber) || !!errors.newPhoneNumber}
+                  onClick={this._handleRequestCode}
```
landonreed, line 341: I'm seeing very strange behavior if I:
1. Add phone number and successfully verify it.
2. Click Change Number.
3. Enter in my same number as before.
4. Click Send verification text.

Basically, the phone number just disappears and the button goes into a disabled state.

lib/components/user/notification-prefs-pane.js
landonreed: Probably, yes. Its placement right now makes it basically invisible anyways. Let's hold off on removing the bottom button row for now, but we should note this in our list of things to revise for the user account refactor.

lib/components/user/notification-prefs-pane.js
```diff
@@ -218,68 +155,244 @@ class NotificationPrefsPane extends Component {
...
+      isPending: false
+    })
+  }
+
+  _handlePhoneNumberKeyDown = e => {
+    if (e.keyCode === 13) {
+      // On the user pressing enter (keyCode 13) on the phone number field,
+      // prevent form submission and request the code.
+      e.preventDefault()
+      this._handleRequestCode()
```
landonreed, line 239: I'm not sure this (handling Enter keydown) is necessary.

lib/util/ui.js
```diff
@@ -1,5 +1,29 @@
...
+  return !(string && string.length)
+}
+
+/**
+ * Formats a phone number according to the structure and format in phoneFormatOptions.
+ * @param {*} rawNumber The raw number (e.g. +155555555) to process
+ * @param {*} phoneFormatOptions The phoneFormatOptions from the configuration.
+ * @return The formatted number, or the raw number if no region code was provided.
+ */
+export function formatPhoneNumber (rawNumber, phoneFormatOptions) {
```
landonreed, line 18: In mongoDB, I updated the phoneNumber field to be removed and the resulting null value for phoneNumber caused this method to throw an uncaught error.

landonreed (changes requested):
Main thing is that the verification flow still has some unexpected behavior and that we should simplify the phone number form input with https://catamphetamine.gitlab.io/react-phone-number-input/

lib/components/user/notification-prefs-pane.js
binh-dam-ibigroup: I think it is necessary for users who have keyboard "Enter" reflexes.

example-config.yml
```diff
@@ -140,3 +140,17 @@ itinerary:
...
+#  # The placeholder displayed in phone number controls
+#  inputPlaceholder: (555) 555-0123
+#  # Region code for converting between raw E.164 numbers and national format.
+#  # (defaults to 'US')
+#  regionCode: US
+#  # awesome-phonenumber is a little bit too strict on validating formatted numbers
+#  # (although converting to raw E.164 number still works fine in such cases),
+#  # so we provide our own regex for validating user input instead.
+#  #   For US numbers: https://www.oreilly.com/library/view/regular-expressions-cookbook/9781449327453/ch04s02.html
+#  validationRegExp: ^\(?([0-9]{3})\)?[-. ]?([0-9]{3})[-. ]?([0-9]{4})$
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Default values moved to the initial redux state in https://github.com/opentripplanner/otp-react-redux/pull/224/commits/31345eaad3aebbda5a36b0a2baeac1720d5e2751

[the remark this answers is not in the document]
binh-dam-ibigroup: Good suggestion on the input (although event handling is a bit different) - Updated in https://github.com/opentripplanner/otp-react-redux/pull/224/commits/6d2f94a9ab8f2cdb8ce90af749c2fd5e2ce3e026

lib/actions/user.js
```diff
@@ -124,11 +144,12 @@ export function createOrUpdateUser (userData, silentOnSuccess = false) {
...
         // Update application state with the user entry as saved
         // (as returned) by the middleware.
-        const newUserData = result.data
-        dispatch(setCurrentUser({ accessToken, user: newUserData }))
+        dispatch(setCurrentUser({ accessToken, user: result.data }))
       } else {
         alert(`An error was encountered:\n${JSON.stringify(result)}`)
       }
+    } else {
+      alert('Corrupted state: User ID not available for exiting user.')
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Should we change the definition of `isNewUser()` to also include condition `!id`, or should we just display the generic "An error was encountered..." when id is null?

lib/components/user/notification-prefs-pane.js
binh-dam-ibigroup: Let's remove these handlers when removing the Save/Cancel buttons for the account page.

lib/actions/user.js
```diff
@@ -30,67 +25,87 @@ function createNewUser (auth0User) {
...
+    const requestUrl = `${apiBaseUrl}${API_MONITORTRIP_PATH}`
 
-    if (otpMiddleware) {
-      const { data: trips, status: fetchStatus } = await getTrips(otpMiddleware, accessToken)
-      if (fetchStatus === 'success') {
-        dispatch(setCurrentUserMonitoredTrips(trips.data))
-      }
+    const { data: trips, status } = await secureFetch(requestUrl, accessToken, apiKey, 'GET')
+    if (status === 'success') {
+      dispatch(setCurrentUserMonitoredTrips(trips))
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Good catch. That was copy/pasta. Fixed in https://github.com/opentripplanner/otp-react-redux/pull/224/commits/643ae432449eaecd2cf8985bd34542aad1bdeba3.

lib/actions/user.js
```diff
@@ -30,67 +25,87 @@ function createNewUser (auth0User) {
...
+    const { accessToken, loggedInUser } = user
+    const { apiBaseUrl, apiKey } = otpMiddleware
+    return {
+      accessToken,
+      apiBaseUrl,
+      apiKey,
+      loggedInUser
+    }
+  } else {
+    throw new Error('This action requires a valid middleware configuration.')
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/224/commits/643ae432449eaecd2cf8985bd34542aad1bdeba3

lib/actions/user.js
```diff
@@ -103,18 +118,23 @@ export function fetchOrInitializeUser (auth) {
...
-    const { otp_middleware: otpMiddleware = null } = otp.config.persistence
+    const { accessToken, apiBaseUrl, apiKey, loggedInUser } = getMiddlewareVariables(getState())
+    const { id } = userData // Middleware ID, NOT auth0 (or similar) id.
+    let requestUrl, method
 
-    if (otpMiddleware) {
-      const { accessToken, loggedInUser } = user
+    // Determine URL and method to use.
+    if (isNewUser(loggedInUser)) {
+      requestUrl = `${apiBaseUrl}${API_USER_PATH}`
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/224/commits/643ae432449eaecd2cf8985bd34542aad1bdeba3.

lib/util/ui.js
binh-dam-ibigroup: No longer an issue with use of react-phone-number-input.

lib/components/user/notification-prefs-pane.js
binh-dam-ibigroup: Good catch. Fixed in https://github.com/opentripplanner/otp-react-redux/pull/224/commits/2c4705c66277b87b6475ec2f895e22c41963e5e2.

lib/components/user/favorite-locations-pane.js
```diff
@@ -83,30 +83,10 @@ class FavoriteLocationsPane extends Component {
...
-    }
-
-    const effectiveLocations = [
-      homeLocation,
-      workLocation,
-      ...savedLocations.filter(notHomeOrWork)
-    ]
+    const { values: userData } = this.props
+    const { savedLocations } = userData
+    const homeLocation = savedLocations.find(isHome)
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Out of scope for this PR.

lib/actions/user.js
```diff
@@ -167,6 +193,8 @@ export function createOrUpdateUserMonitoredTrip (tripData, isNew, silentOnSucces
       } else {
         alert(`An error was encountered:\n${JSON.stringify(result)}`)
       }
+    } else {
+      alert('Corrupted state: Trip ID not available for exiting trip.')
```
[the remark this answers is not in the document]
binh-dam-ibigroup: I went ahead and removed these corrupted state branches in https://github.com/opentripplanner/otp-react-redux/pull/224/commits/a9dc87ae6cf80fe03b4ab246cf75473ba7484321. I don't think we have encountered the situation when an id "disappears" before, have we?

lib/components/user/notification-prefs-pane.js
```diff
@@ -28,85 +36,393 @@ const allowedNotificationChannels = [
...
+              Please check the SMS messaging app on your mobile phone
+              for a text message with a verification code, and enter the code below.
+            </p>
+            <ControlLabel>Verification code:</ControlLabel>
+            <ControlStrip>
+              <Field as={InlineTextInput}
+                name='validationCode'
+                onKeyDown={this._handleValidationCodeKeyDown}
+                placeholder='123456'
+                // onBlur, onChange, and value are passed automatically by Formik
```
landonreed, line 409: Sorry, I didn't catch this earlier during testing, but this Input should have the props `type='number'` and `maxLength=6`

lib/components/user/notification-prefs-pane.js
```diff
@@ -28,85 +36,393 @@ const allowedNotificationChannels = [
...
-              name='phoneNumber'
-              onBlur={handleBlur}
-              onChange={handleChange}
-              type='tel'
-              value={phoneNumber}
-            />
-            <FormControl.Feedback />
-            {errors.phoneNumber && <HelpBlock>{errors.phoneNumber}</HelpBlock>}
+                Verify
+              </Button>
```
landonreed, line 417: With commit 6052d6f04f4fa72753788a2c49f8d9071ad4c36e on otp-middleware, I get this error when I click verify:

image

lib/components/user/notification-prefs-pane.js
```diff
@@ -28,85 +36,393 @@ const allowedNotificationChannels = [
...
+                  currentNumber && <Button onClick={this._handleCancelEditNumber}>Cancel</Button>}
+                {showPhoneError && <HelpBlock>{INVALID_PHONE_MSG}</HelpBlock>}
+              </ControlStrip>
+            </FormGroup>
+          ) : (
+            <FormGroup>
+              <ControlLabel>SMS notifications will be sent to:</ControlLabel>
+              <ControlStrip>
+                <InlineStatic>
+                  {isPending
```
landonreed, line 383: If I 
1. add a phone number and click Send Verification Text
2. switch to `Don't Notify Me`
3. Switch back to SMS

The phone number and pending status disappears. I think there's an issue with `isPending`. I think its value should be calculated based on the presence of a phone number in the user object and isVerified = false. I'm guessing there are some weird things happening with state management in this component.

lib/components/user/user-account-screen.js
```diff
@@ -72,21 +66,93 @@ function cloneWithHomeAndWorkAsTopLocations (loggedInUser) {
...
+        lastPhoneNumberRequested: numberToVerify,
+        lastPhoneRequestTime: timestamp
+      })
+      await this.props.requestPhoneVerificationSms(numberToVerify)
+    }
+
+    // Return the phone number and state from state.user.loggedInUser.
+    // If the SMS request was successful, the pending phone number and unverified status
+    // will be reflected in the state, otherwise they will be unchanged.
+    const { isPhoneNumberVerified, phoneNumber } = this.props.loggedInUser
```
evansiroky, line 138: This makes me a little uneasy because this method relies on the `this.props` variable changing during the `await` on line 132. This seems to be way too much application logic being put into a component. Ideally, components should do as little application logic as possible and focus solely on rendering and delegating event handling to actions. Furthermore, the `lastPhoneNumberRequested` and `lastPhoneRequestTime` state variables are only used here and not passed along to any components to be rendered. To me this is another indication of variables that belong in the redux store (application state) instead of component state since they have no effect on how a component renders.

lib/components/user/user-account-screen.js
```diff
@@ -72,21 +66,93 @@ function cloneWithHomeAndWorkAsTopLocations (loggedInUser) {
...
+   * Sends the phone verification code.
+   * @returns The effective phone number and verification status.
+   */
+  _handleSendPhoneVerificationCode = async code => {
+    await this.props.verifyPhoneNumber(code)
+
+    // Return the phone number and state from state.user.loggedInUser.
+    // If the verification was successful, isPhoneNumberVerified will be true,
+    // otherwise the phone information will remain unchanged.
+    const { isPhoneNumberVerified, phoneNumber } = this.props.loggedInUser
```
evansiroky, line 152: Another funky example of waiting for application state to change within a component method. This method should all probably be contained within an action.

lib/components/user/notification-prefs-pane.js
binh-dam-ibigroup: That is from a missing `!` (`!isPhoneNumberValidE164(...)` in the backend...

binh-dam-ibigroup: Should be addressed by using https://github.com/ibi-group/otp-middleware/pull/83/commits/6088b315d3e47d419c6f388cca00ad5c51ab942f.

lib/components/user/notification-prefs-pane.js
binh-dam-ibigroup: Good find! Should the verification code control also be visible when the user switches back to SMS?

The Formik state is updated with the pending number in https://github.com/opentripplanner/otp-react-redux/pull/224/commits/aab453065a721588fb90219d5e9d99cbab0fb4d7.

lib/components/user/user-account-screen.js
binh-dam-ibigroup: I agree on moving `lastPhoneNumberRequested/Time` to the redux state. What would be a better way to capture the outcome of the actions? `loggedInUser` is already part of the redux state.

lib/components/user/user-account-screen.js
```diff
@@ -72,21 +66,93 @@ function cloneWithHomeAndWorkAsTopLocations (loggedInUser) {
...
 class UserAccountScreen extends Component {
-  _updateUserPrefs = async userData => {
+  constructor (props) {
+    super(props)
+
+    this.state = {
+      // Capture whether user is a new user at this stage, and retain that value as long as this screen is active.
+      // Reminder: When a new user progresses through the account steps,
+      // isNewUser(loggedInUser) will change to false as the database gets updated.
+      isNewUser: isNewUser(props.loggedInUser),
```
evansiroky, line 76: I get why this is being done, but it just feels hacky to use component state here. Since this value never changes due to some event that gets triggered during the use of the component, it feels like this would be better handled in the application state.

lib/components/user/user-account-screen.js
```diff
@@ -72,21 +66,93 @@ function cloneWithHomeAndWorkAsTopLocations (loggedInUser) {
...
+   * Creating the user record before the user finishes the account creation steps
+   * is required by the middleware in order to perform phone verification.
+   *
+   * @param {*} userData The user data state to persist.
+   * @returns The new user id the the caller can use.
+   */
+  _handleCreateNewUser = async userData => {
+    await this._updateUserPrefs(userData, true)
+
+    return this.props.loggedInUser.id
```
evansiroky, line 106: Another pattern of awaiting a change in application state. Move this to an action.

lib/components/user/new-account-wizard.js
```diff
@@ -1,66 +1,76 @@
...
-    },
-    finish: {
-      pane: panes.finish,
-      prevId: 'places',
-      props,
-      title: 'Account setup complete!'
-    }
+    // After user is initially persisted and reloaded to the redux state,
+    // update the 'id' in the Formik state.
+    setFieldValue('id', newId)
```
evansiroky, line 20: This function is probably unnecessary and is probably better handled in the application state. It appears that it's purpose is to update a hidden field to associate further saves of the user with it's ID once it is created. I'm not sure how this occurs, but wouldn't Formik already get updated with the user's ID via the `initialValues` prop that is sent the `loggedInUser` in the `user-account-screen#render` method?

lib/components/user/notification-prefs-pane.js
binh-dam-ibigroup: On Chrome, Firefox and Safari on Mac, the browser enforces `maxLength` on `<input type='text'>` but not on `<input type='number'>.

lib/components/user/new-account-wizard.js
binh-dam-ibigroup: I thought Formik would get updated too, however it is not the case as seen with https://github.com/opentripplanner/otp-react-redux/pull/224#discussion_r510429924.

lib/components/user/user-account-screen.js
```diff
@@ -131,28 +195,30 @@ class UserAccountScreen extends Component {
             // can be wired to be managed by Formik.
             props => {
```
evansiroky, line 196: Can you rename this to be `formikProps`?

lib/components/user/notification-prefs-pane.js
```diff
@@ -28,85 +36,393 @@ const allowedNotificationChannels = [
...
+      this._handleCancelEditNumber()
+    } else {
+      // Empty the input box (and disable the Send text button).
+      this.setState({
+        newPhoneNumber: ''
+      })
+
+      const { isPhoneNumberVerified, phoneNumber } = await onRequestCode(newPhoneNumber)
+
+      if (phoneNumber === newPhoneNumber && !isPhoneNumberVerified) {
```
evansiroky, line 283: This again has a lot of code that waits for application state to change and relies on the component's props to become updated. This is somewhat confusing and I think would be better handled by splitting this logic in 2. Move the code requesting a code into an action. For the rest of this code, create a `componentDidUpdate` method where you can read in the newly updated props and call `setFieldValue` and `setFieldTouched` appropriately.

lib/components/user/notification-prefs-pane.js
```diff
@@ -28,85 +36,393 @@ const allowedNotificationChannels = [
...
+          newPhoneNumber: phoneNumber
+        })
+
+        // Set empty code field in Formik's state and prompt for code.
+        setFieldValue('validationCode', '')
+        setFieldTouched('validationCode', false)
+        this.setState({
+          isEditing: false,
+          isPending: true
+        })
```
evansiroky, line 295: Since this component state update relies on application state updates, I think it might make sense to move these state variables into the application state.

lib/components/user/notification-prefs-pane.js
```diff
@@ -28,85 +36,393 @@ const allowedNotificationChannels = [
...
+      // For new users, render component in editing state.
+      isEditing: isBlank(initialPhoneNumber),
+
+      // If true, a phone verification request has been sent and the UI is awaiting the user sending the code.
+      isPending: false,
+
+      // Holds the new phone number (+15555550123 format) entered by the user
+      // (outside of Formik because (Phone)Input does not have a standard onChange event or simple valitity test).
+      newPhoneNumber: ''
+    }
```
evansiroky, line 223: Pretty much all of these component state variables are changed when application state changes. Therefore, they should probably be application state variables instead.

lib/components/user/new-account-wizard.js
evansiroky: Seems like this is what should be used to overcome this: https://formik.org/docs/api/withFormik#mappropstovalues-props-props--values

lib/components/user/notification-prefs-pane.js
```diff
@@ -28,85 +36,393 @@ const allowedNotificationChannels = [
...
+
+  _handleSubmitCode = async () => {
+    const { onSubmitCode, onVerified, setFieldValue, values } = this.props
+    const { validationCode } = values
+
+    // Clear the code field and disable the verify button.
+    // (The code to send is already captured in const above.)
+    setFieldValue('validationCode', '')
+
+    const result = await onSubmitCode(validationCode)
```
evansiroky, line 314: Similar comment as in `_handleRequestCode`. Everything after this appears to be the result of application state changes. It might make sense to end this method here and handle application state in the render methods or via https://formik.org/docs/api/withFormik#mappropstovalues-props-props--values.

evansiroky (changes requested):
It seems pretty clear that there is a lot of mixing of component state with application state. I believe this could use a refactor to use [Formik's mappropstovalues](https://formik.org/docs/api/withFormik#mappropstovalues-props-props--values) in order to keep application state and component rendering logic separate.

lib/components/user/user-account-screen.js
binh-dam-ibigroup: last phone number states move to redux state in https://github.com/opentripplanner/otp-react-redux/pull/224/commits/bc43d29931e3907bca16067f26d718075ad0e6e3.

lib/components/user/new-account-wizard.js
binh-dam-ibigroup: Updated (a bit differently) in https://github.com/opentripplanner/otp-react-redux/pull/224/commits/3c4802751a090c2ad4e95a887b7497476a12ba22.

lib/components/user/user-account-screen.js
binh-dam-ibigroup: Removed in https://github.com/opentripplanner/otp-react-redux/pull/224/commits/3c4802751a090c2ad4e95a887b7497476a12ba22.

lib/components/user/user-account-screen.js
binh-dam-ibigroup: Hmm how about placing that in a class member instead? `this.isNewUser = isNewUser(...)`

lib/components/user/user-account-screen.js
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/224/commits/b1d3d7c06753a056c14bdee4c5289b222462a7e6.

lib/components/user/notification-prefs-pane.js
binh-dam-ibigroup: I ended up setting the notification channel as SMS in https://github.com/ibi-group/otp-middleware/pull/83/commits/8b11107252b3b93d710eca6913d241a47456bb4a as we are now relying on the redux state fetched from the db.

lib/components/user/user-account-screen.js
evansiroky: That's a little better, but still feel like it could be more straightforward for all of the state data could be handled within the application state somehow.

lib/components/user/notification-prefs-pane.js
```diff
@@ -85,28 +146,236 @@ const NotificationPrefsPane = ({
...
+   * Send phone verification request with the entered values.
+   */
+  _handleRequestCode = () => {
+    const { initialPhoneNumber, initialPhoneNumberVerified, onRequestCode } = this.props
+    const { newPhoneNumber } = this.state
+
+    this._handleCancelEditNumber()
+
+    // Send the sms request, unless the user
+    // entered the same phone number as their current verified number.
```
evansiroky, line 250: These comments could have better line breaks.

lib/actions/user.js
```diff
@@ -134,56 +158,116 @@ export function createOrUpdateUser (userData) {
...
+export function requestPhoneVerificationSms (newPhoneNumber) {
+  return async function (dispatch, getState) {
+    const state = getState()
+    const { number, timestamp } = state.user.lastPhoneSmsRequest
+    const now = new Date()
+
+    // Request a new verification code if we are requesting a different number.
+    // or enough time has ellapsed since the last request (1 minute?).
+    // TODO: Should throttling be handled in the middleware?
+    if (number !== newPhoneNumber || timestamp + 60000 >= now) {
```
evansiroky, line 226: If it should be, then there should be a way to communicate to a user that they are being throttled. Perhaps it might be easier to handle this in the backend. As it is, there is this hardcoded 60000-millisecond throttle that might need to be updated for each deployment.

lib/actions/user.js
```diff
@@ -1,19 +1,17 @@
...
 import { isNewUser } from '../util/user'
 
+// Middleware API paths.
+const API_MONITORTRIP_PATH = '/api/secure/monitoredtrip'
+const API_OTPUSER_PATH = '/api/secure/user'
+const API_OTPUSER_VERIFYSMS_PATH = '/verify_sms'
+
 const setCurrentUser = createAction('SET_CURRENT_USER')
 const setCurrentUserMonitoredTrips = createAction('SET_CURRENT_USER_MONITORED_TRIPS')
+const setLastPhoneSmsRequestInfo = createAction('SET_LAST_PHONE_SMS_REQUEST_INFO')
```
evansiroky, line 14: This action is never handled by a reducer. Is it needed?

lib/actions/user.js
```diff
@@ -134,56 +158,116 @@ export function createOrUpdateUser (userData) {
...
 
-      if (deleteResult.status === 'success') {
-        // Reload user's monitored trips after deletion.
-        await dispatch(fetchUserMonitoredTrips(accessToken))
+    // If the check is successful, status in the returned data will be "approved".
+    if (status === 'success' && data) {
+      if (data.status === 'approved') {
+        // Refetch user and update application state with new phone number and verification status.
+        // (This also refetches the user's monitored trip, and that's ok.)
+        await dispatch(fetchOrInitializeUser({ accessToken }))
```
evansiroky, line 264: Unnecessary `await`

lib/actions/user.js
```diff
@@ -134,56 +158,116 @@ export function createOrUpdateUser (userData) {
...
       } else {
-        alert(`An error was encountered:\n${JSON.stringify(deleteResult)}`)
+        // 'invalid' captures cases when users enter wrong/incorrect codes or expired codes.
+        alert('You entered in invalid code. Please try again.')
```
evansiroky, line 267: Is there a way to distinguish whether the code has expired or if it is just invalid? I don't feel very strongly about it either way, but it'd be nice to communicate invalid codes due to expiration.

lib/components/user/notification-prefs-pane.js
```diff
@@ -85,28 +146,236 @@ const NotificationPrefsPane = ({
...
+                  {formatPhoneNumber(initialPhoneNumber)}
+                  {' '}
+                  {isPending
+                    // eslint-disable-next-line jsx-a11y/label-has-for
+                    ? <Label bsStyle='warning'>Pending</Label>
+                    // eslint-disable-next-line jsx-a11y/label-has-for
+                    : <Label bsStyle='success'>Verified</Label>
+                  }
+                </InlineStatic>
+                <Button onClick={this._handleEditNumber}>Change number</Button>
```
evansiroky, line 341: There is no way to delete a phone number. Not sure if needed since a user could switch to either email or do not notify.

evansiroky (approved):
State management looks much better now. I have a few comments on probably nice-to-have items, but this seems good to go as-is.

lib/components/user/notification-prefs-pane.js
binh-dam-ibigroup: all except `isEditing` were removed by https://github.com/opentripplanner/otp-react-redux/pull/224/commits/bc43d29931e3907bca16067f26d718075ad0e6e3.

lib/components/user/notification-prefs-pane.js
binh-dam-ibigroup: `isPending` was removed in https://github.com/opentripplanner/otp-react-redux/pull/224/commits/bc43d29931e3907bca16067f26d718075ad0e6e3.

lib/actions/user.js
binh-dam-ibigroup: That action is used here: https://github.com/opentripplanner/otp-react-redux/blob/b1d3d7c06753a056c14bdee4c5289b222462a7e6/lib/actions/user.js#L232. Is there something else I missed?

lib/actions/user.js
binh-dam-ibigroup: Basically when the code expires, the backend returns a 500 error status, so that could be handled in the else block that deals with generic errors.

lib/components/user/notification-prefs-pane.js
binh-dam-ibigroup: Putting this in a separate issue: Setting the input type to 'number' causes Formik to convert the value of the input control to a number. This means strings with leading zeros such as `001234` are deemed invalid. If we get validation codes that start with a zero this will be an issue. (I have not received validation codes that start with zero so far.)

binh-dam-ibigroup: Correction. Twilio does send validation codes that start with zero.

lib/actions/user.js
binh-dam-ibigroup: Addressed in https://github.com/opentripplanner/otp-react-redux/pull/224/commits/cbf2d86940c5b4c51d9eb03969c22926c1a2ba7b.

lib/actions/user.js
binh-dam-ibigroup: Removed in https://github.com/opentripplanner/otp-react-redux/pull/224/commits/cbf2d86940c5b4c51d9eb03969c22926c1a2ba7b

lib/actions/user.js
binh-dam-ibigroup: For now I have added displaying an alert in https://github.com/opentripplanner/otp-react-redux/pull/224/commits/cbf2d86940c5b4c51d9eb03969c22926c1a2ba7b.

lib/actions/user.js
binh-dam-ibigroup: Oh, I totally forgot about the actual reducer file...

binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/224/commits/705919a63981a453c4eceb10efd5211c547d19c1.

lib/components/user/notification-prefs-pane.js
binh-dam-ibigroup: Addressed in https://github.com/opentripplanner/otp-react-redux/pull/224/commits/cbf2d86940c5b4c51d9eb03969c22926c1a2ba7b.

lib/components/user/notification-prefs-pane.js
binh-dam-ibigroup: In light of the number handling above, I changed the input type back to `text` in https://github.com/opentripplanner/otp-react-redux/pull/224/commits/cbf2d86940c5b4c51d9eb03969c22926c1a2ba7b, which allows native enforcement of `maxLength`.

landonreed: @binh-dam-ibigroup, if type="number" doesn't work, let's just use `pattern="[0-9]*"` (https://stackoverflow.com/a/9318515/915811)

binh-dam-ibigroup: The regex check is already in place at https://github.com/opentripplanner/otp-react-redux/blob/f40f4a363612c3e635c7ec593fb647eaaf8e19eb/lib/components/user/notification-prefs-pane.js#L86.

I really like to have the numeric keypad on mobile devices too. To do that, we could "cheat" and use `<input type='tel'>`, although the validation code is not really a phone number.

landonreed: Yea, if `tel` seems to work, let's try it.

lib/actions/user.js
```diff
@@ -134,56 +158,122 @@ export function createOrUpdateUser (userData) {
...
+        // (This also refetches the user's monitored trip, and that's ok.)
+        await dispatch(fetchOrInitializeUser({ accessToken }))
       } else {
-        alert(`An error was encountered:\n${JSON.stringify(result)}`)
+        alert(`An error was encountered:\n${JSON.stringify(message)}`)
       }
+    } else {
+      // Alert user if they have been throttled.
+      // TODO: improve this alert.
+      alert('A verification SMS was sent to the indicated phone number less than a minute ago. Please try again in a few moments.')
```
landonreed, line 248: I'm getting this message even when the last code was requested over a minute ago.

binh-dam-ibigroup: There was a logic error in the delay check, https://github.com/opentripplanner/otp-react-redux/pull/224/commits/40871d2ad441eabf66785ccb1cd39548b65e7c36 fixes it.

New issue for server-side throttling: https://github.com/ibi-group/otp-middleware/issues/88.

binh-dam-ibigroup: I added the expiration text in https://github.com/opentripplanner/otp-react-redux/pull/224/commits/738e9cf9cb5e8dcc09baf147f4249a49046b562f.

landonreed (approved):
Changes look good. This is ready to merge.

PR #227, opened 2020-09-08
Add link to resend verification email.

binh-dam-ibigroup:
This PR adds a link to resend an email verification message hooked to the proper OTP middleware endpoint.

To test, use the backend code from https://github.com/ibi-group/otp-middleware/pull/95.

image


binh-dam-ibigroup:
Blocking because https://github.com/ibi-group/otp-middleware/issues/71.

lib/components/user/verify-email-screen.js
```diff
@@ -10,25 +16,57 @@ const _handleClick = () => window.location.reload()
...
+          to verify your email address before finishing your account setup.
+        </DivSpacer>
+        <p>
+          Once you're verified, click the button below to continue.
+        </p>
+        <DivSpacer>
+          <Button
+            bsSize='large'
+            bsStyle='primary'
+            onClick={this._handleClick}
```
evansiroky, line 42: Since there are now 2 buttons, can this handler be renamed to something more specific?

evansiroky (approved):
Approved assuming the 1 small change I requested is implemented.

lib/components/user/verify-email-screen.js
binh-dam-ibigroup: Good catch. Updated in https://github.com/opentripplanner/otp-react-redux/pull/227/commits/7152d277edd003953ccc3fb8d5921ebfd9f93b67.

lib/actions/user.js
```diff
@@ -152,6 +152,25 @@ export function createOrUpdateUser (userData, silentOnSuccess = false) {
   }
 }
 
+/**
+ * Requests the verification email for the new user to be resent.
+ */
+export function resendVerificationEmail (auth0) {
+  return async function (dispatch, getState) {
+    const { apiBaseUrl, apiKey } = getMiddlewareVariables(getState())
+    const accessToken = await auth0.getAccessTokenSilently()
```
landonreed, line 161: Isn't accessToken stored in the reducer now? See line 123 above.

lib/components/user/verify-email-screen.js
```diff
@@ -10,25 +16,57 @@ const _handleClick = () => window.location.reload()
...
+  _handleEmailVerified = () => window.location.reload()
+
+  render () {
+    return (
+      <div>
+        <h1>Verify your email address</h1>
+        <DivSpacer>
+          Please check your email inbox and follow the link in the message
+          to verify your email address before finishing your account setup.
+        </DivSpacer>
```
landonreed, line 34: I'd prefer to use a `p` tag here.

landonreed (changes requested):
Looks good, but can we grab accessToken similar to the other methods?

lib/actions/user.js
binh-dam-ibigroup: `accessToken` is not stored yet because it is set that the same time as the user, and the user didn't get set yet. Should that get split?

landonreed: Hmm, I'm not sure. Perhaps we should split the set access token functionality out from set logged in user. What do you think about having an additional check in `UserLoaderScreen` that checks if `accessToken` is available from the store? If not, it can just call `auth0.getAccessTokenSilently()` and store it. That would allow us to then fetch or initialize user without passing `auth0`.

```javascript
// UserLoaderScreen
componentDidUpdate () {
    const { accessToken, auth0, fetchOrInitializeUser, setAccessToken } = this.props
    if (!accessToken) {
      let token = await auth0.getAccessTokenSilently()
      setAccessToken({accessToken: token, auth0User: auth0.user})
    }
    if (accessToken && this.loggedInUserIsUnfetched()) {
      fetchOrInitializeUser() // can now get accessToken and auth0User from store
    }
}
```

binh-dam-ibigroup: @landonreed let me think what you think of this: https://github.com/opentripplanner/otp-react-redux/pull/227/commits/b9b0ba35ab20c854ca8c40b9229c79ab2b85f092. We still need the `auth0` argument to initialize a default user (needs the auth0 id and email).

lib/actions/user.js
```diff
@@ -49,65 +49,63 @@ function getMiddlewareVariables (state) {
...
+
 /**
  * Attempts to fetch user preferences (or set initial values if the user is being created)
  * into the redux state, under state.user.
- * @param auth0 If provided, the auth0 login object used to initially obtain the auth0 access token
- *   for subsequent middleware fetches (and also to initialize new users from auth0 email and id).
- *   If absent, state.user.accessToken will be used for fetches.
+ * @param auth0 If provided, the auth0 login object used to initialize the default user object (with email and auth0 id).
  */
 export function fetchOrInitializeUser (auth0) {
```
landonreed, line 74: I think this arg should be changed to `auth0User`.

lib/actions/user.js
landonreed: That looks good to me. But `resendVerificationEmail` no longer needs the auth0 param now, right?

landonreed (approved):
Looks good, conditional approval after addressing a couple of final comments.

lib/actions/user.js
binh-dam-ibigroup: How about `auth0Context`? Because the actual auth0 user is in `auth0.user`, which is passed to method `createNewUser`.

binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/227/commits/77a1a2c0f422a76331c75fbaea09579946a51692 to `auth0User`.

lib/actions/user.js
binh-dam-ibigroup: Good catch. Updated in https://github.com/opentripplanner/otp-react-redux/pull/227/commits/77a1a2c0f422a76331c75fbaea09579946a51692.

PR #229, opened 2020-09-10
Fix elevation marker on map

### LINKED ISSUE #228: Elevation marker not showing on map
Post-otp-ui refactor, the elevation marker does not show on the map when hovering over the elevation leg diagram for the example otp-rr. Also, `showElevationProfile` in the connected itin body does not read in the config value:

https://github.com/opentripplanner/otp-react-redux/blob/6f7aebe65023c5e5012d3d217d29088a30bcf661/lib/components/narrative/line-itin/connected-itinerary-body.js#L60-L64

landonreed:
This fixes #228 by correctly reading in the elevationProfile config setting.

It also renames `showLegDiagram` to `setLegDiagram` to match the prop name in otp-ui.

Finally, it replaces its own instance of `LegDiagramPreview` with the one imported from otp-ui.

### DIFF
--- a/lib/actions/map.js
+++ b/lib/actions/map.js
@@ -109,7 +109,7 @@ export function switchLocations () {
   }
 }
 
-export const showLegDiagram = createAction('SHOW_LEG_DIAGRAM')
+export const setLegDiagram = createAction('SET_LEG_DIAGRAM')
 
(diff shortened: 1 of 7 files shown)


binh-dam-ibigroup (approved):
LGTM

evansiroky (approved):

PR #231, opened 2020-09-15
Add trip editing commands to saved trips list

binh-dam-ibigroup:
This PR makes an incremental improvement to the monitored trip management flow
by adding the following trip commands for each trip in the saved trips list:
- Monitor/Pause: Pauses or resume monitoring the trip.
- Edit: Edits the trip.
- Delete: Deletes the trip. When editing individual trips, the delete button is no longer available.

Placement and appearance of these buttons is up for discussion, let me know.

image

### DIFF
--- a/lib/actions/user.js
+++ b/lib/actions/user.js
@@ -128,10 +128,10 @@ export function createOrUpdateUser (userData) {
 
 /**
  * Updates a logged-in user's monitored trip,
- * then, if that was successful, refreshes the redux monitoredTrips
- * with the updated trip.
+ * then, if that was successful, alerts (optional)
+ * and refreshes the redux monitoredTrips with the updated trip.
  */
-export function createOrUpdateUserMonitoredTrip (tripData, isNew) {
+export function createOrUpdateUserMonitoredTrip (tripData, isNew, silentOnSuccess) {
   return async function (dispatch, getState) {
     const { otp, user } = getState()
     const { otp_middleware: otpMiddleware = null } = otp.config.persistence
@@ -148,7 +148,9 @@ export function createOrUpdateUserMonitoredTrip (tripData, isNew) {
 
       // TODO: improve the UI feedback messages for this.
       if (result.status === 'success' && result.data) {
-        alert('Your preferences have been saved.')
+        if (!silentOnSuccess) {
+          alert('Your preferences have been saved.')
+        }
 
         // Reload user's monitored trips after add/update.
         await dispatch(fetchUserMonitoredTrips(accessToken))
--- a/lib/components/user/saved-trip-list.js
+++ b/lib/components/user/saved-trip-list.js
@@ -1,9 +1,11 @@
+import clone from 'clone'
 import React, { Component } from 'react'
-import { Button, ButtonGroup } from 'react-bootstrap'
+import { Button, ButtonGroup, Panel } from 'react-bootstrap'
 import { connect } from 'react-redux'
 import { withLoginRequired } from 'use-auth0-hooks'
 
 import * as uiActions from '../../actions/ui'
+import * as userActions from '../../actions/user'
 import DesktopNav from '../app/desktop-nav'
 import LinkButton from './link-button'
 import TripSummaryPane from './trip-summary-pane'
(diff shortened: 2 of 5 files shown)


evansiroky (approved):
Good to have this functionality for now. We can worry about the UX of it later.

lib/actions/user.js
```diff
@@ -148,7 +148,9 @@ export function createOrUpdateUserMonitoredTrip (tripData, isNew) {
 
       // TODO: improve the UI feedback messages for this.
       if (result.status === 'success' && result.data) {
-        alert('Your preferences have been saved.')
+        if (!silentOnSuccess) {
+          alert('Your preferences have been saved.')
+        }
```
landonreed, line 153: Why is the alert even part of the design here? To me its earlier inclusion felt like the result of early development, but I don't think it should be formalized in the code. Let's just rip the alert out.

lib/components/user/saved-trip-list.js
```diff
@@ -17,11 +19,32 @@ class SavedTripList extends Component {
...
+  /**
+   * Pauses or resumes the specified trip.
+   */
+  _handlePauseOrResumeMonitoring = trip => async () => {
+    const newTrip = clone(trip)
+    newTrip.isActive = !newTrip.isActive
+
+    // Silent update of existing trip.
+    await this.props.createOrUpdateUserMonitoredTrip(newTrip, false, true)
+  }
```
landonreed, line 36: This does not need to be async. There's no statement after the `await`.

lib/components/user/saved-trip-list.js
```diff
@@ -44,13 +67,24 @@ class SavedTripList extends Component {
           {accountLink}
           <h1>My saved trips</h1>
           <p>Click on a saved trip below to modify it.</p>
```
landonreed, line 69: Is this text still needed?

lib/components/user/saved-trip-list.js
```diff
@@ -44,13 +67,24 @@ class SavedTripList extends Component {
...
-          <ButtonGroup vertical block>
-            {trips.map((trip, index) => (
-              <Button key={index} onClick={this._handleTripSelect(trip)} style={{textAlign: 'left'}}>
+
+          {trips.map((trip, index) => (
+            <Panel key={index}>
```
landonreed, line 72: I prefer the trip Panel to be its own component (doesn't necessarily need its own file).

landonreed (changes requested):

lib/actions/user.js
binh-dam-ibigroup: Moved the discussion to https://github.com/opentripplanner/otp-react-redux/issues/245

lib/components/user/saved-trip-list.js
binh-dam-ibigroup: Addressed in https://github.com/opentripplanner/otp-react-redux/pull/231/commits/19709bed5ca9d77f7915003870a9f4c233ff33ca

lib/components/user/saved-trip-list.js
```diff
@@ -44,13 +67,24 @@ class SavedTripList extends Component {
...
+                <Panel.Title componentClass='h3'>{trip.tripName}</Panel.Title>
+              </Panel.Heading>
+              <Panel.Body>
                 <TripSummaryPane monitoredTrip={trip} />
-              </Button>
-            ))}
-          </ButtonGroup>
+                <ButtonGroup>
+                  <Button bsSize='small' onClick={this._handlePauseOrResumeMonitoring(trip)}>
+                    {trip.isActive ? 'Pause' : 'Monitor' /* TODO: Find better word for 'Monitor' */ }
```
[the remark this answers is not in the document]
binh-dam-ibigroup: I have added play/pause icons in https://github.com/opentripplanner/otp-react-redux/pull/231/commits/19709bed5ca9d77f7915003870a9f4c233ff33ca... but would it be better to use the bell icon instead?

lib/components/user/saved-trip-list.js
```diff
@@ -1,72 +1,130 @@
...
+        <Panel.Heading>
+          <Panel.Title componentClass='h3'>{trip.tripName}</Panel.Title>
+        </Panel.Heading>
+        <Panel.Body>
+          <TripSummaryPane monitoredTrip={trip} />
+          <ButtonGroup>
+            <Button bsSize='small' onClick={this._handlePauseOrResumeMonitoring}>
+              {trip.isActive
+                ? <><Glyphicon glyph='pause' /> Pause notifications</>
+                : <><Glyphicon glyph='play' /> Resume notifications</>
```
landonreed, line 63: Remove "notifications" from each.

lib/components/user/saved-trip-list.js
```diff
@@ -12,95 +12,119 @@ import TripSummaryPane from './trip-summary-pane'
...
+  deleteUserMonitoredTrip: userActions.deleteUserMonitoredTrip,
+  routeTo: uiActions.routeTo
+}
+
+const ConnectedTripListItem = connect(itemMapStateToProps, itemMapDispatchToProps)(TripListItem)
+
+/**
+ * This component displays the list of saved trips for the logged-in user.
+ */
+const SavedTripList = ({ trips }) => {
```
landonreed, line 93: I think the code is more readable with the component for which the file is named is at the top of the file.

landonreed (approved):
Looks good just a couple of conditions for approval.

lib/components/user/saved-trip-list.js
binh-dam-ibigroup: If the buttons just say "Pause" or "Resume", could someone read the buttons as "Pause trip" or "Resume trip"?

binh-dam-ibigroup: Maybe with this layout it's ok as the buttons are just underneath the "Notifications" label.

lib/components/user/saved-trip-list.js
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/231/commits/43d545a1dda5f9308667004d7562119533a3229c

PR #232, opened 2020-09-17
Do not close trip settings screen on error saving trip.

binh-dam-ibigroup:
If there is an error while saving a new or existing trip, the trip settings screen will close after the user dismisses the error message. I find this annoying, and this PR prevents the trip settings screen from closing in such situations. In some cases, you want the user to be able to make changes and let them attempt to change again.

To test:
- Use https://github.com/ibi-group/otp-middleware/pull/75 in its original form.
- Try to save a new trip. An error message will be displayed because of a bug, and the saved trip screen will closed automatically.


evansiroky (approved):

lib/components/user/saved-trip-screen.js
```diff
@@ -102,12 +103,35 @@ class SavedTripScreen extends Component {
...
+    const { activeSearchId } = this.props
+
+    const savedTrip = this.props.monitoredTrips.find(trip => {
+      const tripQueryParams = qs.parse(trip.queryParams.split('?')[1])
+      return tripQueryParams.ui_activeSearch === activeSearchId
+    })
+
+    if (savedTrip) {
+      this._goToTripPlanner()
+    }
```
landonreed, line 119: I think I mentioned in a recent PR that we probably shouldn't go to the trip planner after saving the trip, but rather the list of saved trips. I can't recall whether that was already merged/implemented.

lib/components/user/saved-trip-screen.js
```diff
@@ -102,12 +103,35 @@ class SavedTripScreen extends Component {
...
+
+    // For new trips, if a trip with the same ui_activeSearch value in queryParams
+    // appears in the monitored trips redux state,
+    // then we know that save was successful and we can proceed.
+    const { activeSearchId } = this.props
+
+    const savedTrip = this.props.monitoredTrips.find(trip => {
+      const tripQueryParams = qs.parse(trip.queryParams.split('?')[1])
+      return tripQueryParams.ui_activeSearch === activeSearchId
+    })
```
landonreed, line 115: This approach seems flimsy. I would prefer that this logic be contained within the action -- that would remove all of this inference that is being applied here. I think the routing logic for these handlers (goToTripPlanner, goToSavedTrips, etc.) should also be contained within those actions.

lib/components/user/saved-trip-screen.js
```diff
@@ -102,12 +103,35 @@ class SavedTripScreen extends Component {
...
+
     await this._updateMonitoredTrip()
-    this._goToSavedTrips()
+
+    // For existing trips, if the lastUpdated value (long) for the trip with the same id in monitored trips redux state
+    // is more recent that our copy, then we know that save was successful and we can proceed.
+    // (The state is not updated upon save because the next normal step is to close this screen.)
+    const savedTrip = this.props.monitoredTrips.find(trip => trip.id === monitoredTrip.id)
+
+    if (savedTrip.lastUpdated > monitoredTrip.lastUpdated) {
```
landonreed, line 132: See above comment about moving this check (and probably the need for comparing lastUpdated) to the action itself.

lib/components/user/saved-trip-screen.js
```diff
@@ -215,6 +239,7 @@ const mapStateToProps = (state, ownProps) => {
   const activeItinerary = activeSearch && activeSearch.activeItinerary
   const itineraries = getActiveItineraries(state.otp) || []
   return {
+    activeSearchId: state.otp.activeSearchId,
```
landonreed, line 242: Unnecessary if we move logic to actions.

landonreed (changes requested):
I think we ought to move this logic to the actions. That would remove the need for the somewhat flimsy checks and instead just check the result of the fetch.

lib/components/user/saved-trip-screen.js
binh-dam-ibigroup: This was handled in #239 (see this file: https://github.com/opentripplanner/otp-react-redux/blob/cf194ec2d2117caf7d60b47fadc7724c606d5800/lib/components/user/saved-trip-screen.js#L99)

lib/components/user/saved-trip-screen.js
binh-dam-ibigroup: Good call. It's cleaner now in https://github.com/opentripplanner/otp-react-redux/pull/232/commits/861f5e8d951548dac61c07d5083f2b60682f604a.

lib/components/user/saved-trip-screen.js
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/232/commits/861f5e8d951548dac61c07d5083f2b60682f604a. Previous logic is no longer necessary.

lib/components/user/saved-trip-screen.js
```diff
@@ -93,49 +94,12 @@ class SavedTripScreen extends Component {
...
-    // is more recent that our copy, then we know that save was successful and we can proceed.
-    // (The state is not updated upon save because the next normal step is to close this screen.)
-    const savedTrip = this.props.monitoredTrips.find(trip => trip.id === monitoredTrip.id)
-
-    if (savedTrip.lastUpdated > monitoredTrip.lastUpdated) {
-      this._goToSavedTrips()
-    }
+  _handleSaveTrip = async monitoredTrip => {
+    await this._updateMonitoredTrip(monitoredTrip)
   }
```
landonreed, line 103: `_handleSaveTrip` does not need to exist (just use `_updateMonitoredTrip`). Also, `_updateMonitoredTrip` does not need to be async (no need to await `createOrUpdateUserMonitoredTrip`).

landonreed (approved):
Wow, that's a lot simpler. Nice work. Just one final condition for approval (see comment).

lib/components/user/saved-trip-screen.js
binh-dam-ibigroup: Good catch. Updated in https://github.com/opentripplanner/otp-react-redux/pull/232/commits/3fd0a0ab2a63ee98522c108e5e31ae031777b83c along with missing import in `actions/user`.

PR #233, opened 2020-09-18
Use auth0-react

### LINKED ISSUE #253: If logging out on local dev environment, does not redirect to localhost:9966
1. Sign in.
2. Log out.
3. App seems to redirect to URL listed first in Allowed Logout URLs in Auth0 console.

This may be fixed in #233 (I have not tested yet)?

binh-dam-ibigroup:
This PR replaces the no-longer-maintained `use-auth0-hooks` library with its successor `auth0-react`.

This PR also improves the following:

- Logout redirects to the same server (previously, you could be redirected from e.g. localhost to a live instance).
- Login from a protected page/route is now supported (e.g. logging-in from /#/account shows the account page instead of the map)

Codewise:

- Auth0-react renames some props and higher-order components.
- Auth0-react requires we fetch Auth0 tokens ourselves.
- Auth0-react inserts some URL parameters, the PR adds code to remove those parameters.
- `appState.returnTo` is used where possible (fixes #253), method `processSignIn` is simplified.
- ~Constant `URL_ROOT` is replaced with the standard `window.location.origin`.~ (unchanged, see https://github.com/opentripplanner/otp-react-redux/pull/233#discussion_r492827070)
- Improved passing props in `withLoggedInUserSupport` higher order component (removed `renderChildrenWithProps` util method.
- `state.user.loggedInUserMonitoredTrips` now asynchronously fetched after fetching `state.user.loggedInUser`.


lib/components/app/responsive-webapp.js
```diff
@@ -291,7 +291,7 @@ class RouterWrapperWithAuth0 extends Component {
             onLoginError={showLoginError}
             onRedirectCallback={processSignIn}
             onRedirecting={BeforeSignInScreen}
-            redirectUri={URL_ROOT}
+            redirectUri={window.location.origin}
```
landonreed, line 294: I think it's probably ok to remove URL_ROOT throughout the app, but this reminds me that we should document somewhere that user accounts/middleware is not currently compatible with hosting the OTP app at a path, e.g., https://newplanner.trimet.org/map. If we're not OK with that limitation we should probably fix it. (I think the only major concern is probably with the TriMet implementation.)

lib/components/user/saved-trip-screen.js
```diff
@@ -229,6 +229,12 @@ const mapDispatchToProps = {
 }
 
 export default withLoggedInUserSupport(
-  withLoginRequired(connect(mapStateToProps, mapDispatchToProps)(SavedTripScreen)),
+  withAuthenticationRequired(
+    connect(mapStateToProps, mapDispatchToProps)(SavedTripScreen),
+    {
+      // Redirect back to this page if a user logs in from the URL for saving a trip.
+      returnTo: () => window.location.hash.substr(1)
+    }
```
landonreed, line 237: This object with `returnTo` is used multiple times. It should probably be encapsulated somewhere for reuse.

lib/components/user/with-logged-in-user-support.js
```diff
@@ -37,26 +37,32 @@ export default function withLoggedInUserSupport (WrappedComponent, requireLogged
...
 class UserLoaderScreen extends Component {
-  componentDidUpdate () {
-    const { auth, fetchOrInitializeUser, loggedInUser } = this.props
+  async componentDidUpdate () {
+    const { auth0, fetchOrInitializeUser, loggedInUser } = this.props
 
-    // Once accessToken is available, proceed to fetch or initialize loggedInUser.
-    if (auth && auth.accessToken && !loggedInUser) {
-      fetchOrInitializeUser(auth)
+    if (auth0 && auth0.isAuthenticated && !loggedInUser) {
```
landonreed, line 43: Please use a comment to describe this condition that we're checking.

landonreed (changes requested):
Haven't fully tested this yet, but it needs a couple of code changes/comments

lib/components/user/with-logged-in-user-support.js
```diff
@@ -37,26 +37,32 @@ export default function withLoggedInUserSupport (WrappedComponent, requireLogged
...
+      try {
+        // Get the token
+        const token = await auth0.getAccessTokenSilently()
+        // Once accessToken is available, proceed to fetch or initialize loggedInUser.
+        fetchOrInitializeUser(auth0, token)
+      } catch (error) {
+        alert('Error obtaining an authorization token.')
+      }
```
evansiroky, line 51: Please move this logic into the fetchOrInitializeUser action so it all stays in one place.

evansiroky (approved):
Just one refactoring comment.

lib/components/app/responsive-webapp.js
binh-dam-ibigroup: That is a good point, so I should probably keep URL_ROOT and add a TODO for supporting a root path that has a subfolder.

lib/components/user/with-logged-in-user-support.js
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/233/commits/0eedf9d0b11454855e64acc0c8c2569f5240b109.

lib/components/user/with-logged-in-user-support.js
binh-dam-ibigroup: This condition appears in multiple places, so replaced with a method (with jsdoc) in https://github.com/opentripplanner/otp-react-redux/pull/233/commits/0eedf9d0b11454855e64acc0c8c2569f5240b109

lib/components/user/saved-trip-screen.js
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/233/commits/4a6c2911bce4e981f5b2162cbb4e4696cb9f4dc0

lib/components/user/with-logged-in-user-support.js
```diff
@@ -37,27 +37,34 @@ export default function withLoggedInUserSupport (WrappedComponent, requireLogged
  * Upon completion (or if no user is logged in or if auth is disabled), it renders children.
  */
 class UserLoaderScreen extends Component {
-  componentDidUpdate () {
-    const { auth, fetchOrInitializeUser, loggedInUser } = this.props
+  /**
+   * Determines whether user data should be fetched.
+   * @returns true if the logged-in user has passed Auth0 authentication and state.user.loggedInUser has not been set; false otherwise.
+   */
+  shouldBeFetchingUser = () => {
```
evansiroky, line 44: Please rename to `shouldFetchUser` since it seems this shouldn't be a continuous action that is performed.

evansiroky (approved):

lib/components/user/with-logged-in-user-support.js
binh-dam-ibigroup: Done. (I picked that name because it read better in the render method, but that was a split call)

lib/actions/user.js
```diff
@@ -49,44 +49,56 @@ export function fetchUserMonitoredTrips (accessToken) {
...
+        // {
+        //    status: 'error',
+        //    message: 'Error get-ing user...'
+        // }
+        // TODO: Improve AWS response.
+
+        const isNewAccount = fetchStatus === 'error' || (user && user.result === 'ERR')
+        if (!isNewAccount) {
+          // Load user's monitored trips before setting the user state.
+          await dispatch(fetchUserMonitoredTrips(accessToken))
```
landonreed, line 96: Sorry I didn't flag this before, but why does the order matter here?

lib/components/user/with-logged-in-user-support.js
```diff
@@ -37,33 +37,34 @@ export default function withLoggedInUserSupport (WrappedComponent, requireLogged
  * Upon completion (or if no user is logged in or if auth is disabled), it renders children.
  */
 class UserLoaderScreen extends Component {
+  /**
+   * Determines whether user data should be fetched.
+   * @returns true if the logged-in user has passed Auth0 authentication and state.user.loggedInUser has not been set; false otherwise.
```
landonreed, line 42: Line seems a bit long (can't tell for sure outside of my code editor).

lib/components/user/with-logged-in-user-support.js
```diff
@@ -37,33 +37,34 @@ export default function withLoggedInUserSupport (WrappedComponent, requireLogged
...
+   */
+  shouldFetchUser = () => {
```
landonreed, line 44: Would a better name be `userIsMissing/Unfetched`?

lib/components/user/with-logged-in-user-support.js
```diff
@@ -37,33 +37,34 @@ export default function withLoggedInUserSupport (WrappedComponent, requireLogged
...
-        return renderChildrenWithProps(children, { auth0 })
-      }
+    if (requireLoggedInUser && this.shouldFetchUser()) {
+      // If a logged-in user is required, then
+      // display a hint while the logged-in user data is being fetched (from componentDidMount).
+      // Don't display this if loggedInUser is not required or is already available.
+      // TODO: Improve this screen.
+      return <AwaitingScreen />
+    } else if (auth0) {
+      return renderChildrenWithProps(children, { auth0 })
```
landonreed, line 67: "else if" condition could use a comment.

landonreed (changes requested):
Changes look good, just a couple more items.

lib/actions/user.js
binh-dam-ibigroup: We want the user's trips to be available when `state.user.loggedInUser` is set, so other parts of the UI don't have to check whether the monitored trips have been loaded or not. This would be most noticeable when you log in directly from the `/#/savedtrips` route.

binh-dam-ibigroup: I added a comment regarding that in https://github.com/opentripplanner/otp-react-redux/pull/233/commits/c9bb706107fffe6b477e73123313da05a5031f74.

lib/components/user/with-logged-in-user-support.js
binh-dam-ibigroup: Fixed in https://github.com/opentripplanner/otp-react-redux/pull/233/commits/c9bb706107fffe6b477e73123313da05a5031f74

lib/components/user/with-logged-in-user-support.js
binh-dam-ibigroup: Good suggestion. Updated in https://github.com/opentripplanner/otp-react-redux/pull/233/commits/c9bb706107fffe6b477e73123313da05a5031f74.

lib/components/user/with-logged-in-user-support.js
binh-dam-ibigroup: Glad you pointed this out. I ended up removing the else-if block as only one component was using the `auth0` prop to get the `email_verified` state. I added an `isEmailVerified` to the user state in https://github.com/opentripplanner/otp-react-redux/pull/233/commits/8fe223ef4bda834432f6ecc59d3162180af76650, I think that looks better.

evansiroky (approved):

lib/actions/user.js
landonreed: What parts of the UI are fetching the monitored trips? I couldn't really find where this would happen. In any case, this doesn't seem like a pattern we want to embed here. In my view we want to update the store with the user as soon as possible so that we're not waiting for an extra request to the backend to be fulfilled. Also, it avoids calling `serCurrentUser` multiple times.

lib/components/user/user-account-screen.js
```diff
@@ -107,35 +109,10 @@ class UserAccountScreen extends Component {
...
-          onCancel={this._handleExit}
-          onComplete={this._handleExitAndSave}
-          panes={this._panes}
-        />
-      )
-    }
+    const { isEmailVerified, loggedInUser } = this.props
+
+    // Capture the handler for use in the Formik callback (where `this` might mean something else.)
+    const handleExit = this._handleExit
```
landonreed, line 115: I'm not sure why this is treated differently than passing `this._panes` or `this._handleSaveAndExit`.

binh-dam-ibigroup: I thought `this` inside the Formik component could have another meaning but it turns out not to be the case. This is handled by commit https://github.com/opentripplanner/otp-react-redux/commit/6902c2b4c80245cb0821fdf0b69ec5406dbb242f.

lib/components/user/with-logged-in-user-support.js
landonreed: Hmm, after taking a closer look at https://github.com/opentripplanner/otp-react-redux/commit/8fe223ef4bda834432f6ecc59d3162180af76650, it feels a little awkward to me. Here's a few suggestions:
- add default state for create-user-reducer. It can be confusing when there is no initial state.
- in `UserLoaderScreen#componentDidUpdate`:
  - remove `async`
  - add `prevProps` arg and compare prevProps.auth0.email_verified to this.props.... etc.
  - if this.props.email verified is true, dispatch new action (SET_EMAIL_VERIFIED) to true.
  - move email verified logic in user reducer to new case

lib/actions/user.js
landonreed: See my comment about setting a default state in the user reducer. I don't think we should be relying on this order of operations for determining whether a user is allowed to create a new monitored trip.

lib/components/user/with-logged-in-user-support.js
binh-dam-ibigroup: Should the Auth0 state be stored in the redux state?

binh-dam-ibigroup: I don't think we should expect the email verified state to change, so there shouldn't be an action for that (unless when the user clicks "My email is verified", we do something other than reloading the page).

landonreed: Yea, perhaps storing the Auth0 state alongside the loggedInUser would be a better approach. Do you see any issues there?

binh-dam-ibigroup: Adding the Auth0 state in redux would help with another refactor I have been considering doing on the login button.

binh-dam-ibigroup: The `isEmailVerified` hack is removed in https://github.com/opentripplanner/otp-react-redux/pull/233/commits/e25ef5d781232cde60c0ac866b357b7610e65d47.

binh-dam-ibigroup: I reversed course and went back to passing the `auth0` prop to `UserAccountScreen` (found a better way to do that in https://github.com/opentripplanner/otp-react-redux/pull/233/commits/d359a8b8930d10602543081ec3d7e16511180dae). I don't think storing `auth0` in the redux state is warranted yet.

landonreed: Nice work!

lib/actions/user.js
binh-dam-ibigroup: If we want to remove fetching trips when fetching the user data, then `state.user.loggedInUserMonitoredTrips` needs to be fetched on first use from `SavedTripList` and `SavedTripScreen`. Each of these components will need to invoke the fetch and display a hint while fetching is taking place. Is that aligned with your thinking @landonreed ?

binh-dam-ibigroup: I changed the loading of user monitored trips in https://github.com/opentripplanner/otp-react-redux/pull/233/commits/99755e59acd2b46332998b7c656436acff555919. It wasn't too big of a change this time, otherwise it would go in a separate PR.

landonreed (changes requested):
See comments from phone discussion.

lib/actions/user.js
```diff
@@ -127,6 +117,39 @@ export function createOrUpdateUser (userData) {
...
+    }
+  }
+}
+
+/**
+ * Fetches the saved/monitored trips for a user.
+ * We use the accessToken to fetch the data regardless of
+ * whether the process to populate state.user is completed or not.
+ */
+export function fetchUserMonitoredTrips (accessToken) {
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/233/commits/166d00f0b4564c8b4bda20e906a5ab764bdbc635. That also removed the extra fetching from within components.

lib/actions/user.js
```diff
@@ -141,6 +117,39 @@ export function createOrUpdateUser (userData) {
...
+/**
+ * Ensures user monitored trips are loaded.
+ * state.user.loggedInUserMonitoredTrips is at least an empty array when loaded.
+ */
+export function ensureUserMonitoredTripsAreLoaded () {
+  return async function (dispatch, getState) {
+    const { user } = getState()
+    const { accessToken, loggedInUser, loggedInUserMonitoredTrips } = user
+    if (accessToken && loggedInUser && !loggedInUserMonitoredTrips) {
+      await dispatch(fetchUserMonitoredTrips(accessToken))
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Reordered and no longer awaiting in https://github.com/opentripplanner/otp-react-redux/pull/233/commits/166d00f0b4564c8b4bda20e906a5ab764bdbc635 per discussion.

lib/actions/user.js
```diff
@@ -203,7 +192,7 @@ export function deleteUserMonitoredTrip (id) {
 
       if (deleteResult.status === 'success') {
         // Reload user's monitored trips after deletion.
-        await dispatch(fetchUserMonitoredTrips(accessToken))
+        dispatch(fetchUserMonitoredTrips(accessToken))
```
landonreed, line 195: accessToken arg needs to be removed.

lib/actions/user.js
```diff
@@ -73,10 +73,12 @@ export function fetchOrInitializeUser (auth0) {
         // TODO: Improve AWS response.
 
         const isNewAccount = fetchStatus === 'error' || (user && user.result === 'ERR')
+        const userData = isNewAccount ? createNewUser(authUser) : user
+        dispatch(setCurrentUser({ accessToken, user: userData }))
+
+        // Also (asynchronously) load monitored trips for existing users.
```
landonreed, line 79: I don't think it needs to be stated that a fetch is asynchronous. I find the clarification more confusing than helpful.

landonreed (approved):
Looks good, just a couple of tiny loose ends.

lib/actions/user.js
binh-dam-ibigroup: Removed in https://github.com/opentripplanner/otp-react-redux/pull/233/commits/9b74a576f40dd13476fa8c0f4a218dfe55dd6dd5

lib/actions/user.js
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/233/commits/9b74a576f40dd13476fa8c0f4a218dfe55dd6dd5

binh-dam-ibigroup:
FYI @landonreed and @evansiroky there are some not-so-obvious merge changes I addressed in https://github.com/opentripplanner/otp-react-redux/pull/233/commits/b27eb1886ca3ee3456a97dd89afdb79ea1a43ae3. This PR should be ready to merge now.

PR #239, opened 2020-09-30
Form Validation

binh-dam-ibigroup:
This PR converts the forms for user accounts and saved trips into Formik, and adds validation using yup.
Previous form state management code is removed, as all state management is handled by Formik.

The main design change resulting from the Formik conversion is the component change for selecting the monitored trip days.
image

The following fields are validated:
- Phone number (accepts dashes and parentheses for US/Canada phone numbers),
- Trip name (must not be empty and not match an existing trip name),
image
- At least one monitored day must be selected.

PR #242, opened 2020-10-01
Check itinerary existence

binh-dam-ibigroup:
This PR depends on #239 and requires changes from ~https://github.com/ibi-group/otp-middleware/pull/77~ https://github.com/ibi-group/otp-middleware/pull/90.

It causes the trip settings UI to
- check for trip availability for each day of the week,
- provide a visual indication for days a trip is not available,
- set the flag to false for days on which a trip is not available, so those updates are persisted upon saving. (If a trip becomes unavailable on a day after the user saves it -- e.g. service is dropped on weekends on a route -- if the user makes a change, those days would no longer be monitored). 

image


lib/components/user/trip-basics-pane.js
```diff
@@ -31,73 +37,134 @@ const StyledLabel = styled.label`
...
+
+  async componentDidMount () {
+    // Check itinerary existence for all days.
+    const { accessToken, config, setFieldValue, values } = this.props
+
+    const itineraryCheckResult = await checkItinerary(
+      config.persistence.otp_middleware,
+      accessToken,
+      values
+    )
```
evansiroky, line 68: Since this check occurs every time the component mounts, it might be possible that an itinerary becomes impossible for an existing itinerary that was possible before. Is there a way to have this run only when creating a monitored trip?

binh-dam-ibigroup: That was my initial thought and It is possible from the UI, but it will currently fail on the backend (the `preUpdateHook` method will trigger an error if the trip does not exist for one of the monitored days). Should the backend be changed to accommodate that?

evansiroky: We should probably think about this a little more in depth. At some point this UI should also show when an itinerary is no longer possible which means updating things in the backend and then communicating that somehow to the frontend.

evansiroky (changes requested):
Until we figure out how to communicate that a Monitored Trip is no longer possible on a particular day of the week, we should have this itinerary existence check run only upon the creation of a new Monitored Trip.

binh-dam-ibigroup:
> Until we figure out how to communicate that a Monitored Trip is no longer possible on a particular day of the week, we should have this itinerary existence check run only upon the creation of a new Monitored Trip.

@evansiroky Did you mean also not enforcing that in the backend?

lib/components/user/trip-basics-pane.js
binh-dam-ibigroup: I went ahead and remove the trip existence enforcement when saving existing trips in https://github.com/ibi-group/otp-middleware/pull/77/commits/78e31e7b50bcda047f65ddf1cbbbe506dbb8e455. The UI will still perform the check and highlight the days an existing trip is not possible, however it will let the user save that trip. Trip existence continues to be enforced for new trips.

lib/components/user/trip-basics-pane.js
```diff
@@ -31,73 +37,140 @@ const StyledLabel = styled.label`
...
+          <FormGroup validationState={tripNameValidationState}>
+            <ControlLabel>Please provide a name for this trip:</ControlLabel>
+            {/* onBlur, onChange, and value are passed automatically. */}
+            <Field as={FormControl} name='tripName' />
+            <FormControl.Feedback />
+            {errors.tripName && <HelpBlock>{errors.tripName}</HelpBlock>}
+          </FormGroup>
+
+          <FormGroup validationState={monitoredDaysValidationState}>
+            <ControlLabel>What days to you take this trip?</ControlLabel>
```
evansiroky, line 123: typo: `What days to you take this trip?`

evansiroky (approved):
Seems good

lib/components/user/trip-basics-pane.js
binh-dam-ibigroup: Thanks, fixed in https://github.com/opentripplanner/otp-react-redux/pull/242/commits/38b22499b883d6ce22d87c8822d2eacb21eb668b.

lib/components/narrative/tabbed-itineraries.js
```diff
@@ -15,7 +15,7 @@ class TabbedItineraries extends Component {
   static propTypes = {
     itineraries: PropTypes.array,
     itineraryClass: PropTypes.func,
-    pending: PropTypes.number,
+    pending: PropTypes.bool,
```
landonreed, line 18: pending should be a number per batch routing changes.

lib/util/middleware.js
```diff
@@ -2,6 +2,7 @@ if (typeof (fetch) === 'undefined') require('isomorphic-fetch')
 
 const API_USER_PATH = '/api/secure/user'
 const API_MONITORTRIP_PATH = '/api/secure/monitoredtrip'
+const API_ITINERARYCHECK_PATH = '/api/secure/itinerarycheck'
```
landonreed, line 5: A general comment for these constants: please insert missing underscores to make them human readable.

lib/util/middleware.js
```diff
@@ -139,3 +140,12 @@ export async function deleteTrip (middlewareConfig, token, id) {
     }
   }
 }
+
+export async function checkItinerary (middlewareConfig, token, monitoredTrip) {
```
landonreed, line 144: jsdoc needed. Also, this is another case where there is a lot of logic in the component state that I would prefer that we move to actions and persist in the reducer if possible.

binh-dam-ibigroup: This function is now an action in https://github.com/opentripplanner/otp-react-redux/pull/242/commits/ea97ce635402164e163c1f46be35c6ac7d971697.

lib/util/middleware.js
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/242/commits/ea97ce635402164e163c1f46be35c6ac7d971697.

lib/components/narrative/tabbed-itineraries.js
binh-dam-ibigroup: I fixed the wrong itineraries list... the correct one is taken care in https://github.com/opentripplanner/otp-react-redux/pull/242/commits/214e266d7b2b56c0e667358d9a5cac689a23212e.

lib/actions/user.js
```diff
@@ -5,14 +5,17 @@ import { secureFetch } from '../util/middleware'
...
 const API_OTPUSER_PATH = '/api/secure/user'
-const API_OTPUSER_VERIFYSMS_PATH = '/verify_sms'
+const API_OTPUSER_VERIFY_SMS_SUBPATH = '/verify_sms'
 
 const setCurrentUser = createAction('SET_CURRENT_USER')
 const setCurrentUserMonitoredTrips = createAction('SET_CURRENT_USER_MONITORED_TRIPS')
 const setLastPhoneSmsRequest = createAction('SET_LAST_PHONE_SMS_REQUEST')
 export const setPathBeforeSignIn = createAction('SET_PATH_BEFORE_SIGNIN')
+export const clearItineraryExistence = createAction('CLEAR_ITINERARY_AVAILABILITY')
+const setitineraryExistence = createAction('SET_ITINERARY_AVAILABILITY')
```
landonreed, line 18: nitpick: I think the action string should match the action name i.e., replace `AVAILABILITY` with `EXISTENCE`.

lib/actions/user.js
```diff
@@ -293,3 +296,26 @@ export function verifyPhoneNumber (code) {
...
   }
 }
+
+/**
+ * Check itinerary existence for the given monitored trip.
+ */
+export function checkItineraryExistence (trip) {
+  return async function (dispatch, getState) {
+    const { accessToken, apiBaseUrl, apiKey } = getMiddlewareVariables(getState())
+    const requestUrl = `${apiBaseUrl}${API_MONITORED_TRIP_PATH}${API_MONITORED_TRIP_CHECK_SUBPATH}`
```
landonreed, line 306: I don't think `API_MONITORED_TRIP_CHECK_SUBPATH` needs to be in a constant if it's only used one time. Having constants far away from the place they're used can make things hard to read.

lib/components/user/trip-basics-pane.js
```diff
@@ -31,73 +37,142 @@ const StyledLabel = styled.label`
...
+      )
+    }
+  }
+}
+
+// Connect to redux store
+const mapStateToProps = (state, ownProps) => {
+  const { accessToken, itineraryExistence } = state.user
+  return {
+    accessToken,
```
landonreed, line 167: Is `accessToken` needed still?

lib/components/user/trip-basics-pane.js
```diff
@@ -31,73 +37,142 @@ const StyledLabel = styled.label`
...
+  { name: 'friday', text: 'Fri.', fullText: 'Fridays' },
+  { name: 'saturday', text: 'Sat.', fullText: 'Saturdays' },
+  { name: 'sunday', text: 'Sun.', fullText: 'Sundays' }
 ]
 
+/**
+ * @returns true if there is a trip matching for the specified availability/existence check.
+ */
+function itineraryExists (dayCheck) {
+  return dayCheck && dayCheck.valid
```
landonreed, line 53: I'm not sure this method is needed any longer, is it? We should be guaranteed to have an entry for each day in the ItineraryExistence response, right?

lib/components/user/trip-basics-pane.js
```diff
@@ -31,73 +37,142 @@ const StyledLabel = styled.label`
...
+            {/* onBlur, onChange, and value are passed automatically. */}
+            <Field as={FormControl} name='tripName' />
+            <FormControl.Feedback />
+            {errors.tripName && <HelpBlock>{errors.tripName}</HelpBlock>}
+          </FormGroup>
+
+          <FormGroup validationState={monitoredDaysValidationState}>
+            <ControlLabel>What days do you take this trip?</ControlLabel>
+            <div>
+              {allDays.map(({ name, fullText, text }, index) => {
```
landonreed, line 127: It seems a bit redundant to have `allDays` and `ALL_DAYS`, which effectively contain the same info.

landonreed (changes requested):
Just a few nitpicks. Otherwise, looks great!

lib/components/user/trip-basics-pane.js
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/242/commits/846892ab816dc32ccd6688b06042f030225c453a.

lib/components/user/trip-basics-pane.js
binh-dam-ibigroup: No it is not. Updated in https://github.com/opentripplanner/otp-react-redux/pull/242/commits/846892ab816dc32ccd6688b06042f030225c453a.

lib/components/user/trip-basics-pane.js
binh-dam-ibigroup: Good find. Updated in https://github.com/opentripplanner/otp-react-redux/pull/242/commits/846892ab816dc32ccd6688b06042f030225c453a.

lib/actions/user.js
binh-dam-ibigroup: Sure. Updated in https://github.com/opentripplanner/otp-react-redux/pull/242/commits/846892ab816dc32ccd6688b06042f030225c453a

lib/actions/user.js
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/242/commits/846892ab816dc32ccd6688b06042f030225c453a.

landonreed (approved):

PR #243, opened 2020-10-02
Don't send notifications for successful builds

evansiroky:
This PR changes the notifications that Travis sends such that only build failures trigger a notification. 

Wait do people think about doing this... for this repo and all others where we get build notifications? I'm pretty sure all of these successful build notifications aren't all that helpful, but it is helpful to know if a build is failing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,10 @@ cache:
   yarn: true
 notifications:
   email: false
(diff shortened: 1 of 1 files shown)


binh-dam-ibigroup:
If we could also send a notification if the build status says 'fixed' (after a failed build), then the proposed change would help cleanup the Teams channel.

landonreed (approved):
I don't think I really pay attention to these notifications, so probably good to limit them to failures (I would be more likely to treat them less as noise). The only caveat is that I do like the "fixed" email notifications I get from Travis. Not sure if this would remove those from Teams notifications.

binh-dam-ibigroup (approved):
Approving, although it would be useful to include the 'build fixed' notifications.

evansiroky:
I believe I just changed this to trigger notifications when a build is fixed, so I'm going to go ahead and merge.

PR #246, opened 2020-10-02
Refactor user actions

binh-dam-ibigroup:
This PR extracts the code to parse middleware configuration vars,
and inlines methods previously in `utils/middleware`, as brought up by https://github.com/opentripplanner/otp-react-redux/pull/224#discussion_r492961846 and https://github.com/opentripplanner/otp-react-redux/pull/224#discussion_r497508647.

### DIFF
--- a/lib/actions/user.js
+++ b/lib/actions/user.js
@@ -29,8 +29,10 @@ function createNewUser (auth0User) {
  *     - apiKey and apiBaseUrl from state.otp.config.persistence.otp_middleware,
  *     - accessToken and loggedIUnUser from state.user.
  * - checks that otp_middleware is set, and throws an error if not.
- * @param functionToExecute the code to execute, with parameters (dispatch, arguments)
- * @return a redux action that will run if a middleware s configured.
+ * @param functionToExecute a function that can be waited upon,
+ *   with parameters (dispatch, arguments), that contains the code to be
+ *   executed if the OTP middleware is configured.
+ * @return a redux action for the code to be executed.
  */
 function executeWithMiddleware (functionToExecute) {
   return async function (dispatch, getState) {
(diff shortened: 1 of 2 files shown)


lib/actions/user.js
```diff
@@ -28,102 +24,135 @@ function createNewUser (auth0User) {
...
 /**
- * Fetches the saved/monitored trips for a user.
- * We use the accessToken to fetch the data regardless of
- * whether the process to populate state.user is completed or not.
+ * Helper function that
+ * - extracts key variables from the state and passes them to the code to execute:
+ *     - apiKey and apiBaseUrl from state.otp.config.persistence.otp_middleware,
+ *     - accessToken and loggedIUnUser from state.user.
+ * - checks that otp_middleware is set, and throws an error if not.
+ * @param functionToExecute the code to execute, with parameters (dispatch, arguments)
```
evansiroky, line 32: Please note that `functionToExecute` can be a function that can be awaited upon.

evansiroky (approved):
Just one comment about a comment

landonreed (changes requested):
This approach is far too clever. I really disagree with how the `executeWithMiddleware` method body is structured and how it is invoked throughout `user.js`.

binh-dam-ibigroup:
> This approach is far too clever. I really disagree with how the `executeWithMiddleware` method body is structured and how it is invoked throughout `user.js`.

It highlights the core logic for these actions and hides repeat code to extract and check common variables.

evansiroky (changes requested):
I'm going to have to change my review to request changes. I kind of agree with Landon in that this introduces a bit too much abstraction/overhead. I think the `executeWithMiddleware` shouldn't exist and instead it should simply be a util function that makes sure the middleware config exists and if so returns the desired middleware data given the getState method.

PR #251, opened 2020-10-08
Fix results screen when planning trip

### LINKED ISSUE #250: Mobile results screen shows "Waiting..." after planning trip
This appears to be due to changes introduced with batch routing/trip comparison (#165).

To reproduce:
1. Go to planyourtrip.ridemetro.org
2. Open devtools, use mobile device.
3. Plan trip.
4. See below:

image

landonreed:
Correctly handle getting itineraries in mobile screens. Fixes #250.

Note: this does not resolve the fact that the expanded mobile screen does not show a full set of results, but that should be a separate PR/issue.


binh-dam-ibigroup (approved):
LGTM. No immediate odd behavior.

PR #257, opened 2020-10-19
Make navbar border color light color.

binh-dam-ibigroup:
This PR makes the border of the navbar the same as current (non-middleware) deployments.
This PR should be merged in order to resolve this discussion: https://github.com/ibi-group/trimet-mod-otp/pull/277#pullrequestreview-493506135 and merge https://github.com/ibi-group/trimet-mod-otp/pull/277.

### DIFF
--- a/lib/components/app/app.css
+++ b/lib/components/app/app.css
@@ -8,6 +8,10 @@
   padding: 0px 0px;
 }
 
+.navbar-inverse {
+  border-color: #e7e7e7; /* Same as non-inverse style. */
+}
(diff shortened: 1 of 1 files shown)


evansiroky (approved):

landonreed (commented):
I think I've been avoiding this PR because the problem isn't well described and this solution doesn't really seem appropriate. If there are custom css changes needed to the navbar, I think they should be handled in the trimet-mod-otp override css.

PR #259, opened 2020-10-20
Get error prop from state; skip sort if batch disabled

landonreed:
This resolves an issue blocking ibi-group/trimet-mod-otp#278 that prevented error responses from showing up in the narrative itinerary.

It also does not sort itineraries (meaning the default sort in the API response would be used) if batch routing is not enabled.


binh-dam-ibigroup (approved):
Looks good and seems to work with trimet-mod-otp.

PR #266, opened 2020-10-30
Only allow saving itineraries with transit and no rentals

binh-dam-ibigroup:
This PR shows the 'Save trip' button in the `NarrativeItineraries` if the selected itinerary has a transit leg and no rental legs, hides the button otherwise. The PR also removes the 'Save this option' trip button from `LineItinerary`.


lib/util/itinerary.js
```diff
@@ -15,3 +15,45 @@ export function getLeafletLegBounds (leg) {
...
+}
+
+/**
+ * Determines whether an itinerary can be monitored.
+ * @returns true if at least one of the legs of the specified itinerary is
+ *   a transit leg, and none of the legs is a rental leg (e.g. CAR_RENT, BICYCLE_RENT, etc.).
+ */
+export function itineraryCanBeMonitored (itinerary) {
+  return itineraryHasTransit(itinerary) && !itineraryHasRental(itinerary)
+}
```
evansiroky, line 59: These 3 functions could probably be combined into one so that a single loop occurs which checks each leg as needed. Furthermore, the `coreUtils.itinerary.hasRental` could be used. Also, this must check if a ridehail leg exists (`coreUtils.itinerary.hasHail` can be used). Also, I noticed that `lib/util/state.js` has a duplicate function of `itineraryHasTransit`. This redundancy should be avoided. Regarding testing, I think making tests for the `itineraryCanBeMonitored` would be good.

evansiroky (changes requested):
See comment.

lib/components/narrative/narrative-itineraries.js
```diff
@@ -145,8 +147,8 @@ class NarrativeItineraries extends Component {
                 onClick={this._toggleDetailedItinerary}>
                 <i className='fa fa-arrow-left' /> View all options
               </button>
-              {/* FIXME: only save if meets requirements (e.g., is transit + non-realtime mode) */}
-              {persistence && persistence.enabled
+              {/* Only save if itinerary has transit and no rental modes that cannot be guaranteed. */}
+              {persistence && persistence.enabled && itineraryCanBeMonitored(itineraries[activeItinerary])
```
landonreed, line 151: So, I think we should maybe take this opportunity to encapsulate this button into its own component. There are a few states that could be represented here, that we haven't quite addressed yet:
1. User is not logged in - in this instance, I think we should have the SaveTripButton render something like: "Please sign in to save trip"
2. Persistence disabled - just return null
3. itin cannot be monitored - we should probably disable the button and for now show a tooltip saying: "To save a trip it must contain a transit leg and no rental vehicles legs." Not sure if that's a great message long-term though...

landonreed (changes requested):
Save trip button probably warrants a component at this stage. Also, what Evan said.

lib/util/itinerary.js
binh-dam-ibigroup: Yeah, I totally missed on the duplicate in `state.js` but that function is no longer used for the purpose of this PR. The changes are in https://github.com/opentripplanner/otp-react-redux/pull/266/commits/df6bf8da862d7ac1bee6f28cdb581c5cb581cb8d, and the tests in https://github.com/opentripplanner/otp-react-redux/pull/266/commits/aa19444390bafc983ddc5ed51c0eb6fbd0aab1ae.

lib/components/narrative/narrative-itineraries.js
binh-dam-ibigroup: 1. I agree it is probably the simpler solution for now.
2. Agree
3. How about for the disabled button text: `🚫 Cannot save`? And for the tooltip, taken from the backend: "Only trips that include transit and no rentals or ride hailing can be saved/monitored."

__tests__/util/itinerary.js
```diff
@@ -0,0 +1,83 @@
...
+      title: 'should be false for a blank itinerary.'
+    }, {
+      expected: false,
+      itinerary: null,
+      title: 'should be false for a null itinerary.'
+    }]
+
+    testCases.forEach(({ expected, itinerary, title }) => {
+      it(title, () => {
+        expect(itineraryCanBeMonitored(itinerary))[expected ? 'toBeTruthy' : 'toBeFalsy']()
```
evansiroky, line 79: Since `itineraryCanBeMonitored` always returns exactly true of false, the readability of this assertion could be improved to be `expect(...).toBe(expected)`.

lib/components/narrative/save-trip-button.js
```diff
@@ -0,0 +1,87 @@
...
+  return {
+    itinerary: activeSearch && itineraries && itineraries[activeSearch.activeItinerary],
+    loggedInUser: state.user.loggedInUser,
+    persistence
+  }
+}
+
+const mapDispatchToProps = (dispatch, ownProps) => {}
+
+export default connect(mapStateToProps, mapDispatchToProps)(SaveTripButton)
```
evansiroky, line 87: `mapDispatchToProps` is not a required argument, so this can be replaced with `connect(mapStateToProps)(SaveTripButton)`

evansiroky (changes requested):
See 2 comments.

lib/components/narrative/save-trip-button.js
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/266/commits/4ccc5221910a02aab41e33d5251ec20459c8f68c.

__tests__/util/itinerary.js
binh-dam-ibigroup: Thanks, it was otherwise copy/paste from another test... Updated in. https://github.com/opentripplanner/otp-react-redux/pull/266/commits/4ccc5221910a02aab41e33d5251ec20459c8f68c.

evansiroky (approved):

lib/components/narrative/save-trip-button.js
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// connect to the redux store
+
+const mapStateToProps = (state, ownProps) => {
+  const activeSearch = getActiveSearch(state.otp)
+  const { persistence } = state.otp.config
+  const itineraries = getActiveItineraries(state.otp)
+  return {
+    itinerary: activeSearch && itineraries && itineraries[activeSearch.activeItinerary],
```
landonreed, line 79: Opportunity to use `getActiveItinerary(state.otp)` which handles all of this other getActiveSearch/getActiveItineraries logic.

lib/components/narrative/line-itin/line-itinerary.js
```diff
@@ -77,18 +75,13 @@ export default class LineItinerary extends NarrativeItinerary {
           timeOptions={timeOptions}
         />
 
-        {user &&
-          <span className='pull-right'><LinkButton to='/savetrip'>Save this option</LinkButton></span>
-        }
-
```
landonreed, line 83: Actually, I know this goes against what I previously stated, but we should keep the new SaveTrip button component here `<SaveTripButton />`. This was decided in conversation with Ritesh earlier last week.

landonreed (approved):
Looks good. Conditional approval on a few final items.

lib/components/narrative/line-itin/line-itinerary.js
binh-dam-ibigroup: Reinstated in https://github.com/opentripplanner/otp-react-redux/pull/266/commits/80b8a912a73dce09a55078b0a243478e14dc783c.

lib/components/narrative/save-trip-button.js
binh-dam-ibigroup: Thanks. Updated in https://github.com/opentripplanner/otp-react-redux/pull/266/commits/80b8a912a73dce09a55078b0a243478e14dc783c.

lib/components/narrative/line-itin/line-itinerary.js
binh-dam-ibigroup: https://github.com/opentripplanner/otp-react-redux/pull/266/commits/dec353af2400b589ee2998747ea450d76d468e4d uses `<SaveTripButton />` in `LineItinerary`.

PR #269, opened 2020-10-30
Pass timeFormat to rendered itin

### LINKED ISSUE #268: Mismatched time values in Itinerary Narrative vs. Tabbed Summary
When planning a trip for a trip planner hosted in a timezone that is offset from your own (e.g., me, in Atlanta, planning a trip in an Ireland OTP instance), the offset is not being correctly applied. This is related to #247, which we attempted to resolve with #248, but due to the absence of static typing in this library, the true underlying issue (a missing `timeFormat` prop) was missed.

image

landonreed:
Fixes #268. A previous attempt had tried to fix this by connecting the RealtimeTimeColumn to the redux store, but the true underlying issue was a bad destructuring of props in `TabbedItineraries`.

To test, use a config.yml for a timezone different than your local computer's. Plan a trip and verify that the times in the narrative itinerary match what is shown in the tabbed-itin summary.

### DIFF
--- a/lib/components/narrative/line-itin/realtime-time-column.js
+++ b/lib/components/narrative/line-itin/realtime-time-column.js
@@ -3,10 +3,9 @@ import {
   legType,
   timeOptionsType
 } from '@opentripplanner/core-utils/lib/types'
-import { formatTime, getTimeFormat } from '@opentripplanner/core-utils/lib/time'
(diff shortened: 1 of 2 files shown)


binh-dam-ibigroup (approved):
Turned out better than what I thought!

evansiroky (approved):

PR #274, opened 2020-11-17
Use dynamic contents via React context

evansiroky:
This PR depends on #201. 

In response to https://github.com/opentripplanner/otp-react-redux/pull/201#discussion_r524368723, this PR refactors the app to avoid needless prop-drilling of custom components sent in by implementing applications by utilizing the React context as needed.

Although this PR doesn't yet have any commits that indicate a breaking change, there are changes that could be considered breaking. The `itineraryClass`, `itineraryFooter`, `LegIcon` and `ModeIcon` are now all supposed to be added to a `components` property in the `ResponsiveWebapp` instead of in individual components. This also introduces the assumption that implementers of otp-react-redux will use the `ResponsiveWebapp` or directly import the `ComponentContext` and wrap components in a Provider in their app.

### DIFF
--- a/example.js
+++ b/example.js
@@ -19,6 +19,7 @@ import {
   CallTakerControls,
   CallTakerPanel,
   CallTakerWindows,
+  DefaultItinerary,
   DefaultMainPanel,
   DesktopNav,
   BatchRoutingPanel,
@@ -49,6 +50,7 @@ if (useCustomIcons) {
 // define some application-wide components that should be used in
 // various places
 const components = {
+  ItineraryBody: DefaultItinerary,
   LegIcon: MyLegIcon,
   ModeIcon: MyModeIcon
 }
--- a/example.js
+++ b/example.js
@@ -48,7 +48,11 @@ if (useCustomIcons) {
 }
 
 // define some application-wide components that should be used in
-// various places
+// various places. The following components can be provided here:
+// - ItineraryBody (required)
+// - ItineraryFooter (optional)
+// - LegIcon (required)
+// - ModeIcon (required)
 const components = {
   ItineraryBody: DefaultItinerary,
   LegIcon: MyLegIcon,
@@ -125,7 +129,32 @@ class OtpRRExample extends Component {
       </main>
     )
 
-    /** the main webapp **/
+    /**
+     * The main webapp.
+     *
+     * Note: the ResponsiveWebapp creates a React context provider
+     * (./util/contexts#ComponentContext to be specific) to supply custom
+     * components to various other subcomponents throughout otp-react-redux. If
+     * the ResponsiveWebapp is not used and instead some subcomponents that use
+     * the components in the `components` variable are imported and rendered
+     * outside of the ResponsiveWebapp component, then the ComponentContext will
+     * need to wrap that component in order for the subcomponents to be able to
+     * access the component context. For example:
+     *
+     * ```js
+     * import RouteViewer from 'otp-react-redux/build/components/viewers/route-viewer'
+     * import { ComponentContext } from 'otp-react-redux/build/util/contexts'
+     *
+     * const components = {
+     *   ModeIcon: MyCustomModeIconComponent
+     * }
+     * const ContextAwareRouteViewer = () => (
+     *   <ComponentContext.Provider value={components}>
+     *     <RouteViewer />
+     *   <ComponentContext.Provider/>
+     * )
+     * ```
+     */
     return (
       <ResponsiveWebapp
         components={components}
--- a/example.js
+++ b/example.js
@@ -46,6 +46,13 @@ if (useCustomIcons) {
   MyModeIcon = CustomIcons.CustomModeIcon
 }
 
+// define some application-wide components that should be used in
+// various places
+const components = {
+  LegIcon: MyLegIcon,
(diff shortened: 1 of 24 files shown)


landonreed (commented):
My primary comment (stated previously on a call, I think) is that perhaps we should allow users of this library to pass in LegIcon, ModeIcon or ItinType as a prop instead of context. And that we should perhaps not change the `itineraryFooter` name to avoid making this a breaking change.

lib/components/viewers/route-viewer.js
```diff
@@ -187,7 +187,9 @@ class RouteRow extends PureComponent {
   }
 
   render () {
-    const {isActive, ModeIcon, route, operator} = this.props
+    const {isActive, route, operator} = this.props
+    const { ModeIcon } = this.context
```
binh-dam-ibigroup, line 191: Inconsistent spacing between braces.

example.js
```diff
@@ -46,6 +47,14 @@ if (useCustomIcons) {
   MyModeIcon = CustomIcons.CustomModeIcon
 }
 
+// define some application-wide components that should be used in
+// various places
+const components = {
+  ItineraryBody: DefaultItinerary,
```
binh-dam-ibigroup, line 53: I think `ItineraryFooter` should be mentioned here, at least as a commented-out attribute, so implementers know that it is possible to provide a footer if they so wish.

binh-dam-ibigroup (approved):
Conditional approval on mentioning `ItineraryFooter` in `example.js`. This PR begs the question on whether `state.otp.config` should be a context too, but I definitely agree with the approach.

example.js
```diff
@@ -46,6 +47,14 @@ if (useCustomIcons) {
...
+  LegIcon: MyLegIcon,
+  ModeIcon: MyModeIcon
+}
+
```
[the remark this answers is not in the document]
evansiroky: I was also thinking that, but it'd be a much larger change to instead include that in the context instead of the redux store. Also, perhaps some parts of the config might be changeable, for example if we ever implement a webapp with multiple languages, maybe it could have a config setting of the language that could change in the app.

example.js
evansiroky: Improved above comment to note ItineraryFooter in 91bdbb5523830976b84ddf35078c2029899d5567.

lib/components/viewers/route-viewer.js
evansiroky: Fixed in 91bdbb5523830976b84ddf35078c2029899d5567.

evansiroky:
@landonreed, in response to your comment

> My primary comment (stated previously on a call, I think) is that perhaps we should allow users of this library to pass in LegIcon, ModeIcon or ItinType as a prop instead of context. And that we should perhaps not change the `itineraryFooter` name to avoid making this a breaking change.

The more I think of this, the more I'm thinking this is kind of a take-it-or-leave-it PR. If we want to keep this behavior for any of the `*Panel` components, then this would re-introduce the prop-drilling problem. If it's for the final components that use the context, it may not be possible for end-users to configure these properly without reintroducing prop-drilling since some components like `RouteRow` in `RouteViewer` aren't exported directly.

However, it is possible for an end-user to still use components directly without being inside the `ResponsiveWebapp` component. I have added some comments about how to do this in 91bdbb5523830976b84ddf35078c2029899d5567. In all reality, it seems unlikely that any end-user would import individual components directly given how tightly tied to redux the components in this repo are.

example.js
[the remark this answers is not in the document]
evansiroky: The more I think about it tho, maybe it would be good to consider the config being a global context thing that is assumed to not change. It could be useful to even explore adding some kind of context provider to otp-ui packages which would reduce the amount of prop-drilling and simplify the inclusion of config items throughout those components.

evansiroky:
Ready for review again, and I'm reassigning @binh-dam-ibigroup to consider https://github.com/opentripplanner/otp-react-redux/pull/274#discussion_r533700109.

binh-dam-ibigroup (approved):
LGTM.

landonreed:
@evansiroky, your comment about providing the config as context to otp-ui components made me think of this: if components like the StopViewer (or others that make use of context) were to migrate to otp-ui, how would we handle providing them with the context? Maybe this is just a bridge we cross when we get there, but just wanted to get the thought down.

evansiroky:
> @evansiroky, your comment about providing the config as context to otp-ui components made me think of this: if components like the StopViewer (or others that make use of context) were to migrate to otp-ui, how would we handle providing them with the context? Maybe this is just a bridge we cross when we get there, but just wanted to get the thought down.

I'm changing my mind again and thinking that maybe slots is actually better for otp-ui after all. Currently, otp-rr already sends appropriate components defined in the context like this: https://github.com/opentripplanner/otp-react-redux/blob/91bdbb5523830976b84ddf35078c2029899d5567/lib/components/form/connected-settings-selector-panel.js#L13-L41

Since each otp-ui package kind of stands by itself, it seems like it could be more work and unfriendly to first-time users to have to wrap it in a context provider in order for the package to work. Perhaps within each package it could make sense to use the context API to reduce prop-drilling inside the components in the package.

landonreed (approved):

PR #280, opened 2020-12-04
Improve save user locations

### LINKED ISSUE #202: Unable to use LinkContainer
Revisit the use of `LinkContainer`. Earlier attempts have resulted in LinkContainer not being able to get the router info (does it have to do with the React router setup?)

_Originally posted by @landonreed in https://github.com/opentripplanner/otp-react-redux/pull/174_

binh-dam-ibigroup:
This PR changes the user's saved locations pane ~using `LocationField` for entering locations and obtaining the locations coordinates, so that coordinates are persisted with the user preferences.~ by updating the widget (`FavoritePlaceRow`) to display each location (place).

~The PR also lets the user~ Clicking a `FavoritePlaceRow` opens a page (`FavoritePlacePage`) where the user can change the type of location (e.g. 'dining', 'custom' - what other predefined types need to be added?). ~A dropdown is added to the left of the location. (It is not possible to edit the nickname of each location yet.)~

Code-wise, each location row is now its own ~`FavoriteLocation`~ `FavoritePlaceRow` component. The `FavoritePlacePage` ~that~ hosts a styled ~`FavoriteLocationField`~ `PlaceLocationField` version of `LocationField`. ~This will allow further refinements of the layout or allowing more advanced editing so let me know if you have any suggestions.~

EDIT 1:
- The PR refactors the code connecting `LocationField` instances to the redux store.
- The PR replaces instances where FontAwesome is used directly with `Icon` (from narrative).

EDIT 2:
- Added a nickname field for the locations
- Updated the styling to reflect mockup
- Added a new `FixedFavoriteLocation` component and removed quirks in the code regarding placeholders
- Moved more shared components to `favorite-location-controls`.
- Updated screenshots below.

EDIT 3
- Moved favorite location components to subfolder `/user/favorite-locations/`

EDIT 4
- Updated PR description
- Renamed `/user/favorite-locations/` => `/user/places` and renamed components within.
- Added page/routing support for editing places.
- Updated visuals

EDIT 5:

- Create new versions of `<Link>` and `<LinkContainer>` that preserve the query parameters (fix #202).
- Use a separate URL for each page in the New account wizard (e.g. `/account/create/terms`), so that when user finishes editing a place, they remain in the wizard mode instead of seeing a single page with all settings.
- Issue: If the new user set the notification preference other than email then edits a place, the notification preference is lost. This issue will be addressed when we remove the Save button for user settings.

EDIT 6:

When editing places:
- while creating new accounts, the "My account" header is not shown.
- the Back/Save buttons are now shown below the place details. 

~Screenshot:~
image

~NEW screenshots:~ 
image

image

NEWEST screenshots

image

image


lib/components/user/favorite-location.js
```diff
@@ -0,0 +1,236 @@
+import React, { Component } from 'react'
+import {
+  Button,
+  DropdownButton,
+  FormGroup,
+  Glyphicon,
+  InputGroup,
+  MenuItem
+} from 'react-bootstrap'
+import FontAwesome from 'react-fontawesome'
```
landonreed, line 10: This is not generally how icons are defined in this app. We should be importing the Icon component:

```
import Icon from '../narrative/icon'
```

It looks like there are a couple of other spots where FontAwesome is used like this. We should replace those as well.

lib/components/user/favorite-location.js
```diff
@@ -0,0 +1,236 @@
...
+      this._handleLocationChange({ location: newLocation })
+    } else {
+      arrayHelpers.remove(index)
+    }
+  }
+
+  _handleLocationChange = ({ location }) => {
+    const { arrayHelpers, index, location: oldLocation } = this.props
+    const { icon, type } = oldLocation
+    arrayHelpers.replace(index, makeFavoriteLocation(location, type, icon))
```
landonreed, line 93: The updated UI mocks for user account show that we're moving away from making changes and then clicking a 'Save Preferences' button. I don't think we should invest more in this current approach. I'd like to see these handlers make PUT calls to update the user account with the list of locations. I'm also wondering if it might be good to have `UserLocation.class` extend Model so that we can get ids on these objects (separate concern from what's at hand).

lib/components/user/favorite-location.js
```diff
@@ -0,0 +1,236 @@
...
+      DropdownBtn = NewLocationDropdownButton
+    } else {
+      location = this.props.location
+      // TODO: lookup icon using location.type
+      icon = location.icon
+      iconTitle = location.type
+      handler = this._handleLocationChange
+      hideDelete = false // isFixed && isBlank(location.address)
+      placeholder = isFixed ? `Add ${location.type}` : 'Enter a favorite address'
+      DropdownBtn = StyledDropdown
```
landonreed, line 143: Since so much is different between new/existing, it might be more readable to have a separate component: 
```jsx
<NewLocation arrayHelpers={arrayHelpers} />
```

lib/components/user/favorite-location.js
```diff
@@ -0,0 +1,236 @@
...
+      title={<FontAwesome name={locationType.icon} style={{width: '10px'}} />}
+    >
+      {locationTypes.map((t, index) => (
+        <MenuItem eventKey={t} key={index} onSelect={onChange}>
+          <FontAwesome name={t.icon} style={{width: '15px'}} /> {t.text}
+        </MenuItem>
+      ))}
+    </DropdownButtonComponent>
+  )
+}
```
landonreed, line 234: This file is big enough that it might be worth moving this selector to its own file.

lib/components/user/favorite-locations-pane.js
```diff
@@ -1,151 +1,48 @@
...
+  return (
+    <div>
+      <ControlLabel>Add the places you frequent often to save time planning trips:</ControlLabel>
+
+      <FieldArray
+        name='savedLocations'
+        render={arrayHelpers => (
+          <>
+            {savedLocations.map((loc, index) => {
+              const isHomeOrWork = loc === homeLocation || loc === workLocation
```
landonreed, line 24: No need to define `homeLocation` and `workLocation` variables, right? I would rather see the `isFixed` prop go away entirely and just have favorite-location handle this internally (no need to define this extra prop as far as I can tell).

landonreed (changes requested):
A couple of changes. We do have some updated mocks that deal with the user account page (see sprint review presentation) that this doesn't align with, but perhaps we should discuss further (maybe we can handle alignment in a separate PR).

lib/components/form/favorite-location-field.js
```diff
@@ -0,0 +1,59 @@
...
+    display: table-cell;
+    font-size: 100%;
+    line-height: 20px;
+    padding: 6px 12px;
+    width: 100%;
+  }
+`
+
+// connect LocationField to redux store
+// TODO: Refactor. Each time we want to apply styles to LocationField, we need to reconnect it.
```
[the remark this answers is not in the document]
binh-dam-ibigroup: I just thought of this: what do you think of a higher-order component that basically does `connect(mapStateToProps, mapDispatchToProps)(LocationFieldArg)`, and we would apply it to the three instances where we use `LocationField`?

lib/components/user/favorite-location.js
binh-dam-ibigroup: I think we will need an overall discussion when we commision the work to remove the 'Save Preferences' button.

lib/components/user/favorite-location.js
binh-dam-ibigroup: Should I replace the `FontAwesome` instance in `NavigationBar` (mobile component)? This would happen:
image

landonreed: Feel free to tweak `icon.js` if the fixedWidth prop is giving you trouble. Easiest thing might be to just pass `fixedWidth={false}` to where it's used in navbar.

lib/components/user/favorite-location.js
landonreed: This current approach (not saving anything until 'Save prefs' is clicked) is not working well from a UX perspective, so it's really not beneficial to build this component around that behavior.

lib/components/user/favorite-location.js
binh-dam-ibigroup: NavigationBar arrow icon is updated in https://github.com/opentripplanner/otp-react-redux/pull/280/commits/eccfa5ccd2006c56e279ee185ea4f48fbbe62008.

lib/components/form/favorite-location-field.js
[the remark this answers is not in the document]
binh-dam-ibigroup: This is now addressed by #284.

lib/components/user/favorite-location.js
binh-dam-ibigroup: I agree on the principle. I think it is best to revise that event handler around the time we make the change to the backend.

lib/components/user/favorite-locations-pane.js
binh-dam-ibigroup: Hmmm no, this would couple FavoriteLocations with isHome and isWork.

binh-dam-ibigroup: FavoriteLocations is tied with isHome and isWork right now, but I might have an idea.

binh-dam-ibigroup: Okay, I have simplified the overall handling of work and home in https://github.com/opentripplanner/otp-react-redux/pull/280/commits/bcce0e4b6ce0cc0606208d6eac75dc7380fd8f77.

lib/components/user/favorite-location.js
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/280/commits/9298f94739c2196c0a6f47717a9481c12ad1e306.

lib/components/user/favorite-location.js
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/280/commits/9298f94739c2196c0a6f47717a9481c12ad1e306

lib/components/user/favorite-locations-pane.js
binh-dam-ibigroup: I eventually did away with that `isFixed` or `placeholder` prop. There is a new component `FixedFavoriteLocation` from https://github.com/opentripplanner/otp-react-redux/pull/280/commits/b5b3739a2a63157c758ea734c3e8507793f94cfc.

binh-dam-ibigroup:
> Looking good, but I do have a couple of changes. We do have some updated mocks that deal with the user account page (see sprint review presentation) that this doesn't align with, but perhaps we should discuss further (maybe we can handle alignment in a separate PR).

@landonreed I updated the layout so it is closer to the mockup now (see updated screenshots in PR description).

landonreed:
@binh-dam-ibigroup, I'm not seeing the location field with the Home/Work label as shown in the screenshot. Here's what I see:

image

binh-dam-ibigroup:
> @binh-dam-ibigroup, I'm not seeing the location field with the Home/Work label as shown in the screenshot. Here's what I see:
> 
> image

@landonreed I was able to reproduce and fixed in. https://github.com/opentripplanner/otp-react-redux/pull/280/commits/4a33ffd52772c06d4302edbdd09c52c9766a1bae for new accounts. On existing accounts, check in the database that the `name` field of the saved home/work locations is populated.

landonreed:
@binh-dam-ibigroup, I think if the type is home or work, we should force the `name` value to be Home/Work regardless of what's persisted in the database.

binh-dam-ibigroup:
> @binh-dam-ibigroup, I think if the type is home or work, we should force the `name` value to be Home/Work regardless of what's persisted in the database.

Updated in https://github.com/opentripplanner/otp-react-redux/pull/280/commits/3787ba84d82b1cbc2bc3528c3f8ab19f9b6e1063.

lib/components/user/new-favorite-location.js
```diff
@@ -0,0 +1,68 @@
...
+} from './favorite-location-controls'
+
+const blankNewLocationState = {
+  newLocation: null
+}
+
+/**
+ * Component that lets the user pick a new favorite location.
+ */
+class NewFavoriteLocation extends Component {
```
landonreed, line 24: It's probably not very clear in the mockups, but this component is intended to be a button (as were the other items in the list). When clicked, it opens a modal where you can set the name and search for geocode results. The main reason for this is so that the mobile experience is not degraded (things don't work well in mobile with the dropdown results, which is why we added the full screen view when setting the from/to location when planning a trip). We should probably discuss more on a call next week to get on the same page.

landonreed (changes requested):
I don't think the mockups were detailed enough to really give you a good picture of how the location items were intended to function -- sorry about that. Let's discuss more altogether next week to walk through the design in some more detail.

lib/components/app/responsive-webapp.js
```diff
@@ -211,6 +212,7 @@ class RouterWrapperWithAuth0 extends Component {
         history={history}>
         <div>
           <Switch>
+            {/* TODO: Review the url schemes for consistency and brevity. */}
```
evansiroky, line 215: Maybe make an issue or something instead of a comment.

binh-dam-ibigroup: I moved the comment to #309.

lib/util/user.js
```diff
@@ -1,4 +1,81 @@
...
+
+/**
+ * Makes a copy of the provided logged-in user data for editing
+ * (e.g. to set the Formik initial state), with:
+ * - the 'home' and 'work' locations at the top of the savedLocations list
+ *   so they are always shown and shown at the top of the FavoriteLocationsPane.
+ *   The name field is set to 'Home' and 'Work' regardless of the value that was persisted.
+ *   Note: In the returned value, savedLocations is always a valid array.
+ */
+export function cloneForEditing (userData) {
```
evansiroky, line 63: Maybe rename to `cloneUserDataForEditing` to be more specific.

lib/util/user.js
```diff
@@ -1,4 +1,81 @@
...
+    icon: 'home',
+    name: 'Home',
+    type: 'home'
+  },
+  work: {
+    icon: 'briefcase',
+    name: 'Work',
+    type: 'work'
+  }
+  // TODO: add more non-home/work types
```
evansiroky, line 42: I don't think this comment adds much value and could be deleted.

lib/components/user/places/favorite-place-page.js
```diff
@@ -0,0 +1,210 @@
...
+    if (placeIndex === 'new') {
+      clonedUser.savedLocations.push(placeToSave)
+    } else {
+      clonedUser.savedLocations[placeIndex] = placeToSave
+    }
+
+    await createOrUpdateUser(clonedUser, true)
+
+    this._handleBack()
+  }
```
evansiroky, line 82: Most of this logic seems like it would be better placed inside an action.

lib/components/user/places/place-editor.js
```diff
@@ -0,0 +1,124 @@
...
+              inputPlaceholder={isFixed ? `Search for ${place.name} location` : 'Search for location'}
+              location={makeLocationFieldLocation(place)}
+              locationType='to'
+              onLocationSelected={this._handleLocationChange}
+              showClearButton={false}
+            />
+            <FormControl.Feedback />
+            {errors.address && <HelpBlock>{errors.address}</HelpBlock>}
+
+            {/* TODO: Implement setting location using map. */}
```
evansiroky, line 116: I'm not a fan of leaving TODO comments in code because there is no guarantee these ideas are ever implemented. If it's really important we can make an issue, or perhaps better yet if it truly is needed we will just remember about it later without having created an issue or writing a code comment.

evansiroky (approved):
Just a few non-blocking review items.

binh-dam-ibigroup:
@landonreed Alright, this is merged with the new account page design, ready for review now.

lib/components/user/places/place-editor.js
binh-dam-ibigroup: Removed in https://github.com/opentripplanner/otp-react-redux/pull/280/commits/53797da17e0d91c1c07659bbd169f4e14ec8683f.

lib/components/user/places/favorite-place-page.js
binh-dam-ibigroup: Good point. Updated in https://github.com/opentripplanner/otp-react-redux/pull/280/commits/53797da17e0d91c1c07659bbd169f4e14ec8683f.

lib/util/user.js
binh-dam-ibigroup: I updated the comment for this variable in https://github.com/opentripplanner/otp-react-redux/pull/280/commits/53797da17e0d91c1c07659bbd169f4e14ec8683f.

lib/util/user.js
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/280/commits/53797da17e0d91c1c07659bbd169f4e14ec8683f.

binh-dam-ibigroup:
> @landonreed Alright, this is merged with the new account page design, ready for review now.

Updated with changes from @evansiroky's comments.

lib/components/user/places/favorite-places-pane.js
```diff
@@ -0,0 +1,60 @@
...
+
+        <FieldArray
+          name='savedLocations'
+          render={arrayHelpers => (
+            <>
+              {savedLocations.map((place, index) => (
+                <FavoritePlaceRow
+                  isFixed={isHomeOrWork(place)}
+                  key={index}
+                  onDelete={this._handleDelete(arrayHelpers, place, index)}
```
landonreed, line 43: This doesn't actually delete the place/persist the user data. So, when a user clicks delete, and then clicks the item to add a new location, the old location has re-appeared. I think this should be fixed (even before we address the potential removal of the Save/Cancel buttons at the bottom of the user settings page).

landonreed (changes requested):
I think we ought to address the bug I noted with the delete button functionality before merging.

lib/components/user/places/favorite-places-pane.js
binh-dam-ibigroup: This is updated in https://github.com/opentripplanner/otp-react-redux/pull/280/commits/c08040eee92a71d5dd0ca253f98e18bddff97112.

binh-dam-ibigroup:
Ready for review again. See EDIT 5 in the PR description for a summary of changes.

lib/components/user/sequential-pane-display.js
```diff
@@ -9,59 +11,50 @@ import { SequentialPaneContainer } from './styled'
...
-
-    this.state = {
-      activePaneId: props.initialPaneId
-    }
+  /**
+   * Routes to the next pane URL.
+   */
+  _routeTo = nextId => {
+    const { currentPath, activePaneId, routeTo } = this.props
+    routeTo(`${currentPath.replace(new RegExp(`/${activePaneId}$`), `/${nextId}`)}`)
```
evansiroky, line 23: Are the outermost ` marks needed anymore? Does this need to have a regex? Seems like it could be simplified a bit.

lib/components/user/places/place-editor.js
```diff
@@ -0,0 +1,116 @@
...
+          </FormGroup>
+        </>
+        )}
+
+        <div style={{ display: 'flex' }}>
+          {/* For fixed places, just show the icon for place type instead of all inputs and selectors */}
+          {isFixed && <FixedPlaceIcon name={place.icon} />}
+
+          <FormGroup style={{flexGrow: 1}} validationState={errorStates.address}>
+            <PlaceLocationField // TODO: mobile version.
```
evansiroky, line 99: Please remove TODO and create an issue or something.

lib/components/user/places/favorite-places-pane.js
```diff
@@ -0,0 +1,49 @@
...
+// connect to the redux store
+
+const mapStateToProps = (state, ownProps) => {
+  return {
+    loggedInUser: state.user.loggedInUser
+  }
+}
+
+const mapDispatchToProps = {
+  deleteUserPlace: userActions.deleteUserPlace
```
evansiroky, line 46: This is only used in the `FavoritePlaceRow` which is already a connected component. Therefore, this can be moved to that component instead.

evansiroky (approved):
Seems mostly good. See comments for a few ideas for improvement.

lib/components/form/connected-links.js
```diff
@@ -0,0 +1,42 @@
...
+
+const mapDispatchToProps = {
+}
+
+// Enhance routing components, connect the result to redux,
+// and export.
+export default {
+  Link: connect(mapStateToProps, mapDispatchToProps)(connectLink(Link)),
+  LinkContainer: connect(mapStateToProps, mapDispatchToProps)(connectLink(LinkContainer)),
+  Redirect: connect(mapStateToProps, mapDispatchToProps)(connectLink(Redirect))
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/280/commits/b207eb64bdd53e48e90abfe5f78d7bb196807603

lib/components/user/places/place-editor.js
binh-dam-ibigroup: I went ahead and accommodated the mobile view instead in https://github.com/opentripplanner/otp-react-redux/pull/280/commits/209318fb291171582ca4ef30180f9969f4cee40c.

lib/components/user/sequential-pane-display.js
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/280/commits/bdcebdff2bab6237bb804ea49e91e58d44bdb00c to extract the parent path instead.

lib/actions/user.js
```diff
@@ -444,3 +444,20 @@ export function saveUserPlace (placeToSave, placeIndex) {
...
   }
 }
+
+/**
+ * Delete the place data at the specified index for the logged-in user.
+ */
+export function deleteUserPlace (placeIndex) {
+  return function (dispatch, getState) {
+    // Get the places as shown (and not as retrieved from db), so that the index passed from URL applies
+    // (indexes 0 and 1 are for 'Home' and 'Work' places).
```
landonreed, line 454: If a user has home saved and an "other" place, this statement about reserved indexes doesn't necessarily hold true, but I'm splitting hairs! I think the way you've stated this is OK.

lib/components/app/responsive-webapp.js
```diff
@@ -243,10 +246,13 @@ class RouterWrapperWithAuth0 extends Component {
             <Route exact path={ACCOUNT_PATH}>
               <Redirect to={TRIPS_PATH} />
             </Route>
+            <Route exact path={CREATE_ACCOUNT_PATH}>
+              <Redirect to={CREATE_ACCOUNT_TERMS_PATH} />
```
landonreed, line 250: I'd prefer that the Verify screen also get its own path. Not sure how many headaches that would cause though...

lib/components/app/responsive-webapp.js
```diff
@@ -243,10 +246,13 @@ class RouterWrapperWithAuth0 extends Component {
...
+            </Route>
             <Route
               // This route lets new or existing users edit or set up their account.
               component={UserAccountScreen}
-              path={[CREATE_ACCOUNT_PATH, ACCOUNT_SETTINGS_PATH]}
+              path={[`${CREATE_ACCOUNT_PATH}/:id`, ACCOUNT_SETTINGS_PATH]}
```
landonreed, line 255: `:id` is a bit confusing, since we're not dealing with an entity id, but rather a stage of account creation. There may be some weirdness with how these params get processed though, so this is non-blocking (especially if headaches are involved).

lib/components/form/connected-links.js
```diff
@@ -0,0 +1,39 @@
...
+    queryParams: state.router.location.search
+  }
+}
+
+// Enhance routing components, connect the result to redux,
+// and export.
+export default {
+  Link: connect(mapStateToProps)(connectLink(Link)),
+  LinkContainer: connect(mapStateToProps)(connectLink(LinkContainer)),
+  Redirect: connect(mapStateToProps)(connectLink(Redirect))
```
landonreed, line 38: I think these component names would benefit from having some mention of their query param-preserving behavior. E.g.:
```
LinkWithQuery
LinkContainerWithQuery
RedirectWithQuery
```
I'm open to other suggestions for names, but I think it would be good to somehow distinguish them from their vanilla counterparts in usage.

lib/components/form/connected-links.js
```diff
@@ -0,0 +1,39 @@
...
+ * when redirecting the user between the main map and account-related pages,
+ * so that when the user returns to the map, the itinerary that was previously
+ * displayed is shown again.
+ * Implementers only need to specify the 'to' route and
+ * do not need to hook to redux store to retrieve the itinerary search query.
+ * @param RoutingComponent The routing component to enhance.
+ * @returns A new component that passes the redux search params to
+ *          the RoutingComponent's 'to' prop.
+ */
+const connectLink = RoutingComponent =>
```
landonreed, line 19: A better name for this connect method might be something like `connectToQueryParams`.

lib/components/user/account-page.js
```diff
@@ -15,18 +16,25 @@ import SubNav from './sub-nav'
...
-  componentDidMount () {
-    const { loggedInUser, routeTo } = this.props
+  /**
+   * If a user signed up in Auth0 and did not complete the New Account wizard
+   * (and they are not on or have not just left the Terms and Conditions page),
+   * make the user finish set up their accounts first.
+   * monitoredTrips should not be null otherwise.
+   * NOTE: This check applies to any route that makes use of this component.
+   */
+  _checkAccountCreated = () => {
```
landonreed, line 26: I'd rather not bury this logic/routing behavior in a method. Is there a reason for this?

lib/components/user/back-link.js
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Back link that navigates to the previous location in browser history.
+ */
+const BackLink = () => (
+  <Container>
+    <StyledButton
+      bsStyle='link'
+      onClick={navigateBack}
+    >
+      <Icon name='arrow-left' />  Back
```
landonreed, line 24: Given that we're reusing this in a few places, I should point out that I prefer the Icon to have a right margin, rather than separating it from the button text with a space. The reason being that when you hover over the button, there is a portion of whitespace between the icon and text that is underlined. Very minor pet peeve, but a nice detail to fix.

lib/components/user/back-to-trip-planner.js
```diff
@@ -1,14 +1,16 @@
...
 import Icon from '../narrative/icon'
 
 const Container = styled.div``
 
 const BackToTripPlanner = () => (
   <Container>
-    <Link to='/'><Icon name='arrow-left' />  Back to trip planner</Link>
+    <Link to={'/'}>
+      <Icon name='arrow-left' />  Back to trip planner
+    </Link>
```
landonreed, line 13: I'm a little worried about the style for this component deviating from the very similar `back-link.js`. Is there a way to reuse components?

lib/components/user/places/favorite-place-page.js
```diff
@@ -52,7 +42,6 @@ class FavoritePlacePage extends Component {
    * Navigates back to the previous browser location,
    * whether it is the accounts page or another page that linked to the current URL.
    */
-  // TODO: Refine this behavior.
   _handleBack = () => {
     window.history.back()
   }
```
landonreed, line 47: Might be ok just to inline the `window.history.back()` now.

lib/util/user.js
```diff
@@ -15,11 +15,11 @@ export const isWork = loc => loc.type === 'work'
...
 /**
- * An index of common place types, used to obtain FontAwesome icons
- * for place types and to initialize place defaults.
+ * An index of common place types (excluding Home and Work),
+ * each type including a display name and the FontAwesome icon name.
  * Add the supported place types below as needed.
  */
-export const PLACE_TYPES = {
+export const CUSTOM_PLACE_TYPES = {
   custom: {
```
landonreed, line 23: Can this just be an array? I feel like these keys are redundant info.

lib/components/user/places/favorite-places-pane.js
```diff
@@ -1,60 +1,49 @@
...
+  const { savedLocations } = cloneUserDataForEditing(loggedInUser)
+  return (
+    <div>
+      <ControlLabel>Add the places you frequent often to save time planning trips:</ControlLabel>
+
+      {savedLocations.map((place, index) => (
+        <FavoritePlaceRow
+          isFixed={isHomeOrWork(place)}
+          key={index}
+          onDelete={() => deleteUserPlace(index)}
```
landonreed, line 24: Let's use `window.confirm` with `Are you sure you would like to delete ${name} from your saved places?` before actually deleting.

lib/components/user/places/favorite-places-pane.js
```diff
@@ -1,60 +1,49 @@
...
-                  place={place}
-                />
-              )
-              )}
-
-              {/* For adding a location. */}
-              <FavoritePlaceRow path='/places/new' />
-            </>
-          )}
+const FavoritePlacesPane = ({ deleteUserPlace, loggedInUser }) => {
```
landonreed, line 14: I'm not a huge fan of the Pane name. How about FavoritePlacesList?

lib/components/user/places/place-editor.js
```diff
@@ -26,16 +28,17 @@ const FixedPlaceIcon = styled(LargeIcon)`
...
+`
 
 const MESSAGES = {
   SET_PLACE_NAME: 'Set place name'
 }
 
-// Custom places without home or work
-const customPlaces = clone(PLACE_TYPES)
-delete customPlaces.home
-delete customPlaces.work
```
landonreed: +1

lib/components/user/places/place-editor.js
```diff
@@ -75,11 +78,11 @@ class PlaceEditor extends Component {
 
           <FormGroup>
             <ToggleButtonGroup
+              defaultValue={place.type}
```
landonreed, line 81: A couple of styling tweaks. Could you:
- add `bsSize='small'` to ToggleButton?
- change LargeIcon font-size to 125%
- change ToggleButton style to `padding: '5px'` (i.e., change sides from 10px to 5px).

lib/components/user/places/place-editor.js
```diff
@@ -97,23 +100,24 @@ class PlaceEditor extends Component {
         </>
         )}
 
-        <div style={{ display: 'flex' }}>
+        <Flex>
```
landonreed, line 103: Maybe FlexContainer instead or just Container? I don't know why this concerns me but `<Flex>` seems dangerously to a component we might build in the future for GTFS-flex.

lib/components/user/places/place-location-field.js
```diff
@@ -45,6 +47,17 @@ const StyledLocationField = styled(LocationField)`
...
+    border: none;
+    width: 100%;
+  }
+  ${InputGroupAddon} {
+    display: none;
+  }
+  ${MenuItemList} {
+    margin-left: ${props => props.static ? '-50px' : '0px'};
+    position: absolute;
+  }
```
landonreed, line 60: Not sure where this is happening, but I noticed that the dropdown menu items are showing text underlined on hover here, whereas in the main location field they are not. Any idea why? Ideally we would have little difference visually.

landonreed (changes requested):
Phew, this is looking really great. Just have a few more comments to close this out.

lib/actions/user.js
binh-dam-ibigroup: Maybe it makes sense to do that. Let me try that out.

binh-dam-ibigroup: Good call on this. https://github.com/opentripplanner/otp-react-redux/pull/280/commits/d121350ed90b22cfbfea83e84aa256be5e5b4e6f has the location sorting done upfront and simplifies the rest of the code a bit.

lib/components/app/responsive-webapp.js
binh-dam-ibigroup: @landonreed  How about having this route redirect to `account/create/verify`, which would briefly flash and redirect to `account/create/terms` if the email has been verified? (The `Verify my email` would still simply refresh the page.)

lib/components/user/account-page.js
binh-dam-ibigroup: I was trying to address the case when we'd have to check in `componentDidUpdate`, e.g. when someone types in the URL for another account page before they agree on the terms, but I didn't find a good way to redirect without flicker...

lib/components/user/places/place-editor.js
binh-dam-ibigroup: I propose something in between. Your suggestion is a little small.
image

lib/components/user/places/place-location-field.js
binh-dam-ibigroup: Not quite sure how Bootstrap applied extra styles... Fixed in https://github.com/opentripplanner/otp-react-redux/pull/280/commits/436b93da0ef77b9c75628f92d29cf8313085d163 and https://github.com/opentripplanner/otp-react-redux/pull/280/commits/65c087f18c0cb4ef1cbdd84e8c4835a6825c1bcc.

lib/components/user/places/place-editor.js
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/280/commits/436b93da0ef77b9c75628f92d29cf8313085d163.

lib/components/user/places/place-editor.js
binh-dam-ibigroup: See the proposed tradeoff in https://github.com/opentripplanner/otp-react-redux/pull/280/commits/436b93da0ef77b9c75628f92d29cf8313085d163.

lib/components/user/places/favorite-places-pane.js
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/280/commits/436b93da0ef77b9c75628f92d29cf8313085d163.

lib/components/user/places/favorite-places-pane.js
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/280/commits/436b93da0ef77b9c75628f92d29cf8313085d163.

lib/components/user/places/favorite-place-page.js
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/280/commits/436b93da0ef77b9c75628f92d29cf8313085d163.

lib/components/user/back-to-trip-planner.js
binh-dam-ibigroup: The problem is that the back button is a `<button>` with an event handler, and the link straight `<a>`s, so the only reusable piece is the icon.

binh-dam-ibigroup: How about https://github.com/opentripplanner/otp-react-redux/pull/280/commits/6c9e778b2718f014a1a872133a3267371adb4b6f? This commit also updates the icon to have a margin with the link text.

lib/components/form/connected-links.js
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/280/commits/35eb848eeb28b878ba59d7fd07e01b18f92fa260.

lib/components/form/connected-links.js
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/280/commits/35eb848eeb28b878ba59d7fd07e01b18f92fa260.

lib/components/app/responsive-webapp.js
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/280/commits/35eb848eeb28b878ba59d7fd07e01b18f92fa260.

lib/components/app/responsive-webapp.js
binh-dam-ibigroup: Addressed the route for the verify screen in https://github.com/opentripplanner/otp-react-redux/pull/280/commits/2df41d8f67089d1a9dd4d77d29975f807b7b7a6b.

binh-dam-ibigroup:
Ready for review once more.

lib/actions/user.js
landonreed: When I said we should pre-sort on the persistence/API side, I more so meant in otp-middleware, rather than in the UI.

binh-dam-ibigroup: Displaying Home and Work on top is solely a UI feature, so it makes most sense to process the locations on the UI side rather than the backend.

lib/actions/user.js
```diff
@@ -78,13 +79,18 @@ export function fetchAuth0Token (auth0) {
...
+ * placing the Home and Work locations at the beginning of the list
+ * of saved places for rendering in several UI components.
+ *
+ * Also, fetches monitored trips if requested, i.e. when
  * - initializing the user state with an existing persisted user, or
  * - POST-ing a user for the first time.
  */
 function setUser (user, fetchTrips) {
   return function (dispatch, getState) {
+    positionHomeAndWorkFirst(user)
```
landonreed, line 92: I still don't think the client should be concerned with this, but I won't make this blocking so we can move forward. I do think we should add an issue to otp-middleware that handles this sorting before persisting an update and ensures that there is only one home/work location for any given user.

lib/components/app/responsive-webapp.js
```diff
@@ -231,17 +235,24 @@ class RouterWrapperWithAuth0 extends Component {
               ]}
               render={() => <WebappWithRouter {...this.props} />}
             />
+            <Route
+              component={FavoritePlaceScreen}
+              path={`${PLACES_PATH}/:id`}
```
landonreed, line 240: When creating a place in the context of a new account, I think we should avoid placing the My account header content at the top of the page. Doing so takes the user out of the create account context and also shows the Trips/Settings links, which should not be available at this stage. I don't want to overcomplicate this though... thoughts on a clean approach here?

image

landonreed (changes requested):
Just one main thing: the 'My account' page header should not show when adding a place while creating account. Perhaps we could use a path like `account/create/places/0`.

lib/components/app/responsive-webapp.js
binh-dam-ibigroup: This is now addressed in https://github.com/opentripplanner/otp-react-redux/pull/280/commits/70f449c2c26d870e4d2a4e6c3a7e77bd4fa5b63a. Per chat, the Back/Save buttons are harmonized in https://github.com/opentripplanner/otp-react-redux/pull/280/commits/411caf13976af1e18b2f2b18a255f4f445547838.

lib/actions/user.js
binh-dam-ibigroup: Noted in https://github.com/ibi-group/otp-middleware/issues/136.

binh-dam-ibigroup:
Alright, ready for review again...

lib/components/user/places/place-editor.js
```diff
@@ -113,11 +113,14 @@ class PlaceEditor extends Component {
...
               inputPlaceholder={isFixed ? `Search for ${place.name} location` : 'Search for location'}
+              // Add a flag for mobile views to adjust dropdown menu position/width.
+              // The component's 'static' prop is not suitable for mobile views here, as it would
+              // show the list of places over the Back/Save buttons, making these buttons unreachable.
+              isMobile={isMobile()}
               location={makeLocationFieldLocation(place)}
               locationType='to'
               onLocationSelected={this._handleLocationChange}
               showClearButton={false}
-              static={isMobile()}
```
landonreed: Sigh... I forgot that showing the field in mobile view is why we opted for the buttons above instead of below. We may want to change this so that in mobile mode the buttons render at the top of the screen and in desktop they're below the location field.

binh-dam-ibigroup: I'll add that. The other thing we could do in mobile (in another PR) is to flush the back/save buttons to the bottom of the screen so that they are always shown, but that may be also be tricky...

binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/280/commits/326199f4e5bdd0e8a67538c01021973b558fc316.

landonreed (approved):

evansiroky (approved):

PR #283, opened 2020-12-08
User locations in main panel

binh-dam-ibigroup:
This PR makes the main panel display the logged-in user's saved locations, including home and work, along with existing favorite and recent locations searched on the browser.

image


lib/components/form/user-settings.js
```diff
@@ -31,9 +33,16 @@ class UserSettings extends Component {
...
 
-  _getLocations = (user) => {
-    const locations = [...user.locations]
-    if (!locations.find(l => l.type === 'work')) {
+  _getLocations = () => {
+    const { loggedInUser, user } = this.props
+    const locations = clone(user.locations)
+    const savedLocations = loggedInUser ? clone(loggedInUser.savedLocations) : []
+    // Identify saved locations for non-'forgettability' purposes.
+    savedLocations.forEach(l => { l.origin = 'loggedInUser' })
```
landonreed, line 41: I'm confused as to the difference between `savedLocations` and `locations`? Are you merging the middleware persisted locations and the localstorage locations? That's not the goal here: we want to only allow one or the other to be enabled for any given deployment (depending on the config file's persistence strategy).

landonreed (changes requested):
@binh-dam-ibigroup, I think there's been some confusion about how the main panel should handle localstorage vs. middleware locations. See comment.

lib/components/form/user-settings.js
binh-dam-ibigroup: I can make the `user` and `loggedInUser` locations exclusive.

landonreed: @binh-dam-ibigroup, it is probably appropriate to handle this at the reducer level. This component should only know about a generic user's locations. In other words, it should not have to determine by itself which set of locations to choose from, the reducer (and/or startup actions) should handle determining which persistence strategy is defined and set the user object from localstorage/middleware accordingly.

binh-dam-ibigroup: > This component should only know about a generic user's locations...

In that case, should the locations be just a prop passed to the component depending on loggedInUser and user?

binh-dam-ibigroup:
Blocking this PR as it was agreed that `state.otp.user` and corresponding actions should eventually migrate and be merged with `state.user` that captures logged-in user data.

PR #284, opened 2020-12-09
Encapsulate code to connect LocationField to redux store

binh-dam-ibigroup:
This P demonstrates a way to encapsulate the common code to connect styled `LocationField` instances to the redux store.
@landonreed let me know what you think of this approach.


landonreed (changes requested):
Looks good. Just a couple of tweaks in https://github.com/opentripplanner/otp-react-redux/pull/285

lib/components/form/connect-location-field.js
```diff
@@ -0,0 +1,51 @@
...
+      currentPosition,
+      geocoderConfig: config.geocoder,
+      nearbyStops,
+      sessionSearches,
+      showUserSettings: getShowUserSettings(state.otp),
+      stopsIndex: transitIndex.stops,
+      userLocationsAndRecentPlaces: [...user.locations, ...user.recentPlaces]
+    }
+    // Leave stateToProps.location completely unset unless includeLocation is true.
+    if (includeLocation) {
```
binh-dam-ibigroup, line 36: @landonreed It is still necessary to leave the `location` prop unset, otherwise the `IntermediatePlaceField` and `FavoriteLocation` components will use the fixed `null` value as location.

landonreed: Ah gotcha, thanks for the info. Could you include that in the comment above?

landonreed (approved):
Looks great, just one comment about updating that comment re `includeLocation`.

lib/components/form/connect-location-field.js
binh-dam-ibigroup: Yes, updated in https://github.com/opentripplanner/otp-react-redux/pull/284/commits/cae2ba746809b92e8b103378e5752c2c385ec492.

PR #289, opened 2020-12-11
Batch routing form

landonreed:
Add batch routing form elements (settings, date/time, mode buttons).

This is visible at: https://fdot-otp-dev.s3.amazonaws.com/index.html#/

Note: the settings panel still needs to be modified a bit to align the settings with the business rules for the batch routing approach, but this likely needs additional design. The intent of this PR is to build out the core functionality (further refinements are expected).

image


lib/components/form/mode-buttons.js
```diff
@@ -0,0 +1,75 @@
...
+  onClick,
+  selectedModes = [],
+  style = {}
+}) => {
+  const {ModeIcon} = useContext(ComponentContext)
+  return MODE_OPTIONS.map((item, index) => {
+    let buttonStyle = style
+    // Add end style to button style if rendering the last item.
+    if (index === MODE_OPTIONS.length - 1) {
+      buttonStyle = {...buttonStyle, ...endStyle}
```
binh-dam-ibigroup, line 42: Use styled-components for this. (This is the same remark as in the batch routing panel.) In doing so, there is no need to define in the JS code a separate style for the last button.

binh-dam-ibigroup: But the layout is very nice.

lib/components/form/mode-buttons.js
```diff
@@ -0,0 +1,75 @@
...
+        style={buttonStyle}
+        title={item.label}
+      >
+        {item.icon
+          ? <Icon className='fa-2x' type={item.icon} />
+          : <ModeIcon height={25} mode={item.mode} />
+        }
+        {isSelected &&
+          <Icon
+            style={{
```
binh-dam-ibigroup, line 62: Use styled-components.

binh-dam-ibigroup (changes requested):
The subscribing to `window.onresize` bothers me a bit (see an alternate approach in the comments). Most styling should also move to styled-components. The layout looks very nice though.

lib/components/app/batch-routing-panel.js
```diff
@@ -3,20 +3,106 @@ import { Button } from 'react-bootstrap'
...
+  componentDidMount () {
+    this.updateWindowWidth()
+    window.addEventListener('resize', this.updateWindowWidth)
+  }
+
+  componentWillUnmount () {
+    window.removeEventListener('resize', this.updateWindowWidth)
+  }
+
+  updateWindowWidth = () => this.setState({width: window.innerWidth})
```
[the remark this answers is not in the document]
landonreed: We do this elsewhere in projects (e.g., Data Tools). Agreed that media queries might provide a good alternative, but that would change how we're passing styles to narrative-itineraries for the CallTaker panel. Out of scope for now.

landonreed:
OK, styled-components have been put in place for the majority of cases noted. The one exception is narrative-itineraries, which is out of scope of this PR (requires changes to call-taker module).

lib/components/app/batch-routing-panel.js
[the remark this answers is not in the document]
binh-dam-ibigroup: I submitted a potential replacement in #298, please consider that for this PR.

lib/components/app/styled.js
```diff
@@ -0,0 +1,94 @@
...
+  }
+`
+
+const expandableBoxCss = css`
+  background-color: rgb(239, 239, 239);
+  box-shadow: rgba(0, 0, 0, 0.32) 7px 12px 10px;
+  height: 245px;
+  border-radius: 5px 5px 5px 5px;
+  position: absolute;
+  width: 96%; // FIXME: come up with a better value
```
binh-dam-ibigroup, line 74: Since this is using absolute positioning, you could do:
```
left: 10px;
right: 10px;
```

lib/components/form/batch-settings-panel.js
```diff
@@ -0,0 +1,111 @@
+// import {DropdownSelector} from '@opentripplanner/trip-form'
```
binh-dam-ibigroup, line 1: Remove commented import.

lib/components/app/styled.js
```diff
@@ -0,0 +1,94 @@
+import styled, {css} from 'styled-components'
+
+import DateTimePreview from '../form/date-time-preview'
+
+const activeCss = css`
+  background: #e5e5e5;
+  -webkit-box-shadow: inset 0px 0px 5px #c1c1c1;
+     -moz-box-shadow: inset 0px 0px 5px #c1c1c1;
+          box-shadow: inset 0px 0px 5px #c1c1c1;
+   outline: none;
```
binh-dam-ibigroup, line 10: Non blocking: some contrast or a different color would help distinguish pushed settings and date buttons from the default state.

lib/components/form/date-time-preview.js
```diff
@@ -19,27 +19,32 @@ class DateTimePreview extends Component {
...
     const {
       caret,
+      className,
       date,
-      editButtonText,
-      time,
+      dateFormat,
       departArrive,
+      endTime,
+      editButtonText,
```
binh-dam-ibigroup, line 41: This should be above `endTime`.

lib/components/form/date-time-preview.js
```diff
@@ -63,19 +68,22 @@ class DateTimePreview extends Component {
       </div>
     )
 
-    const button = (
-      <div className='button-container'>
+    const button = hideButton
```
binh-dam-ibigroup, line 71: `button = !hideButton && (<div... >)`

lib/components/app/styled.js
```diff
@@ -0,0 +1,94 @@
+import styled, {css} from 'styled-components'
+
+import DateTimePreview from '../form/date-time-preview'
+
+const activeCss = css`
+  background: #e5e5e5;
+  -webkit-box-shadow: inset 0px 0px 5px #c1c1c1;
```
binh-dam-ibigroup, line 7: I would extract the common shadow CSS value.

binh-dam-ibigroup (approved):
Looks great! Conditional approval subject to
- merging #298,
- properly sizing the date/settings popups and other minor comments.

lib/components/form/batch-settings-panel.js
landonreed: We may end up using these selectors instead, so I'd rather leave the reference instead of having to dig up where that component lives (it took a few minutes last time).

lib/components/form/date-time-preview.js
landonreed: Going to leave this how it is. This is just a matter of preference and we use both throughout the app.

landonreed:
This doesn't have two approvals, but we really need to move this code forward. I'm going to go ahead with merging.

PR #292, opened 2020-12-15
Stop viewer new schedules

binh-dam-ibigroup:
This PR fixes #286 and fixes #287 as follows:
- The schedule view by pattern is replaced with a `StopScheduleTable` sorted by departures across all patterns.
- The schedule view scrolls to the first departure from now ~(it will be underneath the stop viewer header)~ and highlights it.
- The underlying API call is changed to use the `/date/` OTP endpoint, removing stop times previously included from the previous service day.
- If `stoptime.headsign` returned by OTP are blank, they are populated using the pattern description from OTP.
- Some common routines are refactored in `utils/viewer.js`, including a new `StopTimeCell` component used by `PatternRow`
- A time zone indicator is shown if user is not in the home timezone config param.
- 

image

Remaining issues
- getStopTimeFormatted needs to be refactored as a component => ~working on this next~ refactored in https://github.com/opentripplanner/otp-react-redux/pull/292/commits/c7e23684606c801429e3ee9b12499f81641293a7.
- The code for schedule view remains in `PatternRow`, will get removed eventiually. => Removed.
- The scroll zone should probably start just under the stop viewer header. Currently, it hides a portion of the scroll zone (the scrollbars extend all the way to the top... any reason to leave as is right now? => Addressed.

image

PR #294, opened 2020-12-22
Stop viewer/itinerary body status

binh-dam-ibigroup:
This PR is in response to https://github.com/opentripplanner/otp-react-redux/pull/292#discussion_r545269646 and does the following refactors:
- Moves the code to draw a status label from `utils/viewer` into a new `BaseStatusLabel` component.
- Refactors code for determining on-time/late/... trip status common to `BaseStatusLabel` and `RealTimeTimeColumn`.
- Simplifies `RealTimeTimeColumn` render code.


landonreed (changes requested):
This generally looks pretty good, but there were some opportunities for encapsulating this logic a bit more into the label component. I was having trouble expressing, so I just made a PR: #296.

landonreed (approved):

lib/util/viewer.js
```diff
@@ -185,3 +157,35 @@ export function getModeFromRoute (route) {
...
+    if (delaySeconds > 60) {
+      // late departure
+      return REALTIME_STATUS.LATE
+    } else if (delaySeconds < -60) {
+      // early departure
+      return REALTIME_STATUS.EARLY
+    } else {
+      // on-time departure
+      return REALTIME_STATUS.ON_TIME
+    }
```
evansiroky, line 186: Consider basing this PR off of #305 and then refactor this function so that it uses the configurable on-time threshold.

evansiroky: Actually, since #305 was introduced after the latest commit here, I think it might be unfair to request this. Therefore, I'm going to merge this and make #305 dependent on this.

evansiroky (approved):

PR #295, opened 2020-12-29
refactor(TripStatus): Refactor rendering methods

binh-dam-ibigroup:
This PR combines the logic to determine what to render that was split across multiple rendering methods.


evansiroky:
I ended up doing a larger refactor in f60e285 and 93f0e5d, so I'm going to close this.

PR #296, opened 2021-01-04
refactor(realtime-status-label): encapsulate more of the status label…

landonreed:
Provides some changes for #294.


lib/components/viewers/realtime-status-label.js
```diff
@@ -0,0 +1,95 @@
+import { formatDuration } from '@opentripplanner/core-utils/lib/time'
+import React from 'react'
+import styled from 'styled-components'
+
+import { getTripStatus, REALTIME_STATUS } from '../../util/viewer'
+
+// If shown, keep the '5 min' portion of the status string on the same line.
+export const DelayText = styled.span`
+  white-space: nowrap;
```
binh-dam-ibigroup, line 9: Almost! I will add a spacer after this element so the rendering is correct in the stop viewer.

image

binh-dam-ibigroup (approved):
I will make further changes to #294.

PR #298, opened 2021-01-05
Replace window event listeners with css @media query

binh-dam-ibigroup:
This PR replaces code that sets up `window.onresize` event handlers with CSS `@media` queries for element positioning.

### DIFF
--- a/lib/components/app/batch-routing-panel.js
+++ b/lib/components/app/batch-routing-panel.js
@@ -211,20 +211,19 @@ class BatchRoutingPanel extends Component {
           !activeSearch && showUserSettings &&
           <UserSettings />
         */}
-        <div className='desktop-narrative-container'>
-          <NarrativeContainer>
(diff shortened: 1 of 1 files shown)


lib/components/app/batch-routing-panel.js
```diff
@@ -207,22 +212,18 @@ class BatchRoutingPanel extends Component {
           <UserSettings />
         */}
         <div className='desktop-narrative-container'>
```
landonreed, line 214: It might make more sense just to replace the `div` above with the new NarrativeContainer and keep the `className` prop intact.

landonreed (approved):
Looks good. Just one minor tweak for conditional approval.

I suppose doing things this way avoids the need for changes to the CallTaker Panel usage of `NarrativeItineraries`, which I had previously thought. Thanks for the rec!

PR #300, opened 2021-01-06
Replace TravisCI with GitHub Actions

evansiroky:
* removes TravisCI file from repo
* adds GitHub Actions workflow for building and deploying
* also adds spellchecking GitHub Action for the fun of it


.github/workflows/codespell.yml
```diff
@@ -0,0 +1,16 @@
+# GitHub Action to automate the identification of common misspellings in text files.
```
binh-dam-ibigroup, line 1: This is useful! a few source files have been flagged already.

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,16 @@
...
+# https://github.com/codespell-project/actions-codespell
+# https://github.com/codespell-project/codespell
+name: codespell
+on: [push, pull_request]
```
binh-dam-ibigroup, line 5: This was the case before, but it looks we are still running twice the same test on each push.
Could this simply be `on: pull_request`?

According to [this](https://frontside.com/blog/2020-05-26-github-actions-pull_request/), the workflow is triggered when a pull request is `synchronize`d, including pushes to the pull request. The `push` entry would no longer be necessary as a result.

.github/workflows/node-ci.yml
```diff
@@ -0,0 +1,59 @@
+name: Node.js CI
+
+on: [push, pull_request]
```
binh-dam-ibigroup, line 3: Could this simply be `on: pull_request`? (See my comment above)

binh-dam-ibigroup (changes requested):
Could we try to not duplicate the same runs and execute the builds/tests `on: pull_request` only?

Also please advise on the need to run tests on multiple node versions.

.github/workflows/node-ci.yml
evansiroky: We definitely need this to run on some pushes so that semantic-release will run. This is primarily for when it runs on the master branch, but with semantic-release v16 it could be on branches with other names so long as those were configured such that semantic-release would perform releases on those branches.

.github/workflows/codespell.yml
evansiroky: I suppose we could just have this run for PRs. I mainly put this here just to try it out, so we'll have to see if it's needed and if we want to have it raise an error for misspellings.

.github/workflows/node-ci.yml
```diff
@@ -0,0 +1,59 @@
...
+
+jobs:
+  test-and-build:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [10.x, 12.x]
```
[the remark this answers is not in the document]
evansiroky: It's good to be checking for compatibility with different versions so long as we want to support each version. I am OK with only supporting one node version at this time though.

evansiroky:
@binh-dam-ibigroup see responses.

.github/workflows/node-ci.yml
binh-dam-ibigroup: Then how about
```
on:
  push:
    branches:
      - master
  pull_request:
```

(see [reference](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows))

evansiroky:
I went ahead and reduced the amount of GitHub Actions that run by only having the Actions run on PRs or pushes to master.

binh-dam-ibigroup (approved):
Nice! The checks results section looks much less cluttered.

.github/workflows/node-ci.yml
```diff
@@ -0,0 +1,41 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js 12.x
+        uses: actions/setup-node@v1
+        with:
+          node-version: 12.x
+      - name: Install npm packages using cache
+        uses: bahmutov/npm-install@v1
```
landonreed, line 21: Should this not use `yarn install`?

landonreed (approved):
Looks good! Just one question on the `npm install` step.

.github/workflows/node-ci.yml
evansiroky: It does. See https://github.com/opentripplanner/otp-react-redux/pull/300/checks?check_run_id=1658566171#step:4:14.

PR #305, opened 2021-01-14
Add configurable on-time value

evansiroky:
This adds the option to set a config value for an on-time threshold in seconds. The default is set to 60 seconds. If any real-time arrival or departure times have a delay more than this value, then they will be considered late. If any real-time arrival or departure times have a delay of less than this negative value, then they will be considered early. Everything else is considered on-time. So far this affects the RealtimeColumn shown in itinerary bodies, the PatternRow shown in the stop viewer and for the TripStatus for monitored trips.


lib/reducers/create-otp-reducer.js
```diff
@@ -71,6 +71,7 @@ export function getInitialState (userDefinedConfig, initialQuery) {
     transitOperators: [],
     realtimeEffectsDisplayThreshold: 120,
     routingTypes: [],
+    onTimeThresholdDeltaSeconds: 60,
```
binh-dam-ibigroup, line 74: `example-config.yml` should include a commented-out line for this new config parameter.

binh-dam-ibigroup (approved):
I think there should be an entry (commented out) in `example-config.yml` for the new parameter, otherwise looks good.

evansiroky:
Just brought in the latest to dev and refactored a bit so that it works with the changes in https://github.com/opentripplanner/otp-react-redux/pull/294.

binh-dam-ibigroup (approved):
Still looks good.

landonreed (approved):
LGTM, but I do have one non-blocking suggestion: change the name of `onTimeThresholdDeltaSeconds` to `onTimeThresholdSeconds`. Threshold delta feels a bit wordy/redundant. Even if it's technically accurate, I think it might sow more confusion than `onTimeThresholdSeconds`.

PR #313, opened 2021-01-21
Fetch trips on user persist

### LINKED ISSUE #312: Blank page shown when trying to save trips
A blank page, only displaying "Processing...", is shown (and the UI stops responding) when trying to save trips or accessing the saved trips screen after creating a new account.

Steps to recreate:
1. Plan trip
2. Create new account
3. App redirects user to previously planned trip
4. Click Save trip
5. Processing.... freeze

Alternate steps to reproduce:
1. Create new account but return to trip planner before accepting terms (press browser back button)
2. Either:
* Plan a trip, then. Click Save Trip
* Navigate to `http://<host:port>/#/savedtrips`
3. Processing.... freeze

binh-dam-ibigroup:
This PR fixes #312 with the following changes:
- Monitored trips are now fetched after POST-ing a new user (in addition to 
- If the user did not finish setting up their account, and they attempt to save a new trip or access the Saved Trips page, the user is redirected to the accounts page.

Internally:
- Separate actions for setting user state and access token
- The action for setting the user state has option to also fetch the user's monitored trips.

### DIFF
--- a/lib/actions/user.js
+++ b/lib/actions/user.js
@@ -9,6 +9,7 @@ const API_MONITORED_TRIP_PATH = '/api/secure/monitoredtrip'
 const API_OTPUSER_PATH = '/api/secure/user'
 const API_OTPUSER_VERIFY_SMS_SUBPATH = '/verify_sms'
 
+const setAccessToken = createAction('SET_ACCESS_TOKEN')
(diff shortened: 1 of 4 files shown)


landonreed:
Closing in favor of #316.

PR #314, opened 2021-01-22
Update README with deploy steps

landonreed:


README.md
```diff
@@ -18,6 +18,25 @@ Install the dependencies and start a local instance using the following script:
...
+3. Build the js/css bundle (which will appear in the `dist/` directory) and upload.
+
+Here's an example set of commands using the AWS CLI:
+```bash
+# The first two steps only need to be performed the first time.
+aws s3 cp index.html s3://my-bucket --acl public-read
+aws s3 cp example.css s3://my-bucket --acl public-read
+# Anytime the otp-react-redux code changes, you'll need to run the following steps:
+yarn build
+aws s3 cp --recursive dist/ s3://my-bucket/dist --acl public-read
```
binh-dam-ibigroup, line 34: It might be better IMO to tell the reader to upload the files organized as follows:
- index.html (modified to point to `dist/index.js`
- example.css
- dist
    - index.js
    - index.js.map
    - index.css
    - index.css.map

Or even this, and just upload the contents of dist folder:
- dist
    - index.html (modified to point to `index.js`
    - example.css
    - index.js
    - index.js.map
    - index.css
    - index.css.map

binh-dam-ibigroup (approved):
See alternate description.

README.md
```diff
@@ -18,6 +18,25 @@ Install the dependencies and start a local instance using the following script:
 yarn start
 ```
 
+## Deploying the UI
+
+1. Change `index.html` to point to `dist/index.js` (instead of `example.js`) and upload to s3 bucket (or other file hosting service).
```
evansiroky, line 23: This seems a little too detailed and oriented towards deploying to AWS S3. I think it would be sufficient to just have this section be about how to load an index.html file with the built assets.

evansiroky (changes requested):

README.md
landonreed: I don't really see what the harm is in including good instructions on how this could actually be used in the real world with AWS.

evansiroky: Because:

1. It's code that needs to be maintained over time. 
2. I don't believe that TriMet uses AWS to host these files.
3. It adds unnecessary instructions to the README that will probably be not useful people who don't use AWS.
4. Deploying like this should be relatively common knowledge for experienced developers and they may even use slightly different or even more advanced AWS settings if they use AWS.

binh-dam-ibigroup (approved):
Approval stands.

README.md
```diff
@@ -20,22 +20,18 @@ yarn start
...
+3. Upload the following files to wherever you're deploying the UI:
+  - `index.html` (modified to point to `dist/index.js`)
+  - `example.css`
+  - `dist/`
+    - `index.js`
+    - `index.js.map`
+    - `index.css`
+    - `index.css.map`
+
+Note: only contents produced during build in the `dist/` directory are likely to change over time (the `index.html` and `example.css` files contain minimal code), so subsequent deployments will typically only need to replace the `dist/` contents.
```
fpurcell, line 34: The instructions are good, Landon.
That said, not sure what to do with call-config.js 

When build was under trimet-mod, this is what I was doing: yarn build --env.JS_CONFIG=$BUILDER/config/call-config.js --env.YAML_CONFIG=$BUILDER/config/config.yml

fpurcell: BTW, I'm also not clear on the whole fieldtrip module.  And also the trinet auth workflow (not that that needs to go here).

evansiroky (approved):

README.md
fpurcell: BTW, Landon explained call-config.js (no purpose with OTP-RR ... was a remnant from trimet-MOD project).  Also have the call app running in our test environ, complete with trinet auth and connections to the call db.

landonreed: @fpurcell, I'll add a brief section pointing people to deployment instructions for trimet-mod-otp. I think that might be helpful for future users of this product.

PR #316, opened 2021-01-22
Fixes to Save Trip routing, check terms

binh-dam-ibigroup:
This PR fixes routing for the Save Trip button and also redirects user to account creation if they try to save a trip or access the trips page and if for some reason they have not accepted the terms yet.

Supersedes #313.


lib/components/user/monitored-trip/saved-trip-list.js
```diff
@@ -10,25 +10,14 @@ import * as userActions from '../../../actions/user'
...
 import withLoggedInUserSupport from '../with-logged-in-user-support'
 
 /**
  * This component displays the list of saved trips for the logged-in user.
  */
 class SavedTripList extends Component {
-  componentDidMount () {
-    const { loggedInUser, routeTo } = this.props
-
-    if (!loggedInUser.hasConsentedToTerms) {
```
binh-dam-ibigroup: This check should remain because this component can be directly accessed from the URL `/account/trips` even if the user did not finish setting up their account.

lib/components/user/account-page.js
```diff
@@ -1,26 +1,67 @@
...
-      </Row>
-    </div>
-  </div>
-)
+class AccountPage extends Component {
+  componentDidMount () {
+    const { loggedInUser, routeTo } = this.props
+
+    if (!loggedInUser.hasConsentedToTerms) {
+      // If a user signed up in Auth0 and did not complete the New Account wizard
```
binh-dam-ibigroup, line 22: Could you add a comment saying this check applies to all routes that use the component?

lib/components/user/monitored-trip/saved-trip-list.js
binh-dam-ibigroup: Ok, please disregard.

lib/components/user/account-page.js
landonreed: Added in https://github.com/opentripplanner/otp-react-redux/pull/316/commits/032070e4e2ad1933a64dda4690310a539fb64cdf

PR #321, opened 2021-01-28
Fix scrolling and overflow in route viewer.

### LINKED ISSUE #320: Route Viewer scrolling issues
Incorrect overflow handling can occur in the Route Viewer if a route has a long label,
resulting in the behavior shown in screenshots below. Also, the lines that separate the routes should
be flush to the edge of the `route-viewer-body` container.
(To reproduce, open the Route Viewer and reduce the width of the browser.)

image

image

binh-dam-ibigroup:
This PR fixes #320.

EDIT: Per request at https://github.com/opentripplanner/otp-react-redux/pull/321#pullrequestreview-583824022, long clipped route labels in the route viewer are rendered with an ellipsis (this changes the rendering code for the route viewer to take advantage of flex boxes).

image

### DIFF
--- a/lib/components/viewers/route-viewer.js
+++ b/lib/components/viewers/route-viewer.js
@@ -113,13 +113,13 @@ const StyledRouteRow = styled.div`
 `
 
 const RouteRowButton = styled(Button)`
-  padding: 8px;
+  align-items: center;
+  display: flex;
+  padding: 6px;
   width: 100%;
 `
 
-const RouteRowElement = styled.div`
-  display: inline-block;
-  vertical-align: middle;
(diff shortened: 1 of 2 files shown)


evansiroky (approved):

landonreed (changes requested):
This is on the right track, but we need to add css to some inner div (not sure which) to get an ellipsis on overflow. Also, the title prop should be populated with the route name value.

landonreed (approved):

landonreed:
@binh-dam-ibigroup, I'm going to go ahead and merge. I don't think the changes since @evansiroky's review really warrant another look from him.

PR #230, opened 2020-09-11
Field trip module

landonreed:
This PR has been updated to contain the following field trip module functionality:
- viewing/filtering field trip requests
- rendering field trip details
- planning trips from request details
- notes functionality (create, delete)
- updating payment/group size info
- cancel request button

It does NOT currently have save trip functionality or splitting groups into specific trips. That will come in another PR (because it is very complex).

PR #326, opened 2021-02-08
Redirect to hash URL if search URL contains no hash

### LINKED ISSUE #325: Web app cannot handle query params if hash (#) is missing
otp-rr uses react-router with hash routing, but currently if a user visits http://localhost:9966?fromPlace=portland, the search will not be interpreted correctly. This leads to unexpected results and an error/limbo state in the app that causes issues in some cases (one example is in the call taker module, where the app loads, but it cannot plan trips).

landonreed:
Fixes #325. To test:
1. Use trimet call config
2. Go to http://localhost:9966?sessionId=test
3. Auto-redirects to http://localhost:9966/#/?sessionId=test
3. Trips are now plannable (previously things errored out in the app).


landonreed:
Closing in favor of #327 (using a branch not based on #230).

PR #328, opened 2021-02-10
field-trip-qbd

binh-dam-ibigroup:
This PR uses Formik to simplify event/form state handling for the EditableSection component.

### DIFF
--- a/lib/components/admin/editable-section.js
+++ b/lib/components/admin/editable-section.js
@@ -25,6 +25,12 @@ export default class EditableSection extends Component {
 
   _onClickSave = data => {
     const {request, onChange} = this.props
+    // Convert all null values received to '',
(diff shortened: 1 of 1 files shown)


landonreed (changes requested):
Generally looks good, but when updating (e.g.) the payment fields, this PR will write a value of "null" to the back end. (empty values should be showing as N/A -- see how this works in #230).

binh-dam-ibigroup:
> Generally looks good, but when updating (e.g.) the payment fields, this PR will write a value of "null" to the back end. (empty values should be showing as N/A -- see how this works in #230).

@landonreed Try the changes from https://github.com/opentripplanner/otp-react-redux/pull/328/commits/4caafb011663dccc4784b55b4628044ab91f8221, feel free to make further changes after merging.

landonreed (approved):

PR #332, opened 2021-02-19
Add mobile search screen for batch routing

landonreed:
This PR partially accomplishes mobile compatibility for batch routing by updating the search screen to:

image


lib/components/mobile/main.js
```diff
@@ -69,8 +71,13 @@ class MobileMain extends Component {
         )
 
       case MobileScreens.SEARCH_FORM:
+        // Render batch search screen if batch routing enabled. Otherwise,
+        // default to standard search screen.
+        const SearchScreen = isBatchRoutingEnabled(config)
+          ? BatchSearchScreen
```
evansiroky, line 77: This kind of goes against the idea that the implementing application is able to choose between either the DefaultMainPanel and the BatchRoutingPanel. In here, the BatchSearchScreen and all its associated dependencies are always imported even if they aren't supposed to be used. Perhaps the SearchScreen can be made into a required prop that must be set by the implementing app.

lib/components/app/batch-routing-panel.js
```diff
@@ -160,52 +47,7 @@ class BatchRoutingPanel extends Component {
...
-        }
-        {expanded === 'SETTINGS' &&
-          <BatchSettingsPanelContainer>
-            <BatchSettingsPanel />
-          </BatchSettingsPanelContainer>
-        }
+        <BatchSettings />
         {/* FIXME: Add back user settings (home, work, etc.) once connected to
             the middleware persistence.
           !activeSearch && showUserSettings &&
```
evansiroky, line 53: Can remove below TODO comment about implementing mobile view.

evansiroky:
There's a lot of unused space in the date/time section.

Screen Shot 2021-02-19 at 5 10 00 PM

It'd look better, especially in mobile, for this area to only go as far down as just below the last element.

evansiroky (changes requested):
See comments. Also, one of the GitHub Actions is failing.

lib/components/mobile/main.js
landonreed: Yea, agreed. I had this thought also. Thanks for flagging. I was considering whether SearchScreen should be a required prop or if we should have `mobileView` be configurable in trimet-mod-otp and have a separate file from `main.js` for batch routing that handles the functionality/logic from this file. Right now, there's only minor differences between the two (the search screen), but the option to define different `mobileViews` in trimet-mod-otp might be nice to have. Thoughts?

landonreed:
@evansiroky, thanks for the comments. Quick question for you above regarding the mobile search screen.

binh-dam-ibigroup (approved):
LGTM. I'll go with @evansiroky's comments and give a pass on props ordering.

lib/components/mobile/main.js
evansiroky: After looking at this more, I'm not sure why there are even slots for `desktopView` and `mobileView` at all, especially now that we're allowing the configuration of custom components via the context API. The example.js and trimet-mod-otp files are nearly identical in how they render all this. Also, all the stuff in `desktopView` and `mobileView` is super connected to all the otp-react-redux internals that I don't think you'd ever want to define a cust mobileView and if you did, you'd just be writing a different app anyways.

I've gone ahead and refactored `ResponsiveWebapp` so that it does away with the `desktopView` and `mobileView` slots completely and now it relies more on the Context API for the things we did want to customize. See https://github.com/opentripplanner/otp-react-redux/pull/337.

evansiroky (changes requested):
See proposed changes in https://github.com/opentripplanner/otp-react-redux/pull/337.

evansiroky:
I've merged in #337 and also fixed the ItineraryCarousel to properly show the expanded ItineraryBody.

lib/components/form/batch-settings.js
```diff
@@ -0,0 +1,202 @@
...
+import React, { Component } from 'react'
+import { connect } from 'react-redux'
+import styled from 'styled-components'
+
+import * as apiActions from '../../actions/api'
+import * as formActions from '../../actions/form'
+import BatchPreferences from './batch-preferences'
+import DateTimeModal from './date-time-modal'
+import ModeButtons, {MODE_OPTIONS, StyledModeButton} from './mode-buttons'
+// import UserSettings from '../form/user-settings'
```
binh-dam-ibigroup, line 11: Let's remove this import.

binh-dam-ibigroup (approved):
Approval stands,  just one tiny item to clean up.

PR #334, opened 2021-02-23
fix(ResponsiveWebapp): Do not reformat URL after auth0 login.

### LINKED ISSUE #333: Unable to login after merging #327
The login status is not updated, and user is not able to access account settings after merging of #327 if persistence strategy is `otp_middleware`. This happens because the auth0 library cannot read the login state.

Observed behavior:
- After login, URL appears in the format `http://localhost:9966/#/?code=xxxxxxx&state=yyyyyyyyy`

Expected behavior:
- After login, URL appears in the format `http://localhost:9966/?code=xxxxxxx&state=yyyyyyyyy`

To reproduce:
- Log out.
- Log in.
- The avatar on the upper-right corner is not displayed.

binh-dam-ibigroup:
Fixes #333 by not reformatting URLs in `/?code=xxxx...`.

### DIFF
--- a/lib/components/app/responsive-webapp.js
+++ b/lib/components/app/responsive-webapp.js
@@ -4,6 +4,7 @@ import { createHashHistory } from 'history'
 import isEqual from 'lodash.isequal'
 import coreUtils from '@opentripplanner/core-utils'
 import PropTypes from 'prop-types'
+import qs from 'qs'
 import React, { Component } from 'react'
 import { connect } from 'react-redux'
 import { Route, Switch, withRouter } from 'react-router'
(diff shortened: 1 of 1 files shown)


landonreed (approved):
Looks good! Thanks for diagnosing.

Only non-blocking comments are that:
1. I'm not a huge fan of the extra whitespace, and especially the commented line with no text.
2. It wouldn't be a huge lift to also check that search contains `&state=`, which I think might give us more assurance that we're not interfering with Auth0 login -- but I don't feel too strongly about this.

binh-dam-ibigroup:
> Looks good! Thanks for diagnosing.
> 
> Only non-blocking comments are that:
> 
> 1. I'm not a huge fan of the extra whitespace, and especially the commented line with no text.
> 2. It wouldn't be a huge lift to also check that search contains `&state=`, which I think might give us more assurance that we're not interfering with Auth0 login -- but I don't feel too strongly about this.

1. I removed the blank line in the comment, but would rather visually separate chunks of comments/logic in the code.
2. I added checking presence of the `state` search param, now using `qs` to parse the search query.

PR #335, opened 2021-02-23
Add 'delete my account' button

landonreed:
This PR adds a delete my account button the user account settings page.

After deletion, the user is logged out (and redirected to the application root URL).

image

### DIFF
--- a/lib/components/user/user-account-screen.js
+++ b/lib/components/user/user-account-screen.js
@@ -50,6 +50,16 @@ class UserAccountScreen extends Component {
     this._updateUserPrefs(userData, true)
   }
 
+  _handleDeleteUser = (evt) => {
+    const {auth0, deleteUser, loggedInUser} = this.props
+    // Avoid triggering onsubmit with formik (which would result in a save user
+    // call).
+    evt.preventDefault()
+    if (confirm('Are you sure you would like to delete your user account? Once you do so, it cannot be recovered.')) {
+      deleteUser(loggedInUser, auth0)
+    }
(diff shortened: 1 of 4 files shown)


binh-dam-ibigroup (approved):

lib/components/user/user-account-screen.js
```diff
@@ -50,6 +50,16 @@ class UserAccountScreen extends Component {
...
   }
 
+  _handleDeleteUser = (evt) => {
+    const {auth0, deleteUser, loggedInUser} = this.props
+    // Avoid triggering onsubmit with formik (which would result in a save user
+    // call).
+    evt.preventDefault()
+    if (confirm('Are you sure you would like to delete your user account? Once you do so, it cannot be recovered.')) {
+      deleteUser(loggedInUser, auth0)
+    }
```
evansiroky, line 60: This confirm block could be moved into the deleteUser action.

evansiroky (approved):
See one suggestion that I feel somewhat opinionated about to make sure that the components don't get cluttered with logic that can just be in actions.

lib/components/user/user-account-screen.js
landonreed: Once we start internationalizing, I think it might be more useful to have this confirm message contained within this component. I view the confirm dialog as more tied with the component itself, whereas the action itself is more of a generic behavior that could presumably be used elsewhere in the app (perhaps in a context where we don't want a confirm dialog to pop up).

PR #337, opened 2021-02-25
ResponsiveWebapp refactor

evansiroky:
This PR is built on top of #332. This PR refactors the ResponsiveWebapp and example.js files a bit to streamline the example.js file and include more component slots as needed. Since this is perhaps a bit more involved than what #332 set out to do, it may be more appropriate to set the base branch to `dev`, but since this PR arose from seeing how #332 was presented, it is proposed to be incorporated into #332 for now. 

Here is a summary of the changed items:

### ResponsiveWebapp

The component has been refactored so that the `desktopView` and `mobileView` are no longer required props. The logic for these props that was in example.js has mostly been moved in here with some components being obtained from the context API as needed.

### example.js

This is refactored so that the ResponsiveWebapp is only called with the `components` prop. The `components` prop has had additional components added to it that are then used elsewhere.

### Mobile components refactor

The props that were passed through to the MobileMain component have been removed in favor of using the context API. This affected a few components as should be seen. The `results-screen.js` file's render function was really big, so I made some sub-renderers to make things a little easier to manage.

### create-otp-reducer

There was an unneeded second argument in the `getInitialState` since the argument for that was always present in the first argument. Therefore, I removed the second argument.

### DIFF
--- a/__tests__/reducers/create-otp-reducer.js
+++ b/__tests__/reducers/create-otp-reducer.js
@@ -6,6 +6,6 @@ describe('lib > reducers > create-otp-reducer', () => {
 
   it('should be able to create the initial state', () => {
     setDefaultTestTime()
-    expect(getInitialState({}, {})).toMatchSnapshot()
+    expect(getInitialState({})).toMatchSnapshot()
   })
 })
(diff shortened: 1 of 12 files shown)


landonreed (approved):
This looks great! Only comment is that this is probably a breaking change, no?

evansiroky:
Well, I guess I could've just merged from this website after all, but I decided to merge and squash on my local machine to be extra sure that I could create a commit message with the breaking change message. So, the relevant commit is a1e63b3cd18544d40eeae993ecaaa4933f88072c.

PR #340, opened 2021-03-08
Terms of service

binh-dam-ibigroup:
This PR defines two new ~optional,~ custom component entries in the `components` context that are required for OTP middleware implementations:
- `TermsOfService` (accessed from the "terms of service..." checkbox in the user account page)
- `TermsOfStorage`(accessed from the "terms of storage" checkbox in the user account page)

The PR implements the routes and wrappers to display these components.
(The contents of these components is out of scope.)


lib/components/app/responsive-webapp.js
```diff
@@ -280,6 +283,14 @@ class RouterWrapperWithAuth0 extends Component {
               component={UserAccountScreen}
               path={[`${CREATE_ACCOUNT_PATH}/:step`, ACCOUNT_SETTINGS_PATH]}
             />
+            <Route
+              component={frame(components.TermsOfService)}
+              path={TERMS_OF_SERVICE_PATH}
+            />
+            <Route
+              component={frame(components.TermsOfStorage)}
```
landonreed, line 291: I'm just wondering if we really need the `AppFrame` component. Could we just keep things how they are in `AccountPage` and use

```jsx
<Route
  render={() => <AccountPage><TermsOfService /></AccountPage>}
  path={TERMS_OF_SERVICE_PATH}
/>
```

landonreed (changes requested):
Looks good! But I think there's some extra abstraction in here that may not be needed.

lib/components/app/responsive-webapp.js
binh-dam-ibigroup: Aren't the terms and conditions routes public?

landonreed: Oh that's true. I forgot about the HOCs that were wrapping AccountPage. In that case, yes I suppose we do need AppFrame.

lib/components/app/app-frame.js
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Creates a simple wrapper component consisting of an AppFrame that surrounds
+ * the provided component. (Displays "no content" if none provided.)
+ */
+export function frame (Component) {
+  return () => (
+    <AppFrame>
+      {Component
+        ? <Component />
+        : <p>No content provided.</p>
```
landonreed, line 34: Maybe we should create a `PageNotFound` component and render that if Component does not exist (it could have a link back to the main trip planner).

landonreed (approved):
OK yep, thanks for clarifying that need for AppFrame. Only one non-blocking comment about creating/using a `PageNotFound` page that gives a little more help to the user to get back to the main trip planner.

lib/components/app/app-frame.js
binh-dam-ibigroup: Hmm I have set these pages to open in new tabs, so does a link make sense with this setup?

landonreed: Yea, perhaps no link in that case.

binh-dam-ibigroup: I enhanced the format for not found content in https://github.com/opentripplanner/otp-react-redux/pull/340/commits/f95638817c7736ad80cb0d6a83f1e52ee3581ebc.

evansiroky (approved):
Good to go pending merge conflict resolution

Issue #338, opened 2021-03-04
setActiveLeg action does not zoom to leg on map


landonreed:
To reproduce:
1. Run the TriMet call taker demo locally.
2. Visit [this trip](http://localhost:9966/#/?sessionId=test&ui_activeSearch=izhpm38tq&ui_activeItinerary=0&fromPlace=Nancy%27s%20Kitchen%2C%20Northwest%20District%20-%20Portland%3A%3A45.526795118388684%2C-122.68758773803712&toPlace=Mama%20Chow%27s%2C%20Downtown%20-%20Portland%3A%3A45.52030060757689%2C-122.67351150512697&date=2021-03-04&time=10%3A13&arriveBy=false&mode=WALK%2CBUS%2CTRAM%2CRAIL%2CGONDOLA&showIntermediateStops=true&maxWalkDistance=1207&optimize=QUICK&walkSpeed=1.34&ignoreRealtimeUpdates=true&companies=&bannedRoutes=&numItineraries=3&otherThanPreferredRoutesPenalty=900&preferredRoutes=)
2. Select the first option
3. Click the `Walk XX miles to XXX` text (shown in red rectangle below)
4. The map should zoom to the selected leg, but the action is not working properly.

I believe the issue is this component is not getting passed the `setActiveLeg` prop (there may be other props it is missing too):

https://github.com/opentripplanner/otp-react-redux/blob/5494729056de26521099cb20a9a35cbda9dd0764/lib/components/narrative/default/default-itinerary.js#L187

![image](https://user-images.githubusercontent.com/2370911/109985503-d0069200-7cd2-11eb-94cd-6ef1669e5837.png)

PR #342, opened 2021-03-11
fix setActiveLeg zoom

robertgregg3:
This PR fixes the bug (#338) where setActiveLeg doesn't zoom to that particular part of the leg. 

Clicking on the text next in the red box should zoom the map to that particular leg using setActiveLeg.

The fix included importing and passing setActiveLeg as a prop to the ItineraryBody component in both:

- narative-itinerary.js
- default-itinierary.js

109985503-d0069200-7cd2-11eb-94cd-6ef1669e5837

### DIFF
--- a/lib/components/narrative/default/default-itinerary.js
+++ b/lib/components/narrative/default/default-itinerary.js
@@ -6,7 +6,6 @@ import ItineraryBody from '../line-itin/connected-itinerary-body'
 import ItinerarySummary from './itinerary-summary'
 import SimpleRealtimeAnnotation from '../simple-realtime-annotation'
 import { getTotalFareAsString } from '../../../util/state'
-import { setActiveLeg } from "../../../actions/narrative";
 
 const { isBicycle, isTransit } = coreUtils.itinerary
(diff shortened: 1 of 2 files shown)


lib/components/narrative/default/default-itinerary.js
```diff
@@ -6,6 +6,7 @@ import ItineraryBody from '../line-itin/connected-itinerary-body'
 import ItinerarySummary from './itinerary-summary'
 import SimpleRealtimeAnnotation from '../simple-realtime-annotation'
 import { getTotalFareAsString } from '../../../util/state'
+import { setActiveLeg } from "../../../actions/narrative";
```
landonreed, line 9: @robertgregg3, looks like the linter flagged some issues here. Just a heads up that for this repo we use single quotes and no semi-colons. Also, this looks to be an unused import, so it can just be deleted.

```suggestion
```

robertgregg3: Thanks Landon. All sorted.

binh-dam-ibigroup (approved):
Looks good, and it works.

PR #344, opened 2021-03-19
Clarify strategies for auto-plan

landonreed:
This PR (in pursuit of changes for Call Taker) clarifies/defines the various strategies for replanning a trip whenever a query param change is detected in the form (AKA auto-plan). This includes a config update to autoPlan (no real breaking changes, unless an implementation had `autoPlan: true` at some point, but this was never documented).

The example config shows the approach:
https://github.com/opentripplanner/otp-react-redux/blob/8d500a2a4437843d3ebd6979469b1d7d49de28bd/example-config.yml#L18-L25

### DIFF
--- a/lib/actions/form.js
+++ b/lib/actions/form.js
@@ -78,53 +78,102 @@ export function parseUrlQueryString (params = getUrlParams()) {
 let debouncedPlanTrip // store as variable here, so it can be reused.
 let lastDebouncePlanTimeMs
 
+/**
+ * This action is dispatched when a change between the old query and new query
+ * is detected. It handles checks for whether the trip should be replanned
+ * (based on autoPlan strategies) as well as updating the UI state (esp. for
+ * mobile).
+ */
 export function formChanged (oldQuery, newQuery) {
   return function (dispatch, getState) {
     const otpState = getState().otp
+    const { config, currentQuery, ui } = otpState
+    const { autoPlan, debouncePlanTimeMs } = config
     const isMobile = coreUtils.ui.isMobile()
-
+    const {
+      fromChanged,
+      oneLocationChanged,
+      shouldReplanTrip,
+      toChanged
+    } = checkShouldReplanTrip(autoPlan, isMobile, oldQuery, newQuery)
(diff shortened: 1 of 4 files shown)


lib/actions/form.js
```diff
@@ -128,3 +130,50 @@ export function formChanged (oldQuery, newQuery) {
     }
   }
 }
+
+/**
+ * Check if the trip should be replanned based on the auto plan strategy,
+ * whether the mobile view is active, and the old/new queries. Response type is
+ * an object containing various booleans.
+ */
+export function checkShouldReplanTrip (autoPlan, isMobile, oldQuery, newQuery) {
```
evansiroky, line 139: Might be nice to have some unit tests for this. This is non-blocking.

evansiroky (approved):

lib/actions/form.js
```diff
@@ -128,3 +130,50 @@ export function formChanged (oldQuery, newQuery) {
...
+ */
+const evaluateAutoPlanStrategy = (strategy, fromChanged, toChanged, oneLocationChanged) => {
+  switch (strategy) {
+    case 'ONE_LOCATION_CHANGED':
+      if (oneLocationChanged) return true
+      break
+    case 'BOTH_LOCATIONS_CHANGED':
+      if (fromChanged && toChanged) return true
+      break
+    case 'ANY': return true
```
binh-dam-ibigroup, line 176: What is this value for? Either remove it or document it in `example-config.yml`.

binh-dam-ibigroup (approved):
Conditional approval on addressing my comment regarding autoplan strategy `ANY`.

lib/actions/form.js
landonreed: https://github.com/opentripplanner/otp-react-redux/blob/8d500a2a4437843d3ebd6979469b1d7d49de28bd/example-config.yml#L21-L22

lib/actions/form.js
landonreed: Added a ticket for revisiting this here: https://github.com/opentripplanner/otp-react-redux/issues/355

PR #346, opened 2021-03-19
Use CRA to build/serve app

landonreed:
This commit reorganizes the project to use Create React App (more specifically the craco tools built
on top of CRA). This required moving the lib dir to a subdir within src, adding/removing some deps,
and some other tweaks. The CSS is not quite working as of this commit.

Note: this also adds Typescript to the app.

BREAKING CHANGE: Replace mastarm with Create React App


src/index.js
```diff
@@ -29,7 +29,7 @@ import {
 } from './lib'
 // load the OTP configuration
 import otpConfig from './config.yml'
-
+console.log(otpConfig)
```
evansiroky, line 32: Remove console.log

PR #352, opened 2021-04-01
Avoid showing duplicate batch itineraries

### LINKED ISSUE #349: Batch routing mode shows duplicate itinerary results
In some cases, planning a trip with batch routing can yield duplicated itineraries in the list of results. This can occur when a mode option is listed more than once with different query parameters. We should check for duplicates when we receive responses back and filter these results out to prevent these showing to the user.

[Example trip plan](https://atl-rides.ibi-transit.com/#/?ui_activeSearch=p76gsjo2v&ui_activeItinerary=0&fromPlace=Inman%20Park%2FReynoldstown%2C%20Atlanta%2C%20GA%2C%20USA%3A%3A33.757822%2C-84.35236&toPlace=Midtown%2C%20Atlanta%2C%20GA%2C%20USA%3A%3A33.781128%2C-84.38636&date=2021-03-25&time=16%3A42&arriveBy=false&mode=WALK%2CBUS%2CSUBWAY%2CTRAM&showIntermediateStops=true&maxWalkDistance=1207&optimize=QUICK&walkSpeed=1.34&ignoreRealtimeUpdates=true&companies=&bannedRoutes=&numItineraries=3&otherThanPreferredRoutesPenalty=900&preferredRoutes=) (see the second and third options in the screenshot)
image

evansiroky:
This PR fixes #349 by not adding duplicate itineraries to the list of active itineraries. Using the already-available library of `object-hash`, we can reliably calculate a hash that can identify duplicate itineraries. However, there is a big performance cost for hashing the complete itinerary - especially on lengthy itineraries. Therefore, I made some extra performance improvements to make sure the app didn't slow down a ton. I added the `reselect` dependency and added reselect-style memorization to the  `getActiveItineraries` function. However, each time a new response was received, the function would run again, so I also memorized the calculation of the hash of each itinerary. It could be possible to further improve the itinerary hashing performance by adding additional black-listed keys that won't contribute to there being different itineraries.

### DIFF
--- a/lib/util/state.js
+++ b/lib/util/state.js
@@ -72,14 +72,35 @@ function getActiveSearchRealtimeResponse (otpState) {
   return search && search.response
 }
 
+// Ignore certain keys that could add significant calculation time to hashing.
+// The alerts are irrelevant, but the intermediateStops, interStopGeometry and
+// steps could have the legGeometry substitute as an equivalent hash value
+const blackListedKeys = [
+  'alerts',
+  'intermediateStops',
+  'interStopGeometry',
+  'steps'
+]
+// make blackListedKeys into an object due to superior lookup performance
+const blackListedKeyLookup = {}
+blackListedKeys.forEach(key => {
+  blackListedKeyLookup[key] = true
+})
+
 /**
  * A memoized function to hash the itinerary.
- * NOTE: It can take a while (>300ms) for the object-hash library to calculate
+ * NOTE: It can take a while (>30ms) for the object-hash library to calculate
  * an itinerary's hash for some lengthy itineraries. If better performance is
- * desired, a custom hash function should be written that only uses certain
- * parts of the itinerary instead of everything.
+ * desired, additional values to blackListedKeys should be added to avoid
+ * spending extra time hashing values that wouldn't result in different
+ * itineraries.
  */
-const hashItinerary = memoize(itinerary => hash(itinerary))
+const hashItinerary = memoize(
+  itinerary => hash(
+    itinerary,
+    { excludeKeys: key => blackListedKeyLookup[key] }
+  )
+)
 
 /**
  * Get the active itineraries for the active search, which is dependent on
--- a/lib/util/state.js
+++ b/lib/util/state.js
@@ -7,7 +7,7 @@ import { createSelector } from 'reselect'
 
 import { MainPanelContent } from '../actions/ui'
 
-const { calculateFares, isTransit } = coreUtils.itinerary
+const { calculateFares } = coreUtils.itinerary
 
 /**
  * Get the active search object
@@ -287,14 +287,6 @@ export function getTotalFareAsString (itinerary) {
   const {centsToString} = calculateFares(itinerary)
   // Return total fare as formatted string.
   return centsToString(getTotalFare(itinerary))
-}
-
-function itineraryHasTransit (itinerary) {
-  let itinHasTransit = false
-  itinerary.legs.forEach(leg => {
-    if (isTransit(leg.mode)) itinHasTransit = true
-  })
-  return itinHasTransit
 }
 
 /**
--- a/lib/util/state.js
+++ b/lib/util/state.js
@@ -7,7 +7,7 @@ import { createSelector } from 'reselect'
 
 import { MainPanelContent } from '../actions/ui'
 
-const { calculateFares, hasBike, isCar, isTransit, isWalk } = coreUtils.itinerary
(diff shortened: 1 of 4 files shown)


lib/util/state.js
```diff
@@ -53,62 +56,133 @@ export function getActiveError (otpState) {
...
-        case 'ALL':
-          return true
-        case 'ACTIVE':
-          // ACTIVE filter checks that the only modes contained in the itinerary
-          // are 'active' (i.e., walk or bike). FIXME: add micromobility?
-          let allActiveModes = true
-          itinerary.legs.forEach(leg => {
-            if (!isWalk(leg.mode) && !hasBike(leg.mode)) allActiveModes = false
+export const getActiveItineraries = createSelector(
+  otpState => otpState.config,
```
binh-dam-ibigroup, line 113: How is `otpState.config` relevant to the selector? (It doesn't change within a session, does it?)

binh-dam-ibigroup (approved):
Seems to work. I just have a comment regarding detecting config changes, but this seems to be good to go as is.

lib/util/state.js
landonreed: It doesn't change, but it's used in the `sortItineraries` func called below.

evansiroky: It is relevant because when creating a selector, only the parts of the state that need to be used in the transform function need to be isolated and obtained via an "input-selector". If all of the values returned from the input selectors are the same, then the memorized value can be returned. This avoids recomputing the return value of the transform function. See [docs about this here](https://www.npmjs.com/package/reselect#creating-a-memoized-selector).

lib/util/state.js
```diff
@@ -53,62 +56,133 @@ export function getActiveError (otpState) {
...
+            let hasCar = false
+            itinerary.legs.forEach(leg => {
+              if (isCar(leg.mode)) hasCar = true
+            })
+            return hasCar
+          default:
+            // return true for unrecognized or falsy filter
+            return true
+        }
+      })
```
landonreed, line 178: I think this filter stuff is dead code that we can remove. I think we just failed to remove it when the dropdown was pulled out.

landonreed (approved):
LGTM! Only thing is that we may want to remove the filtering stuff since it doesn't correspond to any workable UI anymore.

landonreed:
Assigning back to myself to check one more thing.

landonreed (changes requested):
I'm seeing an issue with FDOT for [this trip](http://localhost:9966/#/?ui_activeSearch=27gjv5595&ui_activeItinerary=0&fromPlace=P%26R%3A%3A28.45119%2C-81.36818&toPlace=Downtownish%3A%3A28.54834%2C-81.37745&date=2021-04-12&time=16%3A28&arriveBy=false&mode=WALK%2CBUS%2CRAIL&showIntermediateStops=true&maxWalkDistance=9656.04&optimize=QUICK&walkSpeed=1.34&ignoreRealtimeUpdates=true&companies=&bannedRoutes=&numItineraries=1&otherThanPreferredRoutesPenalty=900&preferredRoutes=). It's showing what appears to be the same Walk to Rail trip twice:
image


With the following config for modes:
```yaml
modes:
  combinations:
      # Standard
    - mode: 'WALK,TRANSIT'
      params:
        numItineraries: 2
      # Bike to transit
    - mode: 'BICYCLE,TRANSIT'
      params:
        numItineraries: 2
      # Bikeshare + transit
    - mode: 'BICYCLE_RENT,TRANSIT'
      params:
        numItineraries: 1
        # companies: 'HOPR'
      # Walk only
    - mode: 'WALK'
    # Bikeshare
    - mode: 'BICYCLE_RENT'
      # Bike only
    - mode: 'BICYCLE'
    # Car only. TODO remove?
    - mode: 'CAR'
      # Park & Ride
    - mode: 'CAR_PARK,WALK,TRANSIT'
  transitModes:
    - mode: BUS
      label: Bus
    - mode: RAIL
      label: SunRail
```

evansiroky:
@landonreed, they are different by 1 second which is the result of one extra point in the final walking leg of each otherwise identical itinerary. 🤣 

Screen Recording 2021-04-12 at 9 51 26 PM

Here is a closeup of the decoded polylines:

Shorter itinerary:

Screen Shot 2021-04-12 at 9 48 58 PM

Longer itinerary:

Screen Shot 2021-04-12 at 9 48 53 PM

landonreed (approved):

landonreed:
Just fixed some merge/linting issues, @evansiroky. Feel free to take a quick look, but I think it's good to merge now.

PR #354, opened 2021-04-09
Bump core-utils for default query override

landonreed:
To test config override, add this bit from `example-config.yml` to your active config:

https://github.com/opentripplanner/otp-react-redux/blob/860d6dc0ab615ae09fba7e8c9305fe09d30811b4/example-config.yml#L18-L21

### DIFF
--- a/example-config.yml
+++ b/example-config.yml
@@ -15,6 +15,11 @@ api:
 #     lon: -122.71607145667079
 #     name: Oregon Zoo, Portland, OR
 
+### The default query parameters can be overridden be uncommenting this object.
+### Note: the override values must be valid values within otp-ui's query-params.js
+# defaultQueryParams:
+#   maxWalkDistance: 3219 # 2 miles in meters
+
 ### The persistence setting is used to enable the storage of places (home, work),
 ### recent searches/places, user overrides, and favorite stops.
 ### Pick the strategy that best suits your needs.
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
   "dependencies": {
     "@auth0/auth0-react": "^1.1.0",
     "@opentripplanner/base-map": "^1.0.5",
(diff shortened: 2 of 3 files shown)


example-config.yml
```diff
@@ -15,6 +15,11 @@ api:
 #     lon: -122.71607145667079
 #     name: Oregon Zoo, Portland, OR
 
+### The default query parameters can be overridden be uncommenting this object.
+### Note: the override values must be valid values within otp-ui's query-params.js
+# defaultQueryParams:
+#   maxWalkDistance: 3219 # 2 miles in meters
```
binh-dam-ibigroup, line 21: The UI does not seem to be set on the default query params, does this PR intend to do that?

As an example, if you set `optimize: TRANSFERS`, the optimization setting will still appear as `SPEED` in the UI (with a red dot indicating that that is not the default setting, though.)

example-config.yml
```diff
@@ -15,6 +15,11 @@ api:
 #     lon: -122.71607145667079
 #     name: Oregon Zoo, Portland, OR
 
+### The default query parameters can be overridden be uncommenting this object.
+### Note: the override values must be valid values within otp-ui's query-params.js
```
binh-dam-ibigroup, line 19: I think because these settings are drop-down items, the comment should rather say:

`... must be one of the values defined in otp-ui's query-params.js for the desired parameter`

binh-dam-ibigroup (approved):
Just a suggestion to reword the comment regarding `defaultQueryParams` but I'm inclined to accept as is.

example-config.yml
landonreed: This is not *necessarily* true. There are some non-dropdown values not in the form that can be overridden (try `otherThanPreferredRoutesPenalty`). I think it's ok how it is for now.

PR #357, opened 2021-04-16
Update call list items

landonreed:
This PR addresses the following call taker items:
- updates the call and query list item labels to be more descriptive (see screenshot).
- seeks to fix an issue where viewing a past call record query initiated a new call (by adding `_CALL` to the `searchId`).

If others have alternative approaches for item 2, I'm happy to entertain them!

image

### DIFF
--- a/lib/actions/form.js
+++ b/lib/actions/form.js
@@ -61,14 +61,15 @@ export function setQueryParam (payload, searchId) {
   }
 }
 
-export function parseUrlQueryString (params = getUrlParams()) {
+export function parseUrlQueryString (params = getUrlParams(), source) {
   return function (dispatch, getState) {
     // Filter out the OTP (i.e. non-UI) params and set the initial query
     const planParams = {}
     Object.keys(params).forEach(key => {
       if (!key.startsWith('ui_')) planParams[key] = params[key]
     })
-    const searchId = params.ui_activeSearch || coreUtils.storage.randId()
+    let searchId = params.ui_activeSearch || coreUtils.storage.randId()
+    if (source) searchId += source
     // Convert strings to numbers/objects and dispatch
     planParamsToQueryAsync(planParams, getState().otp.config)
       .then(query => dispatch(setQueryParam(query, searchId)))
--- a/lib/actions/form.js
+++ b/lib/actions/form.js
@@ -61,6 +61,16 @@ export function setQueryParam (payload, searchId) {
   }
 }
 
+/**
+ * An action that parses the active URL's query params (or whatever is passed in)
+ * and updates the state with the OTP plan query params. A new search will be
+ * performed according to the behavior of setQueryParam (i.e., if the passed
+ * searchId is not null).
+ * @param  {Object} params an object containing URL query params
+ * @param  {string} source optionally contains the source of the query params
+ *                         (e.g., _CALL indicates that the source is from a past
+ *                         query made during a call taker session)
+ */
 export function parseUrlQueryString (params = getUrlParams(), source) {
   return function (dispatch, getState) {
     // Filter out the OTP (i.e. non-UI) params and set the initial query
--- a/lib/components/admin/call-record.js
+++ b/lib/components/admin/call-record.js
@@ -14,6 +14,20 @@ export default class CallRecord extends Component {
     expanded: false
   }
 
+  _getCallDuration = () => {
+    const {call} = this.props
+    const startTimeMoment = moment(call.startTime)
+    const endTimeMoment = moment(call.endTime)
+    const [hours, minutes, seconds] = moment.utc(
+      moment.duration(endTimeMoment.diff(startTimeMoment)).asMilliseconds()
+    ).format('HH:mm:ss').split(':').map(val => +val)
+    const parts = []
+    if (seconds) parts.push(`${seconds} sec`)
+    if (minutes) parts.push(`${minutes} min`)
+    if (hours) parts.push(`${hours} hr`)
+    return parts.join(', ')
+  }
+
   _toggleExpanded = () => {
     const {call, fetchQueries} = this.props
     const {expanded} = this.state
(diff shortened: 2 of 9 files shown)


binh-dam-ibigroup:
Would a layout like this one be more efficient?
image

binh-dam-ibigroup (changes requested):
For now, please see the JSDoc and visual layout recommendations. There is an odd behavior with the `ui_activeSearch` param but I don't have a recommendation at this point.

lib/components/admin/call-record.js
```diff
@@ -14,6 +15,20 @@ export default class CallRecord extends Component {
...
+    const startTimeMoment = moment(call.startTime)
+    const endTimeMoment = moment(call.endTime)
+    const [hours, minutes, seconds] = moment.utc(
+      moment.duration(endTimeMoment.diff(startTimeMoment)).asMilliseconds()
+    ).format('HH:mm:ss').split(':').map(val => +val)
+    const parts = []
+    if (hours) parts.push(`${hours} hr`)
+    if (minutes) parts.push(`${minutes} min`)
+    if (seconds) parts.push(`${seconds} sec`)
+    return parts.join(', ')
```
evansiroky, line 29: In datatools-ui we use [humanize-duration](https://www.npmjs.com/package/humanize-duration). Should we use it here?

lib/reducers/call-taker.js
```diff
@@ -103,18 +103,20 @@ function createCallTakerReducer () {
...
           // If call is in progress, record search ID when a routing response is
           // fulfilled.
           // TODO: How should we handle routing errors.
           return update(state, {
-            activeCall: { searches: { $push: [action.payload.searchId] } }
+            activeCall: { searches: { $push: [searchId] } }
           })
-        } else if (state.callHistory.visible) {
+        } else if (state.callHistory.visible && searchId.indexOf('_CALL') === -1) {
+          console.log(searchId)
```
evansiroky, line 115: Remove console.log

evansiroky (changes requested):
I share the same concerns as Binh and will wait to see what changes result of the changes Binh requested.

lib/reducers/call-taker.js
```diff
@@ -103,18 +103,20 @@ function createCallTakerReducer () {
...
         if (state.activeCall) {
           // If call is in progress, record search ID when a routing response is
           // fulfilled.
           // TODO: How should we handle routing errors.
           return update(state, {
-            activeCall: { searches: { $push: [action.payload.searchId] } }
+            activeCall: { searches: { $push: [searchId] } }
           })
-        } else if (state.callHistory.visible) {
+        } else if (state.callHistory.visible && searchId.indexOf('_CALL') === -1) {
```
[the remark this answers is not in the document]
landonreed: Yes, I'm not entirely sure how to fix this.

landonreed:
Comments addressed... except for the odd case that @binh-dam-ibigroup noted in https://github.com/opentripplanner/otp-react-redux/pull/357#discussion_r615949744. We could pull an issue for this if no one has ideas about the root cause.

binh-dam-ibigroup:
> Comments addressed... except for the odd case that @binh-dam-ibigroup noted in [#357 (comment)](https://github.com/opentripplanner/otp-react-redux/pull/357#discussion_r615949744). We could pull an issue for this if no one has ideas about the root cause.

Depends. Has TriMet raised this issue yet?

lib/components/admin/call-record.js
```diff
@@ -47,29 +57,42 @@ export default class CallRecord extends Component {
...
+        <CallRecordButton
           className='clear-button-formatting'
           onClick={this._toggleExpanded}
         >
-          <Icon type='phone' className='fa-flip-horizontal' />
-          Call {index} ({moment(call.endTime).fromNow()})
-        </button>
+          <CallRecordIcon
+            className='fa-flip-horizontal'
+            style={{width: '5%'}}
```
binh-dam-ibigroup, line 81: Maybe use the same value for the `width` as `margin-left` for the `<ul>` element when expanded? In this case that would be `22px`.

lib/components/admin/call-record.js
```diff
@@ -47,29 +57,42 @@ export default class CallRecord extends Component {
...
+            type='phone' />
+          <span style={{flex: '0 0 120px'}}>
+            {startTimeMoment.format('h:mm a, MMM D')}
```
binh-dam-ibigroup, line 84: (Nitpick: the minutes portion of the time should ideally be flushed right... non blocking)

binh-dam-ibigroup (commented):
One minor comment for now. Let me see if I can find out more about the call being started automatically.

lib/reducers/call-taker.js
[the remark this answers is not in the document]
binh-dam-ibigroup: This may be worse:
1. Start from scratch
2. Open call taker windows and expand an entry
3. Plan a trip on the map, indenpently from the call window.
4. The call button state changes to 'recording'.

Screen Recording 2021-04-27 at 5 24 11 PM

evansiroky:
Unassigning myself until the automatic call starting issue @binh-dam-ibigroup is talking about is resolved.

lib/reducers/call-taker.js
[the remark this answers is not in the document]
binh-dam-ibigroup: I submitted #362 to prevent a call from starting when clicking entries in the call log. It should not affect anything that is not dealing with the `_CALL` suffix in search ids. (Note: #362 does NOT address the issue in the animation above.)

lib/components/admin/query-record.js
```diff
@@ -1,37 +1,51 @@
...
+        </CallRecordButton>
       </li>
     )
   }
 }
 
 const mapStateToProps = (state, ownProps) => {
-  return {}
+  return {
+    config: state.otp.config,
```
evansiroky, line 46: Is this needed?

landonreed: removed in 9527a56db41e2a7cb9e28ae553a669b4fc206a2b

binh-dam-ibigroup (approved):
Approving, now that the call starting mystery has been resolved.

lib/actions/form.js
```diff
@@ -61,14 +61,25 @@ export function setQueryParam (payload, searchId) {
...
+export function parseUrlQueryString (params = getUrlParams(), source) {
   return function (dispatch, getState) {
     // Filter out the OTP (i.e. non-UI) params and set the initial query
     const planParams = {}
     Object.keys(params).forEach(key => {
       if (!key.startsWith('ui_')) planParams[key] = params[key]
     })
-    const searchId = params.ui_activeSearch || coreUtils.storage.randId()
+    let searchId = params.ui_activeSearch || coreUtils.storage.randId()
+    if (source) searchId += source
```
evansiroky, line 82: This always seemed a bit hacky to me to modify an ID to append on more information. Furthermore, this method fits within a complex chain of interactions that eventually the ResponsiveWebapp ends up parsing in its componentDidUpdate method. I can't think of a better way to do it that wouldn't involve a larger refactor, so I'm fine with it for now.

evansiroky (approved):

lib/actions/form.js
landonreed: Minor grievance noted. :)

PR #362, opened 2021-05-05
refactor(actions/form): Pass source to routing query.

binh-dam-ibigroup:
Introduce a fix to prevent starting Calltaker calls when clicking between itinerary searches in the call logs.


landonreed:
Closing in favor of recent changes in #357 (36e8e4501469bf3eaf9bd0b8a3267627bf4107d8) that fix the time parameter setting (which I believe was causing the unpredictable results where a new active call would begin after clicking a past query in the call history window).

PR #369, opened 2021-05-20
Calltaker Recording Tweaks

binh-dam-ibigroup:
Co-authored with @robertgregg3.

This PR makes the following tweaks to the call UI:
- Clear the map/form when ending a call
- Add an optional config parameter to show the call history window on app load
- Start a call on clicking the Plan button when the call module is enabled, and the field trip window is not shown.

### DIFF
--- a/lib/reducers/call-taker.js
+++ b/lib/reducers/call-taker.js
@@ -4,7 +4,8 @@ import moment from 'moment'
 import {constructNewCall} from '../util/call-taker'
 import {FETCH_STATUS} from '../util/constants'
 
-function createCallTakerReducer () {
+function createCallTakerReducer (config) {
+  const calltakerConfig = config.modules.find(m => m.id === 'call')
   const initialState = {
     activeCall: null,
     callHistory: {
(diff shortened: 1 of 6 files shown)


evansiroky (approved):

lib/reducers/call-taker.js
```diff
@@ -4,15 +4,18 @@ import moment from 'moment'
...
     activeCall: null,
     callHistory: {
       calls: {
         status: FETCH_STATUS.UNFETCHED,
         data: []
       },
-      visible: false
+      visible: calltakerConfig &&
+        calltakerConfig.options &&
+        calltakerConfig.options.showCallHistoryOnLoad
```
landonreed, line 18: There's some new JS syntax called [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) that makes this a lot cleaner:
```suggestion
      visible: calltakerConfig?.options?.showCallHistoryOnLoad
```

landonreed (approved):
Looks good, just one minor comment. Also, did you resolve the issue with pasting in a URL?

binh-dam-ibigroup:
> Also, did you resolve the issue with pasting in a URL?

Not at this time.

binh-dam-ibigroup:
@landonreed I made some changes in https://github.com/opentripplanner/otp-react-redux/pull/369/commits/4638a55af190eed58b27d4026a1fda508421afc9 so that a call only start from a click of the Plan button in the calltaker panel (It does not work if autoplan is set to true right now). That way a call does not start when pasting a URL and does not start when viewing previous call records (so the `_CALL` suffix hack is useless and removed now).

landonreed (approved):
Great stuff, thanks @binh-dam-ibigroup.

PR #366, opened 2021-05-19
Transitive styles

binh-dam-ibigroup:
This PR modifies code to use new styling props introduced in `@opentripplanner/core-utils/transitive-overlays` v 1.0.7 (see https://github.com/opentripplanner/otp-ui/pull/246):
- A new, optional section for transitive is added under `example-config.yml#map`.
- A new, optional `getTransitiveRouteLabel` function to render a route label is introduced in the `ComponentContext`,
- Rendering for `TransitiveCanvasOverlay` is updated accordingly.

EDIT: Example configuration for testing: see https://github.com/ibi-group/configurations/pull/141.

EDIT: This PR should fix #374.

PR #370, opened 2021-05-21
Use selector to compute transitive data

evansiroky:
This is a proposed refactor of #366 that seeks to be more efficient and selective on when to recompute and rerender transitive data.

### DIFF
--- a/lib/components/map/connected-transitive-overlay.js
+++ b/lib/components/map/connected-transitive-overlay.js
@@ -1,37 +1,26 @@
-import isEqual from 'lodash.isequal'
-import coreUtils from '@opentripplanner/core-utils'
-import TransitiveCanvasOverlay from '@opentripplanner/transitive-overlay'
+import TransitiveOverlay from '@opentripplanner/transitive-overlay'
 import React, { Component } from 'react'
 import { connect } from 'react-redux'
 
-import { ComponentContext } from '../../util/contexts'
-import { getActiveItinerary, getActiveItineraries, getActiveSearch } from '../../util/state'
+import { getActiveSearch, getTransitiveData } from '../../util/state'
 
 /**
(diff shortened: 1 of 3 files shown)


lib/components/map/connected-transitive-overlay.js
```diff
@@ -1,71 +1,43 @@
...
-class TransitiveCanvasOverlayWithContext extends Component {
-  static contextType = ComponentContext
-
+class TransitiveCanvasOverlay extends Component {
   shouldComponentUpdate (nextProps) {
-    return (
-      !isEqual(nextProps.activeSearchVisibleItinerary, this.props.activeSearchVisibleItinerary) ||
-      !isEqual(nextProps.activeSearchResponseOtp, this.props.activeSearchResponseOtp)
-    )
+    return nextProps.transitiveData !== this.props.transitiveData
```
binh-dam-ibigroup, line 13: The object references can be different although the value and contents can be identical. I thought in such cases a rendering could be avoided.

lib/components/map/default-map.js
```diff
@@ -164,7 +167,12 @@ class DefaultMap extends Component {
               Without it, transitive resolution will not match the map,
               and transitive will appear blurry after e.g. the narrative is expanded.
            */}
-          <TransitiveOverlay key={itineraryView ? `transitive-${itineraryView}` : 'transitive-default'} />
+          <TransitiveOverlay
+            getTransitiveRouteLabel={this.context.getTransitiveRouteLabel}
+            key={itineraryView
```
binh-dam-ibigroup, line 172: Hmmm itineraryView seems to always be `undefined`.

binh-dam-ibigroup (approved):
Thanks @evansiroky for suggesting the selector approach.

PR #375, opened 2021-05-27
Add mini react-intl i18n implementation.

philip-cline:
Add a mini internationalization implementation to the default itinerary changing the mode names,
mode separator, the time format and the duration format. At the moment only english (en-US) and
french (fr-FR) translations are provided and more languages will be added in later pull requests.

Translated messages are defined in language-specific yml configuration files (e.g. en-US.yml) and all language yml files are loaded as dynamically assigned file names are not compatible with node.js. 

For future development, the language setting is retrieved from the URL param 'locale' and may be used to change the displayed language.


lib/util/i18n.js
```diff
@@ -0,0 +1,21 @@
+import objectPath from 'object-path'
+
+// Load the localized strings for all languages here.
+// (We cannot really load languages on demand using fetch... that would require tinkering with mastarm build,
+// so here we build an index of translated messages.)
+const translatedMessages = {
+  'en-US': require('../../i18n/en-US.yml'),
```
landonreed, line 7: One thing that stands out to me here is that we may need to sub out these strings on a per-implementation basis. E.g., the names of otp modes in Portland OR may be different than in Atlanta (e.g. light rail vs streetcar for TRAM). Not really a requirement for right now, but any ideas for this at the moment?

lib/actions/user.js
```diff
@@ -31,10 +31,12 @@ function createNewUser (auth0User) {
     auth0UserId: auth0User.sub,
     email: auth0User.email,
     hasConsentedToTerms: false, // User must agree to terms.
+    language: 'fr',
```
evansiroky, line 34: This should probably not be hardcoded. It seems like something like this would be derived from the current locale setting that the user selected in the app.

lib/actions/user.js
```diff
@@ -134,6 +136,8 @@ export function fetchOrInitializeUser (auth0User) {
 
     const isNewAccount = status === 'error' || (user && user.result === 'ERR')
     const userData = isNewAccount ? createNewUser(auth0User) : user
+    userData.language = 'fr'
+    userData.use24HourFormat = true
```
evansiroky, line 140: These should probably not be hardcoded. It seems like something like this would be derived from the current locale setting that the user selected in the app.

lib/components/narrative/default/default-itinerary.js
```diff
@@ -129,66 +174,83 @@ export default class DefaultItinerary extends NarrativeItinerary {
...
+      </IntlProvider>
     )
   }
 }
+
+const mapStateToProps = (state, ownProps) => {
+  return {
+    locale: getUrlParams().locale || 'en-US',
+    userLanguage: state.user.loggedInUser == null ? 'en-US' : state.user.loggedInUser.language,
+    use24HourFormat: state.user.loggedInUser == null ? false : state.user.loggedInUser.use24HourFormat
```
evansiroky, line 252: These all deserve a little more thought on how they are derived. I think a good place a lot of this could live is somewhere in the `ui` object in the redux state. Even if a user has a default language or time format they might still want to change it to some other language.

lib/components/narrative/default/default-itinerary.js
```diff
@@ -48,18 +90,20 @@ const ITINERARY_ATTRIBUTES = [
...
   },
   {
     id: 'cost',
     order: 2,
-    render: (itinerary, options) => getTotalFareAsString(itinerary)
+    render: (itinerary, options) => {
+      // Get unformatted transit fare portion only (in cents).
+      const { transitFare } = calculateFares(itinerary)
+      return (
+        <FormattedNumber value={transitFare / 100} style='currency' currencyDisplay='narrowSymbol' currency='USD' />
```
evansiroky, line 104: The currency should be dynamically populated, perhaps with a fallback to USD or maybe even some configurable value. I think if OTP is configured properly, it will output the currency code as part of an itinerary response.

lib/components/narrative/default/default-itinerary.js
```diff
@@ -1,34 +1,76 @@
...
       primaryTransitDuration = duration
-      transitMode = mode.toLowerCase()
+      transitMode = <FormattedMessage id={mode} />
     }
-    if (isBicycle(mode)) mainMode = 'Bike'
-    if (rentedBike) mainMode = 'Bikeshare'
-    if (mode === 'CAR') mainMode = 'Drive'
+    if (isBicycle(mode)) mainModeStringId = 'bike'
+    if (rentedBike) mainModeStringId = 'bikeshare'
+    if (mode === 'CAR') mainModeStringId = 'drive'
```
evansiroky, line 34: needs an entry for Micromobility.

i18n/en-US.yml
```diff
@@ -0,0 +1,30 @@
+_id: en-US
+_name: English
+components:
+  DefaultItinerary:
```
evansiroky, line 4: I've never been a fan of organizing messages by components. Often times messages can come from util functions shared across multiple components. I recommend organizing messages by overall category and avoiding deep hierarchical organizations if at all possible.

lib/util/i18n.js
evansiroky: I agree that this is needed, but think it'd be better to figure this out now rather than later. Also, I think it should be possible to load language files on-demand using fetch - the app should just wait until the base language file is loaded before the first main render. I think it'd make a lot of sense to have some kind of config file that can be used to override a default set of language files, however that process would likely be part of the trimet-mod-otp project (at least until we merge those two projects together).

evansiroky (changes requested):
See various comments, most of them are blocking. I think this is a good first demonstration of how things could work, but the overall structure of where the translations are loaded from and then stored in the redux state needs a more comprehensive design that will work everywhere else.

lib/components/app/responsive-webapp.js
```diff
@@ -268,10 +278,31 @@ const WebappWithRouter = withRouter(
...
+    this.updateMessages(prevProps)
+  }
+
+  async updateMessages (prevProps) {
+    // If the locale changes (or no prevProps are provided), update the messages for that locale.
+    const locale = this.props.locale
+    if (locale && (!prevProps || locale !== prevProps.locale)) {
+      this.setState({ messages: await loadLocaleData(locale, this.props.customMessages) })
+    }
+  }
```
landonreed, line 299: Only the `updateMessages` method really needs to be async (the others don't use the `await` keyword). Here's how I would change this (along with some other slight tweaks):

```suggestion
  componentDidMount () {
    this.updateMessages()
  }

  componentDidUpdate (prevProps) {
    this.updateMessages(prevProps.locale)
  }

  async updateMessages (previousLocale) {
    // If the locale changes (or no prevProps are provided), update the messages
    // for that locale.
    const {customMessages, locale} = this.props
    if (locale && (!previousLocale || locale !== previousLocale)) {
      const messages = await loadLocaleData(locale, customMessages)
      this.setState({messages})
    }
  }
```

lib/util/i18n.js
```diff
@@ -0,0 +1,31 @@
...
+ * merge with the provided customized strings from state.otp.config.language
+ * (overrides can be defined per language or for all languages)
+ * and flatten to a map of
+ *   id => string.
+ * FIXME: Load languages on demand using fetch.
+ */
+export async function loadLocaleData (locale, customMessages) {
+  let messages
+  switch (locale) {
+    case 'fr-FR':
```
landonreed, line 14: This should be tolerant of incomplete (`fr` rather than just `fr-FR`) or miscapitalized locale strings. So perhaps a more appropriate approach for now would be to do something like:
```javascript
// Call getMatchingLocale in this method
const matchedLocale = getMatchingLocaleString(locale)
// Define locales and other method above this method.
const locales = {
  en: {
    us: 'en-US'
  },
  fr: {
    fr: 'fr-FR'
  }
}

function getMatchingLocaleString (locale = '', defaultLocale = 'en-US') {
  const [lang, region] = locale.toLowerCase().split('-')
  const language = locales[lang]
  if (!language) return defaultLocale
  const defaultRegion = Object.keys(language)[0]
  return language[region] || language[defaultRegion]
}
```

landonreed (approved):
This is looking good. See final comments about default lang/tolerance of different locale strings.

lib/util/i18n.js
philip-cline: Updated that code in this commit: https://github.com/opentripplanner/otp-react-redux/pull/375/commits/f5a74fabe0e292e3a3f6cb17278548d063945c14

lib/components/app/responsive-webapp.js
philip-cline: Updated that code in this commit: https://github.com/opentripplanner/otp-react-redux/pull/375/commits/f5a74fabe0e292e3a3f6cb17278548d063945c14

lib/components/app/responsive-webapp.js
```diff
@@ -280,79 +311,81 @@ class RouterWrapperWithAuth0 extends Component {
...
-              component={PrintLayout}
-              path='/print'
-            />
-            {/* For any other route, simply return the web app. */}
-            <Route
-              render={() => <WebappWithRouter {...this.props} />}
-            />
-          </Switch>
-        </ConnectedRouter>
+        <IntlProvider locale={locale} messages={this.state.messages} defaultLocale='en-US'>
```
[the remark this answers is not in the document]
philip-cline: updated that code in this commit: https://github.com/opentripplanner/otp-react-redux/pull/375/commits/f5a74fabe0e292e3a3f6cb17278548d063945c14

lib/components/narrative/default/default-itinerary.js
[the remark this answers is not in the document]
philip-cline: updated in https://github.com/opentripplanner/otp-react-redux/pull/375/commits/f5a74fabe0e292e3a3f6cb17278548d063945c14

lib/components/app/responsive-webapp.js
```diff
@@ -280,79 +314,81 @@ class RouterWrapperWithAuth0 extends Component {
...
-              component={PrintLayout}
-              path='/print'
-            />
-            {/* For any other route, simply return the web app. */}
-            <Route
-              render={() => <WebappWithRouter {...this.props} />}
-            />
-          </Switch>
-        </ConnectedRouter>
+        <IntlProvider locale={locale} messages={this.state.messages} defaultLocale={defaultLocale}>
```
evansiroky, line 317: Line too long.

lib/util/i18n.js
evansiroky: I disagree with @landonreed and don't this adds any value. The only way this makes sense is if raw user input is expected. Although this is how things are done with this very alpha branch of code, I fully do not expect users to be typing in raw locale-lang strings with more matured code. Please revert.

lib/components/narrative/default/default-itinerary.js
```diff
@@ -1,34 +1,76 @@
...
-const { isBicycle, isTransit } = coreUtils.itinerary
-const { formatDuration, formatTime } = coreUtils.time
+import ItinerarySummary from './itinerary-summary'
+
+const { calculateFares, isBicycle, isTransit, isMicromobility } = coreUtils.itinerary
 
+/**
+ * Obtains the description of an itinerary in the given locale.
+ */
 // FIXME move to core utils
```
evansiroky, line 18: I think this is specific enough especially now that it is using the `FormattedMessage` component that it should not be moved to the `@opentripplanner/core-utils` package.  However, it definitely seems like it could be moved into a util file within the otp-rr project.

lib/components/narrative/default/default-itinerary.js
```diff
@@ -1,34 +1,76 @@
...
+  let accessModeId = 'walk'
   let transitMode
   itinerary.legs.forEach((leg, i) => {
-    const {duration, mode, rentedBike} = leg
+    const {duration, mode, rentedBike, rentedVehicle} = leg
     if (isTransit(mode) && duration > primaryTransitDuration) {
       // TODO: convert OTP's TRAM mode to the correct wording for Portland
       primaryTransitDuration = duration
-      transitMode = mode.toLowerCase()
+      transitMode = <FormattedMessage id={`common.otpTransitModes.${mode.toLowerCase()}`} />
```
evansiroky, line 28: Line too long.

lib/components/narrative/default/default-itinerary.js
```diff
@@ -1,34 +1,76 @@
...
+    if (isMicromobility(mode)) accessModeId = 'micromobility'
+    if (rentedVehicle) accessModeId = 'micromobilityRent'
+    if (rentedBike) accessModeId = 'bikeshare'
+    if (mode === 'CAR') accessModeId = 'drive'
   })
-  let description = mainMode
-  if (transitMode) description += ` to ${transitMode}`
-  return description
+
+  const mainMode = <FormattedMessage id={`common.accessModes.${accessModeId}`} />
```
evansiroky, line 37: Line too long.

lib/components/narrative/default/default-itinerary.js
```diff
@@ -1,34 +1,76 @@
...
+  return transitMode
+    ? <FormattedMessage id='components.DefaultItinerary.multiModeSummary' values={{ accessMode: mainMode, transitMode }} />
```
evansiroky, line 39: Line too long.

lib/components/narrative/default/default-itinerary.js
```diff
@@ -1,34 +1,76 @@
...
+        id='components.DefaultItinerary.tripDurationFormat'
+        values={{ hours, minutes }}
+      />
+    )
+  }
+}
+
+function formatTime (startTime, endTime, timeFormat) {
+  return (
+    <FormattedMessage id='components.DefaultItinerary.departureArrivalTimes' values={{
```
evansiroky, line 69: Line too long.

lib/components/narrative/default/default-itinerary.js
```diff
@@ -1,34 +1,76 @@
...
+    return (
+      <FormattedMessage
+        id='components.DefaultItinerary.tripDurationFormat'
+        values={{ hours, minutes }}
+      />
+    )
+  }
+}
+
+function formatTime (startTime, endTime, timeFormat) {
```
evansiroky, line 67: I recommend changing this into a proper functional component:
```suggestion
function FormatTime ({ startTime, endTime, timeFormat }) {
```

lib/components/narrative/default/default-itinerary.js
```diff
@@ -1,34 +1,76 @@
...
+  const mainMode = <FormattedMessage id={`common.accessModes.${accessModeId}`} />
+  return transitMode
+    ? <FormattedMessage id='components.DefaultItinerary.multiModeSummary' values={{ accessMode: mainMode, transitMode }} />
+    : mainMode
+}
+
+/**
+ * Formats the given duration according to the selected locale.
+ */
+function formatDuration (duration) {
```
evansiroky, line 46: I recommend changing this into a proper functional component:
```suggestion
function FormattedDuration ({ duration }) {
```

lib/components/narrative/default/default-itinerary.js
```diff
@@ -1,34 +1,76 @@
...
-const { formatDuration, formatTime } = coreUtils.time
+import ItinerarySummary from './itinerary-summary'
+
+const { calculateFares, isBicycle, isTransit, isMicromobility } = coreUtils.itinerary
 
+/**
+ * Obtains the description of an itinerary in the given locale.
+ */
 // FIXME move to core utils
 function getItineraryDescription (itinerary) {
```
evansiroky, line 19: I recommend changing this into a proper functional component:
```suggestion
function ItineraryDescription ({ itinerary }) {
```

lib/components/narrative/default/default-itinerary.js
```diff
@@ -48,18 +90,20 @@ const ITINERARY_ATTRIBUTES = [
...
   },
   {
     id: 'cost',
     order: 2,
-    render: (itinerary, options) => getTotalFareAsString(itinerary)
+    render: (itinerary, options) => {
+      // Get unformatted transit fare portion only (in cents).
+      const { transitFare } = calculateFares(itinerary)
+      return (
+        <FormattedNumber value={transitFare / 100} style='currency' currencyDisplay='narrowSymbol' currency={options.currency} />
```
evansiroky, line 104: Line too long.

lib/components/narrative/default/default-itinerary.js
```diff
@@ -167,6 +213,8 @@ export default class DefaultItinerary extends NarrativeItinerary {
                   options.selection = this.props.sort.type
                 }
                 options.LegIcon = LegIcon
+                options.timeFormat = this.props.use24HourFormat ? 'H:mm' : 'h:mm a'
```
evansiroky, line 216: Please destructure the `use24HourFormat` prop at the beginning of the `render` method with the rest of the props.

lib/components/app/responsive-webapp.js
```diff
@@ -268,10 +278,34 @@ const WebappWithRouter = withRouter(
...
+
+  async updateMessages (previousLocale) {
+    // If the locale changes (or no prevProps are provided), update the messages
+    // for that locale.
+    const {customMessages, locale} = this.props
+    if (locale && (!previousLocale || locale !== previousLocale)) {
+      const messages = await loadLocaleData(locale, customMessages)
+      this.setState({messages})
+    }
+  }
```
evansiroky, line 301: I know @landonreed wants to avoid putting the messages within the redux state, but I really don't understand why. This is just adding unnecessary boilerplate since the state of this component is derived exclusively and unconditionally from the redux state. This is [a react anti-pattern](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#anti-pattern-unconditionally-copying-props-to-state).

evansiroky (changes requested):
See various comments.

lib/components/app/responsive-webapp.js
binh-dam-ibigroup: Does that advice apply to state derived using async calls?

lib/util/i18n.js
binh-dam-ibigroup: There is truth to @evansiroky's comment above, as for each deployment, the user would eventually only choose from a predefined set of languages/locales. How about keeping the proposed code until we find out after adding the UI language selector whether it is still useful?

lib/components/narrative/default/default-itinerary.js
philip-cline: Would the same be true for the this.props.sort.type above?

lib/components/narrative/default/default-itinerary.js
philip-cline: Just out of curiosity, can you explain the advantages of functional components here? More readable?

lib/util/i18n.js
evansiroky: I'd wager that it won't be useful at the point, but I'm fine with this proposal.

lib/components/narrative/default/default-itinerary.js
evansiroky: Even though a generically named function like this can return JSX no problem, having this function named like a component implies an understanding to the developer (at least it does for me) that this method isn't returning simple text or something. Furthermore, by changing the signature to take in an object it can end up reducing the need to refactor the arguments from positional arguments if it is already in a format that a Component is called with.

lib/components/app/responsive-webapp.js
landonreed: I was resistant at first because it seemed somewhat unnecessary since the messages are not really shared state (they're only used by this component), but I'm happy to be overridden if we think the code would be cleaner by stashing them in the redux store/state.

lib/components/narrative/default/default-itinerary.js
evansiroky: Yes please!

lib/components/app/responsive-webapp.js
evansiroky: Even though the value of the messages is only used within this specific component, this value is influenced and can change as a result of an event originating from another component. Regarding Binh's question, my objection doesn't have to do with whether this is async. If some kind of component-specific value uses an async function to obtain it's value, but that initiation event of the call to that async function comes from that same component and the state is not shared in any other parallel component hierarchy then it might be fine to keep it in that component. However, I find that putting logic into actions/reducers makes the code more organized and can keep components decluttered.

binh-dam-ibigroup: I have created an action that loads the localized strings when the locale is set (https://github.com/opentripplanner/otp-react-redux/pull/375/commits/6ea4204258ce54cf8da5a953043aeaffefdbef67).

lib/actions/user.js
binh-dam-ibigroup: Removed in https://github.com/opentripplanner/otp-react-redux/pull/375/commits/2f66ec4f23cc34feb530cac5621277dc7f82e6cb.

lib/actions/user.js
binh-dam-ibigroup: These were removed in https://github.com/opentripplanner/otp-react-redux/pull/375/commits/d0f962fcf847a7ba0d3b9a9b9fa33501ce41e00a.

lib/components/narrative/default/default-itinerary.js
binh-dam-ibigroup: Was updated in https://github.com/opentripplanner/otp-react-redux/pull/375/commits/d0f962fcf847a7ba0d3b9a9b9fa33501ce41e00a.

lib/components/narrative/default/default-itinerary.js
binh-dam-ibigroup: Was updated in https://github.com/opentripplanner/otp-react-redux/pull/375/commits/d0f962fcf847a7ba0d3b9a9b9fa33501ce41e00a.

lib/components/narrative/default/default-itinerary.js
binh-dam-ibigroup: Addressed in https://github.com/opentripplanner/otp-react-redux/pull/375/commits/5e2b6708003925b1163418bd629f3870889ec6a2.

lib/components/narrative/default/default-itinerary.js
binh-dam-ibigroup: Addressed in https://github.com/opentripplanner/otp-react-redux/pull/375/commits/5e2b6708003925b1163418bd629f3870889ec6a2

lib/components/narrative/default/default-itinerary.js
binh-dam-ibigroup: Addressed in https://github.com/opentripplanner/otp-react-redux/pull/375/commits/5e2b6708003925b1163418bd629f3870889ec6a2.

lib/components/narrative/default/default-itinerary.js
binh-dam-ibigroup: Addressed in https://github.com/opentripplanner/otp-react-redux/pull/375/commits/5e2b6708003925b1163418bd629f3870889ec6a2

lib/components/narrative/default/default-itinerary.js
binh-dam-ibigroup: Addressed in https://github.com/opentripplanner/otp-react-redux/pull/375/commits/5e2b6708003925b1163418bd629f3870889ec6a2

lib/components/narrative/default/default-itinerary.js
binh-dam-ibigroup: Addressed in https://github.com/opentripplanner/otp-react-redux/pull/375/commits/5e2b6708003925b1163418bd629f3870889ec6a2

lib/components/narrative/default/default-itinerary.js
binh-dam-ibigroup: Addressed in https://github.com/opentripplanner/otp-react-redux/pull/375/commits/5e2b6708003925b1163418bd629f3870889ec6a2.

lib/components/narrative/default/default-itinerary.js
binh-dam-ibigroup: Addressed in https://github.com/opentripplanner/otp-react-redux/pull/375/commits/5e2b6708003925b1163418bd629f3870889ec6a2.

lib/components/narrative/default/default-itinerary.js
binh-dam-ibigroup: Addressed in https://github.com/opentripplanner/otp-react-redux/pull/375/commits/5e2b6708003925b1163418bd629f3870889ec6a2

lib/components/app/responsive-webapp.js
binh-dam-ibigroup: Addressed in https://github.com/opentripplanner/otp-react-redux/pull/375/commits/5e2b6708003925b1163418bd629f3870889ec6a2

lib/components/narrative/default/default-itinerary.js
```diff
@@ -192,3 +271,14 @@ export default class DefaultItinerary extends NarrativeItinerary {
     )
   }
 }
+
+const mapStateToProps = (state, ownProps) => {
+  return {
+    currency: state.otp.config.localization?.currency || 'USD',
+    locale: state.otp.ui.locale,
+    messages: state.otp.ui.messages,
```
evansiroky, line 279: These are no longer used, please remove.

evansiroky (approved):
Please address 1 comment, then this is good to go.

lib/components/narrative/default/default-itinerary.js
binh-dam-ibigroup: Good catch, updated in https://github.com/opentripplanner/otp-react-redux/pull/375/commits/a9fdcf11538255d1d8ace74fd7fe1ee0fbd7cd64.

lib/components/narrative/default/default-itinerary.js
```diff
@@ -192,3 +271,12 @@ export default class DefaultItinerary extends NarrativeItinerary {
     )
   }
 }
+
+const mapStateToProps = (state, ownProps) => {
+  return {
+    currency: state.otp.config.localization?.currency || 'USD',
```
landonreed, line 277: Should this not be derived from GTFS or the relevant GBFS (or other) providers? I'm not sure if there's a strong justification as to why we need this property.

landonreed (approved):
Looks pretty good. Just a couple of non-blocking comments.

lib/components/narrative/default/default-itinerary.js
binh-dam-ibigroup: We are displaying a cost line item for car-only, walk-only and own-bike-only itineraries, even if no transit itinerary is found and the cost shown is $0.00. We need an "ambient" currency that we don't know if no transit itinerary is shown.
The alternative is to not show the cost item if `itinerary.fare` is not set. Thoughts?

binh-dam-ibigroup: @landonreed says
> Ah, yea, good point, Binh. I'm fine with the ambient currency. That's helpful info that might be useful to throw into a comment though.

Comments regarding this added in https://github.com/opentripplanner/otp-react-redux/pull/375/commits/9b7d311187b3fb919200faf427251bb58668cf0e.

binh-dam-ibigroup:
Adding the Blocked label, I would prefer we cut a release of OTP-RR before we start adding the internationalization features.

PR #377, opened 2021-06-03
chore(lint): Add pre-commit hook to enforce formatting.

binh-dam-ibigroup:
This PR adds a pre-commit hook to run `eslint` on files specified using `git add`. `mastarm` rules are used, ~plus additional rules regarding sorting of prop keys and variable declarations~ (see also https://github.com/conveyal/mastarm/pull/310).

Assumptions:
- You have eslint installed system-wide (otherwise, probably run `yarn global add eslint`)
- ~You have cloned the [mastarm repo](https://github.com/conveyal/mastarm), and repos for your `mastarm` and `otp-react-redux` are in the same directory.~ => Not necessary, the PR uses the existing mastarm dependency.

To test this PR after checkout:
1. Run `yarn install`
2. Add blank spaces/lines to a `.js` file, ~change the order of props or var declarations,~ etc.
3. Commit the changes
4. Observe that your extra spaces ~and prop disordering~ have been fixed, e.g.
image
5. Run `git reset HEAD~` to undo your commit and undo your file changes.

This PR uses a version of husky that is more recent than in OTP-UI, and the way it is configured is thus different. Specifically, husky introduces:
- hook files and its own config directory
- a `postinstall` script that is ran after install, except when `prepublish` is involved. For `prepublish` steps, the library authors recommend disabling installing husky hooks (see [these docs](https://typicode.github.io/husky/#/?id=install-1)).


eslint-precommit.json
```diff
@@ -0,0 +1,36 @@
+{
+  "extends": [
+    "../mastarm/lib/eslintrc.json"
```
evansiroky, line 3: This should be changed to point to the node_modules mastarm since the mastarm project isn't necessarily going to exist in an adjacent directory outside of an IBI development environment.

eslint-precommit.json
```diff
@@ -0,0 +1,36 @@
...
+        "caseSensitive": false
+      }
+    ],
+    "sort-vars": [
+      "error",
+      {
+        "ignoreCase": true
+      }
+    ]
+  }
```
evansiroky, line 35: I didn't realize or forgot about the existence of https://github.com/conveyal/mastarm/pull/308. We shouldn't be added extra linting rules in here if they can be part of the overall linting rules. Furthermore, our tests won't pick up these linting rules. I'll try to contribute to https://github.com/conveyal/mastarm/pull/308 so that a new release of mastarm can be made that includes these new rules.

eslint-precommit.json
binh-dam-ibigroup: Given/pending https://github.com/conveyal/mastarm/pull/310, I have removed this extra eslint settings file in https://github.com/opentripplanner/otp-react-redux/pull/377/commits/e66a45da926abaae948d9460fd49eeee52417e70.

eslint-precommit.json
binh-dam-ibigroup: Great, so the extra rules and this file are no longer needed (https://github.com/opentripplanner/otp-react-redux/pull/377/commits/e66a45da926abaae948d9460fd49eeee52417e70).

evansiroky (approved):

landonreed (approved):

PR #380, opened 2021-06-04
Added the reverse directions button to the Batch routing panel

### LINKED ISSUE #371: Add reverse directions button to batch trip planning form
The button is needed in this version of the trip planning form as well, but was accidentally left out.

robertgregg3:
Fix #371

The Batch routing panel was missing the reverse directions button (switchButton).  

It was present on any configuration that **didn't** require batch routing but not present on configurations that do require batch routine.

Here with the FDOT config the reverse direction button was not showing:
image

Where as here on the ATL Config that doesn't use batch routing it is present:
image

The files that include (or should include) the reverse directions button are: 

- batch-routing-panel.js
- default-main-panel.js

You may need to clear your cache to see the changes. 

I tested on the FDOT and Miami config and the button is now showing. 
image

### DIFF
--- a/lib/components/app/batch-routing-panel.js
+++ b/lib/components/app/batch-routing-panel.js
@@ -36,7 +36,7 @@ class BatchRoutingPanel extends Component {
     const {mobile} = this.props
     const actionText = mobile ? 'tap' : 'click'
     return (
-      <ViewerContainer>
+      <ViewerContainer className='batch-routing-panel'>
         <div className='locations'>
           <LocationField
             inputPlaceholder={`Enter start location or ${actionText} on map...`}
--- a/lib/components/app/batch-routing-panel.js
+++ b/lib/components/app/batch-routing-panel.js
@@ -9,6 +9,7 @@ import LocationField from '../form/connected-location-field'
 import NarrativeItineraries from '../narrative/narrative-itineraries'
 import { getActiveSearch, getShowUserSettings } from '../../util/state'
 import ViewerContainer from '../viewers/viewer-container'
+import switchButton from '../form/switch-button'
 
 // Style for setting the top of the narrative itineraries based on the width of the window.
 // If the window width is less than 1200px (Bootstrap's "large" size), the
@@ -46,6 +47,9 @@ class BatchRoutingPanel extends Component {
           locationType='to'
           showClearButton={!mobile}
         />
+        <div className='switch-button-container'>
+          <switchButton content={<i className='fa fa-exchange fa-rotate-90' />} />
+        </div>
         <BatchSettings />
         {/* FIXME: Add back user settings (home, work, etc.) once connected to
             the middleware persistence.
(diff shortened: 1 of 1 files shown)


lib/components/app/batch-routing-panel.js
```diff
@@ -9,6 +9,7 @@ import LocationField from '../form/connected-location-field'
 import NarrativeItineraries from '../narrative/narrative-itineraries'
 import { getActiveSearch, getShowUserSettings } from '../../util/state'
 import ViewerContainer from '../viewers/viewer-container'
+import switchButton from '../form/switch-button'
```
landonreed, line 12: Component names should generally always be capitalized:
```suggestion
import SwitchButton from '../form/switch-button'
```

lib/components/app/batch-routing-panel.js
```diff
@@ -46,6 +47,9 @@ class BatchRoutingPanel extends Component {
           locationType='to'
           showClearButton={!mobile}
         />
+        <div className='switch-button-container'>
+          <switchButton content={<i className='fa fa-exchange fa-rotate-90' />} />
```
landonreed, line 51:
```suggestion
          <SwitchButton content={<i className='fa fa-exchange fa-rotate-90' />} />
```

landonreed (changes requested):
See suggestions about component naming.

Also, I don't see anything where the button should be when I started the app up. A suggestion for future PRs containing changes to layout: it's often helpful for reviewers (and posterity's sake) to include a screengrab of your changes so reviewers know what they're looking for.

image

binh-dam-ibigroup (changes requested):
Same as @landonreed, I don't see the reverse locations button on the batch routing form.

lib/components/app/batch-routing-panel.js
robertgregg3: Thanks for the feedback. 

I am not sure what happened there. It was working on my machine.  When I loaded it up today it wasn't working...

I changed the code a little, made your change and cleared my cache and it is now working.

binh-dam-ibigroup:
@robertgregg3 can you change the description of the PR to say "Fix #371" please? (so that issue will be closed automatically with the PR)

binh-dam-ibigroup (approved):
LGTM! One housekeeping item: please update the PR description per https://github.com/opentripplanner/otp-react-redux/pull/380#issuecomment-856212501.

binh-dam-ibigroup (approved):

lib/components/app/batch-routing-panel.js
```diff
@@ -35,17 +36,22 @@ class BatchRoutingPanel extends Component {
     const {mobile} = this.props
     const actionText = mobile ? 'tap' : 'click'
     return (
-      <ViewerContainer className='batch-routing-panel'>
```
landonreed, line 38: It looks like this className prop was removed. Is there some reason for that? I suspect it will mess up some styling if we remove it, so I would suggest that we add it back in.

landonreed (changes requested):
See comment about classname.

landonreed (approved):

evansiroky (approved):

PR #385, opened 2021-06-10
Initialize trip monitored days using transit leg serviceDate

### LINKED ISSUE #372: Monitor Trip: Saving weekend trip causes weekend days of the week to be unchecked
1. Plan trip on ATL Rides: https://atl-rides.ibi-transit.com/#/?fromPlace=Peachtree%20Center&toPlace=Midtown
2. Change date to Saturday (make sure you're doing this on a weekday)
2. Log in (so you can save a trip)
3. Click Save trip
4. View the weekend boxes unchecked:
image

## Expected Behavior

The dow for which the trip is planned (i.e., Saturday) definitely should be checked. I imagine that Sunday also should likely be checked.

binh-dam-ibigroup:
Fix #372.

Monitored days for a given itinerary are proposed to be initialized using the `serviceDate` field of the first transit leg as follows:

| If first transit leg happens on | Then initialize monitored days to | Remarks |
| --- | --- | --- |
| Weekday | Monday thru Friday | currently happens for all cases |
| Saturday | Saturday, Sunday | New case |
| Sunday | Saturday, Sunday | New case |

~(only itineraries with transit legs can be monitored).~ If an itinerary does not have transit, the determination above falls back using the itinerary `startTime`.

### DIFF
--- a/lib/util/itinerary.js
+++ b/lib/util/itinerary.js
@@ -58,18 +58,20 @@ function getFirstTransitLeg (itinerary) {
 
 /**
  * Returns the set of monitored days that will be initially shown to the user
- * for the given transit itinerary (itinerary with transit leg).
+ * for the given itinerary.
  * @param itinerary The itinerary from which the default monitored days are extracted.
  * @returns ['monday' thru 'friday'] if itinerary happens on a weekday(*),
  *          ['saturday', 'sunday'] if itinerary happens on a saturday/sunday(*).
- * (*) The first transit leg will be used to make the determination.
+ * (*) For transit itineraries, the first transit leg is used to make
+ * the determination. Otherwise, the itinerary startTime is used.
  */
-export function getTransitItineraryDefaultMonitoredDays (itinerary) {
+export function getItineraryDefaultMonitoredDays (itinerary) {
   const firstTransitLeg = getFirstTransitLeg(itinerary)
-  // Assume a transit leg exists.
-  if (!firstTransitLeg) return null
+  const startMoment = firstTransitLeg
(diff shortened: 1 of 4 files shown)


lib/util/itinerary.js
```diff
@@ -47,3 +48,29 @@ export function itineraryCanBeMonitored (itinerary) {
...
+ * for the given transit itinerary (itinerary with transit leg).
+ * @param itinerary The itinerary from which the default monitored days are extracted.
+ * @returns ['monday' thru 'friday'] if itinerary happens on a weekday(*),
+ *          ['saturday', 'sunday'] if itinerary happens on a saturday/sunday(*).
+ * (*) The first transit leg will be used to make the determination.
+ */
+export function getTransitItineraryDefaultMonitoredDays (itinerary) {
+  const firstTransitLeg = getFirstTransitLeg(itinerary)
+  // Assume a transit leg exists.
+  if (!firstTransitLeg) return null
```
landonreed, line 70: Is returning null helpful here? I think a better alternative would be to just check the start time of the trip and not worry about the service day if this is not a transit itinerary.

landonreed (changes requested):
Looks good, but instead of returning null, we should just future proof this and revert to the start time for non-transit itineraries.

lib/util/itinerary.js
binh-dam-ibigroup: Sure, updated in https://github.com/opentripplanner/otp-react-redux/pull/385/commits/4eaa00046b5dbcda886968309c4452f49cda1237.

landonreed (approved):

evansiroky (approved):

PR #386, opened 2021-06-16
Remove otpState middleman

evansiroky:
This refactor does away with a [middle man code smell ](https://refactoring.guru/smells/middle-man). In a lot of places the `otp` part of the redux state was being extracted and operated upon in various ways. There isn't any advantage to extract just this one part of the redux state and furthermore, often times this `otpState` variable was then being passed into selector functions. This presents a problem for work on the field trip module as I want to modify a certain selector to be able to use part of the redux state outside of the `otp` part of the redux state. 

Therefore, this PR removes all intermediate extractions of the `otp` variable from the redux state. Furthermore, this PR modifies every function that accepted the `otpState` as an input and instead opts to receive the whole redux state object. This should help reduce the amount of change that selector function signatures have in the future if we decide to refactor the structure of the redux state variable.

The following functions have had their signatures changed from taking in the `otpState` variable to the `state` variable:

- actions.api
  - `constructRoutingQuery`
  - `formatRecentSearch`
  - `getActiveItinerary`
  - `getOtpFetchOptions`
  - `updateOtpUrlParams`
- util.state
  - `getActiveError`
  - `getActiveItineraries`
  - `getActiveItinerary`
  - `getActiveSearch`
  - `getActiveSearchNonRealtimeResponse`
  - `getActiveSearchRealtimeResponse`
  - `getRealtimeEffects`
  - `getResponsesWithErrors`
  - `getShowUserSettings`
  - `getStopViewerConfig`
  - `getUiUrlParams`
  - `queryIsValid`


lib/actions/api.js
```diff
@@ -8,10 +8,11 @@ import queryParams from '@opentripplanner/core-utils/lib/query-params'
 import { createAction } from 'redux-actions'
 import qs from 'qs'
 
-import { rememberPlace } from './map'
 import { getStopViewerConfig, queryIsValid } from '../util/state'
 import { getSecureFetchOptions } from '../util/middleware'
 
+import { rememberPlace } from './map'
+
```
evansiroky, line 15: Husky did this to a few files.

landonreed (commented):
In general, this makes sense, but this is the kind of refactor that I would be much more comfortable with applying *after* we add typescript to this repo.

evansiroky:
> In general, this makes sense, but this is the kind of refactor that I would be much more comfortable with applying _after_ we add typescript to this repo.

I agree that if TypeScript support were available, this would be easier to verify (more tests would help too). However, it seems like TypeScript support for this library may still be a ways away since a lot of things in otp-ui could use TypeScript definitions to fill in a lot of gaps with adding TypeScript in this package. Lmk if your comment is blocking.

binh-dam-ibigroup (approved):
I'm inclined to accept, although it is not obvious which instances of middle-man code were removed ([per this](https://refactoring.guru/remove-middle-man)) as part of this process.

landonreed (approved):

PR #387, opened 2021-06-17
fix(routingQuery): Properly revert to list view after planning trip.

### LINKED ISSUE #373: Batch Trips: Revert to list view after re-planning trip
1. Plan trip: https://atl-rides.ibi-transit.com/#/?fromPlace=Peachtree%20Center&toPlace=Midtown
2. Select option.
2. Change date. Click plan button.
3. View that the first option is selected (we should return to list view).

philip-cline:
### Checklist

- [x] Appropriate branch selected _(all PRs must first be merged to `dev` before they can be merged to `master`)_
- [x] Any modified or new methods or classes have helpful JavaDoc and code is thoroughly commented
- [x] The description lists all applicable issues this PR seeks to resolve
- [x] The description lists any configuration setting(s) that differ from the default settings
- [x] All tests and CI builds passing

### Description
fix #373 

Currently, if an option is selected, the batch trips Itinerary View will not revert to the default list view but will instead show the full expanded option. To duplicate these steps:
Plan trip: https://atl-rides.ibi-transit.com/#/?fromPlace=Peachtree%20Center&toPlace=Midtown
Select option.
Change date. Click plan button.
View that the first option is selected (we should return to list view).

To fix this, when a valid routingQuery is made, the itinerary view is now set to DEFAULT. Note the ItineraryView options include two options for 'list' view (DEFAULT and LIST).

### DIFF
--- a/lib/actions/ui.js
+++ b/lib/actions/ui.js
@@ -240,8 +240,6 @@ export const MobileScreens = {
  * (currently only used in batch results).
  */
 export const ItineraryView = {
-  /** FIXME: Duplicate value for 'list', DEFAULT and LIST */
-  DEFAULT: 'list',
   /** One itinerary is shown. (In mobile view, the map is hidden.) */
   FULL: 'full',
   /** One itinerary is shown, itinerary and map are focused on a leg. (The mobile view is split.) */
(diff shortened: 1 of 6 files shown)


binh-dam-ibigroup (approved):
Thanks for fixing this bug!

philip-cline:
Just tagging @evansiroky if you have any input here.

lib/actions/ui.js
```diff
@@ -240,6 +240,7 @@ export const MobileScreens = {
  * (currently only used in batch results).
  */
 export const ItineraryView = {
+  /** FIXME: Duplicate value for 'list', DEFAULT and LIST */
   DEFAULT: 'list',
```
landonreed, line 244: @evansiroky and @binh-dam-ibigroup, do you see any reason to maintain `LIST` and `DEFAULT` (which both have `list` as their value)? I'm of the mind that we should maybe just entirely replace `DEFAULT` with `LIST`, but wanted to get your feedback in case there was a good reason to keep `DEFAULT`.

binh-dam-ibigroup: The alternative would be to replace the DEFAULT entry with
```js
const DEFAULT_ITINERARY_VIEW = ItineraryView.LIST
```
(Whether the view state is default determines whether the `ui_itineraryView` parameter is added to the URL).

evansiroky: I'm in favor of removing DEFAULT completely as it'd be more explicit.

evansiroky (approved):

lib/actions/ui.js
landonreed: @philip-cline, could you move forward with removing the `DEFAULT` option and replacing with `LIST`?

landonreed (approved):
Approved on the condition that `DEFAULT` is removed from `ItineraryView`.

PR #388, opened 2021-06-18
Field trip group size assignment, planning and saving

evansiroky:
This PR ports the relevant code from the OpenTripPlanner front-end that'll split up large field trip groups across multiple itineraries and plan accordingly given constraints found with any other field trips that are already planned. This PR also implements the ability to save itineraries as the inbound or outbound journey of a field trip request. See screenshot:

Screen Shot 2021-06-18 at 10 32 24 AM

The groups are split into subgroups according to a configurable (and set of default) modal capacities.

To set custom modal configs add the following yml to the field trip module in the `modules` section in `config.yml`:

```yml
- id: ft
  modeCapacities:
    TRAM: 60
    BUS: 30
```


evansiroky:
I'm assigning this back to myself as I intend to rewrite this after learning about more of the existing logic in the legacy OTP frontend code.

lib/components/admin/trip-status.js
```diff
@@ -47,9 +48,15 @@ class TripStatus extends Component {
...
+  _onPlanTrip = () => {
+    const { outbound, planTrip, request } = this.props
+    planTrip(request, outbound)
+  }
 
-  _onSaveTrip = () => this.props.saveRequestTrip(this.props.request, this.props.outbound)
+  _onSaveTrip = () => {
+    const { outbound, request, saveRequestTripItineraries } = this.props
+    saveRequestTripItineraries(request, outbound)
+  }
```
landonreed, line 59: It wasn't super clear in the ticket, but I think we also need a `viewTrip` method/button that will load the an already saved trip into the main panel.

lib/util/state.js
```diff
@@ -255,7 +259,8 @@ export const getActiveItineraries = createSelector(
     const {direction, type} = sort
     // If no sort type is provided (e.g., because batch routing is not enabled),
     // do not sort itineraries (default sort from API response is used).
-    return !type
+    // Also, do not sort itineraries if the a field trip request is active
+    return (!type || Boolean(activeFieldTripRequest))
```
landonreed, line 263: I think type is set to null unless batch routing is enabled. Batch routing and the field trip shouldn't generally be enabled at the same time. This change is probably ok, but I don't necessarily think it's needed.

lib/actions/field-trip.js
```diff
@@ -249,17 +406,355 @@ function planInbound (request) {
...
+        `Error fetching trips for field trip travel date: ${JSON.stringify(e)}`
+      )
+      return
+    }
+
+    // create a new searchId to use for making all requests
+    const searchId = randId()
+
+    // set numItineraries param for field trip requests
+    dispatch(setQueryParam({ numItineraries: 1 }))
```
landonreed, line 446: I assume there's a technical reason for doing 10 requests with one itinerary rather than one request with ten itineraries? We should probably document that here (or somewhere else that's appropriate).

lib/actions/field-trip.js
```diff
@@ -249,17 +406,355 @@ function planInbound (request) {
...
+ *  response index and list of itineraries that can be used to update the state
+ *  with modified itinerary objects that can help display the field trip group
+ *  size assigned to each itinerary.
+ * @return {Object} result.remainingGroupSize The remaining about of people in
+ *  the field trip group that have yet to be assigned to an itinerary.
+ * @return {Object} result.tripsToBanInSubsequentSearches An array of strings
+ *  representing tripIds that should be added to the list of tripIds to ban when
+ *  making requests for additional itineraries from OTP.
+ */
+function checkValidityAndCapacity (state, request) {
```
landonreed, line 598: Nit: should this be in field trip utils?

lib/actions/field-trip.js
```diff
@@ -249,17 +406,355 @@ function planInbound (request) {
...
+    assignedItinerariesByResponse,
+    remainingGroupSize,
+    tripsToBanInSubsequentSearches
   }
 }
 
+/**
+ * Formats a field trip's travel date into the given format.
+ */
+function getFormattedRequestTravelDate (
```
landonreed, line 712: Nit: should this be in field trip utils file?

lib/actions/field-trip.js
```diff
@@ -249,17 +406,355 @@ function planInbound (request) {
...
+
+/**
+ * Calculates the mode capacity based on the field trip module mode capacities
+ * (if it exists) or from the above default lookup of mode capacities or if
+ * given an unknown mode, then the unknownModeCapacity is returned.
+ *
+ * @param {Object} fieldTripModuleConfig the field trip module config
+ * @param {string} mode the OTP mode
+ */
+function getFieldTripGroupCapacityForMode (fieldTripModuleConfig, mode) {
```
landonreed, line 740: Nit: should this be in field trip utils file?

lib/actions/field-trip.js
```diff
@@ -249,17 +406,355 @@ function planInbound (request) {
...
     }
     dispatch(setQueryParam(queryParams))
-    dispatch(routingQuery())
+    dispatch(makeFieldTripPlanRequests(request))
+  }
+}
+
+/**
+ * Makes appropriate OTP requests until enough itineraries have been found to
+ * accommodate the field trip group.
```
landonreed, line 421: There's some complexity to this action that I think would be helpful if included in this method's jsdoc.

lib/actions/field-trip.js
```diff
@@ -249,17 +406,355 @@ function planInbound (request) {
...
+  })
+  await dispatch(receivedTravelDateTrips(trips))
+}
+
+/**
+ * Makes the needed requests to the OTP index API to obtain semantic trip hashes
+ * for any tripIds in the active itineraries that haven't already been obtained
+ * from the OTP index API.
+ */
+function getTripHashesFromActiveItineraries () {
```
landonreed, line 557: Method name is slightly confusing. It sounds like the hashes are coming from the itin objects. Simple name change to `getMissingTripHashesForActiveItineraries` might be helpful.

lib/actions/field-trip.js
```diff
@@ -249,17 +406,355 @@ function planInbound (request) {
...
+    const prePlanState = getState()
+    const fieldTripModuleConfig = getModuleConfig(
+      prePlanState,
+      Modules.FIELD_TRIP
+    )
+
+    // request other known trip IDs that other field trips are using on the
+    // field trip request date
+    try {
+      await getTripIdsForTravelDate(dispatch, prePlanState, request)
```
landonreed, line 434: Since we're awaiting this statement, is it really necessary to pass dispatch and prePlanState in this way? I would prefer to make getTripIdsForTravelDate a standard thunk action.

lib/actions/field-trip.js
```diff
@@ -249,17 +406,355 @@ function planInbound (request) {
...
+        // tripIds still stored as 'agencyAndId' in DB, so convert them to
+        // be compatible with OTP responses
+        gtfsTrip.tripId = gtfsTrip.agencyAndId.replace('_', ':')
+        // Add passengers to each trip from group itinerary
+        gtfsTrip.passengers = groupItinerary.passengers
+        trips.push(gtfsTrip)
+      })
+    })
+  })
+  await dispatch(receivedTravelDateTrips(trips))
```
landonreed, line 549: Does this need to be awaited? Seems like it wouldn't.

lib/actions/field-trip.js
```diff
@@ -249,17 +406,355 @@ function planInbound (request) {
...
+  return (configModeCapacities && configModeCapacities[mode]) ||
+      defaultFieldTripModeCapacities[mode] ||
+      unknownModeCapacity
+}
+
+/**
+ * Dispatches the appropriate cleanup actions after making requests to find the
+ * itineraries for an inbound or outbound field trip journey.
+ */
+function doFieldTripPlanRequestCleanup (dispatch, searchId) {
```
landonreed, line 751: I don't see too much value in passing dispatch as a param vs. just using the boilerplate thunk syntax.

lib/actions/field-trip.js
```diff
@@ -249,17 +406,355 @@ function planInbound (request) {
...
+    const searchId = randId()
+
+    // set numItineraries param for field trip requests
+    dispatch(setQueryParam({ numItineraries: 1 }))
+
+    // create a new set to keep track of trips that should be banned
+    const bannedTrips = new Set()
+
+    // track number of requests made such that endless requesting doesn't occur
+    const maxRequests = fieldTripModuleConfig?.maxRequests || 10
```
landonreed, line 452: Perhaps we should just return a default field trip config from `getModuleConfig` if it does not exist, so that we can safely destructure (and avoid the optional chaining)?

lib/actions/field-trip.js
```diff
@@ -249,17 +406,355 @@ function planInbound (request) {
...
+ * @return {Object} result.remainingGroupSize The remaining about of people in
+ *  the field trip group that have yet to be assigned to an itinerary.
+ * @return {Object} result.tripsToBanInSubsequentSearches An array of strings
+ *  representing tripIds that should be added to the list of tripIds to ban when
+ *  making requests for additional itineraries from OTP.
+ */
+function checkValidityAndCapacity (state, request) {
+  const fieldTripModuleConfig = getModuleConfig(state, Modules.FIELD_TRIP)
+  const minimumAllowableRemainingCapacity =
+    fieldTripModuleConfig?.minimumAllowableRemainingCapacity || 10
```
landonreed, line 601: Perhaps we should just return a default field trip config from `getModuleConfig` if it does not exist, so that we can safely destructure (and avoid the optional chaining)?

lib/actions/field-trip.js
```diff
@@ -249,17 +406,355 @@ function planInbound (request) {
...
+          // not used by another vehicle, so not used by this other field trip
+          if (!sameVehicleTrip) return
+
+          // check if the stop ranges overlap. It is OK if one field trip begins
+          // where the other ends.
+          if (
+            leg.from.stopIndex >= tripInUse.toStopIndex ||
+              leg.to.stopIndex <= tripInUse.fromStopIndex
+          ) {
+            // legs do not overlap, so not used by this other field trip
```
landonreed, line 655: Anywhere we have a return statement in this forEach, I think the comment should be changed from: 
```
... so not used by this other field trip
// to
... so this trip/vehicle is free to use
```

lib/actions/field-trip.js
```diff
@@ -249,17 +406,355 @@ function planInbound (request) {
...
+        // TODO: a more advanced way of doing things might be to ban trip
+        //        sequences to not find the same exact sequence, but also
+        //        individual trips that are too full.
+        tripsToBanInSubsequentSearches.push(tripId)
+
+        itineraryCapacity = Math.min(itineraryCapacity, remainingTripCapacity)
+      })
+
+      if (itineraryCapacity > 0) {
+        // itinerary is possible, add to list and update remaining group size.
```
landonreed, line 689:
```suggestion
        // itinerary has capacity, add to list and update remaining group size.
```

lib/reducers/create-otp-reducer.js
```diff
@@ -312,6 +327,33 @@ function createOtpReducer (config) {
...
+        return update(state, {
+          searches: {
+            [searchId]: {
+              pending: { $set: action.payload.pending }
+            }
+          }
+        })
+      case 'SET_ACTIVE_ITINERARIES':
+        const responseUpdate = {}
+        Object.entries(action.payload.assignedItinerariesByResponse)
```
landonreed, line 340: I have a slight preference for iterating over `Object.keys` here rather than `Object.entries`, just because we have a habit of iterating over keys much more often.

lib/actions/api.js
```diff
@@ -107,15 +114,25 @@ export function routingQuery (searchId = null) {
         ({mode, params}) => ({mode, ...params})
       )
       : [{}]
-    dispatch(routingRequest({ activeItinerary, routingType, searchId, pending: iterations.length }))
-    iterations.forEach((injectedParams, i) => {
+    dispatch(routingRequest({
+      activeItinerary,
+      updateSearchInReducer,
```
landonreed, line 119: If we're alphabetizing object fields,`updateSearchInReducer` should go at the end.

lib/actions/api.js
```diff
@@ -77,8 +81,11 @@ function getActiveItinerary (state) {
  * that are no longer contained in the state we don't confuse the search IDs
  * with search IDs from the new session. If we were to use sequential numbers
  * as IDs, we would run into this problem.
+ *
+ * The updateSearchInReducer instructs the reducer to update an existing search
+ * if it exists. This is used by the field trip module.
  */
-export function routingQuery (searchId = null) {
+export function routingQuery (searchId = null, updateSearchInReducer = false) {
   return async function (dispatch, getState) {
```
landonreed, line 89: This `async` keyword looks to be unnecessary.

```suggestion
  return function (dispatch, getState) {
```

lib/reducers/create-otp-reducer.js
```diff
@@ -256,22 +256,37 @@ function createOtpReducer (config) {
...
-                timestamp: getTimestamp()
-              }
-            }
-          },
+        const {
+          activeItinerary,
+          pending,
+          updateSearchInReducer
+        } = action.payload
+        const searchUpdate = updateSearchInReducer
```
landonreed, line 264: Non-blocking: Might be good to throw this ternary into a method that's stored elsewhere?

landonreed (changes requested):
Just a few changes here and there, but this is looking really good overall.

lib/actions/field-trip.js
```diff
@@ -249,17 +406,355 @@ function planInbound (request) {
...
+  'TRAM': 80,
+  'SUBWAY': 120,
+  'RAIL': 80,
+  'BUS': 40,
+  'FERRY': 100,
+  'CABLE_CAR': 20,
+  'GONDOLA': 15,
+  'FUNICULAR': 20
+}
+const unknownModeCapacity = 15
```
binh-dam-ibigroup, line 730: Should this constant be spelled uppercase?

lib/reducers/create-otp-reducer.js
binh-dam-ibigroup: Is there a point in iterating over keys if not using them?

binh-dam-ibigroup (approved):
Looks good and seems to work, but do fix the typos. My other formatting comments are non-blocking.

lib/actions/field-trip.js
evansiroky: It could, but it is only ever used in this file and is a core part to the `makeFieldTripPlanRequests` function.

lib/actions/field-trip.js
evansiroky: It could, but is only ever used in this file.

lib/actions/field-trip.js
evansiroky: I'd prefer the `getModuleConfig` method to return something non-truthy if a module doesn't exist.

lib/actions/field-trip.js
evansiroky: I'd prefer the `getModuleConfig` method to return something non-truthy if a module doesn't exist.

lib/reducers/create-otp-reducer.js
evansiroky: I think the use of `Object.entries` is appropriate and clean in this instance.

lib/reducers/create-otp-reducer.js
evansiroky: I'm not quite following. What would this look like?

lib/actions/field-trip.js
```diff
@@ -249,17 +406,355 @@ function planInbound (request) {
...
+
+  // initialize the remaining group size to be the total group size
+  let remainingGroupSize = getGroupSize(request)
+  const assignedItinerariesByResponse = {}
+  const tripsToBanInSubsequentSearches = []
+
+  // iterate through responses as we need to keep track of the response that
+  // each itinerary came from so that they can be updated
+  activeSearch.response.forEach((response, responseIdx) => {
+    if (!response.plan) {
```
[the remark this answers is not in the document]
evansiroky: Actually, the responseIdx is relevant, so filtering out items would potentially change the `responseIdx` variable which would result in not being able to assign items properly in the `setActiveItineraries` action.

lib/actions/field-trip.js
evansiroky: [According to Airbnb, no](https://github.com/airbnb/javascript/#naming--uppercase)?

lib/actions/field-trip.js
```diff
@@ -249,17 +406,355 @@ function planInbound (request) {
...
+
+    // iterate through itineraries to check validity and assign field trip
+    // groups
+    response.plan.itineraries.forEach((itinerary, idx) => {
+      let itineraryCapacity = Number.POSITIVE_INFINITY
+
+      // check each individual trip to see if there aren't any conflicts
+      itinerary.legs.forEach(leg => {
+        // non-transit legs have infinite capacity
+        if (!isTransit(leg.mode)) {
```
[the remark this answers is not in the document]
evansiroky: Refactored in a4480936984a2a1cb2d9dd6f97adc4ecd3a9a12c

lib/actions/field-trip.js
```diff
@@ -249,17 +406,355 @@ function planInbound (request) {
...
+          let sameVehicleTrip = false
+          if (tripId in tripHashLookup && tripInUse.tripHash) {
+            // use the trip hashes if available
+            sameVehicleTrip = (tripHashLookup[tripId] === tripInUse.tripHash)
+          } else {
+            // as fallback, compare the tripId strings
+            sameVehicleTrip = (tripId === tripInUse.tripId)
+          }
+          // not used by another vehicle, so not used by this other field trip
+          if (!sameVehicleTrip) return
```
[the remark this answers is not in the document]
evansiroky: Extracted to another function in a4480936984a2a1cb2d9dd6f97adc4ecd3a9a12c

lib/actions/field-trip.js
```diff
@@ -249,17 +406,355 @@ function planInbound (request) {
...
+      // An error might have occurred!
+      return
+    }
+
+    // iterate through itineraries to check validity and assign field trip
+    // groups
+    response.plan.itineraries.forEach((itinerary, idx) => {
+      let itineraryCapacity = Number.POSITIVE_INFINITY
+
+      // check each individual trip to see if there aren't any conflicts
```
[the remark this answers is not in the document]
evansiroky: Expanded on comment in a4480936984a2a1cb2d9dd6f97adc4ecd3a9a12c

lib/actions/field-trip.js
```diff
@@ -144,82 +165,215 @@ export function editSubmitterNotes (request, submitterNotes) {
...
+  })
+  return serialize(data, { indices: true })
+}
 
-  return { isValid: true, message: null }
+/**
+ * Creates an OTP-style location string based on the location name, lat and lon.
+ */
+function getOTPLocationString (location) {
+  const llString = `${location.lat},${location.lon}`
```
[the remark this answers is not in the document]
evansiroky: Renamed in a4480936984a2a1cb2d9dd6f97adc4ecd3a9a12c

lib/actions/field-trip.js
```diff
@@ -144,82 +165,215 @@ export function editSubmitterNotes (request, submitterNotes) {
...
+    data.gtfsTrips.push(gtfsTripsForItinerary)
+  })
+  return serialize(data, { indices: true })
+}
 
-  return { isValid: true, message: null }
+/**
+ * Creates an OTP-style location string based on the location name, lat and lon.
+ */
+function getOTPLocationString (location) {
```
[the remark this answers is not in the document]
evansiroky: Renamed in a4480936984a2a1cb2d9dd6f97adc4ecd3a9a12c

lib/actions/field-trip.js
```diff
@@ -26,6 +43,10 @@ export const setActiveFieldTrip = createAction('SET_ACTIVE_FIELD_TRIP')
 export const setGroupSize = createAction('SET_GROUP_SIZE')
 export const toggleFieldTrips = createAction('TOGGLE_FIELD_TRIPS')
```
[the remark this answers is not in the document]
evansiroky: Added comment about this in a4480936984a2a1cb2d9dd6f97adc4ecd3a9a12c

lib/actions/api.js
evansiroky: Removed in a4480936984a2a1cb2d9dd6f97adc4ecd3a9a12c

lib/actions/api.js
evansiroky: Alphabetized in a4480936984a2a1cb2d9dd6f97adc4ecd3a9a12c

lib/actions/field-trip.js
evansiroky: Refactored into a thunk in a4480936984a2a1cb2d9dd6f97adc4ecd3a9a12c

lib/actions/field-trip.js
evansiroky: Changed to return statement after refactoring to be within a thunk in a4480936984a2a1cb2d9dd6f97adc4ecd3a9a12c

lib/actions/field-trip.js
evansiroky: Refactored in a4480936984a2a1cb2d9dd6f97adc4ecd3a9a12c

lib/actions/field-trip.js
evansiroky: Changed in a4480936984a2a1cb2d9dd6f97adc4ecd3a9a12c

lib/actions/field-trip.js
evansiroky: Added more comments in a4480936984a2a1cb2d9dd6f97adc4ecd3a9a12c

lib/actions/field-trip.js
evansiroky: Inlined in a4480936984a2a1cb2d9dd6f97adc4ecd3a9a12c

lib/actions/field-trip.js
evansiroky: Added more comments in a4480936984a2a1cb2d9dd6f97adc4ecd3a9a12c

lib/actions/field-trip.js
evansiroky: Refactored in a448093

lib/components/admin/trip-status.js
evansiroky: Implemented this functionality in the last few commits.

evansiroky:
Ready for review again.

lib/reducers/create-otp-reducer.js
landonreed:
```js
function getSearchUpdate (payload) {
  const {
    activeItinerary,
    pending,
    updateSearchInReducer
  } = payload
  return updateSearchInReducer ? {
    activeItinerary: { $set: activeItinerary },
    activeLeg: { $set: null },
    activeStep: { $set: null },
    pending: { $set: pending },
    // FIXME: get query from action payload?
    query: { $set: clone(state.currentQuery) },
    // omit requests reset to make sure requests can be added to this
    // search
    timestamp: { $set: getTimestamp() }
  }
  : {
    $set: {
      activeItinerary,
      activeLeg: null,
      activeStep: null,
      pending,
      // FIXME: get query from action payload?
      query: clone(state.currentQuery),
      response: [],
      timestamp: getTimestamp()
    }
  }
}
```

binh-dam-ibigroup:
There is a conflict as a result of a recent merge of #375.

evansiroky:
> There is a conflict as a result of a recent merge of #375.

Merge conflict fixed.

lib/components/admin/trip-status.js
```diff
@@ -6,53 +6,80 @@ import { connect } from 'react-redux'
...
+          <span>
+            No itineraries planned! Click Plan to plan trip.
+          </span>
+        </Para>
+      )
+    }
+    return (
+      <>
+        <Para>
+          <span>
```
binh-dam-ibigroup, line 68: Remove the extra `<span>`.

lib/components/admin/trip-status.js
```diff
@@ -6,53 +6,80 @@ import { connect } from 'react-redux'
...
 
-  _onPlanTrip = () => this.props.planTrip(this.props.request, this.props.outbound)
-
-  _onSaveTrip = () => this.props.saveRequestTrip(this.props.request, this.props.outbound)
+  _renderTripStatus = () => {
+    const { status, trip } = this.props
+    if (!status || !trip) {
+      return (
+        <Para>
+          <span>
```
binh-dam-ibigroup, line 59: Remove the extra `<span>`.

binh-dam-ibigroup (approved):
Remove extra `<span>` tags and address one tiny comment tweak, but approval still stands.

lib/components/admin/trip-status.js
evansiroky: Removed in e381aa802e86af0e6889e57e83cec419af1d5589

lib/components/admin/trip-status.js
evansiroky: Removed in e381aa802e86af0e6889e57e83cec419af1d5589

lib/actions/field-trip.js
```diff
@@ -748,11 +769,65 @@ function getFieldTripGroupCapacityForMode (fieldTripModuleConfig, mode) {
...
+
+    // set the itineraries in a new OTP response
+    dispatch(setActiveItinerariesFromFieldTrip({
+      response: [{ plan: { itineraries } }],
+      searchId
+    }))
+
+    // appropriately initialize the Url params. If this doesn't happen, it won't
+    // be possible to set an active itinerary properly due to funkiness with the
+    // change in URL
```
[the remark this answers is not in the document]
evansiroky: Matched in e381aa802e86af0e6889e57e83cec419af1d5589

landonreed (approved):

PR #389, opened 2021-06-23
Add daily report feature; improve perf

landonreed:
This PR handles a couple of items needed for the field trip module:
- adds the daily report button (in footer of field trip lists, see screenshot)
- aims to improve perf in field trip list by memoizing the filter function
- adds travel date to field trip list item

image

### DIFF
--- a/lib/components/admin/field-trip-list.js
+++ b/lib/components/admin/field-trip-list.js
@@ -1,19 +1,28 @@
+import memoize from 'lodash.memoize'
+import moment from 'moment'
 import React, { Component } from 'react'
-import { Badge } from 'react-bootstrap'
+import { Badge, Button } from 'react-bootstrap'
 import { connect } from 'react-redux'
 
 import * as fieldTripActions from '../../actions/field-trip'
-import DraggableWindow from './draggable-window'
 import Icon from '../narrative/icon'
 import Loading from '../narrative/loading'
-import {FieldTripRecordButton, WindowHeader} from './styled'
 import {getVisibleRequests, TABS} from '../../util/call-taker'
 import {FETCH_STATUS} from '../../util/constants'
+
+import {FieldTripRecordButton, WindowHeader} from './styled'
+import DraggableWindow from './draggable-window'
 
 /**
  * Displays a searchable list of field trip requests in a draggable window.
  */
 class FieldTripList extends Component {
+  constructor (props) {
+    super(props)
+    this.state = {
+      date: moment().startOf('day').format('YYYY-MM-DD')
+    }
+  }
   _onClickFieldTrip = (request) => {
     const {callTaker, fetchFieldTripDetails, setActiveFieldTrip} = this.props
     if (request.id === callTaker.fieldTrip.activeId) {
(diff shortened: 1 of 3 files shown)


lib/components/admin/field-trip-list.js
```diff
@@ -127,6 +159,32 @@ class FieldTripList extends Component {
...
+        key={tab.id}
+        name={tab.id}
+        onClick={onClick}
+        style={style}
+      >
+        {tab.label} <Badge>{this.getCount(data, tab.filter)}</Badge>
+      </button>
+    )
+  }
+}
```
evansiroky, line 186: I guess using `lodash.memoize` works here. However, since `lodash.memoize` caches based on the first argument, it's accomplishing the memorization without the second argument affecting the cache key calculation, which is fine since that argument wouldn't get updated.  Another way of doing this without lodash.memoize would be to use React's `useMemo`:

```suggestion
function Tab ({data, filter, onClick, tab}) {
  const count = React.useMemo(() => data.filter(tab.filter).length, [data, tab])
  const active = tab.id === filter.tab
  const style = {
    backgroundColor: active ? 'navy' : undefined,
    borderRadius: 5,
    color: active ? 'white' : undefined,
    padding: '2px 3px'
  }
  return (
    <button
      className='clear-button-formatting'
      key={tab.id}
      name={tab.id}
      onClick={onClick}
      style={style}
    >
      {tab.label} <Badge>{count}</Badge>
    </button>
  )
}
```

However, in either of the `lodash.memoize` or `useMemo` approaches, if the FieldTripList component is unmounted then these Tab instances and the cache saved in those `lodash.memoize`/`useMemo` instances are lost. I'm not sure how often call-takers would close and then reopen the window and how much of a performance hit this would be to recalculate everything. If that were a big deal, then another way to do this would be to use reselect to compute the values for all of the tab counts and return an object where the count values could be looked up by `tab.id`.

lib/components/admin/field-trip-list.js
```diff
@@ -174,11 +234,11 @@ class FieldTripRequestRecord extends Component {
               {this._getStatusIcon(outboundTripStatus)} Outbound
             </span>
           </span>
-          <span style={{display: 'block'}}>
+          <span style={{display: 'block', fontSize: '.9em'}}>
```
evansiroky, line 237: Converting a lot of these components into styled components instead of having inline css might also improve performance. However, having a virtualized list for all these items would probably be even better.

evansiroky (approved):
I'm not exactly sure what performance issues are going on that we're trying to resolve, but putting that aspect aside it looks good.

lib/components/admin/field-trip-list.js
landonreed: The main perf case I was trying to resolve was when users search or switch tabs. This looks great, thanks for sharing the suggestion.

lib/components/admin/field-trip-list.js
```diff
@@ -30,6 +39,14 @@ class FieldTripList extends Component {
     this.props.setActiveFieldTrip(null)
   }
 
+  _getReportUrl = () => {
+    const {datastoreUrl} = this.props
+    const {date} = this.state
+    const [year, month, day] = date.split('-')
+    const params = {month, day, year}
+    return `${datastoreUrl}/fieldtrip/opsReport?month=${qs.stringify(params)}`
```
binh-dam-ibigroup, line 47: Is this a bug on my side?

image

landonreed: No idea. But I do see that the month= needs to be removed from that URL string. Fixed in 72cd7de52e3bd5e38606f7fb7ffd3210ef034d3f

binh-dam-ibigroup (approved):
Inclined to approve as is.

PR #378, opened 2021-06-03
Mailables

landonreed:
Add mailables functionality for TriMet.

PR #391, opened 2021-06-24
Separate MailableOption into 2 components

evansiroky:
Following up on https://github.com/opentripplanner/otp-react-redux/pull/378#discussion_r658287419, this PR proposes to use 2 separate components for the previous MailableOption. Also, it converts some inline styles to styled components.

### DIFF
--- a/lib/components/admin/styled.js
+++ b/lib/components/admin/styled.js
@@ -163,7 +163,7 @@ export const MailablesList = styled.div`
   overflow-y: scroll;
 `
 
-export const MailableItem = css`
+const mailableItemCss = css`
   background-color: #eaeaea;
   display: block;
   margin: 0px 0px 2px 0px;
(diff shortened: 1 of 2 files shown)


lib/components/admin/styled.js
```diff
@@ -161,3 +162,26 @@ export const MailablesList = styled.div`
   max-height: 120px;
   overflow-y: scroll;
 `
+
+export const MailableItem = css`
```
landonreed, line 166: I don't think this needs to be exported. Also, I think we generally don't capitalize these css variables.
```suggestion
const mailableItemCss = css`
```

lib/components/admin/styled.js
```diff
@@ -161,3 +162,26 @@ export const MailablesList = styled.div`
...
+  background-color: #eaeaea;
+  display: block;
+  margin: 0px 0px 2px 0px;
+  max-width: 290px;
+  padding: 3px 2px;
+  width: 100%;
+`
+
+export const SelectableMailableButton = styled.button`
+  ${MailableItem}
```
landonreed, line 176:
```suggestion
  ${mailableItemCss}
```

lib/components/admin/styled.js
```diff
@@ -161,3 +162,26 @@ export const MailablesList = styled.div`
...
+`
+
+export const SelectedMailableContainer = styled.div`
+  ${MailableItem}
```
landonreed, line 180:
```suggestion
  ${mailableItemCss}
```

landonreed (approved):
Looks good. Just a couple of tweaks. If you agree, feel free to apply suggestions and merge.

PR #395, opened 2021-06-28
Fix calltaker recording

### LINKED ISSUE #392: Calltaker: Clicking a search from a previous call should not add it to current call
When a user clicks a previous search in the call history window, it should:
- plan that trip immediately
- not start a new call nor log a query in an existing call.

Original message:

> Regarding the potential call history recording bug, I was not able to duplicate it (or it wasn't ever an issue to begin with), but I did note the following [issue]:
> I think we've discussed this before, but while I'm in an active recording, if I click on one of the prior entries, it creates a new entry.  (That is, if I click on the first trip I planned in the current recording, because maybe the customer had a question about it, it creates a new entry in that recording (so maybe I had done 3, then went back to the first, and now it shows a total of 4 for that call session/recording)).

### LINKED ISSUE #393: Calltaker: Clicking "End call" button should update list of queries for ended call.
Original message:

> Also, I just noticed a call history quirk.  I selected a previously recorded call, clicked on one of the recorded trips, changed a field and then pressed plan.  This auto-started a new call.  Then I clicked the button to end the recording.  When I did this, it added a new entry to the history, but said no queries were recorded.  The 8:20 entry for today in the screenshot below (the 9 min one) is the one I selected, and the three above it are results of what I described above.  Let me know if you'd like intermediate screenshots.

### LINKED ISSUE #394: Calltaker: "arriveBy" queries in past calls should plan as arriveBy.
Original message:

> If I plan an arrive by trip, this does not get recorded properly; rather it is recorded as a depart by (and the resultant itinerary is the depart by as well, not arrive by).

binh-dam-ibigroup:
This pull request addresses the issues below:

fix #392, fix #393, 
fix #394.


lib/components/admin/query-record.js
```diff
@@ -13,14 +14,22 @@ import {CallRecordButton, CallRecordIcon} from './styled'
 class QueryRecordLayout extends Component {
   _getParams = () => {
     const {query} = this.props
-    return planParamsToQuery(JSON.parse(query.queryParams))
+    const originalQueryParams = JSON.parse(
+      query.queryParams,
+      // convert boolean for the 'arriveBy' field to strings
+      // per https://github.com/opentripplanner/otp-ui/blob/master/packages/core-utils/src/query.js#L285
+      // FIXME: move this elsewhere?
+      (key, value) => key === 'arriveBy'
```
landonreed, line 22: @binh-dam-ibigroup, I think this needs to become a fiield-trip util because it is also used here: https://github.com/opentripplanner/otp-react-redux/blob/master/lib/actions/field-trip.js#L217

lib/actions/call-taker.js
```diff
@@ -73,9 +74,10 @@ export function endCall () {
       {method: 'POST', body: callData}
     )
       .then(res => res.json())
-      .then(id => {
+      .then(async id => {
         // Inject call ID into active call and save queries.
-        dispatch(saveQueriesForCall({...activeCall, id}))
+        await dispatch(saveQueriesForCall({...activeCall, id}))
+        // Wait until query was saved before re-fetching queries for this call.
         dispatch(fetchCalls())
```
landonreed, line 81: I find that this doesn't work for me. Also, if we're introducing async/await, I would prefer that we don't nest that inside calls to `then`.
```javascript
    const id = await fetch(`${otp.config.datastoreUrl}/calltaker/call`,
      {method: 'POST', body: callData}
    )
      .then(res => res.json())
      .catch(err => {
        console.error(err)
        alert(`Could not save call: ${JSON.stringify(err)}`)
      })
    // Inject call ID into active call and save queries.
    await dispatch(saveQueriesForCall({...activeCall, id}))
    // Wait until query was saved before re-fetching queries for this call.
    await dispatch(fetchCalls())
```

landonreed (changes requested):
Mostly good, just needs a couple of tweaks (reusable parsing method and unnesting of async stuff).

lib/components/admin/query-record.js
binh-dam-ibigroup: Thanks for pointing that out, code updated in https://github.com/opentripplanner/otp-react-redux/pull/395/commits/a2cfb7341f39d940744a3c02d4c637193df6eb97.

lib/actions/call-taker.js
binh-dam-ibigroup: Refactored in https://github.com/opentripplanner/otp-react-redux/pull/395/commits/789205533d4a85ad0ec30dfe71c19b84f60cd4ae. @landonreed if after this refactor this is still not working for you, could you share the steps or could we have a walkthrough? Thanks.

landonreed (changes requested):
One jsdoc suggestion and one series of steps that I think still causing confusing results:
1. Open call history
2. View past call
3. Start new call
4. Plan trip
5. Stop call
6. Note that the expanded/visible call records show "No queries recorded."

To resolve, I think we should close all visible call records when `fetchCalls` is dispatched. Maybe we should agree on an approach before implementing.

lib/actions/call-taker.js
binh-dam-ibigroup: Hopefully this is fixed with https://github.com/opentripplanner/otp-react-redux/pull/395/commits/076454ba7c7d0ba9e43e3338e941e2bd6c396307 and the dev merge.

landonreed (approved):
Looks good! Thanks, @binh-dam-ibigroup.

evansiroky (approved):

PR #397, opened 2021-06-29
Plan first last

landonreed:
**Note: this depends on changes in https://github.com/opentripplanner/otp-ui/pull/258 (so that the `startTransitStopId` param will be applied).**

This PR:
- Adds buttons to plan first/previous/next/last trip to the `default-itinerary` for the Call Taker project.
- shows errors if there are no itineraries returned (see [this line](https://github.com/opentripplanner/otp-react-redux/pull/397/files#diff-70bb8f4ecdd96dd933a31deefe87c401dadfc35cfa8d40f82575a2a4ba08f1edR169))

This is enabled via the config.yml (`itinerary#showPlanFirstLastButtons`):
https://github.com/opentripplanner/otp-react-redux/blob/d0a9b8b8a6f6ef69677a6407565ef5aa6a7cfdd2/example-config.yml#L169-L176

image


lib/components/narrative/plan-first-last-buttons.js
```diff
@@ -0,0 +1,159 @@
...
+      startTransitStopId: getFirstStopId(itinerary),
+      date: moment(currentQuery.date).add('days', 1).format(OTP_API_DATE_FORMAT),
+      time: SERVICE_BREAK,
+      departArrive: 'ARRIVE'
+    }
+    this._updateParamsAndPlan(params)
+  }
+
+  render () {
+    const {enabled, itineraries} = this.props
```
evansiroky, line 99: There are only 2 props in this render method, yet 14 props provided via the `mapStateToProps` method. A more decluttered design would be to make each of these buttons directly call an action, especially since they don't require any user input.

evansiroky (changes requested):
See 1 blocking comment for now.

lib/components/narrative/plan-first-last-buttons.js
```diff
@@ -0,0 +1,48 @@
...
+      </Button>
+      <Button bsSize='small' onClick={planLast}>
+        Last
+      </Button>
+    </FlexButtonGroup>
+  )
+}
+
+const mapStateToProps = (state, ownProps) => {
+  const itineraries = getActiveItineraries(state.otp)
```
binh-dam-ibigroup, line 34: This should be after merging #386:
```suggestion
  const itineraries = getActiveItineraries(state)
```

lib/actions/api.js
```diff
@@ -321,14 +305,49 @@ export function getRoutingParams (query, config, ignoreRealtimeUpdates) {
...
+function paramIsDefined (query, key) {
+  return key in query && query[key] !== '' && query[key] !== null
+}
 
-  // hack to add walking to driving/TNC trips
-  if (hasCar(params.mode)) {
-    params.mode += ',WALK'
-  }
+/**
+ * A given parameter is included in the request if all of the following:
```
binh-dam-ibigroup, line 315:
```suggestion
 * A given parameter is included in the request if all of the following conditions are met:
```

binh-dam-ibigroup (changes requested):
Main changes needed are:
- It looks like #386 is included, so calls to `getActiveItinerary.ies` are broken and I cannot test the rest.
- A clarification of the definition of 'first/next/prev/last' trips is needed, right now these seem to be any possible trips between the start/end locations using any routing.
- Probably rename the pagination control.

See also my other comments.

lib/components/narrative/styled.js
```diff
@@ -0,0 +1,11 @@
+import {ButtonGroup} from 'react-bootstrap'
+import styled from 'styled-components'
+
+export const FlexButtonGroup = styled(ButtonGroup)`
+  display: flex;
+  flex: 1;
+  margin-top: 5px;
+  button {
+    flex: 1;
+  }
```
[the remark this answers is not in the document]
landonreed: Just tried and it doesn't appear to work out of the box, so I'll skip.

lib/actions/api.js
landonreed: Evan pointed out that getRoutingParams is already exported from otp-ui, so I'm removing this code in favor of that version of the method.

example-config.yml
```diff
@@ -193,6 +193,9 @@ itinerary:
   # Show fares for each transit leg (false if omitted).
   # (Requires using LineItinerary.)
   showRouteFares: false
+  # Whether the plan first/previous/next/last buttons should be shown along with
+  # plan trip itineraries.
+  showPlanFirstLastButtons: false
```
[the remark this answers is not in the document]
landonreed: I'm not fond of TripTimePagination. If @evansiroky has a good suggestion, we can go with that. Otherwise, I'm inclined to keep as is.

landonreed:
@binh-dam-ibigroup, until the `startTransitStopId` param is included in https://github.com/opentripplanner/otp-ui/pull/258, this PR won't be able to plan the first/last/etc. for the exact same trip.

binh-dam-ibigroup (approved):
Seems to work, so just remove the console statement. Also see my other suggestions.

lib/components/narrative/plan-first-last-buttons.js
```diff
@@ -0,0 +1,48 @@
...
+  if (!enabled || itineraries.length === 0) {
+    return null
+  }
+  return (
+    <FlexButtonGroup>
+      <Button bsSize='small' onClick={planFirst}>
+        First
+      </Button>
+      <Button bsSize='small' onClick={planPrevious}>
+        Previous
```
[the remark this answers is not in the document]
landonreed: These buttons were labeled Previous/Next in the previous version, so I'd prefer to keep them the same.

Issue #376, opened 2021-05-27
Clicking "Back" button on "Edit trip" page after logging in causes loop effect


landonreed:
When clicking the "Back" button after having just logged in, it will send me back to the URL from which I was just redirected.

miles-grant-ibigroup:
Have run into this problem again when trying to prevent navigation away from a dirty trip edit page. Using `onbeforeunload` is supposed to prevent navigation with the browser back and reload buttons. The reload button is blocked, but back is not intercepted. This to me suggests that react-router has already intercepted it and isn't working with it properly.

Resolving this issue will likely involve fixing behavior within the higher level component which is doing the routing 

https://github.com/opentripplanner/otp-react-redux/blob/f18fa7edc116b7804c35c273b4f117ec35824b96/lib/components/app/responsive-webapp.js#L137 

and including a method to set blocking routing, or removing an item from history. 

Would likely involve working on this function https://github.com/opentripplanner/otp-react-redux/blob/f18fa7edc116b7804c35c273b4f117ec35824b96/lib/actions/ui.js#L113-L151

PR #400, opened 2021-06-30
Trip page: Add warning about navigating away with unsaved changes

miles-grant-ibigroup:
When a user is adding or editing a trip, they will now be prompted to confirm they really want to leave if they try to navigate away. 

If they haven't made any changes or if they haven't made any changes, they aren't prompted.

### DIFF
--- a/lib/components/user/monitored-trip/trip-basics-pane.js
+++ b/lib/components/user/monitored-trip/trip-basics-pane.js
@@ -89,8 +89,24 @@ class TripBasicsPane extends Component {
   }
 
   render () {
-    const { errors, isCreating, itineraryExistence, values: monitoredTrip } = this.props
+    const {
+      canceled,
+      dirty,
+      errors,
+      isCreating,
+      isSubmitting,
+      itineraryExistence,
+      values: monitoredTrip
+    } = this.props
     const { itinerary } = monitoredTrip
+
+    // Prevent user from leaving when form has been changed,
+    // but don't show it when they click submit or cancel.
+    const unsavedChanges = dirty && !isSubmitting && !canceled
+    // Message changes depending on if the new or existing trip is being edited
+    const unsavedChangesMessage = `You have haven't saved your ${
+      isCreating ? 'new ' : ''
+    }trip yet. If you leave, ${isCreating ? 'it' : 'changes'} will be lost.`
 
     if (!itinerary) {
       return <div>No itinerary to display.</div>
@@ -109,11 +125,9 @@ class TripBasicsPane extends Component {
 
       return (
         <div>
-          {/* Prevent user from leaving when form has been changed, but
-          don't show it when they click submit or cancel. */}
           <Prompt
-            when={this.props.dirty && !this.props.isSubmitting && !this.props.canceled}
-            message={`You have haven't saved your ${this.props.isCreating ? 'new ' : ''}trip yet. If you leave, ${this.props.isCreating ? 'it' : 'changes'} will be lost.`}
+            when={unsavedChanges}
+            message={unsavedChangesMessage}
           />
 
           {/* Do not show trip status when saving trip for the first time
--- a/lib/components/user/monitored-trip/trip-basics-pane.js
+++ b/lib/components/user/monitored-trip/trip-basics-pane.js
@@ -109,10 +109,11 @@ class TripBasicsPane extends Component {
 
       return (
         <div>
-          {/* Prevent user from leaving when form is dirty */}
+          {/* Prevent user from leaving when form has been changed, but
+          don't show it when they click submit.  */}
           <Prompt
-            when={this.props.dirty}
-            message="You have haven't saved your new trip yet. If you leave, it will be lost."
+            when={this.props.dirty && !this.props.isSubmitting}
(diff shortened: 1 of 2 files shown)


lib/components/user/monitored-trip/trip-basics-pane.js
```diff
@@ -107,6 +109,13 @@ class TripBasicsPane extends Component {
 
       return (
         <div>
+          {/* Prevent user from leaving when form has been changed, but
+          don't show it when they click submit or cancel. */}
+          <Prompt
+            when={this.props.dirty && !this.props.isSubmitting && !this.props.canceled}
+            message={`You have haven't saved your ${this.props.isCreating ? 'new ' : ''}trip yet. If you leave, ${this.props.isCreating ? 'it' : 'changes'} will be lost.`}
```
evansiroky, line 116: Lines are too long. Please split up these long lines. A boolean variable could be created for the first one. The second one could also be calculated separately before returning this JSX. Also, please restructure these props at the very beginning of the render method.

evansiroky (changes requested):
See 1 comment.

lib/components/user/monitored-trip/trip-basics-pane.js
miles-grant-ibigroup: Thanks! I'm not entirely happy with the `unsavedChangesMessage` line break, but that's the only way I could think of that doesn't create a break in the popup.

lib/components/user/monitored-trip/trip-basics-pane.js
```diff
@@ -87,9 +89,25 @@ class TripBasicsPane extends Component {
...
+      itineraryExistence,
+      values: monitoredTrip
+    } = this.props
     const { itinerary } = monitoredTrip
 
+    // Prevent user from leaving when form has been changed,
+    // but don't show it when they click submit or cancel.
+    const unsavedChanges = dirty && !isSubmitting && !canceled
+    // Message changes depending on if the new or existing trip is being edited
+    const unsavedChangesMessage = `You have haven't saved your ${
```
landonreed, line 107:
```suggestion
    const unsavedChangesMessage = `You haven't saved your ${
```

lib/components/user/monitored-trip/trip-basics-pane.js
```diff
@@ -87,9 +89,25 @@ class TripBasicsPane extends Component {
...
+      isCreating ? 'new ' : ''
+    }trip yet. If you leave, ${isCreating ? 'it' : 'changes'} will be lost.`
```
landonreed, line 109: I might just generify this a bit to something like:
```suggestion
const unsavedChangesMessage = 'You have unsaved changes to this trip. If you leave before saving, your changes will be lost'
```

lib/components/user/monitored-trip/trip-basics-pane.js
landonreed, line 107:
```suggestion
    const unsavedChangesMessage = `You haven't saved your ${
```

binh-dam-ibigroup: For internationalization, we should stay away from these string concats. I recommend we just display one string for when you create a trip and one other string for when you modify a trip.

landonreed (approved):
This looks great. Just one tweak for a typo. Also, I noticed an issue when clicking the <- Back button, where it updated the URL even if I press "Cancel" in the prompt, but that may be more related to https://github.com/opentripplanner/otp-react-redux/issues/376

lib/components/user/stacked-pane-display.js
```diff
@@ -1,36 +1,40 @@
...
+          <StackedPaneContainer key={index}>
+            <h3>{title}</h3>
+            <div><Pane canceled={isBeingCanceled} {...props} /></div>
+          </StackedPaneContainer>
+        ))
+      }
+
+      <FormNavigationButtons
+        backButton={{
+          onClick: () => { updateBeingCanceled(true); onCancel() },
```
binh-dam-ibigroup, line 28: What formatting do @landonreed and @evansiroky prefer regarding the semicolon? Spread over multiple lines?

binh-dam-ibigroup (approved):
Conditional approval subject to using different prompt messages for new and existing trips.

lib/components/user/stacked-pane-display.js
miles-grant-ibigroup: This was created by prettier -- ambiguities like this should be resolved in the future once we get the new version of mastarm up and running (which should include auto formatting).

landonreed: Yea, I was stuck on this a bit also, @binh-dam-ibigroup. Let's just go ahead and move to multiple lines, since we tend not to include semicolons anywhere in this code base.

miles-grant-ibigroup:
@landonreed have played around with this and I definitely think it is to do with #376. I can register the `onbeforeunload` listener which is supposed to intercept browser navigation. Reload is intercepted, but back is ignored. This to me says that somewhere higher up something is doing something fishy with react-router? 

Should I try to address this in this pr or should we just expand #376?

landonreed:
@miles-grant-ibigroup, can you document your findings in #376? We can handle the back button stuff altogether in that issue (so we can continue with the merge for this one).

landonreed (approved):

Issue #414, opened 2021-07-12
Rental vehicle overlay still renders after unchecking in layer selector.


binh-dam-ibigroup:
See screenshot below.

![image](https://user-images.githubusercontent.com/56846598/125309816-0c40a380-e300-11eb-9568-a846cf23863c.png)

PR #419, opened 2021-07-19
Bump vehicle overlay/base map

landonreed:
Fixes:

- #414.
- A bug that arises when looking at trip planning issues and there is a company with an issue that does not have an icon defined
- A bug that prevented the proper rendering of transitive data with an itinerary with an eScooter leg


miles-grant-ibigroup (approved):
This looks good thanks for going into the weeds like this to solve these issues!

landonreed:
Latest changes have my approval! (I can't approve my own PR the standard way.)

binh-dam-ibigroup (approved):

PR #424, opened 2021-07-30
Field trip fixes July 30

landonreed:
This PR resolves a couple of field trip items:
- Fixes ibi-group/trimet-mod-otp#303
- Along with https://github.com/opentripplanner/otp-datastore/pull/7, fixes ibi-group/trimet-mod-otp#302

It also fixes a css issue with display of date time options in Firefox  by using flex (rendering is also improved with Chrome)
- Fixes ibi-group/trimet-mod-otp#304

### DIFF
--- a/lib/components/admin/field-trip-status-icon.js
+++ b/lib/components/admin/field-trip-status-icon.js
@@ -2,10 +2,10 @@ import React from 'react'
 
 import Icon from '../narrative/icon'
 
-const FieldTripStatusIcon = ({ status }) => {
-  return status
+const FieldTripStatusIcon = ({ status }) => (
+  status
     ? <Icon className='text-success' type='check' />
     : <Icon className='text-warning' type='exclamation-circle' />
-}
+)
 
 export default FieldTripStatusIcon
--- a/lib/components/admin/field-trip-status-icon.js
+++ b/lib/components/admin/field-trip-status-icon.js
@@ -1,0 +1,11 @@
+import React from 'react'
+
+import Icon from '../narrative/icon'
+
+const FieldTripStatusIcon = ({ status }) => {
+  return status
+    ? <Icon className='text-success' type='check' />
+    : <Icon className='text-warning' type='exclamation-circle' />
+}
+
+export default FieldTripStatusIcon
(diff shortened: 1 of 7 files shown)


lib/components/admin/field-trip-status-icon.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+
+import Icon from '../narrative/icon'
+
+const FieldTripStatusIcon = ({ status }) => {
```
binh-dam-ibigroup, line 5: The direct return `=> (` notation could be used here.

lib/components/admin/trip-status.js
```diff
@@ -92,7 +90,7 @@ class TripStatus extends Component {
     return (
       <Full>
         <Header>
-          {this._renderStatusIcon()}
+          <FieldTripStatusIcon status={this._tripIsPlanned()} />
```
binh-dam-ibigroup, line 93: The value passed to status is ambiguous here... are we passing a trip or a status? (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND#using_and)

lib/components/form/call-taker/date-time-options.js
```diff
@@ -190,40 +190,36 @@ export default class DateTimeOptions extends Component {
...
+            }}
+            value={timeInput || dateTime.format('H:mm')}
+          />
+        </OverlayTrigger>
         <input
           className='datetime-slim'
           onChange={this.handleDateChange}
           onKeyDown={onKeyDown}
           style={{
             border: 'none',
```
binh-dam-ibigroup, line 218: With the latest round of visual changes, could the date selector get a border too? (if so, make this control 2px wider for Firefox, and the `Plan` button needs a new height property)
image

binh-dam-ibigroup (approved):
Looks good with minor suggestions. Passing `_tripIsPlanned()` to the status prop seems ambiguous, maybe that needs to be clarified.

landonreed:
Thanks @binh-dam-ibigroup. Comments addressed in a99744fd52aec3870de904faa616f4b42a5d90c8

evansiroky (approved):

landonreed:
@binh-dam-ibigroup I changed it to `ok` in e8722a507e0ec9b959c571882c0adb7bc1d8d6be

PR #408, opened 2021-07-08
WIP: React intl ramp up

binh-dam-ibigroup:
This PR adds localization support to OTP-RR.

PR #426, opened 2021-08-05
Fix Accessibility and Test for Accessibility Issues

miles-grant-ibigroup:
In tandem with https://github.com/opentripplanner/otp-ui/pull/283 makes colors within OTP-RR conform with WCAG.

A few notes:

- the URL used to generate the journey plan is Houston specific, and may not render a proper trip response. What should we do here? How can we guarantee a trip option is rendered?
- The tests **will fail** until https://github.com/ibi-group/otp-ui/pull/8 is merged and published.


lib/components/app/app-frame.js
```diff
@@ -9,7 +9,7 @@ import NotFound from './not-found'
  * content and an optional sub-navigation component can be inserted.
  */
 const AppFrame = ({ children, SubNav }) => (
-  <div className='otp'>
+  <div className='otp' id='otp' role='main'>
```
binh-dam-ibigroup, line 12: Why does it need an `id`?

index.html
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
```
binh-dam-ibigroup, line 2: Add a FIXME for placing the correct language code when doing i18n. (I will also add it to the active i18n branch too.)

lib/components/viewers/realtime-status-label.js
```diff
@@ -34,11 +34,11 @@ const STATUS = {
...
-    color: '#d9534f',
+    color: '#D92923',
     label: 'late'
   },
   ON_TIME: {
-    color: '#5cb85c',
+    color: '#028602',
     label: 'on time'
   },
   SCHEDULED: {
```
binh-dam-ibigroup, line 44: What color are we using for scheduled departures? We may need to check that we are using the same as in https://github.com/ibi-group/otp-ui/pull/8/files for consistency.

binh-dam-ibigroup (approved):
I suppose there is more to uncover as we go but this looks like a good start.

lib/components/viewers/realtime-status-label.js
miles-grant-ibigroup: From what I can gather, it's inherited from otp-ui? Or am I misunderstanding what's going on here

lib/components/app/app-frame.js
miles-grant-ibigroup: The rule stated that the aria main element also needs an id to guarantee uniqueness

index.html
miles-grant-ibigroup: Good call. Added in 994a5fd0e1e5bb59571ea28c81b8f951e9f9a482

miles-grant-ibigroup:
Thank you for the feedback! All is resolved aside from a small question about the scheduled colors. From what I can tell based on looking at the inspector, if no color is set the color is inherited from the otp-ui css.

lib/components/app/app-frame.js
binh-dam-ibigroup: There is a `<main>` element rendered somewhere in the hierarchy already, is that the element that the rule is looking for?

lib/components/viewers/realtime-status-label.js
binh-dam-ibigroup: The color is `#bbb` and defined underneath and gets merged in the CSS by the `Container` component. Not a critical item, and I don't think the Lighthouse devtools plugin can catch that as is as these labels are not part of the initial render of stop viewers.

binh-dam-ibigroup:
> Thank you for the feedback! All is resolved aside from a small question about the scheduled colors. From what I can tell based on looking at the inspector, if no color is set the color is inherited from the otp-ui css.

Should be ok for now, so you can pass it to the next reviewer.

lib/components/viewers/realtime-status-label.js
miles-grant-ibigroup: If `withBackground` is false, isn't it inherited? Otherwise yes I agree. May have to do this on a project by project basis. For what it's worth, Firefox dev tools can do color contrast checking even on subsequent renders!

lib/components/app/app-frame.js
miles-grant-ibigroup: I didn't know that. But I think it would have to be set as `<main role='main' id='main'>`. I think the rules are designed to ensure things work on as many screen readers as possible. I guess this is a question for @evansiroky should we move the `role='main'` to the `<main>` element?

miles-grant-ibigroup:
Assigning this back to myself after today's discussion, will add the puppeteer fixes first.

miles-grant-ibigroup:
As said in the PR description, tests currently fail (because otp-ui isn't updated yet). However, if I manually update the dependencies, the new tests pass. Assigning to Evan for second review.

package.json
```diff
@@ -3,9 +3,10 @@
   "description": "A library for writing modern OpenTripPlanner-compatible multimodal journey planning web applications using React and Redux",
   "main": "build/index.js",
   "scripts": {
+    "a11y-test": "mastarm test a11y --force-exit",
     "build": "mastarm build --env production",
     "cover": "mastarm test -e test --coverage",
-    "jest": "mastarm test -e test",
+    "jest": "mastarm test -e test __tests__",
     "lint": "mastarm lint lib __tests__ --quiet",
```
evansiroky, line 10: Please also lint the `a11y` directory.

lib/components/narrative/narrative.css
```diff
@@ -195,7 +195,7 @@
 
 .otp .tabbed-itineraries .tab-button .details {
   font-size: 12px;
-  color: gray;
+  color: #685c5c; /* from "gray" to meet WCAG */
```
evansiroky, line 198: We probably don't need a comment about the previous value here.

lib/components/admin/styled.js
```diff
@@ -145,7 +145,7 @@ export const Text = styled.span`
 
 export const Val = styled.span`
   :empty:before {
-    color: grey;
+    color: #685C5C; /* from "gray" to meet WCAG */
```
evansiroky, line 148: We probably don't need a comment about the previous value here.

index.html
evansiroky: I'd be interested to see how this applies to a website that can change languages, but I guess it's good enough to have the fixme there. Maybe just add a note that the `lang` attribute is there to make a11y happy.

a11y/test-config.yml
```diff
@@ -0,0 +1,168 @@
...
+  - id: 405
+    msg: Sorry, we couldn't find any transit or rideshare/carshare options at the time and/or location you chose. Please try again later, or change the settings of your trip.
+    modes:
+      - CAR_HAIL
+      - CAR_RENT
+  - id: 480
+    msg: No available transit routes or rideshare/carshare service at origin.
+
+reportIssue:
+  mailto: webmaster@ridemetro.org
```
evansiroky, line 156: Change to something not actually valid.

a11y/test-config.yml
```diff
@@ -0,0 +1,168 @@
...
+      minLon: -97.4197
+      maxLon: -93.2674
+      minLat: 28.6093
+      maxLat: 30.9685
+  # fetch options
+  options:
+    headers:
+      # AWS API Gateway key specific to Houston (IBI AWS account):
+      # https://console.aws.amazon.com/apigateway/home?region=us-east-1#/api-keys/fchsdgyys9
+      x-api-key: lvJXqETokn3T338y1c8U843Fjh1ONdgzK3Ar0pl4
```
evansiroky, line 94: Can these API gateway keys be populated from somewhere else and not actually live in this file?

a11y/test-config.yml
```diff
@@ -0,0 +1,168 @@
...
+
+  # Default map center
+  initLat: 29.7604
+  initLon: -95.3698
+  initZoom: 12
+
+  # Base layers for default map
+  baseLayers:
+    - name: Streets
+      url: //api.mapbox.com/styles/v1/mapbox/outdoors-v11/tiles/{z}/{x}/{y}{retina}?access_token=pk.eyJ1IjoiaWJpLXRyYW5zaXQtZGF0YS10ZWFtIiwiYSI6ImNqeTY3Zms2bzBkbjMzbGs5bWpoemRxY3UifQ.IrkMhr7JMhUNcb2qXuzwig
```
evansiroky, line 37: This is a special token not tied to a domain that we control! Always check before committing any tokens like this! I have deleted this token now that it is out in the open.

a11y/test-config.yml
```diff
@@ -0,0 +1,168 @@
...
+    modes:
+      - CAR_HAIL
+      - CAR_RENT
+  - id: 480
+    msg: No available transit routes or rideshare/carshare service at origin.
+
+reportIssue:
+  mailto: webmaster@ridemetro.org
+
+# NOTE: Houston GA for production is set in the index.html file because it
```
evansiroky, line 158: Remove note about Houston.

a11y/a11y.test.js
```diff
@@ -0,0 +1,67 @@
...
+      OTP_RR_CONFIG_FILE_PATH,
+      OTP_RR_CONFIG_BACKUP_PATH
+    )
+    console.log('Backed up current OTP-RR config file')
+  }
+  // copy over test config file
+  fs.copyFileSync(
+    OTP_RR_TEST_CONFIG_PATH,
+    OTP_RR_CONFIG_FILE_PATH
+  )
```
evansiroky, line 26: This bit of code should be the place where tokens and live endpoints should be inserted from environment variables.

a11y/a11y.test.js
```diff
@@ -0,0 +1,67 @@
...
+  console.log('Copied a11y test config file')
+
+  // Build OTP-RR main.js using new config file
+  execa.sync('yarn', ['build'])
+  console.log('Built OTP-RR')
+
+  // Launch mock OTP server
+  const MOCK_SERVER_PORT = 9999
```
evansiroky, line 34: This port won't work with the ports allowed with our localhost Mapbox token.

a11y/a11y.test.js
```diff
@@ -0,0 +1,67 @@
...
+
+afterEach(() => {
+  fs.unlinkSync(OTP_RR_CONFIG_FILE_PATH)
+  if (fs.existsSync(OTP_RR_CONFIG_BACKUP_PATH)) {
+    fs.renameSync(
+      path.resolve(OTP_RR_CONFIG_BACKUP_PATH),
+      path.resolve(OTP_RR_CONFIG_FILE_PATH)
+    )
+  }
+  console.log('Restored original OTP-RR config file')
```
evansiroky, line 48: Should the mockServer be shut down too here?

evansiroky (changes requested):
Needs an update so secrets and live endpoints are obtained from environment variables and not from plain-text files.

package.json
miles-grant-ibigroup: Added in 6188d3ac981359220491cc83995fe7f351d9efb2

lib/components/narrative/narrative.css
miles-grant-ibigroup: Removed in 6188d3ac981359220491cc83995fe7f351d9efb2

lib/components/admin/styled.js
miles-grant-ibigroup: Removed in 6188d3ac981359220491cc83995fe7f351d9efb2

a11y/test-config.yml
miles-grant-ibigroup: None of this is actually needed. Removed in b9200a6d143e74ae12e63673d6acc8a44ac88b08. Apologies.

a11y/test-config.yml
miles-grant-ibigroup: None of these keys are needed. Removed in b9200a6d143e74ae12e63673d6acc8a44ac88b08. Apologies.

a11y/test-config.yml
miles-grant-ibigroup: None of these keys are needed. Removed in b9200a6d143e74ae12e63673d6acc8a44ac88b08. Apologies.

a11y/test-config.yml
miles-grant-ibigroup: None of these keys are needed. Removed in b9200a6d143e74ae12e63673d6acc8a44ac88b08. Apologies.

a11y/a11y.test.js
miles-grant-ibigroup: The mock server removes the need for any of these keys. Sorry again for leaking them. Will keep my guard up more next time.

a11y/a11y.test.js
miles-grant-ibigroup: Leaflet fails a11y tests and can't really be fixed from within this repo (the copyright at the bottom) so it should be alright?

a11y/a11y.test.js
miles-grant-ibigroup: Have added a server shut down here but am still had to keep the `--force-exit` to get jest to exit properly.

miles-grant-ibigroup:
Have addressed these pressing items. Sorry again about writing that config to this repository.

evansiroky (changes requested):
Looks good, just needs those updated otp-ui versions.

index.html
binh-dam-ibigroup: FYI I have added the code to update the `lang` attribute in https://github.com/opentripplanner/otp-react-redux/pull/408/commits/01797599815ddde031061e9908abd267156c020b in the #408 branch.

landonreed:
Assigned to @miles-grant-ibigroup for integrating new otp-ui releases.

evansiroky (approved):

PR #434, opened 2021-08-18
Further a11y fixes

miles-grant-ibigroup:
The first round of a11y fixes failed to test trimet-mod-otp components or the stop viewer. This PR makes the remaining changes, and includes tests for the stop viewer. trimet-mod-otp isn't tested, as this would require adding the same puppeteer tests to that repo.


a11y/a11y.test.js
```diff
@@ -54,18 +54,30 @@ afterEach(async () => {
...
-    ]
+    disabledRules: disabledRules
   })
+  // Test stop viewer
+  page = await browser.newPage()
+  await page.goto(`file://${path.resolve(__dirname, '../index-for-puppeteer.html')}#/stop/exampleStop?ui_activeSearch=u9dwdhmyo&ui_activeItinerary=2&fromPlace=945 Columbia Street%2C Houston%2C TX%2C USA%3A%3A29.78881282532108%2C-95.3932571411133&toPlace=Hardy Street Yard%2C Houston%2C TX%2C USA%3A%3A29.772125846370574%2C-95.3551483154297&date=2021-08-18&time=17%3A07&arriveBy=false&mode=WALK%2CBUS%2CTRAM&showIntermediateStops=true&maxWalkDistance=1207&optimize=QUICK&walkSpeed=1.34&ignoreRealtimeUpdates=true&numItineraries=3&otherThanPreferredRoutesPenalty=900`)
+  await page.waitForTimeout(4000)
+  await page.click('.expansion-button')
+  await expect(page).toPassAxeTests({
+    disabledRules: disabledRules
```
landonreed, line 79: Here and on line 71, we could just use the shorthand `{ disabledRules }`. We could also just store that in a variable `const axeOptions = { disabledRules }` for reuse.

a11y/a11y.test.js
```diff
@@ -54,18 +54,30 @@ afterEach(async () => {
...
-    disabledRules: [
-      'region', // Leaflet does not comply
-      'meta-viewport', // Leaflet does not comply
-      'page-has-heading-one' // Heading is provided by logo
-    ]
+    disabledRules: disabledRules
   })
+  // Test stop viewer
+  page = await browser.newPage()
+  await page.goto(`file://${path.resolve(__dirname, '../index-for-puppeteer.html')}#/stop/exampleStop?ui_activeSearch=u9dwdhmyo&ui_activeItinerary=2&fromPlace=945 Columbia Street%2C Houston%2C TX%2C USA%3A%3A29.78881282532108%2C-95.3932571411133&toPlace=Hardy Street Yard%2C Houston%2C TX%2C USA%3A%3A29.772125846370574%2C-95.3551483154297&date=2021-08-18&time=17%3A07&arriveBy=false&mode=WALK%2CBUS%2CTRAM&showIntermediateStops=true&maxWalkDistance=1207&optimize=QUICK&walkSpeed=1.34&ignoreRealtimeUpdates=true&numItineraries=3&otherThanPreferredRoutesPenalty=900`)
```
landonreed, line 75: I don't think we need to dirty this URL with the query params. Otherwise, folks might think that full URL is necessary for the test to pass.
```suggestion
  await page.goto(`file://${path.resolve(__dirname, '../index-for-puppeteer.html')}#/stop/Agency:12345`)
```

a11y/a11y.test.js
```diff
@@ -54,18 +54,30 @@ afterEach(async () => {
...
   await expect(page).toPassAxeTests({
-    disabledRules: [
-      'region', // Leaflet does not comply
-      'meta-viewport', // Leaflet does not comply
-      'page-has-heading-one' // Heading is provided by logo
-    ]
+    disabledRules: disabledRules
   })
+  // Test stop viewer
+  page = await browser.newPage()
```
landonreed, line 74: I would feel a little better if we weren't reusing this variable. Rec: use two variables:
```javascript
const planTripPage = ...
await planTripPage.goto(....)
const stopViewerPage = ...
await stopViewerPage.goto(....)
```

landonreed (approved):
This looks good, just a couple of nits. Also, while we're at it, does it make sense to test all of the routes? Also, it seems like we would want `yarn a11y-test` to run as part of the `yarn test` script.

a11y/a11y.test.js
miles-grant-ibigroup: Resolved in https://github.com/opentripplanner/otp-react-redux/pull/436

a11y/a11y.test.js
miles-grant-ibigroup: Thank you for resolving in https://github.com/opentripplanner/otp-react-redux/pull/436

a11y/a11y.test.js
miles-grant-ibigroup: Thank you for refactoring in https://github.com/opentripplanner/otp-react-redux/pull/436

miles-grant-ibigroup:
Got things working in https://github.com/opentripplanner/otp-react-redux/pull/436 and left a comment. `a11y-test` was split out of the main test command because of how much slower it is. There is a new step in the ci which runs it though!

lib/components/viewers/pattern-row.js
```diff
@@ -72,7 +72,7 @@ export default class PatternRow extends Component {
 
           {/* expansion chevron button */}
           <div className='expansion-button-container'>
-            <button className='expansion-button' onClick={this._toggleExpandedView}>
+            <button aria-label='View all departures' className='expansion-button' onClick={this._toggleExpandedView}>
```
binh-dam-ibigroup, line 75: `aria-label` should rather say "Expand|Collapse all departures for route xxx" depending on `this.state.expanded` so that users of screen readers can know what that state is and what route they are touching.

lib/components/viewers/route-viewer.js
```diff
@@ -230,7 +230,7 @@ class RouteRow extends PureComponent {
           {isActive && (
             <RouteDetails>
               {route.url
-                ? <a href={route.url} target='_blank'>Route Details</a>
+                ? <a className='routeDetails' href={route.url} target='_blank'>Route Details</a>
```
binh-dam-ibigroup, line 233: Should that color be the default for all links?

lib/components/viewers/viewers.css
```diff
@@ -26,6 +26,11 @@
   animation-name: yellowfade;
 }
 
+/* Route Details Link a11y compatibility */
+a.routeDetails {
```
binh-dam-ibigroup, line 30: Should that apply to all `a` links?

lib/components/viewers/pattern-row.js
binh-dam-ibigroup: The `aria-expanded` and `aria-controls` attribute should be used too, see https://www.w3.org/WAI/GL/wiki/Using_the_WAI-ARIA_aria-expanded_state_to_mark_expandable_and_collapsible_regions. (`aria-controls` is the id of the element which will be shown/hidden by actuating the button.)

binh-dam-ibigroup (changes requested):
I would recommend refining the a11y attributes related to the `PatternRow` expand/collapse button. Also should the `routeDetails` color applied to all links?

lib/components/viewers/pattern-row.js
binh-dam-ibigroup: And maybe a `role` attribute can be set to the `VelocityTransitionGroup` at L82...

a11y/a11y.test.js
```diff
@@ -49,23 +75,38 @@ afterEach(async () => {
...
+  console.log('Closed mock server and headless browser')
 })
 
-test('checks the test page with Axe', async () => {
-  jest.setTimeout(600000)
-  // Web security is disabled to allow requests to the mock OTP server
-  const browser = await puppeteer.launch({args: ['--disable-web-security']})
-  const page = await browser.newPage()
-  await page.goto(`file://${path.resolve(__dirname, '../index-for-puppeteer.html')}#/?ui_activeSearch=0qoydlnut&ui_activeItinerary=0&fromPlace=1900%20Main%20Street%2C%20Houston%2C%20TX%2C%20USA%3A%3A29.750144%2C-95.370998&toPlace=800%20Congress%2C%20Houston%2C%20TX%2C%20USA%3A%3A29.76263%2C-95.362178&date=2021-08-04&time=08%3A14&arriveBy=false&mode=WALK%2CBUS%2CTRAM&showIntermediateStops=true&maxWalkDistance=1207&optimize=QUICK&walkSpeed=1.34&ignoreRealtimeUpdates=true&numItineraries=3&otherThanPreferredRoutesPenalty=900`)
+jest.setTimeout(600000)
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Added in bc710192935bea3619dca1522de31788edee0a49

lib/components/app/responsive-webapp.js
```diff
@@ -64,6 +64,7 @@ class ResponsiveWebapp extends Component {
 
   /** Lifecycle methods **/
 
+  /* eslint-disable-next-line complexity */
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Sadly yes...

lib/components/viewers/pattern-row.js
miles-grant-ibigroup: Have expanded `aria-label`, added `aria-expanded` and `aria-controls` in bc710192935bea3619dca1522de31788edee0a49, but am not sure what role you're thinking of for the `VelocityTransitionGroup`

lib/components/viewers/route-viewer.js
miles-grant-ibigroup: These links only have to be darker because of the non-white background they're on.

lib/components/viewers/viewers.css
miles-grant-ibigroup: These links only have to be darker because of the non-white background they're on.

miles-grant-ibigroup:
Thanks for the suggestions Binh -- all are addressed aside from one question about `role`.

lib/components/viewers/pattern-row.js
miles-grant-ibigroup: aria attributes added in bc710192935bea3619dca1522de31788edee0a49, roles added in 34da5db712af0cddc7242c81e64b00f04a752aeb

lib/components/viewers/pattern-row.js
```diff
@@ -72,7 +72,15 @@ export default class PatternRow extends Component {
 
           {/* expansion chevron button */}
           <div className='expansion-button-container'>
-            <button aria-label='View all departures' className='expansion-button' onClick={this._toggleExpandedView}>
+            <button
+              aria-controls={`route-${routeName}`}
+              aria-expanded={this.state.expanded}
+              aria-label={`${
+                this.state.expanded ? 'Collapse' : 'Expand'
+              } all departures for ${routeName}`}
```
binh-dam-ibigroup, line 80:
```suggestion
              } all departures for Route ${routeName}`}
```

binh-dam-ibigroup (approved):
Maybe one little tweak but otherwise good to go for me.

lib/components/viewers/pattern-row.js
miles-grant-ibigroup: Resolved in 2a60d39fcbc6689ff14a447a1046bb44558d98e4

miles-grant-ibigroup:
Thanks Binh! Language is changed. Assigning to @landonreed for merge.

PR #436, opened 2021-08-18
refactor: move routes to array for testing

landonreed:
This PR has changes for #434. It breaks out all routes in the app into an array so that we can TEST ALL THE THINGS! 

But, the tests are all broken! So assigning to @miles-grant-ibigroup to see if he can pick this up and fix things. Might also be worth creating a parameterized framework.


miles-grant-ibigroup:
Had to wrangle jest a little bit but now have this working. Most routes test nicely even without mock data, but I (for now) disabled all account paths from being tested -- the amount of calls we'd have to mock would take a long time to set up. However, it is easy to reverse, making future expansion here easy.

Also ran into a weird codespell bug -- the word ignore list [does NOT work](https://github.com/opentripplanner/otp-react-redux/runs/3369336916?check_suite_focus=true). All it does is make the false positive error twice (once in the file, once in the config file)

landonreed:
Amazing stuff! @binh-dam-ibigroup, could you take a look at these new tests?

binh-dam-ibigroup:
@landonreed did you mean to assign me to #434?

landonreed:
Sure, I'll just merge this one into #434 and assign you there.

PR #444, opened 2021-08-31
Use apiKey if configured for certain requests

### LINKED ISSUE #442: OTP-RR should forward x-api-key to OTP endpoints
From @landonreed:
Some OTP-middleware implementations require that OTP proxied endpoints be protected via the `x-api-key` header that we pass with our plan trip API requests. 
We need otp-react-redux to pass the headers from the `getOtpFetchOptions` method to the fetch that happens within createQueryAction: https://github.com/opentripplanner/otp-react-redux/blob/dev/lib/actions/api.js#L755

evansiroky:
Fixes #442.

### DIFF
--- a/lib/actions/api.js
+++ b/lib/actions/api.js
@@ -723,18 +723,7 @@ function createQueryAction (endpoint, responseAction, errorAction, options = {})
   return async function (dispatch, getState) {
     const state = getState()
     const { config } = state.otp
-    let url
-    if (
-      options.serviceId &&
-        config.alternateTransitIndex &&
-        config.alternateTransitIndex.services.includes(options.serviceId)
-    ) {
-      console.log('Using alt service for ' + options.serviceId)
-      url = config.alternateTransitIndex.apiRoot + endpoint
-    } else {
-      const api = config.api
-      url = `${api.host}${api.port ? ':' + api.port : ''}${api.path}/${endpoint}`
-    }
+    const url = makeApiUrl()
 
     if (!options.noThrottle) {
       // don't make a request to a URL that has already seen the same request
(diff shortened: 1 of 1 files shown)


lib/actions/api.js
```diff
@@ -775,6 +765,54 @@ function createQueryAction (endpoint, responseAction, errorAction, options = {})
...
+  const apiKey = config?.persistence?.otp_middleware?.apiKey
+  if (apiKey) {
+    const headers = fetchOptions.headers || {}
+    fetchOptions.headers = {
+      'x-api-key': apiKey,
+      ...headers
+    }
+    if (!fetchOptions.mode) {
+      fetchOptions.mode = 'cors'
+    }
```
landonreed, line 811: There's a bit of shared code here with `getSecureFetchOptions`. Did you consider any reuse? Given that they inject a lot of the same headers, I think that might be worthwhile.

landonreed (commented):
Looks good. I pushed 527c6dc9422326f6cabfc70bf018e78dd65cb1d3 to fix a minor omission, but have a question about reusing some code from `getSecureFetchOptions`.

lib/actions/api.js
evansiroky: I did consider reuse, but only the apiKey and cors thing are needed, so I figured it wasn't worth the effort to add extra configuration arguments to that method.

landonreed: See #445 for some proposed changes.

landonreed (commented):
@evansiroky, I just made some changes in #445. Let me know what you think (feel free to merge here and re-assign this PR if they look OK to you).

evansiroky:
Hmm, #445 this looks a bit better.

landonreed (approved):

landonreed:
@binh-dam-ibigroup, once you approve, could you merge and create a release PR?

binh-dam-ibigroup (approved):

PR #449, opened 2021-09-08
Batch itinerary cost fix

evansiroky:
This PR fixes a potential regression introduced in https://github.com/opentripplanner/otp-react-redux/pull/375.

It reverts to using the calculations in `lib/util/state#getTotalFare` for displaying the fare in the `lib/components/narrative/default/default-itinerary` component. Also, it refactors `getTotalFare` to not take in the whole config object.

Additionally, there is a separate bug fix to `lib/components/narrative/narrative-itineraries-header` to make sure the correct value is passed to the select element that displays the sort type.


landonreed (approved):
LGTM, great work, @evansiroky!

miles-grant-ibigroup (approved):
This looks great!

PR #450, opened 2021-09-08
fix: remove dead links

miles-grant-ibigroup:
Removes help link that doesn't contain content and makes terms of storage more info link hidden by default unless explicitly enabled in the config.

### DIFF
--- a/lib/components/user/terms-of-use-pane.js
+++ b/lib/components/user/terms-of-use-pane.js
@@ -54,13 +54,9 @@ const TermsOfUsePane = ({
     </div>
   )
 }
-const mapStateToProps = (state, ownProps) => {
+const mapStateToProps = (state) => {
   return {
     termsOfStorageSet: state.otp.config.persistence?.terms_of_storage
   }
 }
-
-const mapDispatchToProps = {
-}
-
-export default connect(mapStateToProps, mapDispatchToProps)(TermsOfUsePane)
+export default connect(mapStateToProps)(TermsOfUsePane)
--- a/lib/components/user/terms-of-use-pane.js
+++ b/lib/components/user/terms-of-use-pane.js
@@ -11,15 +11,13 @@ const TermsOfUsePane = ({
   disableCheckTerms,
   handleBlur,
   handleChange,
-  otpConfig,
+  termsOfStorageSet,
   values: userData
 }) => {
   const {
     hasConsentedToTerms,
     storeTripHistory
   } = userData
-  const { persistence } = otpConfig
-  const { terms_of_storage: termsOfStorageSet } = persistence
 
   return (
     <div>
@@ -58,7 +56,7 @@ const TermsOfUsePane = ({
 }
 const mapStateToProps = (state, ownProps) => {
   return {
-    otpConfig: state.otp.config
+    termsOfStorageSet: state.otp.config.persistence?.terms_of_storage
   }
 }
 
--- a/lib/components/user/terms-of-use-pane.js
+++ b/lib/components/user/terms-of-use-pane.js
@@ -1,4 +1,5 @@
 import React from 'react'
+import { connect } from 'react-redux'
 import { Checkbox, ControlLabel, FormGroup } from 'react-bootstrap'
 
 import { TERMS_OF_SERVICE_PATH, TERMS_OF_STORAGE_PATH } from '../../util/constants'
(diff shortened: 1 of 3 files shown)


lib/components/user/terms-of-use-pane.js
```diff
@@ -46,11 +50,19 @@ const TermsOfUsePane = ({
...
-            improve transit services in my area. <a href={`/#${TERMS_OF_STORAGE_PATH}`} target='_blank'>More info...</a>
+            improve transit services in my area. {termsOfStorageSet && <a href={`/#${TERMS_OF_STORAGE_PATH}`} target='_blank'>More info...</a>}
         </Checkbox>
       </FormGroup>
     </div>
   )
 }
+const mapStateToProps = (state, ownProps) => {
+  return {
+    otpConfig: state.otp.config
```
landonreed, line 61: Our general practice is to only supply the specific field to a component through mapStateToProps (rather than the whole config object).

landonreed (approved):
LGTM! Just one nit.

lib/components/user/terms-of-use-pane.js
miles-grant-ibigroup: As I was writing it I knew something felt off... Thanks for catching that! Refactored in 930a4a7

miles-grant-ibigroup:
Thanks @landonreed ! Assigning to @binh-dam-ibigroup for final review of this small PR

lib/components/user/terms-of-use-pane.js
```diff
@@ -46,11 +48,19 @@ const TermsOfUsePane = ({
...
   )
 }
+const mapStateToProps = (state, ownProps) => {
+  return {
+    termsOfStorageSet: state.otp.config.persistence?.terms_of_storage
+  }
+}
+
+const mapDispatchToProps = {
+}
```
binh-dam-ibigroup, line 64: Remove this var and the second argument of the `connect` function which is optional.

lib/components/user/terms-of-use-pane.js
```diff
@@ -46,11 +48,19 @@ const TermsOfUsePane = ({
...
-            improve transit services in my area. <a href={`/#${TERMS_OF_STORAGE_PATH}`} target='_blank'>More info...</a>
+            improve transit services in my area. {termsOfStorageSet && <a href={`/#${TERMS_OF_STORAGE_PATH}`} target='_blank'>More info...</a>}
         </Checkbox>
       </FormGroup>
     </div>
   )
 }
+const mapStateToProps = (state, ownProps) => {
+  return {
+    termsOfStorageSet: state.otp.config.persistence?.terms_of_storage
```
binh-dam-ibigroup, line 59: Add a comment in `example-config.yml` to show this parameter works.

binh-dam-ibigroup: Do we need a similar config variable for the terms of use?

binh-dam-ibigroup (changes requested):
Please update `example-config.yml` with the new config parameter, plus a minor refactoring.

lib/components/user/terms-of-use-pane.js
miles-grant-ibigroup: Didn't know that thanks! Refactored in 62278865204a45e1afa33483a5742f794c7f5b5e

lib/components/user/terms-of-use-pane.js
miles-grant-ibigroup: Have added an example in 62278865204a45e1afa33483a5742f794c7f5b5e. I don't think a similar config item for terms of use makes sense since otherwise the terms of use checkbox would not refer to any terms of use to accept!

miles-grant-ibigroup:
Thanks Binh! All changes addressed

lib/components/user/terms-of-use-pane.js
binh-dam-ibigroup: Also, there is an existing component in `ComponentContext` in which the contents for the optional Terms of Storage is defined. Maybe you can use it in this PR to check for the existence of terms of storage instead of a new config property, or combine that in another PR. Thoughts?

miles-grant-ibigroup: I gave this some thought, having an attribute on the "dummy" components shipped within otp-rr, and then if that attribute is missing showing the link. However, I found that all our configurations ship with the dummy components anyway, so that wouldn't work unless we changed all the configurations, which I wanted to avoid.

binh-dam-ibigroup: Can't we just check for `components.TermsOfStorage !== null`?

binh-dam-ibigroup: Given the rush I'm inclined to accept the approach, with issue #452 to deduplicate configuration.

binh-dam-ibigroup (approved):
Accepted subject to resolving #452 in some near future.

PR #443, opened 2021-08-27
Traveler tools menu

robertgregg3:
Blocked by #435 which provides internationalization for the app menu.

Refactored the current menu located in app-menu.js into it's own component with the following uses:

- Allow sub menus
- Allow sub menus within submenus
- optional icons from url or types
- hyperlink option

## Points to note:
- The code uses the same setup as the patternRow.js expandable mechanics.
- The menu is currently set up to go 2 sub menus deep: Top level > sub menu > sub menu
- The additional menu items will sit below the standard menu items
- The associated PR for the config repo [can be found here](https://github.com/ibi-group/configurations/pull/177)

## WIP notes:
- I added the css to the index.css file which I can put into it's own file. 
- The hyperlinks for ATL RIDES are not in place and need to be added
- The icons are just a first draft, mostly hyperlinks. Would need a review for sure
- Any other improvements or things I have done wrong in the code etc I would be happy to hear

PR #456, opened 2021-09-17
Add Accessibility Labels to Legs and Itineraries 

miles-grant-ibigroup:
This PR depends on https://github.com/opentripplanner/otp-ui/pull/289 to render the labels. Until these updates are released, imports (and therefore tests) will fail.

Adds labels to legs if they are provided by OTP, and generates a combined score for the itinerary 
Screen Shot 2021-09-17 at 3 09 12 PM


example-config.yml
```diff
@@ -269,6 +269,14 @@ itinerary:
 #     defaultRouteTextColor: '000000'
 #     longNameSplitter: ' - '
 #     order: 2
+# Use this config to overwrite the  accessibility score gradation
+# map that ships with otp-ui
+#accessibilityScore:
+#    gradationMap:
+#        0.0:
+#            color: "#ffb5b9"
+#            text: 'Not Accessible'
```
binh-dam-ibigroup, line 278: This should also support message ids for i18n. Message ids for each locale can be defined in this file (see what was done in #443 for the custom app menu entries).

example-config.yml
```diff
@@ -269,6 +269,14 @@ itinerary:
 #     defaultRouteTextColor: '000000'
 #     longNameSplitter: ' - '
 #     order: 2
+# Use this config to overwrite the  accessibility score gradation
```
binh-dam-ibigroup, line 272:
```suggestion
# Use this config to overwrite the accessibility score gradation
```

lib/components/narrative/default/default-itinerary.js
```diff
@@ -11,6 +12,11 @@ import SimpleRealtimeAnnotation from '../simple-realtime-annotation'
 import FormattedDuration from '../../util/formatted-duration'
 import FormattedTimeRange from '../../util/formatted-time-range'
 import { getTotalFare } from '../../../util/state'
+import {
+  itineraryHasAccessibilityScores,
+  getAccessibilityScoreForItinerary
```
binh-dam-ibigroup, line 17: Sort imports.

binh-dam-ibigroup (changes requested):
The accessibility label texts need to support i18n.

example-config.yml
miles-grant-ibigroup: Patched in 83e10d7

lib/components/narrative/default/default-itinerary.js
miles-grant-ibigroup: Sorted in 723e386

example-config.yml
miles-grant-ibigroup: Resolved in bc19c7b

lib/components/narrative/default/default-itinerary.js
```diff
@@ -194,53 +203,86 @@ class DefaultItinerary extends NarrativeItinerary {
...
+  // Generate icons based on fa icon keys in config
+  // Override text fields if translation set
+  Object.keys(gradationMap).forEach(key => {
+    const {icon} = gradationMap[key]
+    if (icon && typeof icon === 'string') {
+      gradationMap[key].icon = <Icon type={icon} />
+    }
+
+    const localizationId = `config.acessibilityScore.gradationMap.${key}`
+    const localizedText = intl.formatMessage({id: localizationId})
```
binh-dam-ibigroup, line 277: Add a comment to remind that message ids must be dynamically generated here because they are found in config (and not in the standard language files).

binh-dam-ibigroup (commented):
One more minor comment for now.

lib/components/narrative/default/default-itinerary.js
miles-grant-ibigroup: Comment added in 1aefcd05!

lib/components/util/accessibility-routing.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * Determine if an itinerary has accessibility scores
+ */
+export const itineraryHasAccessibilityScores = (itinerary) => {
```
philip-cline, line 4: For the i18n work, I was organising util functions into `lib/util`, and utility components like FormattedDuration into `lib/components/util`. Should these be moved to lib/util?

philip-cline (approved):
Looks great to me, just one small comment about file structure.

lib/components/util/accessibility-routing.js
binh-dam-ibigroup: Good point, if the function is used from multiple files then it should go to `lib/util`, if not, it should become private to this file.

miles-grant-ibigroup: Sorry I was merging too many things this morning and accidentally forgot about this! Very sorry. The fix is in #487

PR #465, opened 2021-10-05
Fix stuff that broke in trimet-mod-otp

binh-dam-ibigroup:
This PR fixes things from OTP-RR 3.5.0 that prevented from executing in trimet-mod-otp.


philip-cline (approved):

lib/components/app/default-main-panel.js
```diff
@@ -26,6 +26,7 @@ class DefaultMainPanel extends Component {
         <div style={{
           bottom: showPlanTripButton ? 55 : 0,
           left: 0,
+          overflowY: 'auto',
```
binh-dam-ibigroup, line 29: There is loss of scrolling ability on the itinerary results without this, so you only see the beginning of long itineraries. Not sure where the loss comes from.

PR #479, opened 2021-10-26
Make Requests to Either OTP 1 or OTP 2 Depending on Request Type

miles-grant-ibigroup:
As OTP instances don't provide all functionality required by OTP-RR, this PR adds support for routing requests to different servers depending on mode. In this case, flex requests go to OTP-2 instances while non-flex requests go to the OTP-1 instance.

Piggyback PR of https://github.com/opentripplanner/otp-react-redux/pull/470. However, I think it should be merged into dev after #470 is merged.

The end result: flex and e-scooters in the same itinerary results!
Screen Shot 2021-10-26 at 3 04 37 PM

### DIFF
--- a/example-config.yml
+++ b/example-config.yml
@@ -2,6 +2,13 @@ api:
   host: http://localhost
   path: /otp/routers/default
   port: 8001
+
+# Support OTP-2 in parallel for certain requests
+# api_v2:
+#  host: http://localhost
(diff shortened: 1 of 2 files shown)


miles-grant-ibigroup:
It looks like this has been superseded by/integrated into #470.

PR #482, opened 2021-10-28
style(EnhancedStopMarker): Tweak calculation of bubble offset

binh-dam-ibigroup:
Tweaks for positioning of the bubbles for #469.


miles-grant-ibigroup (approved):
This doesn't seem to work perfectly when changing values, but I prefer it to what we had before.

Screen Shot 2021-11-01 at 10 27 58 AM

PR #484, opened 2021-10-28
fix(deps): Upgrade OTP-UI location field to 1.7.0

binh-dam-ibigroup:
This PR adds subtitles to search results in the location fields. No changes or additions to configurations are needed.


philip-cline (approved):
Looks great! I think this is a great improvement for the user

PR #486, opened 2021-11-01
update focus point based on geolocation

miles-grant-ibigroup:
This PR adjusts the geocoder focus point when the user's geolocation is updated. This means that mobile users will get results based on where they are! 

The standard focus point is preserved if geolocation is not present/permitted.

### DIFF
--- a/lib/components/form/connect-location-field.js
+++ b/lib/components/form/connect-location-field.js
@@ -24,8 +24,8 @@ export default function connectLocationField (StyledLocationField, options = {})
     const query = activeSearch ? activeSearch.query : currentQuery
 
     const geocoderConfig = config.geocoder
-    if (state.otp.location.currentPosition.coords) {
-      const {latitude: lat, longitude: lon} = state.otp.location.currentPosition.coords
+    if (currentPosition?.coords) {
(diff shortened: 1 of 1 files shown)


daniel-heppner-ibigroup (approved):

lib/components/form/connect-location-field.js
```diff
@@ -23,9 +23,15 @@ export default function connectLocationField (StyledLocationField, options = {})
     const activeSearch = getActiveSearch(state)
     const query = activeSearch ? activeSearch.query : currentQuery
 
+    const geocoderConfig = config.geocoder
+    if (state.otp.location.currentPosition.coords) {
+      const {latitude: lat, longitude: lon} = state.otp.location.currentPosition.coords
```
binh-dam-ibigroup, line 28: Reuse variable `currentPosition` already declared on L22.

binh-dam-ibigroup (approved):
LGTM, with one small refactor.

lib/components/form/connect-location-field.js
miles-grant-ibigroup: Can't believe I missed that! Thanks for catching and updated in 50deca7318c6444da0ded3e0b97a3d5c23dd8aec

PR #491, opened 2021-11-08
fix(deps): update otp-ui location field to 1.8.0

daniel-heppner-ibigroup:
The 1.7.0 location field update introduced street names to business/place search results, however the street names didn't show up in other locations, like the itinerary or saved trips. This update fixes that, so those results will propagate everywhere. For old saved trips, or for URLs (with query params) saved before this update, users will still see the old names.

### DIFF
--- a/__tests__/components/viewers/__snapshots__/stop-viewer.js.snap
+++ b/__tests__/components/viewers/__snapshots__/stop-viewer.js.snap
@@ -369,7 +369,7 @@ exports[`components > viewers > stop viewer should render countdown times after 
                               size="0.9em"
                               title="From Location Icon"
                             >
-                              <Styled(Component)
+                              <StyledIconBase
                                 className="styled__FromIcon-n5xcvc-0 fTLtLq"
                                 iconAttrs={
                                   Object {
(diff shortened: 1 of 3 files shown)


daniel-heppner-ibigroup:
Can someone take a look at the yarn.lock? Not sure about the geocoder- looks like one version got removed in the git merge.

miles-grant-ibigroup (changes requested):
Looks good, thanks for updating this! The yarn.lock issues might come down to one of the otp-ui dependencies having an older version of geocoder as a dependency. I ran `yarn install` locally and it made some changes to the lock file that looked productive. That might resolve the issue? Otherwise we can try to run [yarn-deduplicate](https://github.com/atlassian/yarn-deduplicate)

daniel-heppner-ibigroup:
yarn.lock looking correct now?

miles-grant-ibigroup:
Lockfile is definitely looking better, but I think now is probably a good a time as any to run `yarn-deduplicate -s fewer` which should clean things up a bit, since I've been having similar lockfile issues elsewhere.

miles-grant-ibigroup (approved):
Looks wonderful now, thanks for all these adjustments!

PR #492, opened 2021-11-08
Enhanced stop viewer bugfixes

### LINKED ISSUE #483: New stop viewer issues/improvements
Issues arising from enhanced stop viewer features (bugs introduced by #469 and maybe #423)
- The map marker (a teal dot on the map) that identifies the selected stop in the stop viewer has been lost.
- Nearby micromobility devices are shown at all zoom levels (even when zoomed far and the micromobility map layers are turned off). They should be shown at the same zoom levels as the enhanced stop bubbles.

miles-grant-ibigroup:
The enhanced stop viewer brought quite a few bugs, outlined in #483. This PR closes #483 by either resolving these bugs or addressing them via feature tweaks.

### DIFF
--- a/lib/actions/map.js
+++ b/lib/actions/map.js
@@ -40,8 +40,8 @@ export function clearLocation (payload) {
 export function zoomToStop (stop) {
   return function (dispatch, getState) {
     if (!stop) return
-    dispatch(setMapCenter({ lat: stop.lat, lon: stop.lon }))
     dispatch(setMapZoom({ zoom: 17 }))
+    dispatch(setMapCenter({ lat: stop.lat, lon: stop.lon }))
   }
 }
 
--- a/lib/actions/map.js
+++ b/lib/actions/map.js
@@ -41,7 +41,7 @@ export function zoomToStop (stop) {
   return function (dispatch, getState) {
     if (!stop) return
     dispatch(setMapCenter({ lat: stop.lat, lon: stop.lon }))
-    dispatch(setMapZoom({ zoom: 18 }))
+    dispatch(setMapZoom({ zoom: 17 }))
   }
 }
 
--- a/lib/components/map/connected-stop-viewer-overlay.js
+++ b/lib/components/map/connected-stop-viewer-overlay.js
@@ -8,6 +8,7 @@ import { connect } from 'react-redux'
 import * as mapActions from '../../actions/map'
 
 import EnhancedStopMarker from './enhanced-stop-marker'
+import StopMarker from './connected-stop-marker'
 
 // Minimum zoom to show stop-viewer overlay content
 const MIN_ZOOM = 17
@@ -54,7 +55,8 @@ class StopViewerOverlay extends MapLayer {
       {
         minZoom: MIN_ZOOM,
         symbol: EnhancedStopMarker
-      }
+      },
+      { minZoom: 2, symbol: StopMarker }
     ]
     const zoom = leaflet.map.getZoom()
     // if (zoom < stopSymbols[0].minZoom) return <FeatureGroup />
--- a/lib/components/map/connected-stop-viewer-overlay.js
+++ b/lib/components/map/connected-stop-viewer-overlay.js
@@ -8,7 +8,6 @@ import { connect } from 'react-redux'
 import * as mapActions from '../../actions/map'
 
 import EnhancedStopMarker from './enhanced-stop-marker'
-import StopMarker from './connected-stop-marker'
 
 // Minimum zoom to show stop-viewer overlay content
 const MIN_ZOOM = 17
@@ -29,9 +28,7 @@ class StopViewerOverlay extends MapLayer {
     const {leaflet, stopData} = this.props
     // If a new stop is clicked, close the stop viewer popup
     if (stopData?.id !== prevProps?.stopData?.id) {
-      leaflet.map && leaflet.map.eachLayer((layer) => {
-        if (layer.options.pane === 'popupPane') layer._closeButton.click()
-      })
+      leaflet.map && leaflet.map.closePopup()
     }
   }
 
@@ -55,8 +52,7 @@ class StopViewerOverlay extends MapLayer {
       {
         minZoom: MIN_ZOOM,
         symbol: EnhancedStopMarker
-      },
-      { minZoom: 2, symbol: StopMarker }
+      }
     ]
     const zoom = leaflet.map.getZoom()
     // if (zoom < stopSymbols[0].minZoom) return <FeatureGroup />
--- a/lib/components/map/connected-stop-viewer-overlay.js
+++ b/lib/components/map/connected-stop-viewer-overlay.js
@@ -24,6 +24,16 @@ class StopViewerOverlay extends MapLayer {
 
   updateLeafletElement () {}
 
+  componentDidUpdate (prevProps) {
+    const {leaflet, stopData} = this.props
+    // If a new stop is clicked, close the stop viewer popup
+    if (stopData?.id !== prevProps?.stopData?.id) {
+      leaflet.map && leaflet.map.eachLayer((layer) => {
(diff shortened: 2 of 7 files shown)


daniel-heppner-ibigroup:
Looks good, only thing I'd say is that these changes don't completely preclude the layer click issue with the enhanced stop bubbles. You can still zoom back in after clicking on a random location and then you are unable to click on the stop bubbles still. I liked your idea of reducing the opacity of the bubbles when they're not clickable, but I'm still not sure this is ideal. It seems like it would be pretty rare that one would need to click "behind" a stop bubble, especially since they only show up at such a low zoom level.

binh-dam-ibigroup (commented):
I think the teal map marker should still be shown when zoomed away, it is useful to locate the selected stop when no stop markers are shown on the map. I saw the little magnifier icon in the Stop Viewer to zoom into the stop, but it does not seem to work, so that should be fixed too.

lib/actions/map.js
```diff
@@ -40,8 +40,8 @@ export function clearLocation (payload) {
 export function zoomToStop (stop) {
   return function (dispatch, getState) {
     if (!stop) return
+    dispatch(setMapZoom({ zoom: 17 }))
     dispatch(setMapCenter({ lat: stop.lat, lon: stop.lon }))
```
binh-dam-ibigroup, line 44: When zoomed out, these actions seemed to be called, but no UI changes happen in response.

lib/components/map/enhanced-stop-marker.js
```diff
@@ -119,7 +120,7 @@ class EnhancedStopMarker extends Component {
       // Instead, we want the tip of the caret at the bottom center of the bubble
       // to be at the stop location.
       // Add some margins so the stop marker (which may be unintentionally offset) remains visible.
-      iconAnchor: [bubblePixels / 2 + 4, bubblePixels + caretVisibleHeight + 8]
+      iconAnchor: [bubblePixels(this.props) / 2 + 4, bubblePixels(this.props) + caretVisibleHeight + 8]
```
binh-dam-ibigroup, line 123: I think the bubble should be shifted down so that the border where it meets the caret doesn't show.
image

binh-dam-ibigroup (changes requested):
I'm going to request a change over https://github.com/opentripplanner/otp-react-redux/pull/492/files#r746751001 (plus the visual tweak). When addressed, feel free to dismiss this review.

lib/actions/map.js
miles-grant-ibigroup: As far as I can tell this is a Leaflet bug where if you take any user action on the map the map updates don't work. The same bug can be seen when planning a trip, going to the route viewer (letting the map pan to the route) and then going back to the itinerary view. This will work, so long as the map wasn't dragged while in the route viewer.

I'm not sure if this behavior was ever intentional.

miles-grant-ibigroup:
All issues mentioned here are caused by the fact that otp-rr has no way to access internal leaflet state. 

For allowing the bubbles to be clickable at all times, or to change opacity based on what the user is doing requires accessing leaflet state which is not possible -- otp-rr does not know which popups are active (beyond doing hacks like we're doing to close all popups)

This lack of info about leaflet state is also what is causing the magnifying glass issue. If the user pans the map the leaflet state diverges from the otp-rr tracked leaflet state. There is no way that I can tell to keep these in sync. Therefore, the magnifying glass has no effect after panning the map manually, as otp-rr does not believe the map has changed.

I have tested this, and both of these bugs are present in the latest dev branch, including before any of the stop viewer changes were made. Since they both seem to be unrelated to this PR, I would ask that we table them into a separate issue and examine how Leaflet and OTP-RR are connected. I have some ideas on how to link the 2, as there is a way to get the current zoom level and position out of the Leaflet object, but no way to read updates from it. However, I think these ideas are best executed in a separate PR which addresses this issue in particular, as again it is not related to this PR or the other stop viewer PR.

miles-grant-ibigroup:
Active stop highlights added in ec8b9e71742ae0efdfbc1f7e8f2811b9e2b33e1b

lib/components/map/enhanced-stop-marker.js
miles-grant-ibigroup: Fixed in abcff8324907d35f9686a327d8c68a4ca9a283f8

miles-grant-ibigroup:
Alright I'm going to assign back to @binh-dam-ibigroup for his thoughts first, then @daniel-heppner-ibigroup.

lib/actions/map.js
binh-dam-ibigroup: Noted in #499.

lib/components/map/connected-stop-viewer-overlay.js
```diff
@@ -21,6 +25,16 @@ class StopViewerOverlay extends MapLayer {
 
   updateLeafletElement () {}
 
+  componentDidUpdate (prevProps) {
+    const {leaflet, stopData} = this.props
+    // If a new stop is clicked, close the stop viewer popup
+    if (stopData?.id !== prevProps?.stopData?.id) {
+      leaflet.map && leaflet.map.eachLayer((layer) => {
+        if (layer.options.pane === 'popupPane') layer._closeButton.click()
```
binh-dam-ibigroup, line 33: There are two existing ways to dismiss a popup, could you try either:
- by calling `setMapPopupLocation({ location: null })` (as in the `DefaultMap` component),
- by calling `leaflet.map.closePopup()` (as in the `SetFromToButtons` component).

binh-dam-ibigroup (approved):
Approving subject to trying the suggested ways to cleanly dismiss a popup.

lib/components/map/connected-stop-viewer-overlay.js
miles-grant-ibigroup: Very good call thank you! Much cleaner now in 718b9acd450d65c9fd1552edd7b464a6c021d387

lib/components/map/connected-stop-viewer-overlay.js
```diff
@@ -62,7 +73,7 @@ class StopViewerOverlay extends MapLayer {
                 {...overlayConfig}
                 configCompanies={configCompanies}
                 stations={bikeRental && bikeRental.stations}
-                visible
+                visible={zoom >= MIN_ZOOM - 1}
```
daniel-heppner-ibigroup, line 76: for my own curiosity: How does this component rerender when the zoom changes? I remember you saying that we couldn't get the zoom updates out of Leaflet, but that must be happening here. Is that because this is a layer that is installed in Leaflet?

lib/components/map/point-popup.js
```diff
@@ -10,10 +14,20 @@ const PopupTitle = styled.div`
...
+function MapPopup ({
   mapPopupLocation,
-  onSetLocationFromPopup
+  onSetLocationFromPopup,
+  setMapZoom,
+  zoom
 }) {
+  // Zoom out if zoomed in very far
+  useEffect(() => {
+    if (zoom > 15) {
```
daniel-heppner-ibigroup, line 25: Is this not just the zoom from config? (looking at mapstatetoprops)

daniel-heppner-ibigroup (approved):

lib/components/map/point-popup.js
miles-grant-ibigroup: Yes it is, but `setMapZoom` updates this config value which then gets passed into leaflet

lib/components/map/connected-stop-viewer-overlay.js
miles-grant-ibigroup: Good catch figuring out how this works! that's exactly it. We can only do `leaflet.map.getZoom()` here because we know that the render method will be called whenever Leaflet's zoom level changes. However, we can't pass this down to the stopSymbols so easily because we need to deal with the `ZoomBasedMarkers` sitting in-between. Not an impossible task, but definitely out of scope for this PR.

PR #496, opened 2021-11-10
Calltaker + Field Trip Craco Fixes

### LINKED ISSUE #490: Call Taker UI: Z-Index of advanced options incorrect
In the Call Taker UI, the advanced options pane goes in front of search results and the drop down button is behind the itinerary options panel. This means that after planning a trip, in some browser windows, the drop down is not clickable. 
image

daniel-heppner-ibigroup:
This fixes #490, and it's probably a good change to make besides that. I couldn't see any reason that the narrative pane needed to be positioned absolutely, with complicated math to calculate its position. Just letting it fall into the UI flow seems to work just fine in every browser I tested. I made the necessary padding adjustments for it to work properly.


lib/components/app/call-taker-panel.js
```diff
@@ -251,15 +251,10 @@ class CallTakerPanel extends Component {
...
             <NarrativeItineraries
               containerStyle={{
-                bottom: '0',
                 display: 'flex',
-                flexDirection: 'column',
-                left: '0',
-                position: 'absolute',
-                right: '0',
+                flexDirection: 'column'
                 // FIXME: This top pixel value can be variable dependent on
```
miles-grant-ibigroup, line 256: This comment can probably be removed now right

miles-grant-ibigroup (approved):
Looks really good! Thanks for the refactor. Once the comment is gone we should be good to go.

daniel-heppner-ibigroup:
I would like to double check what the batch-routing-panel looks like before merging this in, but I can't get ATL RIDES running to see it.

miles-grant-ibigroup:
I can confirm that the batch routing panel looks good.

lib/components/app/call-taker-panel.js
```diff
@@ -251,15 +251,8 @@ class CallTakerPanel extends Component {
...
-                bottom: '0',
                 display: 'flex',
-                flexDirection: 'column',
-                left: '0',
-                position: 'absolute',
-                right: '0',
-                // FIXME: This top pixel value can be variable dependent on
-                // height of the form above. It may need to be specified differently
-                top: 210 + intermediatePlaces.length * 45
+                flexDirection: 'column'
```
binh-dam-ibigroup, line 255: One side effect of this change is that the results are no longer on the edge of the panel:
image

miles-grant-ibigroup:
This should be ready for testing again with craco!

miles-grant-ibigroup:
Have fixed calltaker to work with craco. @binh-dam-ibigroup the issue you raised with the padding is created by external CSS, not the built-in otp-rr CSS.

@daniel-heppner-ibigroup I will assign back to you to approve the latest changes.

daniel-heppner-ibigroup:
I honestly can't tell what's going in the PR because of all the changes, but I checked it out in browser and it looks good.

binh-dam-ibigroup (approved):
Let's go with these changes for the purposes of this PR, but some more visual dusting needs to be handled separately.

PR #467, opened 2021-10-06
Make use of updated `trip-details` component

miles-grant-ibigroup:
Blocked by https://github.com/opentripplanner/otp-ui/pull/294. **Tests will fail until new packages are released.**

This PR uses the updated `trip-details` component (https://github.com/opentripplanner/otp-ui/pull/294) to display all fares returned by OTP.
Screen Shot 2021-10-05 at 4 23 03 PM

PR #498, opened 2021-11-10
OTP-RR Release (November 10, 2021)

miles-grant-ibigroup:
This release contains the following PRs:
- #482
- #477
- #484
- #473 
- #456 
- #487 
- #486 
- #467


binh-dam-ibigroup (approved):

daniel-heppner-ibigroup (changes requested):
Is it too late to include location-field 1.8.0 in this release, which includes the update to include the street name for places everywhere that you see the place rather than just in the search results?

miles-grant-ibigroup:
Ah yes good call! Will add it to the list and mark as blocked until we get that merged in to dev.

philip-cline (approved):

PR #500, opened 2021-11-15
Add user setting for accessible trips by default

miles-grant-ibigroup:
This PR adds a new setting in user preferences to use accessible routes by default.

The user store is also updated to update the queryparams accordingly if the option is set.

Relies on https://github.com/ibi-group/otp-middleware/pull/157 to set the user preference correctly.


daniel-heppner-ibigroup (approved):

lib/actions/user.js
```diff
@@ -100,6 +101,11 @@ function setUser (user, fetchTrips) {
     if (fetchTrips) {
       dispatch(fetchMonitoredTrips())
     }
+
+    const { accessibilityRoutingByDefault } = user
+    if (accessibilityRoutingByDefault) {
+      dispatch(setQueryParam({wheelchair: accessibilityRoutingByDefault}))
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Yeah that's true that's a bit unclear. Have made it a bit more clear in e7578083ff6ddbdb2d25f1a7ca60964484260a07. I'm weary of doing an explicit true so that it's clear that the variable can be false too rather than just undefined.

daniel-heppner-ibigroup (approved):

lib/actions/user.js
```diff
@@ -360,18 +404,27 @@ export function toggleSnoozeTrip (trip, intl) {
  * Deletes a logged-in user's monitored trip,
  * then, if that was successful, refreshes the redux monitoredTrips state.
  */
-export function confirmAndDeleteUserMonitoredTrip (tripId, intl) {
+export function confirmAndDeleteUserMonitoredTrip(tripId, intl) {
   return async function (dispatch, getState) {
     if (
+      // eslint-disable-next-line no-restricted-globals
       !confirm(
```
binh-dam-ibigroup, line 411: Can writing `window.confirm` help get rid of the linting exception?

lib/actions/user.js
```diff
@@ -528,11 +605,11 @@ export function saveUserPlace (placeToSave, placeIndex, intl) {
...
  * Delete the place data at the specified index for the logged-in user.
  */
-export function deleteUserPlace (placeIndex, intl) {
+export function deleteUserPlace(placeIndex, intl) {
   return function (dispatch, getState) {
-    if (
-      !confirm(intl.formatMessage({ id: 'actions.user.confirmDeletePlace' }))
-    ) return
+    // eslint-disable-next-line no-restricted-globals
+    if (!confirm(intl.formatMessage({ id: 'actions.user.confirmDeletePlace' })))
```
binh-dam-ibigroup, line 611: Same here, does `window.confirm` let us remove the linting exception?

binh-dam-ibigroup (approved):
LGTM!

lib/actions/user.js
miles-grant-ibigroup: Great idea thank you! Totally worked in 90efea159fb187b6d36a42607c00df74e0880d54

lib/actions/user.js
miles-grant-ibigroup: Absolutely does! Good shout

Done in 90efea159fb187b6d36a42607c00df74e0880d54

PR #503, opened 2021-11-17
fix(switch-button): automatically re-plan after clicking

miles-grant-ibigroup:
The switch button was doing everything correctly, but because didn't wait for dispatches to finish, had an internal race condition that sometimes resulted in it not planning the new trip.

This PR fixes this.

### DIFF
--- a/lib/actions/map.js
+++ b/lib/actions/map.js
@@ -102,19 +102,19 @@ export function setLocationToCurrent (payload) {
 }
 
 export function switchLocations () {
-  return function (dispatch, getState) {
+  return async function (dispatch, getState) {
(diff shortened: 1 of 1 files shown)


philip-cline (approved):

daniel-heppner-ibigroup:
All you needed to do was make this return a promise? Does that mean this was originally the intended functionality but wasn't working properly?

daniel-heppner-ibigroup (approved):

miles-grant-ibigroup:
I think some change must have happened somewhere that allowed the race condition to occur? But yes I do believe that this was the intended functionality.

PR #504, opened 2021-11-19
feat: show errors next to results if results also present

miles-grant-ibigroup:
There was a bug with the tabbed itineraries which caused a bike share network not being available to hide all itineraries! This was resolved by allowing errors to be displayed above results if both results and errors are present.


daniel-heppner-ibigroup:
What about the other itinerary views?

lib/components/narrative/tabbed-itineraries.js
```diff
@@ -64,49 +66,57 @@ class TabbedItineraries extends Component {
...
-              showRealtimeAnnotation={showRealtimeAnnotation}
-              timeFormat={timeFormat}
-              {...itineraryBodyProps}
-            />
-          )
-          : null
-        }
+          {/* Show active itin if itineraries exist and active itin is defined. */}
+          {(itineraries.length > 0 && activeItinerary >= 0)
+            ? (
```
daniel-heppner-ibigroup, line 101: pretty sure this could just be `(condition) && <component>`

daniel-heppner-ibigroup: although it looks like the only change here was indentation, so maybe you don't want to change this.

daniel-heppner-ibigroup (approved):

lib/components/narrative/tabbed-itineraries.js
miles-grant-ibigroup: yeah usually we don't refactor unless we're directly touching that line. I try to not commit auto-formatted noise but it looks like this one slipped through!

miles-grant-ibigroup:
The other itinerary view handles the error better already! The error gets included in the big error list that I don't think anyone ever clicks.

lib/components/narrative/tabbed-itineraries.js
```diff
@@ -64,49 +66,57 @@ class TabbedItineraries extends Component {
...
-                index={index}
-                isActive={index === activeItinerary}
-                itinerary={itinerary}
-                key={index}
-                onClick={setActiveItinerary}
-              />
-            )
-          })}
-        </div>
+      <>
```
binh-dam-ibigroup, line 69: Do we need this fragment `<>` tag?

lib/components/narrative/tabbed-itineraries.js
```diff
@@ -64,49 +66,57 @@ class TabbedItineraries extends Component {
...
+        {error && itineraries && <div className='error-message'>
+          <div className='header'>
+            <i className='fa fa-exclamation-circle' /> Warning
```
binh-dam-ibigroup, line 72: Use `<Icon type="..." withSpace />`, and move the "Warning" text inside the i18n message YML.

lib/components/narrative/tabbed-itineraries.js
```diff
@@ -64,49 +66,57 @@ class TabbedItineraries extends Component {
...
+          </div>
+          <div className='message'>{error.msg}</div>
```
binh-dam-ibigroup, line 74: Display the localized error message based on error code, then fallback on the error message. See https://github.com/opentripplanner/otp-react-redux/blob/05db79c6e1852f8cab72467fd8bbac94c9198304/lib/components/form/error-message.js#L11.

binh-dam-ibigroup: Or use the `<ErrorMessage>` component here instead!

binh-dam-ibigroup (changes requested):
Please use the the proper way to get localized error messages. Also, for future work, I feel this type of change should be under a `fix` commit rather than `feat`.

lib/components/narrative/tabbed-itineraries.js
miles-grant-ibigroup: Removed in ae51c07e1e6bbf86d5e1edd1c32408e0ef025295

lib/components/narrative/tabbed-itineraries.js
miles-grant-ibigroup: Good find thank you! Much much cleaner in ae51c07e1e6bbf86d5e1edd1c32408e0ef025295

miles-grant-ibigroup:
Good catch on the `<ErrorMessage>`! And I agree with the `feat` vs `fix`, I anticipated this would be a larger change than it was.

lib/components/form/error-message.js
```diff
@@ -15,10 +22,10 @@ const ErrorMessage = ({ error, message }) => {
     <div className='error-message'>
       <div className='header'>
         <Icon type='exclamation-circle' />
-        <FormattedMessage id='components.ErrorMessage.header' />
+        <FormattedMessage id={warning ? 'components.ErrorMessage.warning' : 'components.ErrorMessage.header'} />
```
binh-dam-ibigroup, line 25: Per a comment in another PR by @evansiroky, it is preferrable to issue a `<FormattedMessage id="..." />` for each message id rather than generate ids dynamically (this is so that `formatjs`' CLI tools can pick up the messages).

miles-grant-ibigroup: Didn't realize it had to be the entire component rather than just the string! Adjusted in c47e62fa3758622229941f3636acbd33136ad89e

binh-dam-ibigroup (approved):

PR #505, opened 2021-11-29
Add missed i18n strings, clean up PR3

philip-cline:
PR to add missed translation string in the StopViewer component. (thanks @miles-grant-ibigroup for the catch)


i18n/en-US.yml
```diff
@@ -100,14 +100,14 @@ components:
     bikesAvailable: >
       {count, plural,
         =0 {No bikes available}
-        one {{count} bike available}
-        other {{count} bikes available}
+        one {# bike available}
```
binh-dam-ibigroup, line 103: Good find!

binh-dam-ibigroup (approved):
Probably one variable to rename but otherwise LGTM. Thanks for pre-translating!

miles-grant-ibigroup (approved):
Thanks for fixing all these straggling issues! Looks great

PR #511, opened 2021-12-02
OTP 2 trip viewer repair

miles-grant-ibigroup:
This is the final fix for running OTP 1 and 2 in parallel.

The data returned from the trip endpoint of OTP 2 doesn't include some of the things the OTP 1 endpoint contained, and this causes it to crash in some cases.

This PR solves the issue in 2 ways. First, it makes a graphQL request to obtain the supplemental data. Second, it patches the trip viewer to not crash if this information is missing.

The extra GraphQL request is not getting that much information, and we may decide during the review process to not make this request and instead leave a comment noting that when we fully move to OTP-2 the full information can be retrieved.


binh-dam-ibigroup:
Any reason the tests aren't showing?

miles-grant-ibigroup:
I think it was because it was a merge commit rather than a "proper" commit. We can try pushing a dummy commit

lib/actions/api.js
```diff
@@ -559,6 +560,56 @@ export function findTrip(params) {
     }
   )
 }
+export function findGraphQLTrip(params) {
+  // Attempt an otp-2 graphQL request for the same trip before failing
+  // since the otp-2 rest api doesn't return the needed fields
```
binh-dam-ibigroup, line 565: Convert this comment to JSDoc (insert blank line before JSDoc).

lib/components/viewers/trip-viewer.js
```diff
@@ -1,29 +1,30 @@
...
 import moment from 'moment'
 import PropTypes from 'prop-types'
 import React, { Component } from 'react'
-import { Button, Label } from 'react-bootstrap'
-import { FormattedMessage, FormattedTime } from 'react-intl'
-import { connect } from 'react-redux'
 
+import { findGraphQLTrip, findTrip } from '../../actions/api'
+import { setLocation } from '../../actions/map'
+import { setViewedTrip } from '../../actions/ui'
```
binh-dam-ibigroup, line 13: Use the following syntax for importing actions that was adopted some time in 2020.
It prevents from accidentally calling an action without mapping it to dispatch:
```js
import * as apiActions from '../../actions/api'
...
const mapDispatchToProps = {
  findTrip: apiActions.findTrip
  ...
}

```

lib/components/viewers/trip-viewer.js
```diff
@@ -32,91 +33,100 @@ class TripViewer extends Component {
...
+          {tripData &&
+            tripData.stops &&
+            tripData.stopTimes &&
             tripData.stops.map((stop, i) => {
               // determine whether to use special styling for first/last stop
               let stripMapLineClass = 'strip-map-line'
               if (i === 0) stripMapLineClass = 'strip-map-line-first'
-              else if (i === tripData.stops.length - 1) stripMapLineClass = 'strip-map-line-last'
+              else if (i === tripData.stops.length - 1)
+                stripMapLineClass = 'strip-map-line-last'
```
binh-dam-ibigroup, line 120: I would use brackets on multi-line if/else blocks.

lib/components/viewers/trip-viewer.js
```diff
@@ -32,91 +33,100 @@ class TripViewer extends Component {
...
               let highlightClass
-              if (i === viewedTrip.fromIndex) highlightClass = 'strip-map-highlight-first'
-              else if (i > viewedTrip.fromIndex && i < viewedTrip.toIndex) highlightClass = 'strip-map-highlight'
-              else if (i === viewedTrip.toIndex) highlightClass = 'strip-map-highlight-last'
+              if (i === viewedTrip.fromIndex)
+                highlightClass = 'strip-map-highlight-first'
+              else if (i > viewedTrip.fromIndex && i < viewedTrip.toIndex)
+                highlightClass = 'strip-map-highlight'
+              else if (i === viewedTrip.toIndex)
+                highlightClass = 'strip-map-highlight-last'
```
binh-dam-ibigroup, line 129: I would use brackets here.

binh-dam-ibigroup (changes requested):
Can you make one import syntax adjustment please? Seems to work fine otherwise. See other minor comments. too.

lib/actions/api.js
miles-grant-ibigroup: Good call! Done in ca08debd338383fb327a3543108c92b3e269a740

lib/components/viewers/trip-viewer.js
miles-grant-ibigroup: That's why we do that! Genius. Refactored in ca08debd338383fb327a3543108c92b3e269a740

lib/components/viewers/trip-viewer.js
miles-grant-ibigroup: Good catch thanks! Prettier's doing

fixed in ca08debd338383fb327a3543108c92b3e269a740

lib/components/viewers/trip-viewer.js
miles-grant-ibigroup: Good catch thanks! Prettier's doing

fixed in ca08deb

lib/components/viewers/trip-viewer.js
```diff
@@ -1,29 +1,30 @@
...
 import moment from 'moment'
 import PropTypes from 'prop-types'
 import React, { Component } from 'react'
-import { Button, Label } from 'react-bootstrap'
-import { FormattedMessage, FormattedTime } from 'react-intl'
-import { connect } from 'react-redux'
 
+import * as mapActions from '../../actions/map'
+import * as uiActions from '../../actions/ui'
+import { findGraphQLTrip, findTrip } from '../../actions/api'
```
binh-dam-ibigroup, line 13: Do the same for API actions.

binh-dam-ibigroup (approved):
One more import refactor for API actions, please then I'm good.

lib/components/viewers/trip-viewer.js
miles-grant-ibigroup: Slept right through that one! Thank you for catching

Extracted in 24d33b2cc7bc6c907163261e887bc141de279f1d

PR #521, opened 2022-01-12
Add Mapillary links in itinerary

miles-grant-ibigroup:
Depends on https://github.com/opentripplanner/otp-ui/pull/328

Adds Mapillary key to config to make use of `itinerary-body` updates. Also uses the callback to render the Mapillary iframe in a popover in the map, rather than as a popup.
Screen Shot 2022-01-12 at 4 23 59 PM

There is a lot of auto formatter noise in this PR, apologies. The actual changes are minimal, and mostly clone the elevation popup reducers to store a Mapillary ID in redux state that is then read by a new component which renders the iframe (and could theoretically render something else)

To enable the new feature, the active config must have a `maxillary -> key` set (see example config). You can register for a free key at [https://www.mapillary.com/dashboard/developers](https://www.mapillary.com/dashboard/developers)


demory (approved):

example-config.yml
```diff
@@ -409,3 +409,7 @@ dateTime:
 #   # Whether to render routes within flex zones of a route's patterns. If set to true,
 #   # routes will not be rendered within flex zones.
 #   hideRouteShapesWithinFlexZones: true
+
+# API key to make Mapillary API calls. These are used to show street imagery.
+# mapillary:
+#   key: <Mapillary Key>
```
daniel-heppner-ibigroup, line 415: I'm not seeing an API key provided from Mapillary on the dev portal. They provide a client token and a client secret, at least with the way I registered the app. What do I put in this field? Can we call this something else? Or clarify in the comment?

daniel-heppner-ibigroup (changes requested):

example-config.yml
miles-grant-ibigroup: I've made a clearer comment in f68f92513ae1214fa5cea5e9cdd287fcd36cc78a!

daniel-heppner-ibigroup (approved):
I think my only problem with this is that it's not immediately clear what I'm looking at which I click on a Mapillary link. 

For example, if I click on the icon for a "Turn left onto X road", am I looking at the imagery for X road or the road I was on right before the turn? I think the current language sounds like I am going to be looking at the intersection where I need to turn, but it seems like it is showing the image from right after the turn.

Is there any way we could clarify this? Maybe that would be better put in the itinerary-body in OTP UI. Or maybe it's not that important.

PR #525, opened 2022-01-17
Detect user locale, add UI override

philip-cline:
This PR adds detection of a default locale, through the following ways (in fallback order): User's language in `preferredLanguage` setting, `window.localstorage` language, `navigator.language`, configuration default locale, 'en-US'. 

** To test this functionality, a `localizationEnabled: true` flag is required in the config.yml file.

This PR relies on additions to otp-middleware to include the new preferredLanguage setting in the OtpUser here:

The PR also adds a UI override for display language (which also updates the preferredLanguage for the user): 
image

Relies on ATL-RIDES QA config from this PR: https://github.com/ibi-group/configurations/pull/230


lib/actions/ui.js
```diff
@@ -87,12 +219,13 @@ export function matchContentToUrl (location) {
         }
         break
       case 'start':
-      case '@':
+      case '@': {
         // Parse comma separated params (ensuring numbers are parsed correctly).
-        let [lat, lon, zoom, routerId] = id ? idToParams(id) : []
+        const [lat, lon, zoom, routerId] = id ? idToParams(id) : []
```
miles-grant-ibigroup, line 224: This change is causing a11y tests to fail. It needs to be changed back to `let`

example-config.yml
```diff
@@ -365,6 +369,37 @@ itinerary:
 #         modes: "CAR_HAIL, CAR_RENT"
 #       480:
 #         msg: No available transit routes or rideshare/carshare service at origin.
+  # fr-FR:
+  #   name: French (France)
+  #   flag: 🇫🇷
```
miles-grant-ibigroup, line 374: We're not using `flag` anymore are we?

lib/components/app/locale-selector.tsx
```diff
@@ -0,0 +1,82 @@
...
+        (key) =>
+          /* Key is locale code, e.g. 'en-US' */
+          key !== currentLocale && (
+            <MenuItem onClick={(e) => handleLocaleSelection(e, key)}>
+              <FlagContainer>{FLAG_ICON_MAPPING[key]}</FlagContainer>
+              {configLanguages[key].name}
+            </MenuItem>
+          )
+      )}
+    </NavDropdown>
```
miles-grant-ibigroup, line 66: There's a lot of serious typescript issues here

lib/config.js
```diff
@@ -25,6 +26,19 @@ const ModeIcon = ClassicModeIcon
...
+ * using emojis would be a simpler solution, but they are not available on Windows
+ */
+export const FLAG_ICON_MAPPING = {
+  'en-US': <US style={{ width: 15 }} />,
+  'es-ES': <ES style={{ width: 15 }} />,
+  'fr-FR': <FR style={{ width: 15 }} />,
+  'ko-KR': <KR style={{ width: 15 }} />,
+  'vi-VN': <VN style={{ width: 15 }} />,
+  'zh-CN': <CN style={{ width: 15 }} />
+}
```
miles-grant-ibigroup, line 40: As mentioned earlier, these defaults need to go somewhere else

miles-grant-ibigroup (changes requested):
This is a good start! For now, I've noted how to make the tests pass and getting the typescript more correct.

There's also some issues when using this on a config without multiple languages, but the typescripting might fix that.

lib/actions/ui.js
philip-cline: changed in https://github.com/opentripplanner/otp-react-redux/pull/525/commits/a6de95114aed8b19ec7c8910f2e5e500a88a524d

example-config.yml
philip-cline: Good catch, changed in https://github.com/opentripplanner/otp-react-redux/pull/525/commits/a6de95114aed8b19ec7c8910f2e5e500a88a524d

lib/components/app/locale-selector.tsx
```diff
@@ -0,0 +1,82 @@
+import { connect } from 'react-redux'
+import { MenuItem, NavDropdown } from 'react-bootstrap'
+import { useIntl } from 'react-intl'
+import React from 'react'
+import styled from 'styled-components'
+
+import * as uiActions from '../../actions/ui'
+import * as userActions from '../../actions/user'
+import { FLAG_ICON_MAPPING } from '../../config'
```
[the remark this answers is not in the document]
philip-cline: Changed in https://github.com/opentripplanner/otp-react-redux/pull/525/commits/a6de95114aed8b19ec7c8910f2e5e500a88a524d

lib/components/app/locale-selector.tsx
philip-cline: Updated in https://github.com/opentripplanner/otp-react-redux/pull/525/commits/a6de95114aed8b19ec7c8910f2e5e500a88a524d

lib/config.js
philip-cline: Updated to be a constant in the locale-selector file in https://github.com/opentripplanner/otp-react-redux/pull/525/commits/a6de95114aed8b19ec7c8910f2e5e500a88a524d

example-config.yml
```diff
@@ -338,12 +338,21 @@ itinerary:
 ### The nested structure should be the same as the language files under the i18n folder.
 ### You can also customize OTP error messages for itinerary searches based on OTP HTTP codes.
 ### A separate message can be set for each language or locale if necessary.
+### All available languages should have an entry with at least a name identifier in this section so that
+### they may be rendered in the locale selector. 
+### Language display icons (flags) should be configured in config.js
+# LocalizationEnabled also provides a method to toggle the new localization functionality, which requires
```
miles-grant-ibigroup, line 344: This should probably be `localeSelectorEnabled`

lib/components/app/locale-selector.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+    // Hack: Reload the page (IntlProvider is at too high a level to listen to the redux store)
+    // TODO: move IntlProvider to Responsive Webapp for reload.
+    document.location.reload()
+  }
+
+  return (
+    <NavDropdown id="locale-selector" title={FLAG_ICON_MAPPING[currentLocale]}>
+      {Object.keys(configLanguages).map(
+        (key) =>
```
miles-grant-ibigroup, line 69: We can probably rename `key` to `language` or something for clarity

lib/components/app/locale-selector.tsx
```diff
@@ -0,0 +1,98 @@
...
+              {configLanguages[key].name}
+            </MenuItem>
+          )
+      )}
+    </NavDropdown>
+  )
+}
+
+// Typescript TODO: create state types and type state, ownProps properly
+const mapStateToProps = (state: any, ownProps: Record<string, any>) => {
```
miles-grant-ibigroup, line 85: I think you can remove `ownProps` entirely

lib/components/app/locale-selector.tsx
```diff
@@ -0,0 +1,98 @@
...
+    <NavDropdown id="locale-selector" title={FLAG_ICON_MAPPING[currentLocale]}>
+      {Object.keys(configLanguages).map(
+        (key) =>
+          /* Key is locale code, e.g. 'en-US' */
+          key !== currentLocale && (
+            <MenuItem
+              onClick={(e: MouseEvent) => handleLocaleSelection(e, key)}
+            >
+              <FlagContainer>{FLAG_ICON_MAPPING[key]}</FlagContainer>
+              {configLanguages[key].name}
```
miles-grant-ibigroup, line 76: If you don't provide a `name` then the menu looks kind of weird. How do you feel about this as a fallback?

```suggestion
              {configLanguages[key].name || key}
```

miles-grant-ibigroup (changes requested):
Looking really good! Thanks for that typescript work, it looks like that was really annoying.

I'm a little divided on the new config property -- on one hand I like that the menu can be disabled, but at the same time I think it's annoying to have to enable it manually since I assume that will be the default. Maybe we switch to it being shown by default unless disabled  by a new config property? Curious for your thoughts

example-config.yml
philip-cline: Updated in https://github.com/opentripplanner/otp-react-redux/pull/525/commits/50614e16b8a4687cb376cdba79923c00c2e76071

lib/components/app/locale-selector.tsx
philip-cline: Changed to `locale` in https://github.com/opentripplanner/otp-react-redux/pull/525/commits/50614e16b8a4687cb376cdba79923c00c2e76071, I think that's more clear

lib/components/app/locale-selector.tsx
philip-cline: Yep, removed in https://github.com/opentripplanner/otp-react-redux/pull/525/commits/50614e16b8a4687cb376cdba79923c00c2e76071

lib/components/app/locale-selector.tsx
philip-cline: With the new functionality, the `LocaleSelector` will not be rendered if any language does not have a `name` defined

miles-grant-ibigroup: Perfect! So simple

lib/components/app/desktop-nav.tsx
```diff
@@ -70,10 +69,15 @@ const DesktopNav = ({ otpConfig }: otpConfigType) => {
...
-            Object.keys(configLanguages).length > 1 && (
+          {configLanguages &&
+            // Avoid rendering the LocaleSelector for configs without languages
+            !Object.keys(configLanguages).find(
+              (key) => key !== 'allLanguages' && !configLanguages[key].name
+            ) &&
+            // Ensure that > 1 valid language is defined
+            Object.keys(configLanguages).filter(
+              (key) => key !== 'allLanguages' && configLanguages[key].name
+            ).length > 1 && (
```
miles-grant-ibigroup, line 80: This filter can probably be removed as its made redundant by the `find` above. The length of the Object.keys should do

lib/components/app/locale-selector.tsx
```diff
@@ -0,0 +1,98 @@
...
+  'es-ES': <ES style={{ width: 15 }} />,
+  'fr-FR': <FR style={{ width: 15 }} />,
+  'ko-KR': <KR style={{ width: 15 }} />,
+  'vi-VN': <VN style={{ width: 15 }} />,
+  'zh-CN': <CN style={{ width: 15 }} />
+}
+
+const FlagContainer = styled.span`
+  &::after {
+    margin: 0 0.125em;
```
miles-grant-ibigroup, line 28: This isn't working for me on safari. We might need `content: ''`?
Screen Shot 2022-01-21 at 9 12 46 AM

miles-grant-ibigroup (approved):
This looks so cool and works so cool! 2 tiny things to fix but once that's done this is good to go in my eyes!

lib/components/app/locale-selector.tsx
miles-grant-ibigroup: We can probably just throw the padding on the span itself right? do we need the pseudo element

lib/components/app/desktop-nav.tsx
philip-cline: Hm I left the filter in to remove the 'allLanguages' key from the count.

lib/components/app/locale-selector.tsx
binh-dam-ibigroup: You have set the `content` attribute too, see for instance `SpanWithSpace`, etc.

philip-cline: Updated in https://github.com/opentripplanner/otp-react-redux/pull/525/commits/46e67c33586931a4bf67960af05245df3520ff94

philip-cline: Although lmk if you think just applying the styling on the span directly is preferable.

lib/components/app/desktop-nav.tsx
miles-grant-ibigroup: Ah in that case let's get rid of the `find` then right? Since the `filter` length will never be `> 1` if the find fails

philip-cline: updated in https://github.com/opentripplanner/otp-react-redux/pull/525/commits/83f2573ec464312673ee9a1317c33f66dce53d11

lib/components/app/locale-selector.tsx
miles-grant-ibigroup: Latest commit works great for me! I think I'd move away from the pseudo element but I'll leave that up to @binh-dam-ibigroup as the final reviewer

example-config.yml
```diff
@@ -338,12 +338,17 @@ itinerary:
...
+### only be rendered if more than one valid language (all languages must have "name" defined) is included
+### in the language config. Locale icons (flags) are currently defined in locale-selector.tsx file. 
+
 # language:
 #   allLanguages
 #     common:
 #       accessModes:
 #         bikeshare: Relay Bike
 #   en-US:
+#     name: English (US)
```
binh-dam-ibigroup, line 351: I think I get the need to list supported languages, so this could be a way to do so. I am puzzled about providing the display names of each language. There is functionality available with react-intl already: https://formatjs.io/docs/react-intl/api#formatdisplayname (we may have to use the polyfill also). Let me think through it.

miles-grant-ibigroup: The polyfill looks like it's massive! As cool as that feature would be, maybe it's a good idea to wait until intl has broad browser support...

binh-dam-ibigroup: Okay. We don't need to have that in this PR.

lib/util/i18n.js
```diff
@@ -34,13 +30,25 @@ function getMatchingLocaleString (locale = '', defaultLocale = 'en-US') {
  *   id => string.
  * FIXME: Load languages on demand using fetch.
  */
-export async function loadLocaleData (locale, customMessages) {
-  const matchedLocale = getMatchingLocaleString(locale)
+export async function loadLocaleData(locale, customMessages, configLocales) {
+  const matchedLocale = getMatchingLocaleString(locale, 'en-US', configLocales)
   let messages
   switch (matchedLocale) {
+    case 'es-ES':
```
binh-dam-ibigroup, line 37: Since we are adding string files for other non-[US English] languages, probably we should remove the country-specific code. `fr-FR` designates French text with France-specific messages (e.g. acronyms or terms only used in the country), but since we are providing examples used in the generic context, the locale and the message file name should just be `fr` going forward. Same for the other languages, and same in `config.yml`.

binh-dam-ibigroup (changes requested):
Good work so far! Please replace the in-code country-to-flag mapping with something done in `config.yml` where the language display text is set.  Also consider showing the selected language in the language dropdown with a checkmark beneath it. See other formatting comments.

lib/components/app/desktop-nav.tsx
```diff
@@ -1,14 +1,14 @@
...
+import { Nav, Navbar } from 'react-bootstrap'
+import React from 'react'
 
-import NavLoginButtonAuth0 from '../user/nav-login-button-auth0.js'
 import { accountLinks, getAuth0Config } from '../../util/auth'
 import { DEFAULT_APP_TITLE } from '../../util/constants'
+import NavLoginButtonAuth0 from '../user/nav-login-button-auth0'
 
 import AppMenu from './app-menu'
+import LocaleSelector from './locale-selector'
```
binh-dam-ibigroup, line 10: Is the locale selector supposed to be available in the mobile view too? (Right now it is not.)

lib/actions/ui.js
```diff
@@ -1,24 +1,88 @@
...
+
+export const setHoveredStop = createAction('SET_HOVERED_STOP')
+
+export const setViewedTrip = createAction('SET_VIEWED_TRIP')
+
+const viewRoute = createAction('SET_VIEWED_ROUTE')
+
+export const toggleAutoRefresh = createAction('TOGGLE_AUTO_REFRESH')
+
+const setPreviousItineraryView = createAction('SET_PREVIOUS_ITINERARY_VIEW')
```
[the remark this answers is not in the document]
philip-cline: Removed in https://github.com/opentripplanner/otp-react-redux/pull/525/commits/2ba2f5eeb9da864347789e418be2bc16ab16ce6b

lib/actions/ui.js
```diff
@@ -1,24 +1,88 @@
...
 
+// UI state enums
+
+export const MainPanelContent = {
+  ROUTE_VIEWER: 1,
+  STOP_VIEWER: 2
+}
+
+export const MobileScreens = {
+  // eslint-disable-next-line sort-keys
```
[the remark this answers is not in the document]
philip-cline: Sorted by name in https://github.com/opentripplanner/otp-react-redux/pull/525/commits/2ba2f5eeb9da864347789e418be2bc16ab16ce6b

lib/actions/ui.js
```diff
@@ -343,12 +355,22 @@ export function showMobileSearchScreen () {
...
-    const messages = await loadLocaleData(effectiveLocale, customMessages)
+    // TODO: Should we get the locales from the config every
+    // time we set the locale or should we destructure once into the
+    // store.
+    const configLocales = getConfigLocales(customMessages)
+    const effectiveLocale = locale || getDefaultLocale(config, loggedInUser)
+    const messages = await loadLocaleData(
+      effectiveLocale,
+      customMessages,
+      configLocales
```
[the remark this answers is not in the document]
philip-cline: Moved inside of `loadLocaleData` in https://github.com/opentripplanner/otp-react-redux/pull/525/commits/2ba2f5eeb9da864347789e418be2bc16ab16ce6b

lib/components/app/desktop-nav.tsx
```diff
@@ -37,49 +43,59 @@ const DesktopNav = ({ otpConfig }) => {
...
+      <Navbar.Collapse>
+        <Nav pullRight>
+          {configLanguages &&
+            // Ensure that > 1 valid language is defined
+            Object.keys(configLanguages).filter(
+              (key) => key !== 'allLanguages' && configLanguages[key].name
+            ).length > 1 && (
+              <LocaleSelector
+                configLanguages={configLanguages}
+                style={{ maxWidth: 60 }}
```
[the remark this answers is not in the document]
philip-cline: The `maxWidth` was a holdover from experimenting with CSS earlier and I removed it in https://github.com/opentripplanner/otp-react-redux/pull/525/commits/2ba2f5eeb9da864347789e418be2bc16ab16ce6b

lib/components/app/locale-selector.tsx
```diff
@@ -0,0 +1,99 @@
...
+ * will need to be found if further customization is required.
+ */
+const FLAG_ICON_MAPPING: Record<string, React.ReactElement> = {
+  'en-US': <US style={{ width: 15 }} />,
+  'es-ES': <ES style={{ width: 15 }} />,
+  'fr-FR': <FR style={{ width: 15 }} />,
+  'ko-KR': <KR style={{ width: 15 }} />,
+  'vi-VN': <VN style={{ width: 15 }} />,
+  'zh-CN': <CN style={{ width: 15 }} />
+}
```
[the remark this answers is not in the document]
philip-cline: Removed this mapping and added a `flag` attribute to the config language section, which renders the appropriate flag from this library. The downside of this is that the entire package is added to the bundle size, as it is imported as `import * as flags from '...'` (all of this in https://github.com/opentripplanner/otp-react-redux/pull/525/commits/2ba2f5eeb9da864347789e418be2bc16ab16ce6b)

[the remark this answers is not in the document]
philip-cline: Width: 15 is supposed to keep the flag sizes consistent

lib/components/app/locale-selector.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+    window.localStorage.setItem('lang', locale)
+
+    if (loggedInUser) {
+      loggedInUser.preferredLanguage = locale
+      createOrUpdateUser(loggedInUser, false, intl)
+    }
+    setLocale(locale)
+
+    // Hack: Reload the page (IntlProvider is at too high a level to listen to the redux store)
```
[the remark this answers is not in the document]
philip-cline: Removed in https://github.com/opentripplanner/otp-react-redux/pull/525/commits/2ba2f5eeb9da864347789e418be2bc16ab16ce6b and we'll have to investigate later and open a separate PR for that fix

lib/components/app/locale-selector.tsx
```diff
@@ -0,0 +1,99 @@
...
+    // TODO: move IntlProvider to Responsive Webapp for reload.
```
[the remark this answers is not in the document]
philip-cline: Removed the TODO in https://github.com/opentripplanner/otp-react-redux/pull/525/commits/2ba2f5eeb9da864347789e418be2bc16ab16ce6b

lib/util/i18n.js
philip-cline: Removed the country codes for most in https://github.com/opentripplanner/otp-react-redux/pull/525/commits/2ba2f5eeb9da864347789e418be2bc16ab16ce6b. Kept the codes for en-US bc some spellings and terms are US specific, and I figured your french was likewise FR specific, @binh-dam-ibigroup

lib/components/app/desktop-nav.tsx
philip-cline: Updated to be included in the mobile view in https://github.com/opentripplanner/otp-react-redux/pull/525/commits/2ba2f5eeb9da864347789e418be2bc16ab16ce6b

binh-dam-ibigroup (changes requested):
One more file and a few more ids to rename, and add the `flag` attribute in the example config, but it looks good.

miles-grant-ibigroup (approved):

binh-dam-ibigroup (approved):
Good work! Two tiny cleanup things but good to go.

lib/components/app/locale-selector.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  return (
+    <NavDropdown
+      id="locale-selector"
+      pullRight
+      title={<CurrentLocaleFlag style={{ width: 15 }} />}
+    >
+      {Object.keys(configLanguages).map((locale) => {
+        const Flag = flags[configLanguages[locale].flag]
+        return (
```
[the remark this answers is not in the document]
philip-cline: This was a prettier suggestion. Ok to keep?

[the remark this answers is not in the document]
binh-dam-ibigroup: Sure

lib/components/user/nav-login-button.css
```diff
@@ -8,24 +8,30 @@
   * In desktop and mobile modes, we change the link color to white for consistency with the AppMenu component.
   */
 
-.otp.mobile .navbar .container-fluid > li {
-  display: block;
-  padding: 14px;
+.otp.mobile .navbar .mobile-navbar-container { 
+  
+}
```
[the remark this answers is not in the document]
philip-cline: Updated in https://github.com/opentripplanner/otp-react-redux/pull/525/commits/f43e11a66b8654b1bb26b944d8634534a327749c

PR #528, opened 2022-01-20
fix(actions/api): don't duplicate query looking for nonRealtime query

miles-grant-ibigroup:
This small PR removes the duplicated queries made when planning a trip. There are some minor implications on the performance of the middleware (as mentioned in the removed comment), but I've tested it with a local middleware instance and all features appear to still be working.


lib/actions/api.js
```diff
@@ -117,11 +117,21 @@ export function routingQuery(searchId = null, updateSearchInReducer = false) {
...
+          .map(({ mode, params }) => ({
+            mode,
+            ...params
+          }))
+          // Remove non-accessible combinations when accessible routing is enabled
+          .filter((combination) => {
+            if (!accessibleRoutingEnabled) return true
+
+            // Explicit check for false is needed to avoid matching
+            // combinations without the pparameter set (undefined)
```
daniel-heppner-ibigroup, line 132: Fix typo

daniel-heppner-ibigroup (approved):
Just found a little typo

lib/actions/api.js
miles-grant-ibigroup: Good catch thanks! Fixed in 47595228122b9d50cefec2bddba18a25e8baa34e

lib/actions/api.js
```diff
@@ -188,50 +206,7 @@ export function routingQuery(searchId = null, updateSearchInReducer = false) {
...
-        )
-          .then(getJsonAndCheckResponse)
-          .then((json) => {
-            // FIXME: This is only performed when ignoring realtimeupdates currently, just
-            // to ensure it is not repeated twice.
-            // FIXME: We should check that the mode combination actually has
-            // realtime (or maybe this is set in the config file) to determine
-            // whether this extra query to OTP is needed.
-            return dispatch(
-              nonRealtimeRoutingResponse({ response: json, searchId })
```
binh-dam-ibigroup, line 226: Delete in the corresponding files:
- this action (this file is the only place it is used)
- `NON_REALTIME_ROUTING_RESPONSE` from the reducer
- instances of `activeSearch.nonRealtimeResponse`
- instances of `state.otp.useRealtime`
- Other vars and code that gets unused from the above.

binh-dam-ibigroup: Would this result in a breaking change commit?

binh-dam-ibigroup (changes requested):
I think we delete a reducer action and propagate the changes (see comment).

lib/actions/api.js
```diff
@@ -188,50 +206,7 @@ export function routingQuery(searchId = null, updateSearchInReducer = false) {
...
-        //
-        // FIXME: The statement below may no longer apply with future work
-        // involving realtime info embedded in the OTP response.
-        // (That action records an entry again in the middleware.)
-        // For users who opted in to store trip request history,
-        // to avoid recording the same trip request twice in the middleware,
-        // only add the user Authorization token to the request
-        // when querying the non-realtime route.
-        //
-        // The advantage of using non-realtime route is that the middleware will be able to
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Yes I think we should do some thinking/investigation about this -- it was my impression that the middleware only took the origin and destination and then made a new trip plan request anyway?

If we do need to make an `ignoreRealtimeUpdates` request then I agree that should only happen from within the user settings.

lib/actions/api.js
miles-grant-ibigroup: See comment above -- I kept these thinking that we would re-use them when we refactor the monitored trips component. I'll add some comments to all these explaining why they're there. 

Something for the backlog though would be to clear out a lot of these reducers. There's plenty of unused ones floating around...

miles-grant-ibigroup: I've gone ahead and removed `useRealtime` as it appeared to be a setting that wasn't used anywhere. That led me to have to add typescript in some places and also make a small refactor to the realtime status label. No breaking change though as `useRealtime` wasn't even in the example-config and hasn't been used in years.

miles-grant-ibigroup: Using `useRealtime` prior to this commit wouldn't have done anything anyway

lib/actions/api.js
```diff
@@ -188,50 +206,7 @@ export function routingQuery(searchId = null, updateSearchInReducer = false) {
...
-              nonRealtimeRoutingResponse({ response: json, searchId })
-            )
-          })
-          .catch((error) => {
-            console.error(error)
-            // do nothing
-          })
-
-        return Promise.all([realTimeFetch, nonRealtimeFetch])
+        return Promise.all([realTimeFetch])
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Yes! Done in ffcd36c3def760c57acd170d6d993bf0764be177

lib/components/mobile/results-header.js
```diff
@@ -100,18 +100,13 @@ class ResultsHeader extends Component {
 
 const mapStateToProps = (state, ownProps) => {
   const activeSearch = getActiveSearch(state)
-  const {useRealtime} = state.otp
-  const response = !activeSearch
-    ? null
-    : useRealtime ? activeSearch.response : activeSearch.nonRealtimeResponse
+  const response = !activeSearch ? null : activeSearch.response
```
binh-dam-ibigroup, line 103: Would this work?
```suggestion
  const response = activeSearch?.response
```

lib/components/mobile/results-screen.js
```diff
@@ -175,18 +168,14 @@ const mapStateToProps = (state, ownProps) => {
...
+        ? itineraries.length
+        : response.otp.profile.length
+      : null
   }
 }
 
-const mapDispatchToProps = {
-  setUseRealtimeResponse: narrativeActions.setUseRealtimeResponse
-}
+const mapDispatchToProps = {}
```
binh-dam-ibigroup, line 179: Delete this var and the 2nd arg passed to `connect`.

lib/components/viewers/realtime-status-label.tsx
```diff
@@ -108,19 +118,27 @@ const RealtimeStatusLabel = ({
...
-              <FormattedDuration duration={Math.abs(delay)} />
-            </DelayText>
-          )}
+          minutes={
+            isEarlyOrLate ? (
+              <DelayText>
+                <FormattedDuration duration={Math.abs(delay)} />
+              </DelayText>
+            ) : (
+              <>{null}</>
```
binh-dam-ibigroup, line 127: Just `null` should be enough, no?

lib/util/state.js
```diff
@@ -265,21 +269,18 @@ export const getActiveItineraries = createSelector(
...
-    useRealtime,
     activeFieldTripRequest
   ) => {
-    // set response to use depending on useRealtime
     const response =
       !nonRealtimeResponse && !realtimeResponse
         ? null
-        : useRealtime
+        : realtimeResponse
         ? realtimeResponse
```
binh-dam-ibigroup, line 284: Can this be simplified?

binh-dam-ibigroup (approved):
A few minor changes but it should be good to go.

lib/components/mobile/results-header.js
miles-grant-ibigroup: Good call thanks! It's so easy to get lost while refactoring something down like this. Cleaned up in 23ef058a46835bd8ce5710176ca1671ffcf7a586

lib/components/mobile/results-screen.js
```diff
@@ -162,10 +158,7 @@ class MobileResultsScreen extends Component {
 
 const mapStateToProps = (state, ownProps) => {
   const activeSearch = getActiveSearch(state)
-  const {useRealtime} = state.otp
-  const response = !activeSearch
-    ? null
-    : useRealtime ? activeSearch.response : activeSearch.nonRealtimeResponse
+  const response = !activeSearch ? null : activeSearch.response
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Yep! Cleaned up in 23ef058a46835bd8ce5710176ca1671ffcf7a586

lib/components/mobile/results-screen.js
miles-grant-ibigroup: Good call thanks! Cleaned up in 23ef058a46835bd8ce5710176ca1671ffcf7a586

lib/components/viewers/realtime-status-label.tsx
miles-grant-ibigroup: If only... Typescript was very upset about the JSX being passed in as a value for FormattedMessage so I had to wrap the null to make it happy

lib/util/state.js
miles-grant-ibigroup: Yep! in 23ef058a46835bd8ce5710176ca1671ffcf7a586

PR #534, opened 2022-01-24
Add `tsc` to `yarn test` and fix typescript

miles-grant-ibigroup:
This PR adds `tsc` to our tests and then fixes bugs found via running `yarn tsc`.

Also cut the ci time in almost half since we're now only running the accessibility tests once!


daniel-heppner-ibigroup (approved):

lib/components/user/stacked-pane-display.tsx
```diff
@@ -10,8 +9,15 @@ import FormNavigationButtons from './form-navigation-buttons'
...
  * TODO: add types once Pane type exists
  */
-// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
-const StackedPaneDisplay = ({ onCancel, paneSequence, title }) => {
+const StackedPaneDisplay = ({
+  onCancel,
+  paneSequence,
+  title
+}: {
+  onCancel: () => void
```
philip-cline, line 17: In the tech call a few weeks ago, I think we decided to consistently include explicit types for the props of React components. This is what we meant by that right (defining after props in a list, rather than exporting to a type like `StackedPaneDisplayProps`?) Not raising an issue here just using this as a use case to make sure I understand.

philip-cline (approved):

lib/components/user/stacked-pane-display.tsx
miles-grant-ibigroup: Yep! 2f4185241c0e6bcd1c7a030edf1b8a2eb074f713 is how this would look following our new agreed-upon pattern

philip-cline: Gucci. Thanks for the explanation

PR #538, opened 2022-02-01
Update issue reporting template

miles-grant-ibigroup:
The previous issue reporting email template left a lot to be desired in terms of user experience. This PR attempts to make the template more user friendly.

French translations are now obsolete and need to be updated @binh-dam-ibigroup

### DIFF
--- a/lib/components/narrative/trip-tools.js
+++ b/lib/components/narrative/trip-tools.js
@@ -92,7 +92,10 @@ class ReportIssueButtonBase extends Component {
       configuredSubject ||
       intl.formatMessage({ id: 'components.TripTools.reportEmailSubject' })
     const bodyLines = [
-      intl.formatMessage({ id: 'components.TripTools.reportEmailTemplate' }),
+      intl.formatMessage(
+        { id: 'components.TripTools.reportEmailTemplate' },
+        { br: '\n' }
+      ),
       '',
       // Search data section is for support and is not translated.
       'SEARCH DATA:',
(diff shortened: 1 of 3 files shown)


lib/components/narrative/trip-tools.js
```diff
@@ -222,9 +151,74 @@ class LinkButton extends Component {
   }
 }
```
daniel-heppner-ibigroup, line 153: This feels like it would work well as a functional component.

lib/components/narrative/trip-tools.js
```diff
@@ -222,9 +151,74 @@ class LinkButton extends Component {
...
+class TripTools extends Component {
+  static defaultProps = {
+    buttonTypes: ['COPY_URL', 'PRINT', 'REPORT_ISSUE', 'START_OVER']
+  }
+
+  render() {
+    const { buttonTypes, reactRouterConfig, reportConfig } = this.props
+
+    const buttonComponents = []
+    buttonTypes.forEach((type) => {
```
daniel-heppner-ibigroup, line 169: How about a map to map buttonTypes to buttonComponents?

daniel-heppner-ibigroup (approved):
Some code structure thoughts... but still approving since my opinion might not be the best here!

miles-grant-ibigroup:
You're definitely right on both fronts and I think we should do both of these fixes when we move this file to Typescript, hopefully soon!

PR #539, opened 2022-02-03
refactor(call-taker): make plan button more prominent

daniel-heppner-ibigroup:
This makes the Calltaker UI plan button more prominent and puts it on its own line. 
image

### DIFF
--- a/lib/components/app/call-taker-panel.js
+++ b/lib/components/app/call-taker-panel.js
@@ -230,8 +230,6 @@ class CallTakerPanel extends Component {
           </div>
           <div className="search-plan-button-container">
             <Button
-              bsSize="default"
-              bsStyle="primary"
               onClick={this._planTrip}
               style={{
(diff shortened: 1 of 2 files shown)


miles-grant-ibigroup (approved):
This looks so much better thank you for doing this!

lib/components/app/call-taker-panel.js
```diff
@@ -227,13 +227,15 @@ class CallTakerPanel extends Component {
                 this.setState({ transitModes })
               }}
             />
+          </div>
+          <div className="search-plan-button-container">
             <Button
-              bsSize="small"
-              bsStyle="default"
+              bsSize="default"
```
binh-dam-ibigroup, line 233: Is this needed if we just say "default"?

binh-dam-ibigroup (approved):
LGTM, maybe one prop on the Plan button could be removed.

PR #541, opened 2022-02-04
feat(DefaultItinerary): show CO2 for each trip option

daniel-heppner-ibigroup:
TODO: Update core-utils version from https://github.com/opentripplanner/otp-ui/pull/346

This update adds CO2 emissions to the default itinerary view, and some new configuration options associated with it. Ensure you have https://github.com/opentripplanner/otp-ui/pull/346 in your modules folder!


example-config.yml
```diff
@@ -332,6 +332,27 @@ itinerary:
 #       - WALK
 #       - BICYCLE
 
+### Configure here the carbon output in grams per meter for each mode.
+### massUnits can be 'pound', 'kilogram', 'ounce', or 'gram'
+# co2:
+#   carbonIntensity:
+#     walk: 0
+#     bicycle: 0
```
miles-grant-ibigroup, line 340: Seeing NaN for bicycle_rent -- we probably should have a better null check and also include a value for bike_rent 

Screen Shot 2022-02-09 at 7 48 55 AM

lib/components/narrative/default/default-itinerary.js
```diff
@@ -148,12 +143,36 @@ const ITINERARY_ATTRIBUTES = [
         // FIXME: For CAR mode, walk time considers driving time.
         <>
           <FormattedDuration duration={itinerary.walkTime} />
-          <LegIconWrapper>
+          <LegIconWrapper noSpace>
```
miles-grant-ibigroup, line 146: Why do we need the noSpace? The icon looks the same as the others

lib/components/narrative/default/default-itinerary.js
```diff
@@ -148,12 +143,36 @@ const ITINERARY_ATTRIBUTES = [
...
           </LegIconWrapper>
         </>
       )
     }
+  },
+  {
+    id: 'carbonEmission',
+    order: 4,
+    render: (itinerary, { co2Config }) => {
+      if(!co2Config) return // If there is no CO2 Config, don't display this element
```
miles-grant-ibigroup, line 157: Would it be possible to add the feature we were talking about where the value is only shown if there is a significant difference (and then making the results really green and really red perhaps based on how big of a difference there is)? These rows are getting really tall... and 15oz vs 14oz is not always the most helpful info...

lib/components/narrative/default/default-itinerary.js
```diff
@@ -148,12 +143,36 @@ const ITINERARY_ATTRIBUTES = [
...
+
+      const { carbonIntensity, massUnit } = co2Config
+      const co2 = calculateEmissions(carbonIntensity, massUnit, itinerary)
+      return (
+        <>
+          <FormattedNumber
```
miles-grant-ibigroup, line 163: Should we maybe consider adding some rounding? The nearest 15 or 30 or even 50 might make the values a bit more glancable

example-config.yml
```diff
@@ -332,6 +332,27 @@ itinerary:
 #       - WALK
 #       - BICYCLE
 
+### Configure here the carbon output in grams per meter for each mode.
+### massUnits can be 'pound', 'kilogram', 'ounce', or 'gram'
```
miles-grant-ibigroup, line 336: I would move all the carbonIntensity values to core-utils and then make massUnit the only valid setting and then we can use typescript to enforce these options

miles-grant-ibigroup (changes requested):
A good start! I love the leaf. Just a few small implementation questions. Also, I'm not seeing the values in tabbed mode -- am I missing something?

example-config.yml
daniel-heppner-ibigroup: You don't think the agencies might want to customize the carbon intensity of each mode?

example-config.yml
daniel-heppner-ibigroup: I thought I fixed that!

lib/components/narrative/default/default-itinerary.js
daniel-heppner-ibigroup: There was a margin around LegIconWrapper and I wasn't sure why, so instead of removing it I just added this prop.

lib/components/narrative/default/default-itinerary.js
daniel-heppner-ibigroup: We have the percentages/relative values, as well as some rounding and hiding values below 10%. Does this feel like enough? The design is not great, but it really just needs a complete redesign. I think adding bright colors would be a bit too much.

example-config.yml
```diff
@@ -332,6 +332,27 @@ itinerary:
...
+# co2:
```
miles-grant-ibigroup, line 337: we should add `enabled: true` here!

i18n/en-US.yml
```diff
@@ -767,6 +767,12 @@ common:
     calories: "{calories, number} Cal"
     noItineraryToDisplay: No itinerary to display.
     transfers: "{transfers, plural, =0 {} one {# transfer} other {# transfers}}"
+    # This string will have a formatted number before it, including units.
+    relativeCo2: >
+     {isMore, select,
+      true {more}
+      other {less}
+      } CO2 than driving alone.
```
miles-grant-ibigroup, line 775: I would shorten this string so that the time isn't pushed awkwardly to the middle. Maybe `XX% of car emissions`? Either that or we should figure out a way to force a line break

lib/components/narrative/default/default-itinerary.js
```diff
@@ -153,12 +148,41 @@ const ITINERARY_ATTRIBUTES = [
...
     }
+  },
+  {
+    id: 'carbonEmission',
+    order: 4,
+    render: (itinerary, { co2VsBaseline, hideCo2IfHigher }) => {
+      if (!co2VsBaseline) return 
+      const ROUND_TO_NEAREST = 5
+      co2VsBaseline = Math.round(co2VsBaseline * 100 / ROUND_TO_NEAREST)*ROUND_TO_NEAREST
+      if(Math.abs(co2VsBaseline) <= 10) return // Only show greater than 10%
```
miles-grant-ibigroup, line 165: Is this auto formatted?

lib/components/narrative/narrative-itineraries.js
```diff
@@ -152,11 +180,27 @@ class NarrativeItineraries extends Component {
...
 
+    const baselineCo2 = this._getBaselineCo2()
+
+    const itinerariesWithCo2 =
+      itineraries?.map((itin) => {
+        const emissions = coreUtils.itinerary.calculateEmissions(
+          itin,
+          co2Config?.carbonIntensity
+        )
+        console.log('Emissions', emissions)
```
miles-grant-ibigroup, line 191: oh no looks like this snuck in!

i18n/en-US.yml
miles-grant-ibigroup: We should also be careful with language. if a car emits 100 grams, and we say 85% less co2 then isn't that 15 grams? is that what we mean or do we mean 85 grams

miles-grant-ibigroup (changes requested):
Looking really good well done! As much as I think we should re-work the batch itinerary list I think having the time pushed to the middle is problematic which is the main thing.

Also, I'm not seeing the co2 results in trip-details even though I've updated trip-details to your version. Not sure what's going on there!

lib/components/narrative/default/default-itinerary.js
daniel-heppner-ibigroup: supposedly it was... I added some spaces

i18n/en-US.yml
daniel-heppner-ibigroup: I think having the word "less" and "more" is pretty important to make it clear to understand at a glance.

daniel-heppner-ibigroup: OK I aligned all the attributes to the right. Pretty big change, but I think I do like it more. Let me know what you think.

miles-grant-ibigroup: This definitely looks better thanks! Feels a bit busy but we can probably leave that for the redesign

lib/components/narrative/default/default-itinerary.js
```diff
@@ -316,7 +347,7 @@ class DefaultItinerary extends NarrativeItinerary {
 const mapStateToProps = (state, ownProps) => {
   const { intl } = ownProps
   const gradationMap = state.otp.config.accessibilityScore?.gradationMap
-
+  const co2Config = state.otp.config.co2
```
miles-grant-ibigroup, line 350: Optional chaining is probably needed here!

miles-grant-ibigroup (changes requested):
Looking really good, although I'm going go request changes since I'm seeing some white screen of death errors with the TripDetails. I guess we have to leave this blocked for now

lib/components/narrative/default/default-itinerary.js
daniel-heppner-ibigroup: Hmmm... I don't think so. If CO2 was null then it would be crashing on the previous line, right?

miles-grant-ibigroup: Are we ever checking for the `co2` field of `config`?

daniel-heppner-ibigroup: It should be checked everywhere that it's used. You did catch a place in the other PR that it wasn't being checked, though, so maybe that was causing the error? That's fixed now.

miles-grant-ibigroup:
I think I found the reason the Percy tests are failing: Screen Shot 2022-08-15 at 7 18 33 AM

As soon as I plan a route the app crashes. My guess is it has something to do with mismatched otp-ui versions? I tried upgrading a few to their latest but that only changed the message I got. I think @daniel-heppner-ibigroup you know better than me which package versions are needed.

example-config.yml
```diff
@@ -374,6 +374,28 @@ itinerary:
...
```
miles-grant-ibigroup, line 379: Are these still accurate? I'm not seeing them have an effect on the results. Additionally, the defaults seem a bit strange: for a certain itinerary I'm getting that walking is 85% of driving alone, but scooter is 100% less? That doesn't make any sense

lib/components/narrative/default/default-itinerary.js
```diff
@@ -132,12 +128,43 @@ const ITINERARY_ATTRIBUTES = [
...
+  },
+  {
+    id: 'carbonEmission',
+    order: 4,
+    render: (itinerary, { co2VsBaseline }) => {
+      if (!co2VsBaseline) return
+      const ROUND_TO_NEAREST = 5
+      co2VsBaseline =
+        Math.round((co2VsBaseline * 100) / ROUND_TO_NEAREST) * ROUND_TO_NEAREST
+      if (Math.abs(co2VsBaseline) <= 10) return // Only show greater than 10%
```
miles-grant-ibigroup, line 146: Should this maybe be configurable? I would also set this to something like 33% or something so it doesn't show up as often

lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -245,6 +259,26 @@ class MetroItinerary<Props> extends NarrativeItinerary {
...
+    const co2VsBaseline = Math.round(itinerary.co2VsBaseline * 100)
+    const emissionsNote = !mini && co2VsBaseline < -20 && (
+      <>
+        <ItineraryNoteIcon type="leaf" />
+        <FormattedNumber
+          style="unit"
+          unit="percent"
+          unitDisplay="narrow"
+          value={Math.abs(co2VsBaseline)}
+        />{' '}
```
miles-grant-ibigroup, line 271: I don't like the `{' '}`. Would it be possible to insert the number as a value inside the message and then pass this entire component in as the value?

lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -245,6 +259,26 @@ class MetroItinerary<Props> extends NarrativeItinerary {
       currency
     )
 
+    const co2VsBaseline = Math.round(itinerary.co2VsBaseline * 100)
+    const emissionsNote = !mini && co2VsBaseline < -20 && (
```
miles-grant-ibigroup, line 263: Is this 20 related to the threshold we saw previously?

miles-grant-ibigroup (changes requested):
Looks really slick! A few small questions but I think this is almost good to go!

Also, this seems to not work with the "old" batch ui or the tabbed UI. Is this intentional?

example-config.yml
daniel-heppner-ibigroup: The mass units should show up in the trip details, where the actual mass of the CO2 is shown. It won't have any effect on the percentages, of course.

lib/components/narrative/default/default-itinerary.js
daniel-heppner-ibigroup: Done!

lib/components/narrative/metro/metro-itinerary.tsx
daniel-heppner-ibigroup: uhhh maybe. it's gone now, uses the config!

daniel-heppner-ibigroup:
You should be able to see this on the old batch UI! 
image

example-config.yml
daniel-heppner-ibigroup: Hm there does appear to be an issue where the units are not changing. I think this is a problem on the OTP-UI side, since the entire CO2 config just gets passed into TripDetails.

daniel-heppner-ibigroup: And finally, can you give me an example itinerary where you're seeing the wacky numbers? That sounds weird.

miles-grant-ibigroup: I'm seeing it on all e-scooter share itineraries. My guess is it may be related to the name changing between `SCOOTER` to `MICROMOBILITY_RENT`? In that case I think our fallback should be infinite carbon rather than 0 carbon

miles-grant-ibigroup: I've now discovered this was a configuration issue. Thanks for the new documentation!

example-config.yml
```diff
@@ -374,6 +374,30 @@ itinerary:
...
+#     ferry: 0.082
+#     cable_car: 0.021
+#     gondola: 0.021
+#     funicular: 0.041
+#     transit: 0.041
+#     leg_switch: 0
+#     airplane: 0.382
+#     micromobility: 0
+#   massUnit: "ounce"
+#   showCo2IfHigher: false // show the relative CO2 even if it's higher than driving
```
miles-grant-ibigroup, line 398: Is it possible to rename this to `showIfHigher`? Since it's already a child of `co2`

example-config.yml
```diff
@@ -374,6 +374,30 @@ itinerary:
...
+#   co2Cutoff: 33 // Only show the CO2 comparison if the % difference from driving is at least this much
```
miles-grant-ibigroup, line 399: Is it possible to rename this to `cutoff`? Since it's already a child of `co2`

lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -245,6 +260,30 @@ class MetroItinerary<Props> extends NarrativeItinerary {
       currency
     )
 
+    const roundedCo2VsBaseline = Math.round(itinerary.co2VsBaseline * 100)
+    const emissionsNote = !mini &&
+      Math.abs(roundedCo2VsBaseline) >= (co2Config?.co2Cutoff || 0) &&
+      co2Config?.enabled && (
```
miles-grant-ibigroup, line 266: Bit of a nitpick, but the `enabled` check can probably come first

miles-grant-ibigroup: Additionally, do we need an explicit `enabled` flag? Can't we just use the presence of one or more `carbonIntensity` keys?

miles-grant-ibigroup (approved):
A few small nitpicks, but overall I think this looks great! Thanks so much for all your work on this I know it's been a long process

lib/components/narrative/metro/metro-itinerary.tsx
daniel-heppner-ibigroup: Not really, because someone (probably most people) will just want to use the defaults. Technically we could check if the `co2` property exists in the config but I think that would look a little weird. You'd have the property like `co2:` in the yaml all on its own with nothing under it.

miles-grant-ibigroup: That's a good point. I'll leave it up to the second reviewer to decide on this!

demory:
Quick question: am I supposed to see CO2 numbers in the trip summary for walk/bike alone trips? I see the "X% lower than driving" bar at the top of these itineraries but nothing in the summary.

daniel-heppner-ibigroup:
@demory The problem appeared to be the outdated `trip-details` package. Not sure why that never got updated, but I've changed it to the latest version which seems to have fixed the issue!

daniel-heppner-ibigroup:
@miles-grant-ibigroup Do you know why all the times switched from AM to PM in the percy snapshots?

miles-grant-ibigroup:
@daniel-heppner-ibigroup The am/pm thing is a really old bug from when we were adjusting date formatting! If you merge latest dev into this branch it should go away.

demory:
@daniel-heppner-ibigroup to recap what we discussed on our last call:
- I am still seeing the issue with no CO2 data in the trip details box for walk/bike only trips. The bug you addressed above appears to be different
- We talked about not showing CO2 at all for those trips (both the header bar and the details box entry), or having that be configurable

daniel-heppner-ibigroup:
I've made some changes to iron out a bunch of bugs and also improve the flow of data. This requires a change to OTP-UI as well, which I've done in a separate PR here. https://github.com/opentripplanner/otp-ui/pull/461
We will need to update the version numbers in this PR once that one is merged in.

daniel-heppner-ibigroup:
@demory This is ready for a review by the way!

demory:
Just need to implement `showIfHigher` in the metro UI and this should be good to go

daniel-heppner-ibigroup:
Good to go, @demory. I fixed up the issues with `showIfHigher`

demory (approved):

PR #542, opened 2022-02-04
Update issue reporting template qbd

binh-dam-ibigroup:
Simplify changes from #538 and update French file.


miles-grant-ibigroup (approved):
This is a great improvement thanks!

Issue #543, opened 2022-02-07
Map rendering errors out randomly.


binh-dam-ibigroup:
I'm getting some errors while making a new account running FDOT. Are these related to this PR? There are a few errors so I'll just put the steps to recreate.
On this URL: http://localhost:9966/#/account/create/terms after creating an account and verifying email.
![image](https://user-images.githubusercontent.com/92878779/152869769-2e64fae2-9b29-4518-970b-835c03c3581b.png)
If this is not due to this PR, I won't block the PR with it. 
EDIT: this is not related to this PR

_Originally posted by @daniel-heppner-ibigroup in https://github.com/opentripplanner/otp-react-redux/issues/329#issuecomment-1031914559_

miles-grant-ibigroup:
Is this possibly related to https://github.com/opentripplanner/otp-react-redux/issues/457?

Issue #547, opened 2022-02-15
Bounds being set to invalid values when switching between itinerary and route viewer


miles-grant-ibigroup:
Steps to reproduce: 
- go to itinerary viewer and plan a trip
- open route viewer and select a route, then a pattern

<img width="1439" alt="Screen Shot 2022-02-15 at 1 08 11 PM" src="https://user-images.githubusercontent.com/86619099/154149544-35081053-a134-4fcd-92e9-244b38babc76.png">
 
This error is likely also related to selecting a route sometimes not updating the active bounds.

PR #548, opened 2022-02-16
fix(route-viewer): don’t zoom to itinerary when route viewer is active

miles-grant-ibigroup:
There were a few regressions brought in over the past few releases that caused some problems with the route viewer zooming to the correct route. All these bugs have been fixed. This should also resolve some "white screen of death" issues we've been seeing in relation to the route viewer.

This PR closes https://github.com/opentripplanner/otp-react-redux/issues/543 and https://github.com/opentripplanner/otp-react-redux/issues/547

### DIFF
--- a/lib/actions/api.js
+++ b/lib/actions/api.js
@@ -1030,6 +1030,7 @@ export function findGeometryForPattern(params) {
         findGeometryForPatternResponse,
         findGeometryForPatternError,
         {
+          noThrottle: true,
           rewritePayload: (payload) => ({
             geometry: payload,
             patternId,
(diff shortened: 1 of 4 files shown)


miles-grant-ibigroup:
Good catch! Thanks for the detailed error report, I'll be sure to put that into an issue.

This issue is resolved by https://github.com/opentripplanner/otp-react-redux/pull/560/commits/d9e1f47334f05c0d9db0a87225595cb46cfa0697

demory (approved):
Confirmed that this is working (tested along w/ the rest of PR #560)

philip-cline (approved):

PR #549, opened 2022-02-18
Add config param to disable flex arc display

philip-cline:
…display

This PR makes use of a configuration parameter (which would be put under the transitive section) to disable displaying a flex route arc (as requested by NYSDOT). 

Relies on OPT-UI PR: https://github.com/opentripplanner/otp-ui/pull/352

### DIFF
--- a/example-config.yml
+++ b/example-config.yml
@@ -112,6 +112,7 @@ map:
   ###   styles.segment_labels: styles attributes recognized by transitive.js.
   ###                         For examples of applicable style attributes, see
   ###                         https://github.com/conveyal/transitive.js/blob/master/stories/Transitive.stories.js#L47.
+  ### - disableFlexArc: optional parameter to disable rendering flex itinerary legs as an arc. 
   # transitive:
   #   labeledModes:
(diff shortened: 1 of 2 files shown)


example-config.yml
```diff
@@ -112,6 +112,7 @@ map:
   ###   styles.segment_labels: styles attributes recognized by transitive.js.
   ###                         For examples of applicable style attributes, see
   ###                         https://github.com/conveyal/transitive.js/blob/master/stories/Transitive.stories.js#L47.
+  ### - disableFlexArc: optional parameter to disable rendering flex itinerary legs as an arc.
```
miles-grant-ibigroup, line 115: wait are there 2 ways to enable this?

miles-grant-ibigroup (changes requested):
Looks good, nice and clean!

Just one quick question about a possible error in the example config

miles-grant-ibigroup (approved):
Looking great! Sorry for the delay on this

daniel-heppner-ibigroup (approved):

PR #554, opened 2022-02-23
fix(formatted-date-time-preview): don't crash on missing route mode

miles-grant-ibigroup:
Avoids crash on missing route mode.


philip-cline (approved):
Hey nice this was my fault from working on this component. Thanks for the fix :)

binh-dam-ibigroup (approved):

PR #556, opened 2022-02-25
Fix schedule table - addon

binh-dam-ibigroup:
This PR uses existing behavior in the Stop Viewer to load the full schedule for a day when switching to the schedule view.
Also included are some layout tweaks.


miles-grant-ibigroup (approved):
Woah very good catch on finding we had that method! This looks so clean thank you

PR #560, opened 2022-02-28
Remove support for OTP1/2 in parallel

miles-grant-ibigroup:
This PR removes the buggy support for making OTP1 and OTP2 requests in parallel.

In its place this PR provides full support for both OTP1 and OTP2 instances on their own.

V2 instances are now noted with adding a `v2: true` field in the api block. `api_v2` is no longer supported.

Closes https://github.com/opentripplanner/otp-react-redux/issues/563

Blocked by https://github.com/opentripplanner/otp-react-redux/pull/548, https://github.com/opentripplanner/otp-react-redux/pull/552

This PR actually wound up touching quite a lot of files.
### Things I'm looking for feedback on:
- As the primary effort here was to get things stable and to achieve feature parity, some architecture decisions are a little unclean. If you have good ideas on how to re-structure some of the api methods, these suggestions will be welcomed! However, please do keep in mind that the long-term goal is to move the entire api module to something far more modular like Apollo, removing the need for any of these methods
- A lot of work was done to match OTP1 response types as much as possible. If you have any ideas on how to remove the explicit re-writing and instead change downstream components to be more lenient in a way that is shorter than the current re-writing code, I would love that!
- OTP2 is missing some nice-to-haves that OTP1 currently enjoys, but I think the two are paired up enough for now (we can make the touch-ups in the future)
- Unrelated to this PR, but `STOPS_IN_BBOX` is firing a **lot**. I was not able to find an easy fix to include here


lib/actions/apiV1.js
```diff
@@ -0,0 +1,233 @@
...
+      serviceId: 'stops'
+    }
+  )
+}
+
+const fetchStopInfo = (stop) =>
+  async function (dispatch, getState) {
+    await dispatch(findStop({ stopId: stop.stopId }))
+    const state = getState()
+    const { nearbyRadius: radius } = state.otp.config.stopViewer
```
binh-dam-ibigroup, line 132: Move to just above `if (radius ...)`. What is the fallback if `stopViewer` is not provided in the config?

lib/actions/apiV1.js
```diff
@@ -0,0 +1,233 @@
...
+            }
+          }
+        }
+      )
+    )
+  }
+}
+
+export function findStop(params) {
+  return createQueryAction(
```
binh-dam-ibigroup, line 113: I kinda like the `const ... => createQueryAction(...)` notation (multiple instances).

lib/actions/apiV2.js
```diff
@@ -0,0 +1,533 @@
...
+      rewritePayload: (payload) => {
+        if (!payload?.data?.trip) return {}
+
+        payload.data.trip.geometry = payload.data.trip.tripGeometry
+        return payload.data.trip
+      }
+    }
+  )
+
+export function vehicleRentalQuery(
```
binh-dam-ibigroup, line 108: I kinda like the notation `const vehicleRentalQuery... => createGraphQLQueryAction(...)` (multiple instances).

lib/actions/apiV2.js
```diff
@@ -0,0 +1,533 @@
...
+                desc: pattern.name,
+                geometry: pattern.patternGeometry,
+                stops: patternStops
+              }
+            })
+            route.patterns = routePatterns
+            // TODO: avoid explicit behavior shift like this
+            route.v2 = true
+
+            return route
```
binh-dam-ibigroup, line 503: In that case it's better to create a shallow copy of `payload.data` that gets modified (like in the other methods above) or, perhaps more preferably, perform a `clone`.

binh-dam-ibigroup (changes requested):
Thanks for this PR, it is much more pleasant to read! I still need to play around more, but the one big change I need is to compute/initialize the correct version of the `actions` variable either using a function and maybe also using a redux state.

binh-dam-ibigroup:
API-wise, things seem to work!

lib/actions/form.js
```diff
@@ -93,18 +95,22 @@ export function setQueryParam (payload, searchId) {
  *                         query made during a call taker session, to prevent
  *                         a search from being added to the current call)
  */
-export function parseUrlQueryString (params = getUrlParams(), source) {
+export function parseUrlQueryString(params = getUrlParams(), source) {
   return function (dispatch, getState) {
+    const state = getState()
+    if (state.otp.ui.mainPanelContent !== null) return state
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Our void returns were causing crashes! It turns out redux requires that each action returns the state object even if no changes are made.

lib/actions/ui.js
```diff
@@ -172,7 +173,12 @@ export function matchContentToUrl(location) {
     switch (root) {
       case 'route':
         if (id) {
-          dispatch(findRoute({ routeId: id }))
+          // This is a bit of a hack to check if the route details have been grabbed
+          // agencyName will only be populated if route details are repsent
+          // Moving away from manual requests should help resolve this
+          if (!state.otp.transitIndex?.routes?.[id]?.agencyName) {
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Good call! Will use `bikesAllowed` instead which if unset in gtfs gets returned as `0`

lib/reducers/create-otp-reducer.js
```diff
@@ -659,7 +659,7 @@ function createOtpReducer(config) {
           }
         })
       case 'SET_VIEWED_STOP':
-        if (action.payload) {
+        if (action?.payload !== null) {
```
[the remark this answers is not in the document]
miles-grant-ibigroup: This action is sometimes called with `undefined` as the value. The explicit check avoids crashes

lib/reducers/create-otp-reducer.js
```diff
@@ -715,18 +715,22 @@ function createOtpReducer(config) {
...
           }
         })
       case 'FIND_STOPS_FOR_PATTERN_RESPONSE':
+        const patternStops = action?.payload?.stops || []
+
         return update(state, {
           transitIndex: {
             routes: {
-              [action.payload.routeId]: {
+              [action.payload.routeId || 'null']: {
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Good call! Added

lib/actions/api.js
```diff
@@ -15,17 +14,20 @@ import { convertToPlace, getPersistenceMode } from '../util/user'
...
+import v2Actions from './apiV2'
 
 if (typeof fetch === 'undefined') require('isomorphic-fetch')
 
 const { getRoutingParams, getTripOptionsFromQuery, getUrlParams } =
   coreUtils.query
 const { randId } = coreUtils.storage
 
+// set actions variable that can be set to v2Actions if v2 server is set
+let actions = v1Actions
```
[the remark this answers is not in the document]
miles-grant-ibigroup: That's a good way to approach it thanks! Agreed it's clean for now and we can re-evaluate in the future.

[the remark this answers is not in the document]
miles-grant-ibigroup: Done in e0bc80dc3dcc9b84645df4d5399b03c8de1788c9!

lib/actions/api.js
```diff
@@ -245,16 +247,14 @@ function getJsonAndCheckResponse(res) {
 function getOtpFetchOptions(state, includeToken = false) {
   let apiBaseUrl, apiKey, token
 
-  const { api: apiV1, api_v2: apiV2, persistence } = state.otp.config
+  const { api, persistence } = state.otp.config
+  if (api.v2) actions = v2Actions
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Refactored in e0bc80dc3dcc9b84645df4d5399b03c8de1788c9

lib/actions/api.js
```diff
@@ -768,241 +530,35 @@ export function findRoutes(params) {
...
-            }
-          }
-        )
-      )
-    }
+    const { api } = state.otp.config
+    // Because findRoute is sometimes called from strange places,
+    // we must explicitly check for V2 as we can not rely on
+    // the check having already happened
+    if (api.v2) actions = v2Actions
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Refactored in e0bc80dc3dcc9b84645df4d5399b03c8de1788c9

lib/actions/api.js
```diff
@@ -1015,30 +571,21 @@ const findGeometryForPatternError = createAction(
...
-        {
-          rewritePayload: (payload) => ({
-            geometry: payload,
-            patternId,
-            routeId
-          })
+  return createQueryAction(
+    `index/patterns/${params.patternId}/geometry`,
+    findGeometryForPatternResponse,
+    findGeometryForPatternError,
```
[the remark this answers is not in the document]
miles-grant-ibigroup: I agree this could use work, but I think our energy here should be more focused on a full graphQL implementation can remove the need for any of these actions at all (and rather be one single action that can request any variable needed)

lib/actions/apiV1.js
miles-grant-ibigroup: Copy pasted this from the old `api.js` but good catch with the `stopViewer` being undefined! I think all methods called in this action have radius defaults

lib/actions/apiV1.js
miles-grant-ibigroup: Me too!

lib/actions/apiV2.js
miles-grant-ibigroup: Good call thanks! Addressed in d2507aa982ecca23a82fb1964c5e7a7099b81d88

lib/actions/apiV2.js
miles-grant-ibigroup: Refactored in d2507aa982ecca23a82fb1964c5e7a7099b81d88

lib/actions/apiV1.js
miles-grant-ibigroup: Refactored in d2507aa982ecca23a82fb1964c5e7a7099b81d88

lib/actions/apiV1.js
miles-grant-ibigroup: Refactored this a bit in d2507aa982ecca23a82fb1964c5e7a7099b81d88

miles-grant-ibigroup:
Thanks for the suggestions!

lib/actions/api.js
```diff
@@ -91,6 +88,24 @@ function getActiveItinerary(state) {
   return activeItinerary
 }
 
+/**
+ * Dispatches a method from either v1actions or v2actions, depending on
+ * which version of OTP is specified in the config.
+ * @param {*} methodName    the method to execute
+ * @param  {...any} params  varargs of params to send to the action
+ */
+function executeOTPVersionSpecificAction(methodName, ...params) {
```
binh-dam-ibigroup, line 97: Pretty cool! Maybe you can rename to just `executeOTPAction`.

lib/actions/apiV2.js
```diff
@@ -496,20 +503,19 @@ export function findRoute(params) {
                 stops: patternStops
               }
             })
-            route.patterns = routePatterns
+            newRoute.patterns = routePatterns
             // TODO: avoid explicit behavior shift like this
```
binh-dam-ibigroup, line 507: Can this comment be removed now?

lib/actions/ui.js
```diff
@@ -174,9 +174,9 @@ export function matchContentToUrl(location) {
       case 'route':
         if (id) {
           // This is a bit of a hack to check if the route details have been grabbed
-          // agencyName will only be populated if route details are repsent
+          // bikesAllowed will only be populated if route details are repsent
```
binh-dam-ibigroup, line 177: Typo "~repsent~" => "present"

binh-dam-ibigroup: Not sure about `bikesAllowed` either - that field is not populated by many Atlanta agencies.

binh-dam-ibigroup: Would the presence of at least one pattern be a more reliable criterion?

binh-dam-ibigroup (commented):
The refactor is nice! I am wondering about using `bikesAllowed` to detect downloading route details.

lib/actions/api.js
miles-grant-ibigroup: We sure can! Good suggestion thank you

lib/actions/apiV2.js
miles-grant-ibigroup: I wish... the problem for me is the `v2` field which I don't like. I think ideally the rest of OTP shouldn't know if it's dealing with OTP1 or OTP2

lib/actions/ui.js
miles-grant-ibigroup: Good catch thank you! and `bikesAllowed` feels strong to me because even if it's not populated OTP still sets it to 0 which allows us to do this null check

miles-grant-ibigroup: One pattern was what I initially had, but ran into problems where sometimes routes wouldn't have patterns (?) although the full route was downloaded.

miles-grant-ibigroup:
Thanks Binh, changes are addressed

lib/actions/api.js
miles-grant-ibigroup: Renamed in a500cbe779f68c61c9e9739c30b640641fdb2b1a

lib/actions/ui.js
miles-grant-ibigroup: Typo fixed in a500cbe779f68c61c9e9739c30b640641fdb2b1a

binh-dam-ibigroup (approved):

demory:
Have you seen this error when loading the stop viewer (either via an itinerary or a map popup) when running against OTP2? I'm seeing it consistently. Works fine w/ OTP1.

image

miles-grant-ibigroup:
Very good catch thank you! That's a regression that came in when I moved the nearby endpoint to OTP2. Fixed in e8ff07d4c60bdb564eeb79b4e4fb28486c1da907

demory:
Thanks Miles. I've been testing w/ the latest changes and now am seeing a couple other OTP2-specific UI issues.

1. In the stop viewer, any departure time for the "related stops" always shows as 12:00am:

image

2. Navigating from the trip viewer to the stop viewer using one of the stop-specific "View" buttons causes a crash:

image

(Also, I'm getting some odd OTP2 results using the ATL Rides instance, e.g. no transit trips when I definitely expect them. But that could be an OTP issue, need to investigate further.)

miles-grant-ibigroup:
Thanks @demory for spotting those regressions! Things should be back to stability in 24adf0b688525050b8c29e7cd68a1d1248698679

miles-grant-ibigroup:
@demory midnight bus departures are resolved in 1633f44c8867253ba13bd0d9c0aed65a4043da92, the stops layer not working error was actually not a client-side issue but a server side issue and one that I've fixed [here](https://github.com/ibi-group/OpenTripPlanner/pull/99). I do still think moving to graphQL for those requests is the way to go, but that can be a separate PR. I'm also taking a look at the snapping, but will push any fixes for that to a new PR. 

Please let me know if there are any other blocking changes!

demory (approved):

PR #562, opened 2022-03-02
Hide zero fares

philip-cline:
This PR ignores fares less than 0.01 in default-itinerary (for situations without GTFS fares provided). 

Blocked by #560 (builds on this PR)

Note for reviewer: 
One awkward situation this can produce is shared fares with bicycling & transit (will display bike fare + 0.00 fare for bus).


miles-grant-ibigroup (approved):
It does look a bit awkward with free modes not showing anything, but since we're redesigning the batch ui soon I think this is alright

PR #564, opened 2022-03-08
Use Stop Code as Stop ID where Applicable

miles-grant-ibigroup:
This PR uses stop ID instead of stop code when it makes sense to. This makes the stop ID more useful for users.


daniel-heppner-ibigroup (approved):
Sorry, missed that this had lost the blocked label!

PR #567, opened 2022-03-10
Remove flags from LocaleSelector

philip-cline:
This PR removes flags from the `LocaleSelector` component. Flags are replaced by the language name, which is bolded if selected (checkbox removed). 

A small improvement to the locale state update is made as well, which only updates the state with a matched (supported) locale. This allows bolding of the used language on initialization, as well as preventing a bug where custom messages are not imported before language selection. 

A related PR (https://github.com/ibi-group/configurations/pull/234) removes the flags from the ATL-RIDES QA config

### DIFF
--- a/lib/actions/ui.js
+++ b/lib/actions/ui.js
@@ -344,7 +344,11 @@ export function setLocale(locale) {
     const { language: customMessages } = config
     const configLocales = getConfigLocales(customMessages)
     const effectiveLocale = locale || getDefaultLocale(config, loggedInUser)
-    const matchedLocale = getMatchingLocaleString(effectiveLocale, 'en-US')
+    const matchedLocale = getMatchingLocaleString(
+      effectiveLocale,
+      'en-US',
+      configLocales
+    )
     const messages = await loadLocaleData(
       matchedLocale,
       customMessages,
--- a/lib/actions/ui.js
+++ b/lib/actions/ui.js
@@ -3,7 +3,12 @@ import { matchPath } from 'react-router'
 import { push } from 'connected-react-router'
 import coreUtils from '@opentripplanner/core-utils'
 
-import { getDefaultLocale, loadLocaleData } from '../util/i18n'
+import {
+  getConfigLocales,
+  getDefaultLocale,
+  getMatchingLocaleString,
+  loadLocaleData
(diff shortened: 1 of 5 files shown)


lib/actions/ui.js
```diff
@@ -336,13 +341,18 @@ export function setLocale(locale) {
   return async function (dispatch, getState) {
     const { config } = getState().otp
     const { loggedInUser } = getState().user
-
     const { language: customMessages } = config
+    const configLocales = getConfigLocales(customMessages)
     const effectiveLocale = locale || getDefaultLocale(config, loggedInUser)
-    const messages = await loadLocaleData(effectiveLocale, customMessages)
+    const matchedLocale = getMatchingLocaleString(effectiveLocale, 'en-US')
```
binh-dam-ibigroup, line 347: Not sure what happened here, the locale/language switching does not seem to work.
Also, using the URL to switch the locale (e.g. by appending `?locale=fr`) no longer works.

binh-dam-ibigroup (changes requested):
Something broke along the way. Please advise regarding https://github.com/opentripplanner/otp-react-redux/pull/567/files#r824797351.

philip-cline:
Thanks, didn't push the change with passing configLocales to getMatchingLocaleString. This should be working now

binh-dam-ibigroup (approved):
Nice!

lib/components/app/locale-selector.tsx
```diff
@@ -65,20 +46,20 @@ const LocaleSelector = (props: LocaleSelectorProps): JSX.Element => {
     <NavDropdown
       id="locale-selector"
       pullRight
-      title={<CurrentLocaleFlag style={{ width: 15 }} />}
+      title={<Icon style={{ color: '#fff' }} type="globe" />}
```
miles-grant-ibigroup, line 49: Super nit, but could we make the color `color: rgba(255, 255, 255, 0.85);`? This lets some of the bg bleed through (while still being accessible!) and matches the view switcher

miles-grant-ibigroup (approved):
Looks good and works for me! Just one small nit about the color. I have feelings about the bold as well but I think that's more productive as part of an upcoming redesign

PR #569, opened 2022-03-10
Fix mode selector

daniel-heppner-ibigroup:


miles-grant-ibigroup:
We can merge dev into this now!

PR #570, opened 2022-03-10
Show date/time picker on focus on Chrome desktop

binh-dam-ibigroup:
This PR hacks the `<input type="date">` and `<input type="time">` components on the desktop versions of Chrome/Chromium, so that, if so configured, when tapping "~anywhere" on these controls on a desktop OS, even using a touchscreen, the date or time picker is displayed automatically.

Note: there is https://developer.chrome.com/blog/show-picker/ that I have considered, but requires a longer review cycle because it touches the OTP-UI repo.

### DIFF
--- a/example-config.yml
+++ b/example-config.yml
@@ -450,3 +450,7 @@ dateTime:
 # Mapillary calls these "Client Tokens". They can be created at https://www.mapillary.com/dashboard/developers
 # mapillary:
 #   key: <Mapillary Key>
+
+### Setting to enable touch-friendly behavior
+### e.g. on touch-screen kiosks that run a desktop OS.
+# isTouchScreenOnDesktop: true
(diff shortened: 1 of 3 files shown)


daniel-heppner-ibigroup (approved):
Is browser support important? It doesn't appear to work for the time picker in Safari, but it does work in Chrome. I am guessing based on your comment in the CSS that it's only for Chromium based browsers. Maybe add a similar comment on the example config?

binh-dam-ibigroup:
> Is browser support important? It doesn't appear to work for the time picker in Safari, but it does work in Chrome. I am guessing based on your comment in the CSS that it's only for Chromium based browsers. Maybe add a similar comment on the example config?

The change is specific to a particular config that uses Google Chrome on a desktop OS. Other configs are not affected.

binh-dam-ibigroup:
> Maybe add a similar comment on the example config?
> 

I picked a generic name for the config setting, as I guess eventually we want to include other browsers as/when they support the requested feature.

miles-grant-ibigroup (approved):
Since you mentioned that this is a bit of a hack fix we could maybe put a `console.warn` complaining about the use of the field? Although our console output is already quite noisy so maybe that's not worth it.

Glad we got this sorted this quickly!

PR #571, opened 2022-03-15
Add new flex message if both call-ahead and continuous service exist in itinerary

philip-cline:
Currently in the flex indicator, if an itinerary includes both call-ahead and continuous service, only one of the two messages will be provided in the `FlexIndicator`. This PR provides a new message, which directs a user to check out the `TripDetails` in order to see that the itinerary includes both call-ahead and continuous service. The `TripDetails` description for these type of flex service is new in this PR, and required adding some flex information to the leg pickup and drop off booking info to recognize the service as traditional flex. 

The PR also adds default flex messages in the en-US translations.

### DIFF
--- a/lib/components/narrative/line-itin/connected-itinerary-body.js
+++ b/lib/components/narrative/line-itin/connected-itinerary-body.js
@@ -62,11 +62,12 @@ class ConnectedItineraryBody extends Component {
     } = this.props
     const { LegIcon } = this.context
 
-    let [pickupAdded, dropOffAdded] = [false, false]
+    const clonedItinerary = clone(itinerary)
+    let pickupAdded = false
+    let dropOffAdded = false
 
     // Support OTP1 flex messages in Trip Details
-    itinerary.legs = itinerary.legs.map((oldLeg) => {
-      const leg = clone(oldLeg)
+    clonedItinerary.legs = clonedItinerary.legs.map((leg) => {
       if (!!leg.boardRule && !leg.pickupBookingInfo && !pickupAdded) {
         pickupAdded = true
         leg.pickupBookingInfo = {
@@ -92,7 +93,7 @@ class ConnectedItineraryBody extends Component {
           accessibilityScoreGradationMap={accessibilityScoreGradationMap}
           config={config}
           diagramVisible={diagramVisible}
-          itinerary={itinerary}
+          itinerary={clonedItinerary}
           LegIcon={LegIcon}
           LineColumnContent={LineColumnContent}
           mapillaryCallback={setMapillaryId}
@@ -114,8 +115,8 @@ class ConnectedItineraryBody extends Component {
           TransitLegSubheader={TransitLegSubheader}
           TransitLegSummary={TransitLegSummary}
         />
-        <TripDetails itinerary={itinerary} />
-        <TripTools itinerary={itinerary} />
+        <TripDetails itinerary={clonedItinerary} />
+        <TripTools itinerary={clonedItinerary} />
       </ItineraryBodyContainer>
     )
   }
--- a/lib/components/narrative/line-itin/connected-itinerary-body.js
+++ b/lib/components/narrative/line-itin/connected-itinerary-body.js
@@ -61,31 +61,35 @@ class ConnectedItineraryBody extends Component {
       timeOptions
     } = this.props
     const { LegIcon } = this.context
-
     const clonedItinerary = clone(itinerary)
-    let pickupAdded = false
-    let dropOffAdded = false
 
     // Support OTP1 flex messages in Trip Details
-    clonedItinerary.legs = clonedItinerary.legs.map((leg) => {
-      if (!!leg.boardRule && !leg.pickupBookingInfo && !pickupAdded) {
-        pickupAdded = true
-        leg.pickupBookingInfo = {
-          latestBookingTime: {
-            daysPrior: 0
-          }
+    // Adding empty pickupBookingInfo and dropOffBookingInfo objects
+    // to a leg will trigger relevant flex pickup / dropoff descriptions in
+    // the Trip Details component.
+    const boardingRuleLeg = clonedItinerary.legs.find(
+      (leg) => !!leg.boardRule && !leg.pickupBookingInfo
+    )
+    const alightingRuleLeg = clonedItinerary.legs.find(
+      (leg) => !!leg.alightRule && !leg.dropOffBookingInfo
+    )
+
+    // Add pickupBookingInfo for any boarding rules
+    if (boardingRuleLeg) {
+      boardingRuleLeg.pickupBookingInfo = {
+        latestBookingTime: {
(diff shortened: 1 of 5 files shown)


lib/components/narrative/line-itin/connected-itinerary-body.js
```diff
@@ -61,6 +62,30 @@ class ConnectedItineraryBody extends Component {
     } = this.props
     const { LegIcon } = this.context
 
+    let [pickupAdded, dropOffAdded] = [false, false]
```
binh-dam-ibigroup, line 65: We don't need an array for these declarations.

lib/components/narrative/line-itin/connected-itinerary-body.js
```diff
@@ -61,6 +62,30 @@ class ConnectedItineraryBody extends Component {
...
+
+    // Support OTP1 flex messages in Trip Details
+    itinerary.legs = itinerary.legs.map((oldLeg) => {
```
binh-dam-ibigroup, line 68: I would rather not touch the `itinerary` prop here. Maybe instead of cloning each leg as below, clone the entire `prop.itinerary` and modify the cloned itinerary at leisure (no need for a `.map()` or `clone`ing the legs in that case, a simple `.forEach()` will do). Then pass the resulting itinerary to `<StyledItineraryBody>` below.

lib/components/narrative/line-itin/connected-itinerary-body.js
```diff
@@ -61,6 +62,30 @@ class ConnectedItineraryBody extends Component {
...
+      const leg = clone(oldLeg)
+      if (!!leg.boardRule && !leg.pickupBookingInfo && !pickupAdded) {
+        pickupAdded = true
+        leg.pickupBookingInfo = {
```
binh-dam-ibigroup, line 72: Do we want this applied to each leg or only the first leg that satisfies the if condition above?

binh-dam-ibigroup (changes requested):
I recommend cloning the entire itinerary and change the result at leisure instead of modifying a prop.

lib/components/narrative/line-itin/connected-itinerary-body.js
binh-dam-ibigroup: Then maybe `clonedItinerary.legs.find` will work better than the `if` and the `let` assignments.

binh-dam-ibigroup: Would the best place to do that itinerary "transformation" be in the first redux `connect` statement that extracts the itinerary from OTP itinerary results.

miles-grant-ibigroup: I think we decided to do it here since it's a "transformation" that's very specific to the components called in this file. This is a bit of a hack, so I thought it makes sense to have these weird extra fields permeate as few points of the code as possible

philip-cline: I think we need to use `.map` since we're updating the leg's `leg.pickupBookingInfo`. If using `.find` here we'd need to update via an index

lib/components/narrative/line-itin/connected-itinerary-body.js
philip-cline: Ok, separated into their own lines in https://github.com/opentripplanner/otp-react-redux/pull/571/commits/dd2eb27dda8441bac039f6bc0095269df86c09cf

lib/components/narrative/line-itin/connected-itinerary-body.js
philip-cline: whole itinerary cloned in https://github.com/opentripplanner/otp-react-redux/pull/571/commits/dd2eb27dda8441bac039f6bc0095269df86c09cf

lib/components/narrative/line-itin/connected-itinerary-body.js
binh-dam-ibigroup: Hmm I disagree, if you use `find` you should be able to update that leg directly. In doing so you won't need the `let` variables you defined above.
```js
const legToUpdate = clonedItinerary.legs.find(<condition>)
if (legToUpdate) {
  // Should update the applicable leg object in clonedItinerary.legs
  legToUpdate.pickupBookingInfo = { ... }
}
```

binh-dam-ibigroup (changes requested):
Please see https://github.com/opentripplanner/otp-react-redux/pull/571#discussion_r831530378.

binh-dam-ibigroup (approved):
Thanks for the changes!

lib/components/narrative/line-itin/connected-itinerary-body.js
```diff
@@ -60,14 +61,43 @@ class ConnectedItineraryBody extends Component {
...
+    const boardingRuleLeg = clonedItinerary.legs.find(
+      (leg) => !!leg.boardRule && !leg.pickupBookingInfo
+    )
+    const alightingRuleLeg = clonedItinerary.legs.find(
+      (leg) => !!leg.alightRule && !leg.dropOffBookingInfo
+    )
+
+    // Add pickupBookingInfo for any boarding rules
+    if (boardingRuleLeg) {
+      boardingRuleLeg.pickupBookingInfo = {
```
miles-grant-ibigroup, line 79: Is this adjusting the `clonedItinerary`? I think this is what's causing this to no longer work

miles-grant-ibigroup (changes requested):
I'm not getting the PR to work anymore -- it's either my fault misconfiguring it or possibly a bug? 

Either way overall this looks good.

lib/components/narrative/line-itin/connected-itinerary-body.js
philip-cline: Things still seem to be working for me? Where are you seeing things not working? 

here's a sample search when running nysdot-qa that works for me: [here](http://localhost:9966/#/?ui_activeSearch=5yrwwpa68&ui_activeItinerary=0&fromPlace=Fitbody%20by%20Krista%2C%20979%20RT-22%2C%20Brewster%2C%20NY%2C%2010509%2C%20USA%3A%3A41.39683320215821%2C-73.60353447614618&toPlace=Terry%20Hill%20Rd%2C%20Carmel%2C%20NY%2C%2010512%2C%20USA%3A%3A41.461234603414816%2C-73.6545398793056&date=2022-03-28&time=13%3A38&arriveBy=false&mode=WALK%2CBUS%2CRAIL%2CSUBWAY%2CTRAM%2CFERRY&showIntermediateStops=true&maxWalkDistance=1207&optimize=QUICK&walkSpeed=1.34&ignoreRealtimeUpdates=true&numItineraries=3&otherThanPreferredRoutesPenalty=900)

miles-grant-ibigroup (approved):
Looks like the object is still getting updated. Javascript is so cool. Looks great!

PR #574, opened 2022-03-18
Inactive UI timeout

binh-dam-ibigroup:
This PR introduces a session timeout, after which the initial URL is reloaded. A warning message is shown a minute before the timeout expires.
For now, the change only applies to desktop environments and does not include OTP Middleware users.


lib/components/app/session-timeout.tsx
```diff
@@ -0,0 +1,118 @@
...
+      if (secondsToTimeout < 0) {
+        // TODO: If OTP middleware persistence is enabled,
+        // log out the logged-in user before resetting the page.
+
+        // Reload initial URL (page state is lost after this point)
+        startOverFromInitialUrl()
+      } else {
+        this.setState({
+          // If within a minute of timeout, display dialog, don't otherwise.
+          showTimeoutWarning: secondsToTimeout >= 0 && secondsToTimeout <= 60
```
caleb-diehl-ibigroup, line 58: This works as long as sessionTimeoutSeconds is configured to at least a few mins. If it's close to a minute it displays the dialog right away. Maybe dynamically calculate the threshold for displaying the dialog as a fraction of the configured time.

lib/reducers/create-otp-reducer.js
```diff
@@ -222,7 +228,13 @@ function createOtpReducer(config) {
   validateInitialState(initialState)
 
   /* eslint-disable-next-line complexity */
-  return (state = initialState, action) => {
+  return (prevState = initialState, action) => {
+    // Update last action timestamp for eligible actions.
+    const state = TIMEOUT_IGNORED_ACTIONS.includes(action.type)
```
caleb-diehl-ibigroup, line 233: Can eligible actions be defined in a config, or are they the same for every deployment? Not sure if that would be important, but maybe different agencies would want to allow different actions to reset the timer

lib/actions/ui.js
```diff
@@ -389,3 +392,15 @@ export function setRouteViewerFilter(filter) {
...
 }
+
+/**
+ * Start over from the initial URL
+ */
+export function startOverFromInitialUrl() {
+  return function (dispatch, getState) {
+    const { initialUrl } = getState().otp
+    window.location.replace(initialUrl)
+    // FIXME: Trigger a reload to reset page, otherwise the UI keeps some previous states.
```
caleb-diehl-ibigroup, line 403: I think you can delete this comment line - page reload seems to be working well

caleb-diehl-ibigroup (approved):

lib/actions/ui.js
binh-dam-ibigroup: I combined with the JsDoc of this method in https://github.com/opentripplanner/otp-react-redux/pull/574/commits/090c0aa72dd1b5c23858c77fa101497e47b509c9.

lib/components/app/session-timeout.tsx
binh-dam-ibigroup: I added a shorter warning window for shorter session timeouts (1/3 of the time) in https://github.com/opentripplanner/otp-react-redux/pull/574/commits/090c0aa72dd1b5c23858c77fa101497e47b509c9.

lib/reducers/create-otp-reducer.js
binh-dam-ibigroup: They will be the same for all deployments that use the optional timeout feature as this is trying to catch actions initiated by the user.

i18n/en-US.yml
```diff
@@ -401,6 +401,10 @@ components:
     signInTooltip: Please sign in to save trip.
   SearchScreen:
     header: Plan Your Trip
+  SessionTimeout:
+    body: Your session will expire within a minute. Press 'Keep Session' to keep your search.
+    header: Session about to timeout!
+    keepSession: Keep Session
```
daniel-heppner-ibigroup, line 407: Maybe "continue" instead of "keep"?

lib/actions/ui.js
```diff
@@ -396,3 +399,14 @@ export function setRouteViewerFilter(filter) {
     }
   }
 }
+
+/**
+ * Start over by setting the initial URL and resetting (reloading) the page.
+ */
+export function startOverFromInitialUrl() {
```
daniel-heppner-ibigroup, line 406: Did you check if there's a function that does this already from the sidebar's "start over" button?

lib/components/app/session-timeout.tsx
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  componentDidMount() {
+    // Wait one second or so after loading before probing for changes
+    // so that initialization actions can complete.
+    setTimeout(this.handleAfterInitialActions, 1500)
+  }
+
+  componentWillUnmount() {
+    // @ts-ignore SessionTimeout is not typed yet
```
daniel-heppner-ibigroup, line 30: Instead of using `ts-ignore` with an `eslint-disable`, you can use `ts-expect-error`.

lib/components/app/session-timeout.tsx
```diff
@@ -0,0 +1,124 @@
...
+  state = {
+    showTimeoutWarning: false,
+    timeoutObject: null,
+    timeoutStartMillis: 0
+  }
+
+  componentDidMount() {
+    // Wait one second or so after loading before probing for changes
+    // so that initialization actions can complete.
+    setTimeout(this.handleAfterInitialActions, 1500)
```
daniel-heppner-ibigroup, line 26: What would happen on a slower computer/connection, if things aren't finished in 1.5 seconds?

daniel-heppner-ibigroup: I guess the risk is that if some Redux actions get set off after this, then the `lastActionMillis` will be greater than `timeoutStartMillis` which would mean the page could time out without a user ever interacting with it?

lib/components/app/session-timeout.tsx
```diff
@@ -0,0 +1,124 @@
...
+      timeoutObject: setInterval(this.handleTimeoutWatch, 10000),
+      timeoutStartMillis: new Date().valueOf()
+    })
+  }
+
+  handleTimeoutWatch = () => {
+    // @ts-ignore SessionTimeout is not typed yet
+    const { lastActionMillis, sessionTimeoutSeconds, startOverFromInitialUrl } =
+      this.props
+    if (lastActionMillis > this.state.timeoutStartMillis) {
```
daniel-heppner-ibigroup, line 45: Is the purpose of this so that the timer doesn't start until the user makes their first action on the page? Can you add a comment about that? Took a bit of thinking on my part to finally figure that out

lib/components/app/session-timeout.tsx
```diff
@@ -0,0 +1,124 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment */
```
daniel-heppner-ibigroup, line 1: Is there something blocking typescripting this component?

daniel-heppner-ibigroup (approved):
One overarching thought is why use the reducer to detect user activity rather than adding an event listener to the body for clicks and keypresses? Doing something like that might solve the issue of having to wait until we think all the initial page load stuff is done before starting the timeout process, since the timer wouldn't be accidentally triggered by initial page load actions. But there could be some other downside to that approach that I am not seeing.

lib/components/app/session-timeout.tsx
binh-dam-ibigroup: Added types in https://github.com/opentripplanner/otp-react-redux/pull/574/commits/292181c4439ad16d50a1e1860fd7504c3b1f2721.

lib/components/app/session-timeout.tsx
binh-dam-ibigroup: Added types instead in https://github.com/opentripplanner/otp-react-redux/pull/574/commits/292181c4439ad16d50a1e1860fd7504c3b1f2721.

binh-dam-ibigroup:
> One overarching thought is why use the reducer to detect user activity rather than adding an event listener to the body for clicks and keypresses? Doing something like that might solve the issue of having to wait until we think all the initial page load stuff is done before starting the timeout process, since the timer wouldn't be accidentally triggered by initial page load actions. But there could be some other downside to that approach that I am not seeing.

That approach would require extra code to wire the events with the top-level elements, and you would still need to use the reducer to store the time since last action/time before the session expires, so the approach uses existing actions instead.

lib/components/app/session-timeout.tsx
binh-dam-ibigroup: This just delays the start of the timer and the timeout check under `this.handleTimeoutWatch`, so that the `lastActionMills` state doesn't get unnecessarily updated with all the stuff that normally occurs during a page load. I'll add a comment about it.

lib/components/app/session-timeout.tsx
binh-dam-ibigroup: With this check, actions that happened before the session timer is started are simply ignored for timeout purposes. I'll add a comment about that.

lib/actions/ui.js
binh-dam-ibigroup: This code is slightly different from the "Start Over" buttons that clear out the original query params used to load the page. We need to preserve those original query params. I'll add a comment regarding that need, thanks for pointing this out.

binh-dam-ibigroup:
@daniel-heppner-ibigroup Please see my responses and clarifying comments in the code in https://github.com/opentripplanner/otp-react-redux/pull/574/commits/08564816c5df4ff839f6af5a6fe6b42e28d7f0bc.

PR #576, opened 2022-03-28
Mode selector customization

daniel-heppner-ibigroup:
This PR moves mode selector options into the config instead of being hard coded. This means that for the batch UI, you need to include mode selector options in the config for the deployment. (edit: now with backwards comaptability so old behavior is restored if the config isn't updated. But we should really update the configs!) Combinations are enabled when every one of their required modes is enabled.


lib/components/form/batch-settings.tsx
```diff
@@ -105,14 +108,18 @@ class BatchSettings extends Component<{
...
     this.setState({ selectedModes: newModes })
     // Update the available mode combinations based on the new modes selection.
-    const possibleModes = getModeOptions(this.props.intl).map((m) => m.mode)
+    const possibleModes = modeOptions.map((m) => m.mode)
     const disabledModes = possibleModes.filter((m) => !newModes.includes(m))
-    // Do not include combination if any of its modes are found in disabled
-    // modes list.
+    // Only include a combination if it every required mode is enabled.
     const newCombinations = possibleCombinations
-      .filter((c) => !combinationHasAnyOfModes(c, disabledModes))
```
miles-grant-ibigroup, line 113: We can probably remove the `combinationHasAnyOfModes` method right?

lib/components/form/batch-settings.tsx
```diff
@@ -216,6 +225,7 @@ class BatchSettings extends Component<{
 const mapStateToProps = (state: any) => ({
   config: state.otp.config,
   currentQuery: state.otp.currentQuery,
+  modeOptions: state.otp.config.modes.modeOptions,
```
miles-grant-ibigroup, line 228: I had a total crash when I tried to load a config without the new options. I know we need the `requiredModes` for the new stuff to work, but what if we set a default here and then extracted the requiredModes from the string if they weren't explicitly specified? I know this would probably result in weird behavior but I think the backwards compatibility is valuable, and it gives a nice "it just works" feeling

miles-grant-ibigroup (changes requested):
I love how clean this is! I was expecting so much more to be needed for this change. However, I'm a bit weary of mandating our new options. Let me know what you think of my suggestion, but I'm open to any idea that allows previous configs to still work (even if behavior changes a bit)

lib/components/form/batch-settings.tsx
daniel-heppner-ibigroup: Okay, some big changes to add in defaults similar to old behavior. (which already was not great and a bit glitchy, mind you)

lib/components/form/batch-preferences.tsx
```diff
@@ -40,12 +43,21 @@ class BatchPreferences extends Component<{
...
+        if (c.requiredModes) {
+          return c.requiredModes.every((m) => enabledModes.includes(m))
+        } else {
+          // This is for backwards compatibility
+          // In case a combination does not include requiredModes.
+          console.warn(
+            `Combination ${c.mode} does not have any specified required modes.`
+          )
+          const modesInCombination = c.mode.split(',')
+          return modesInCombination.every((m) => enabledModes.includes(m))
```
miles-grant-ibigroup, line 59: Would it be possible to extract this back into the util file that you deleted those old methods from? Let me know if that didn't make sense

miles-grant-ibigroup (approved):
Works great and looks great! Thanks so much for cleaning this all up. I have one last suggestion about avoiding some repeated code but it's not that essential.

lib/components/form/mode-buttons.tsx
```diff
@@ -65,17 +40,18 @@ const ModeButton = ({
...
-      placement="bottom"
-    >
+    <OverlayTrigger overlay={overlayTooltip} placement="bottom">
       <button className={className} onClick={() => onClick(mode)}>
         {icon ? (
           <Icon className="fa-2x" type={icon} />
         ) : (
           <ModeIcon height={25} mode={mode} />
         )}
+        <FormattedMode mode={mode} />
```
binh-dam-ibigroup, line 54: Is that rendered in the final HTML?

lib/components/form/mode-buttons.tsx
```diff
@@ -97,19 +73,23 @@ export const StyledModeButton = styled(ModeButton)`
 
 const ModeButtons = ({
   className,
+  modeOptions,
```
binh-dam-ibigroup, line 76: Sort props.

example-config.yml
```diff
@@ -174,6 +174,16 @@ geocoder:
...
 modes:
+  modeOptions:
+    - mode: TRANSIT
+    - mode: WALK
+    - mode: CAR
+    - mode: BICYCLE
+    - mode: RENT
+      icon: mobile
+      # This overrides the label generated from the i18n. It's not recommended.
+      # label: Rental
```
binh-dam-ibigroup, line 185: `label` may probably not be necessary, as the label for the mode can already be overwritten in this file, from the `language` section.

binh-dam-ibigroup (commented):
Looks mostly good! Could you advise regarding rendering the mode within the `ModeButton` `<button>` elements, and whether another means to overwrite a mode `label` is needed.

lib/components/form/mode-buttons.tsx
daniel-heppner-ibigroup: Shouldn't be there, good catch.

example-config.yml
daniel-heppner-ibigroup: That's a good point, I'll go ahead and remove.

binh-dam-ibigroup (approved):

PR #577, opened 2022-03-28
feat(i18n): import strings from otp-ui packages

miles-grant-ibigroup:
This PR adds support for importing language strings from otp-ui. Strings copied from the repo have been removed. The deepmerge dependency is new in `package.json`, but not in `yarn.lock` (it was previously a sub dependency)


lib/util/i18n.js
```diff
@@ -1,5 +1,7 @@
 import flatten from 'flat'
 
+const merge = require('deepmerge')
```
binh-dam-ibigroup, line 3: Can this be a regular `import`?

lib/util/i18n.js
```diff
@@ -45,6 +47,33 @@ export function getConfigLocales(configLanguages) {
...
+ * Loads in all localized strings from @opentripplanner packages.
+ * Package list is loaded from the package.json and each yml file then imported.
+ *
+ * All strings are then merged
+ */
+async function loadOtpUiLocaleData(matchedLocale) {
+  const packageJson = await import('../../package.json')
+  const otpUiMessages = await Promise.all(
+    Object.keys(packageJson.dependencies)
+      .filter((pkg) => pkg.includes('@opentripplanner'))
```
binh-dam-ibigroup, line 60: `.startsWith`?

lib/util/i18n.js
```diff
@@ -76,9 +105,12 @@ export async function loadLocaleData(matchedLocale, customMessages) {
       break
   }
 
+  const otpUiMessages = await loadOtpUiLocaleData(matchedLocale)
+
   // Merge custom strings into the standard language strings.
   const mergedMessages = {
     ...flatten(messages),
+    ...flatten(otpUiMessages),
```
binh-dam-ibigroup, line 113: There might be a case where we would need OTP-RR messages to overwrite default OTP-UI messages, so ideally `otpUiMessages` should be assigned before `messages`.

binh-dam-ibigroup (approved):
Nice! See the few minor comments.

lib/util/i18n.js
miles-grant-ibigroup: Sadly no: https://github.com/TehShrike/deepmerge#include

lib/util/i18n.js
miles-grant-ibigroup: Good call! Refactored in bfe701697e8cbceef1b004a198352f7504f4e6b0

lib/util/i18n.js
miles-grant-ibigroup: That feels good! Refactored in bfe701697e8cbceef1b004a198352f7504f4e6b0

lib/util/i18n.js
binh-dam-ibigroup: Place a comment in the code referencing that link.

miles-grant-ibigroup: Done!

philip-cline (approved):

PR #583, opened 2022-04-04
Batch UI Improvements 

miles-grant-ibigroup:
This PR doesn't fix all of the UX issues with the batch UI, but gets a start.

Shows multiple departure times for the same itinerary instead of showing the itinerary multiple times. 
 

### Looking for feedback on
- ~~Cleaning up the merge logic, if possible~~

### DIFF
--- a/lib/components/narrative/default/default-itinerary.js
+++ b/lib/components/narrative/default/default-itinerary.js
@@ -90,7 +90,7 @@ function ItineraryDescription({ itinerary }) {
   return transitMode
     ? <FormattedMessage
       id='components.DefaultItinerary.multiModeSummary'
-      values={{ accessMode: mainMode, transitMode, primaryRoute }}
+      values={{ accessMode: mainMode, transitMode }}
     />
     : mainMode
 }
--- a/lib/components/narrative/default/default-itinerary.js
+++ b/lib/components/narrative/default/default-itinerary.js
@@ -126,9 +126,9 @@ const ITINERARY_ATTRIBUTES = [
         ) 
       }
         if (options.selection === 'ARRIVALTIME') {
-          return <FormattedTime value={formattedStartTime} />
-        } else {
           return <FormattedTime value={formattedEndTime} />
+        } else {
+          return <FormattedTime value={formattedStartTime} />
         }
     }
   },
--- a/lib/components/narrative/default/default-itinerary.js
+++ b/lib/components/narrative/default/default-itinerary.js
@@ -110,11 +110,11 @@ const ITINERARY_ATTRIBUTES = [
     id: 'arrivalTime',
     order: 1,
     render: (itinerary, options) => {
-      const formattedStartTime = itinerary.startTime + options.offset
-      const formattedEndTime = itinerary.endTime + options.offset
+      const startTimeWithOffset = itinerary.startTime + options.offset
+      const endTimeWithOffset = itinerary.endTime + options.offset
 
       if (options.selection === 'ARRIVALTIME') {
-        return <FormattedTime value={formattedEndTime} />
+        return <FormattedTime value={endTimeWithOffset} />
       }
       if (options.selection !== 'DEPARTURETIME' && itinerary.allStartTimes) {
         const allStartTimes = Array.from(itinerary.allStartTimes).sort()
@@ -125,7 +125,7 @@ const ITINERARY_ATTRIBUTES = [
           />
         )
       }
-      return <FormattedTime value={formattedStartTime} />
+      return <FormattedTime value={startTimeWithOffset} />
     }
   },
   {
--- a/lib/components/narrative/default/default-itinerary.js
+++ b/lib/components/narrative/default/default-itinerary.js
@@ -109,22 +109,25 @@ const ITINERARY_ATTRIBUTES = [
     render: (itinerary, options) => {
       const formattedStartTime = itinerary.startTime + options.offset
       const formattedEndTime = itinerary.endTime + options.offset
-      if (options.isSelected) {
+
+      if (itinerary.extraStartTimes) {
+        const extraStartTimes = Array.from(itinerary.extraStartTimes)
+        return (
+          <FormattedMessage
+            id='common.time.multiDepartureTimes'
+            values={{
+              startTime: formattedStartTime,
+              secondStartTime: extraStartTimes?.[0] + options.offset || 0,
+              thirdStartTime: extraStartTimes?.[1] + options.offset || 0
+            }}
+          />
+        ) 
+      }
         if (options.selection === 'ARRIVALTIME') {
           return <FormattedTime value={formattedStartTime} />
         } else {
           return <FormattedTime value={formattedEndTime} />
         }
-      }
-      return (
-        <FormattedMessage
-          id='common.time.departureArrivalTimes'
-          values={{
-            endTime: formattedEndTime,
-            startTime: formattedStartTime
-          }}
-        />
-      )
     }
   },
   {
--- a/lib/components/narrative/default/default-itinerary.js
+++ b/lib/components/narrative/default/default-itinerary.js
@@ -4,6 +4,7 @@
 import coreUtils from '@opentripplanner/core-utils'
 import React from 'react'
 import {
+  FormattedList,
   FormattedMessage,
   FormattedNumber,
   FormattedTime,
@@ -112,24 +113,19 @@ const ITINERARY_ATTRIBUTES = [
       const formattedStartTime = itinerary.startTime + options.offset
       const formattedEndTime = itinerary.endTime + options.offset
 
-      if (itinerary.extraStartTimes) {
-        const extraStartTimes = Array.from(itinerary.extraStartTimes)
-        return (
-          <FormattedMessage
-            id='common.time.multiDepartureTimes'
-            values={{
-              startTime: formattedStartTime,
-              secondStartTime: extraStartTimes?.[0] + options.offset || 0,
-              thirdStartTime: extraStartTimes?.[1] + options.offset || 0
-            }}
-          />
-        ) 
+      if (options.selection === 'ARRIVALTIME') {
+        return <FormattedTime value={formattedEndTime} />
       }
-        if (options.selection === 'ARRIVALTIME') {
-          return <FormattedTime value={formattedEndTime} />
-        } else {
-          return <FormattedTime value={formattedStartTime} />
-        }
+      if (options.selection !== 'DEPARTURETIME' && itinerary.allStartTimes) {
+        const allStartTimes = Array.from(itinerary.allStartTimes).sort()
+        return (
+          <FormattedList
+            type='conjunction'
+            value={allStartTimes.map(time => <FormattedTime value={time} />)}
+          />
+        )
+      }
+      return <FormattedTime value={formattedStartTime} />
     }
   },
   {
--- a/lib/components/narrative/default/default-itinerary.js
+++ b/lib/components/narrative/default/default-itinerary.js
@@ -70,12 +70,14 @@ function ItineraryDescription({ itinerary }) {
   let primaryTransitDuration = 0
   let accessModeId = 'walk'
   let transitMode
+  let primaryRoute
   itinerary.legs.forEach((leg, i) => {
-    const { duration, mode, rentedBike, rentedVehicle } = leg
+    const { duration, mode, rentedBike, rentedVehicle, routeShortName } = leg
     if (isTransit(mode) && duration > primaryTransitDuration) {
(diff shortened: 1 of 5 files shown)


lib/components/narrative/default/default-itinerary.js
```diff
@@ -109,22 +111,25 @@ const ITINERARY_ATTRIBUTES = [
...
+          />
+        ) 
+      }
         if (options.selection === 'ARRIVALTIME') {
-          return <FormattedTime value={formattedStartTime} />
-        } else {
           return <FormattedTime value={formattedEndTime} />
+        } else {
+          return <FormattedTime value={formattedStartTime} />
         }
```
binh-dam-ibigroup, line 132: Fix indent.

lib/components/narrative/narrative-itineraries.js
```diff
@@ -148,6 +150,33 @@ class NarrativeItineraries extends Component {
...
+    // Merge duplicate itineraries together and save multiple departure times
+    let nextIndex = 0
+    const mergedItineraries = itineraries.reduce((prev, cur) => {
+      const updated = clone(prev)
+      cur.index = nextIndex++
+
+      const dupe = updated.findIndex((itin) => itinerariesAreEqual(itin, cur))
+      // If no duplicate, push full itinerary to output
+      if (dupe === -1) {
+        updated.push(cur)
```
binh-dam-ibigroup, line 162: If we are going to update indexes or change other fields, do `.push(clone(cur))`

lib/components/narrative/narrative-itineraries.js
```diff
@@ -148,6 +150,33 @@ class NarrativeItineraries extends Component {
...
+      } else {
+        // Add only new start time to existing itinerary
+        if (!updated[dupe].extraStartTimes) {
```
binh-dam-ibigroup, line 165: Extract var for `updated[dupe]`.

lib/components/narrative/narrative-itineraries.js
```diff
@@ -148,6 +150,33 @@ class NarrativeItineraries extends Component {
...
+          updated[dupe].extraStartTimes = new Set()
```
binh-dam-ibigroup, line 166: I think this array should hold all the start times, especially since we are building a list for display. So probably rename to `allStartTimes` and include the first start time.

lib/components/narrative/narrative-itineraries.js
```diff
@@ -148,6 +150,33 @@ class NarrativeItineraries extends Component {
...
+        }
+
+        // Ensure that the lowest start time is always first
+        if (updated[dupe].startTime > cur.startTime) {
+          updated[dupe].extraStartTimes.add(updated[dupe].startTime)
+          updated[dupe].startTime = cur.startTime
```
binh-dam-ibigroup, line 172: I think you wouldn't need to update this field if building a list of `allStartTimes`.

lib/util/itinerary.js
```diff
@@ -74,14 +76,26 @@ function getFirstTransitLeg (itinerary) {
...
-    : WEEKDAYS
+  return dayOfWeek === 0 || dayOfWeek === 6 ? WEEKEND_DAYS : WEEKDAYS
+}
+
+export function itinerariesAreEqual(itinerary, other) {
+  return (
+    itinerary.legs.length === other.legs.length &&
+    itinerary.legs.every(
+      (leg, index) =>
+        other?.legs?.[index].mode === leg.mode &&
```
binh-dam-ibigroup, line 94: Extract var for `other?.legs?.[index]`.

binh-dam-ibigroup: And perhaps for the `from`s and `to`s, or extract a function to compare the coords of two places.

binh-dam-ibigroup (changes requested):
I think we should use `<FormattedList>` and build a list of departure times accordingly.

lib/components/narrative/default/default-itinerary.js
```diff
@@ -109,22 +111,25 @@ const ITINERARY_ATTRIBUTES = [
     render: (itinerary, options) => {
       const formattedStartTime = itinerary.startTime + options.offset
       const formattedEndTime = itinerary.endTime + options.offset
-      if (options.isSelected) {
+
+      if (itinerary.extraStartTimes) {
+        const extraStartTimes = Array.from(itinerary.extraStartTimes)
+        return (
+          <FormattedMessage
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Was looking for and failed to find exactly this. Thanks for the tip! Much cleaner now in 7969cc8c08590f6daf18c60716df01360bcd547b

lib/components/narrative/default/default-itinerary.js
miles-grant-ibigroup: Fixed in 7969cc8c08590f6daf18c60716df01360bcd547b

lib/components/narrative/narrative-itineraries.js
```diff
@@ -148,6 +150,33 @@ class NarrativeItineraries extends Component {
 
     if (!activeSearch) return null
 
+    // Merge duplicate itineraries together and save multiple departure times
+    let nextIndex = 0
+    const mergedItineraries = itineraries.reduce((prev, cur) => {
+      const updated = clone(prev)
+      cur.index = nextIndex++
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Good call! Have cloned the entire thing in 7969cc8c08590f6daf18c60716df01360bcd547b

lib/components/narrative/narrative-itineraries.js
miles-grant-ibigroup: Redundant because we don't use cur anymore in 7969cc8c08590f6daf18c60716df01360bcd547b

lib/components/narrative/narrative-itineraries.js
```diff
@@ -148,6 +150,33 @@ class NarrativeItineraries extends Component {
...
+
+      const dupe = updated.findIndex((itin) => itinerariesAreEqual(itin, cur))
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Renamed in 7969cc8c08590f6daf18c60716df01360bcd547b

lib/components/narrative/narrative-itineraries.js
miles-grant-ibigroup: Extracted in 7969cc8c08590f6daf18c60716df01360bcd547b

miles-grant-ibigroup: Always scared of javascript's auto-pointers... Works perfectly until suddenly you're writing to a copy

lib/components/narrative/narrative-itineraries.js
miles-grant-ibigroup: Really good call! Refactored in 7969cc8c08590f6daf18c60716df01360bcd547b

lib/components/narrative/narrative-itineraries.js
miles-grant-ibigroup: Love it. Refactored in 7969cc8c08590f6daf18c60716df01360bcd547b

lib/util/itinerary.js
miles-grant-ibigroup: Cleaned up in 7969cc8c08590f6daf18c60716df01360bcd547b

miles-grant-ibigroup:
Thanks for some really good suggestions! Much cleaner now

lib/components/narrative/default/default-itinerary.js
```diff
@@ -109,22 +112,20 @@ const ITINERARY_ATTRIBUTES = [
     render: (itinerary, options) => {
       const formattedStartTime = itinerary.startTime + options.offset
       const formattedEndTime = itinerary.endTime + options.offset
```
binh-dam-ibigroup, line 114: Rename these vars to `startTimeWihOffset`, `endTimeWithOffset`.

lib/components/narrative/narrative-itineraries.js
```diff
@@ -148,6 +150,32 @@ class NarrativeItineraries extends Component {
 
     if (!activeSearch) return null
 
+    // Merge duplicate itineraries together and save multiple departure times
+    let nextIndex = 0
+    const mergedItineraries = itineraries.reduce((prev, cur) => {
```
binh-dam-ibigroup, line 155: Rename these vars. `prev` to `prevItineraries`... `cur` to `currentItinerary` or `currentItin`?

lib/components/narrative/default/default-itinerary.js
```diff
@@ -109,22 +112,20 @@ const ITINERARY_ATTRIBUTES = [
...
-        } else {
-          return <FormattedTime value={formattedEndTime} />
-        }
+
+      if (options.selection === 'ARRIVALTIME') {
+        return <FormattedTime value={formattedEndTime} />
       }
-      return (
-        <FormattedMessage
-          id='common.time.departureArrivalTimes'
```
binh-dam-ibigroup, line 121: Do we still use this message id elsewhere? If not remove it from language files.

binh-dam-ibigroup (approved):
Thanks for the changes! Some more var renaming and this should be good to go.

lib/components/narrative/narrative-itineraries.js
miles-grant-ibigroup: I think that would be more confusing than keeping the `reduce` standard. It's not really the previous itineraries, isn't it more the result from the previous iteration of the loop? I tried to think of better names but `prev` and `cur` are quite standard

lib/components/narrative/default/default-itinerary.js
miles-grant-ibigroup: Yes it's still used in the tabbed itineraries and I think it makes sense there!

lib/components/narrative/default/default-itinerary.js
miles-grant-ibigroup: Renamed in 2dd6711fa9cb06428815df5ab9d1515fe95c1826

lib/components/narrative/narrative-itineraries.js
```diff
@@ -148,6 +150,32 @@ class NarrativeItineraries extends Component {
...
+      const updatedItineraries = clone(prev)
```
daniel-heppner-ibigroup, line 156: What's the reason for cloning prev into a new array? It seems like it's safe to just modify prev.

lib/components/narrative/narrative-itineraries.js
```diff
@@ -148,6 +150,32 @@ class NarrativeItineraries extends Component {
 
     if (!activeSearch) return null
 
+    // Merge duplicate itineraries together and save multiple departure times
+    let nextIndex = 0
```
daniel-heppner-ibigroup, line 154: Would you be able to use the current index provided as the third argument to the reduce callback?

lib/util/itinerary.js
```diff
@@ -74,14 +76,29 @@ function getFirstTransitLeg (itinerary) {
...
+  return legLocation.lat === other.lat && legLocation.lon === other.lon
+}
+export function itinerariesAreEqual(itinerary, other) {
+  return (
+    itinerary.legs.length === other.legs.length &&
+    itinerary.legs.every((leg, index) => {
+      const otherLeg = other?.legs?.[index]
+      return (
+        otherLeg.mode === leg.mode &&
+        legLocationsAreEqual(otherLeg?.to, leg?.to) &&
```
daniel-heppner-ibigroup, line 99: If `otherLeg` or `leg` is null, does `legLocationsAreEqual` still get called but with a null object? Because it looks like that could cause a crash when it tries to access lat and lon on the null object. This might just be a gap on my understanding of JS trivia.

daniel-heppner-ibigroup (changes requested):
A few little suggestions here on code structure. Overall the change looks good!

lib/components/narrative/narrative-itineraries.js
miles-grant-ibigroup: Binh convinced me that this is a case of better safe than sorry

lib/components/narrative/narrative-itineraries.js
miles-grant-ibigroup: This is so good I totally forgot about that argument thank you!

lib/util/itinerary.js
miles-grant-ibigroup: Yeah this feels like a pop quiz javascript behavior question... My understanding is `legLocationsAreEqual` will receive `undefined` instead of an error being thrown. Which means we do need a null check! Added in 8d0c158e

daniel-heppner-ibigroup (approved):
Thanks for the fixes! Looks good.

PR #586, opened 2022-04-06
Don't display fares explictly set <0

miles-grant-ibigroup:
This fix feels like it may be more appropriate to make at the OTP level. I would be open to closing this PR and making one in OTP instead.

Either way, this PR prevents the following object from being rendered "correctly"
MicrosoftTeams-image (1)


philip-cline (approved):
Changes makes sense to me. I'll defer to others on moving this to an OTP change

binh-dam-ibigroup (approved):

PR #588, opened 2022-04-12
feat(mode-buttons): disable modes by default

### LINKED ISSUE #590: Accessibility Routing No Longer Works
It looks like our mode selector changes have broken the method used to append `trip-form` query parameters to the URL.

daniel-heppner-ibigroup:
This adds the ability to disable modes by default on page load. 

Closes #590


lib/components/form/mode-buttons.tsx
```diff
@@ -6,10 +6,10 @@ import { buttonCss } from './batch-styled'
 import { ComponentContext } from '../../util/contexts'
 import { getFormattedMode } from '../../util/i18n'
 import { injectIntl, IntlProvider, IntlShape } from 'react-intl'
-import FormattedMode from '../util/formatted-mode'
 import Icon from '../util/icon'
 
 export type Mode = {
```
miles-grant-ibigroup, line 11: I think you need to do address what the auto formatter is complaining about in this file

lib/reducers/create-otp-reducer.js
```diff
@@ -142,9 +143,21 @@ export function getInitialState(userDefinedConfig) {
   if (currentQuery.routingType === 'ITINERARY') {
     queryModes = ensureSingleAccessMode(queryModes)
   }
-  if (config.modes && config.modes.combinations) {
+  if (config.modes?.combinations) {
     // Inject combinations from config for batch routing.
-    currentQuery.combinations = clone(config.modes.combinations)
+    // Filter options based on defaults if modeOptions is specified in config
+    let enabledCombinations
```
miles-grant-ibigroup, line 149: Might we be able to avoid a let here if we do the default modes above this line and then use a `||`? Usually I'm against such shorthands but there's really only 2 cases here

lib/reducers/create-otp-reducer.js
```diff
@@ -142,9 +143,21 @@ export function getInitialState(userDefinedConfig) {
   if (currentQuery.routingType === 'ITINERARY') {
     queryModes = ensureSingleAccessMode(queryModes)
   }
-  if (config.modes && config.modes.combinations) {
+  if (config.modes?.combinations) {
```
miles-grant-ibigroup, line 146:
```suggestion
  if (config?.modes?.combinations) {
```
I think this missing question mark is causing the crash

miles-grant-ibigroup (changes requested):
Looks like all the unit tests failures are fixable. Just one note about the actual implementation. Once it runs without crashing I'll give it a test run, but overall quite excited about this!

miles-grant-ibigroup:
This is probably the venue to fix #590.

miles-grant-ibigroup (approved):
We should probably get 2 more reviewers for this since I have contributed

lib/reducers/create-otp-reducer.js
```diff
@@ -142,11 +143,18 @@ export function getInitialState(userDefinedConfig) {
...
     queryModes = ensureSingleAccessMode(queryModes)
   }
-  if (config.modes && config.modes.combinations) {
+  if (config?.modes?.combinations) {
     // Inject combinations from config for batch routing.
-    currentQuery.combinations = clone(config.modes.combinations)
+    // Filter options based on defaults if modeOptions is specified in config
+    const defaultModes = config.modes.modeOptions
+      ?.filter((m) => !m.defaultUnselected)
+      .map((m) => m.mode)
```
binh-dam-ibigroup, line 151: Extract method and place in a file under `/lib/utils/` (maybe `itinerary.js`?). In there, when converting to TypeScript, it could be a good thing to perform type checks.

binh-dam-ibigroup (changes requested):
See various comments.

daniel-heppner-ibigroup:
I think all the comments are for @miles-grant-ibigroup.

lib/components/form/batch-settings.tsx
```diff
@@ -33,12 +33,19 @@ import type { Mode } from './mode-buttons'
 
 /**
  * A function that generates a filter to be used to filter a list of combinations.
- * @param enabledModes A list of the modes enabled in the UI
+ * @param enabledModesMysteryType A list of the modes enabled in the UI
  * @returns Filter function to filter combinations
  */
 export const combinationFilter =
-  (enabledModes: string[]) =>
+  (enabledModesMysteryType: string[] | { mode: string }[]) =>
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Good call. Renamed in 314374d254f4e3746d93d21d263b36a8514f93af

lib/components/form/mode-buttons.tsx
```diff
@@ -1,15 +1,16 @@
...
-import { buttonCss } from './batch-styled'
 import { ComponentContext } from '../../util/contexts'
 import { getFormattedMode } from '../../util/i18n'
-import { injectIntl, IntlProvider, IntlShape } from 'react-intl'
-import FormattedMode from '../util/formatted-mode'
 import Icon from '../util/icon'
 
+import { buttonCss } from './batch-styled'
+
 export type Mode = {
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Added in 314374d254f4e3746d93d21d263b36a8514f93af

lib/reducers/create-otp-reducer.js
```diff
@@ -142,11 +143,18 @@ export function getInitialState(userDefinedConfig) {
   if (currentQuery.routingType === 'ITINERARY') {
     queryModes = ensureSingleAccessMode(queryModes)
   }
-  if (config.modes && config.modes.combinations) {
+  if (config?.modes?.combinations) {
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Extracted in 314374d254f4e3746d93d21d263b36a8514f93af

lib/components/form/batch-settings.tsx
```diff
@@ -100,7 +114,9 @@ class BatchSettings extends Component<{
 }> {
   state = {
     expanded: null,
-    selectedModes: this.props.modeOptions.map((m) => m.mode)
+    selectedModes: this.props.modeOptions
+      .filter((m) => !m.defaultUnselected)
+      .map((m) => m.mode)
```
binh-dam-ibigroup, line 119: Use new method `getDefaultModes` here too? (this block looks very similar)

binh-dam-ibigroup: Is this block a new change? I can't get it to pick up the unselected modes. Should probably be moved to the class constructor.

binh-dam-ibigroup (changes requested):
One (small) change is needed to make it work, I think, and one more refactor.

lib/components/form/batch-settings.tsx
miles-grant-ibigroup: The bug was a result of a merge conflict that we fixed in 647708e50a79e9185017b051388ff377b214dd83. We've also used the new method in 6ffad50f66e885e0f7db863d42cfe857bc14291b

lib/components/form/batch-settings.tsx
```diff
@@ -100,7 +115,7 @@ class BatchSettings extends Component<{
 }> {
   state = {
     expanded: null,
-    selectedModes: this.props.modeOptions.map((m) => m.mode)
+    selectedModes: getDefaultModes(this.props.modeOptions)
```
binh-dam-ibigroup, line 118: As mentioned in the previous review, this assignment to `state` needs to be done in the class constructor. As written, it seems that `this.props` is not properly set at the time of execution.

binh-dam-ibigroup (changes requested):
I think this https://github.com/opentripplanner/otp-react-redux/pull/588/files#r860149626 still needs to be addressed?

lib/components/form/batch-settings.tsx
miles-grant-ibigroup: Not sure how this one slipped through! Fixed in 820468c62705513a5cab0b9e635fd0ab50b1d54c

binh-dam-ibigroup (approved):
LGTM, thanks for the changes!

daniel-heppner-ibigroup:
@binh-dam-ibigroup, could you take one last look at 0e22c7878a07624d1eee09d878b42f8795dc6892 where I made some changes to get rid of that intermediate "mystery". The problem was that the filter was sometimes getting a string and sometimes an object with the string in it, so I fixed the usages so it will only ever receive a string, which got rid of the weird `or` type.

philip-cline (approved):
Works really well for me!

PR #589, opened 2022-04-13
fix(user-profiles): various regressions surrounding react-intl

miles-grant-ibigroup:
A few bugs were reported which seem to all stem from changes we've made over the past few months with relation to react-intl. **I'm very unsure that my fixes are the ideal way to resolve these issues, so would appreciate a careful review from the original authors of the affected files!**

Specifically:
- adding a monitored trip only works for Saturday and Sunday
- monitored trip icons are vertical instead of horizontal
- adding a custom place doesn't work (can't select anything that `location-field` returns)

### DIFF
--- a/lib/components/user/places/place-editor.tsx
+++ b/lib/components/user/places/place-editor.tsx
@@ -66,7 +66,7 @@ class PlaceEditor extends Component<
   } & WrappedComponentProps
 > {
   _handleLocationChange = (
-    _: IntlShape, // Ignore intl object. TODO: localize the name field of the returned location object?
+    _: IntlShape, // Ignore intl object.
     { location }: { location: Location }
   ) => {
     const { setValues, values } = this.props


lib/components/user/monitored-trip/trip-basics-pane.tsx
```diff
@@ -200,12 +200,11 @@ class TripBasicsPane extends Component<TripBasicsProps> {
                   : monitoredTrip[day]
                   ? 'bg-primary'
                   : ''
-                const notAvailableText = isDayDisabled
-                  ? intl.formatMessage(
-                      { id: 'components.TripBasicsPane.tripNotAvailableOnDay' },
-                      { repeatedDay: getFormattedDayOfWeekPlural(day, intl) }
-                    )
-                  : null
+                const notAvailableText = intl.formatMessage(
```
binh-dam-ibigroup, line 203: What was the bug that you observed here? It looks like we are currently showing a popup that says "Trip not available on <day>" when the user hovers on a day for which a trip is available, but this change does not seem to fix this particular issue.

lib/components/user/places/place-editor.tsx
```diff
@@ -64,15 +64,18 @@ class PlaceEditor extends Component<
...
 > {
-  _handleLocationChange = ({
-    location
-  }: {
-    location: {
-      lat: number
-      lon: number
-      name: string
+  _handleLocationChange = (
+    _: unknown, // Ignore intl object. TODO: localize name?
```
binh-dam-ibigroup, line 68: Use `IntlShape` for the arg type.

lib/components/user/places/place-editor.tsx
```diff
@@ -64,15 +64,18 @@ class PlaceEditor extends Component<
...
+    {
+      location
+    }: {
+      location: {
```
binh-dam-ibigroup, line 72: For `location`, you can use the `Location` type from `@opentripplanner/types`.

binh-dam-ibigroup (commented):
See the different comments

lib/components/user/monitored-trip/trip-basics-pane.tsx
miles-grant-ibigroup: The bug here was that only Saturday and Sunday was showing up. I think that might be related to the bug you describe, but I think it's more serious to not have the days show then to not have them show in the tooltip

lib/components/user/places/place-editor.tsx
miles-grant-ibigroup: Good call thanks. Refactored in 71c5a88

lib/components/user/places/place-editor.tsx
miles-grant-ibigroup: Added in 71c5a88!

lib/components/user/monitored-trip/trip-basics-pane.tsx
binh-dam-ibigroup: Okay, I see. Can you try the original code on this block and remove the `notAvailableText &&` condition in the `return` statement below? I think that's where things broke originally. In `dev` it reads that if a trip **is** available on a particular day, `notAvailable` would be `null` and the label/checkbox for the day is thus not shown.

lib/components/user/places/place-editor.tsx
```diff
@@ -64,15 +65,10 @@ class PlaceEditor extends Component<
...
-    location
-  }: {
-    location: {
-      lat: number
-      lon: number
-      name: string
-    }
-  }) => {
+  _handleLocationChange = (
+    _: IntlShape, // Ignore intl object. TODO: localize name?
```
binh-dam-ibigroup, line 69: What do you mean by "localize name"?

lib/components/user/monitored-trip/trip-basics-pane.tsx
miles-grant-ibigroup: Screen Shot 2022-04-19 at 10 38 49 AM

Switching to what you describe causes this to happen for all trips. Not sure if there's more refactoring you want to do here or if you'd rather move this debugging to a different PR

lib/components/user/places/place-editor.tsx
miles-grant-ibigroup: I've clarified the comment in 36909e9b441bf6b9dc7bea471474ca1d80b39ee2. I'm not sure if this makes sense to do, but thought I could clarify what the `intl` object could possibly be used for one day

lib/components/user/monitored-trip/trip-basics-pane.tsx
binh-dam-ibigroup: I'm not seeing the blanket unavailability.

binh-dam-ibigroup: Again, remove the condition on original line 210, and revert original lines 203-208.

miles-grant-ibigroup: Resolved in ec4dd4ab

miles-grant-ibigroup:
Things seem to be working now

PR #591, opened 2022-04-14
Better screen reader support

miles-grant-ibigroup:
Some elements had issues with screen readers. This PR addresses these by adding relevant aria tags.

### TODO
- [ ] bikes allowed label must be labeled in its entirety and then the children `labelledby`
- [ ] the app-menu-icon should become a button, and then styled appropriately (don't forget that the lines will need a `width: 100%`)
- [ ] app menu is described strangely

### DIFF
--- a/i18n/en-US.yml
+++ b/i18n/en-US.yml
@@ -456,6 +456,7 @@ components:
     hideSettings: " Hide Settings"
   TabbedItineraries:
     optionNumber: "Option {optionNum, number}"
+    optionDetails: "Itinerary details"
     # Note to translator: This text is width-constrained.
     mustCallAhead: Must call ahead!
     fareCost: >
@@ -593,6 +594,7 @@ components:
   TripViewer:
     accessible: Accessible
     bicyclesAllowed: Allowed
+    description: Trip Viewer showing all stops and stop times on route with stops served in trip itinerary highlighted
     header: Trip Viewer
     routeHeader: "Route: <strong>{routeShortName}</strong> {routeLongName}"
     viewStop: View
--- a/i18n/en-US.yml
+++ b/i18n/en-US.yml
@@ -594,7 +594,7 @@ components:
   TripViewer:
     accessible: Accessible
     bicyclesAllowed: Allowed
-    description: Trip Viewer showing all stops and stop times on route with stops served in trip itinerary highlighted
+    description: The Trip Viewer highlights the subset of stops and stop times of a transit route traveled on the itinerary you selected.
     header: Trip Viewer
     routeHeader: "Route: <strong>{routeShortName}</strong> {routeLongName}"
     viewStop: View
--- a/lib/components/app/app-menu.tsx
+++ b/lib/components/app/app-menu.tsx
@@ -185,6 +185,7 @@ class AppMenu extends Component<
               ? intl.formatMessage({ id: 'components.AppMenu.closeMenu' })
               : intl.formatMessage({ id: 'components.AppMenu.openMenu' })
           }
+          aria-roledescription="button"
           className="app-menu-icon"
           onClick={this._togglePane}
           onKeyDown={this._togglePane}
--- a/lib/components/app/app-menu.tsx
+++ b/lib/components/app/app-menu.tsx
@@ -180,6 +180,7 @@ class AppMenu extends Component<
     return (
       <>
         <div
+          aria-expanded={isPaneOpen}
           aria-label={
             isPaneOpen
               ? intl.formatMessage({ id: 'components.AppMenu.closeMenu' })
--- a/lib/components/narrative/tabbed-itineraries.js
+++ b/lib/components/narrative/tabbed-itineraries.js
@@ -89,11 +89,11 @@ class TabButton extends Component {
     )
     return (
       <Button
-        aria-selected={isActive}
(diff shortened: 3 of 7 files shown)


i18n/en-US.yml
```diff
@@ -593,6 +594,7 @@ components:
   TripViewer:
     accessible: Accessible
     bicyclesAllowed: Allowed
+    description: Trip Viewer showing all stops and stop times on route with stops served in trip itinerary highlighted
```
binh-dam-ibigroup, line 597:
```suggestion
    description: The Trip Viewer highlights the subset of stops and stop times of a transit route traveled on the itinerary you selected.
```

lib/components/app/app-menu.tsx
```diff
@@ -180,11 +180,13 @@ class AppMenu extends Component<
     return (
       <>
         <div
+          aria-expanded={isPaneOpen}
           aria-label={
             isPaneOpen
               ? intl.formatMessage({ id: 'components.AppMenu.closeMenu' })
               : intl.formatMessage({ id: 'components.AppMenu.openMenu' })
           }
+          aria-roledescription="button"
```
binh-dam-ibigroup, line 189: The best thing would be to convert the `<div>` into a `<button>` and remove the extra roles.

lib/components/app/app-menu.tsx
```diff
@@ -180,11 +180,13 @@ class AppMenu extends Component<
     return (
       <>
         <div
+          aria-expanded={isPaneOpen}
```
binh-dam-ibigroup, line 183: Might need to associate this component with the sliding pane with `aria-controls` (non-blocking).

lib/components/narrative/tabbed-itineraries.js
```diff
@@ -88,9 +89,11 @@ class TabButton extends Component {
     )
     return (
       <Button
+        ariaRole="tab"
         className={classNames.join(' ')}
         key={`tab-button-${index}`}
         onClick={this._onClick}
+        tabIndex={isActive ? '0' : '-1'}
```
binh-dam-ibigroup, line 96: This would make inactive tabs unreachable from the tab key, right? Do we want to set the value to zero for all of them?

lib/components/util/icon.tsx
```diff
@@ -28,12 +32,20 @@ const FontAwesomeWithSpace = styled(FontAwesome)`
...
+  intl,
   type,
   withSpace = false,
   ...props
 }: IconProps): JSX.Element => {
   const FontComponent = withSpace ? FontAwesomeWithSpace : FontAwesome
-  return <FontComponent fixedWidth={fixedWidth} name={type} {...props} />
+  return (
+    <FontComponent
+      aria-label={intl ? getFormattedMode(type, intl, true) : ''}
```
binh-dam-ibigroup, line 43: Should we fallback to `type` if `intl` is absent?

lib/components/viewers/trip-viewer.js
```diff
@@ -91,7 +97,7 @@ class TripViewer extends Component {
                 <SpanWithSpace margin={0.25} />
                 {tripData.bikesAllowed === 1 && (
                   <Label bsStyle="success">
```
binh-dam-ibigroup, line 99: Would it make sense to place an `aria-label` for this entire component? (non-blocking)

binh-dam-ibigroup (changes requested):
Main thing is to convert the hamburger "button" to an actual `<button>` and make inactive itinerary tabs tab-able.

lib/components/narrative/tabbed-itineraries.js
miles-grant-ibigroup: I'm not sure... I followed https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role#example

lib/components/util/icon.tsx
miles-grant-ibigroup: Yes this was very much a temporary patch. I'm not sure however what the "correct" solution looks like. What would the fallback be?

lib/components/viewers/trip-viewer.js
miles-grant-ibigroup: Yes that might be a better option and then have the icon be `labelledby` the parent

lib/components/app/app-menu.tsx
```diff
@@ -180,11 +180,13 @@ class AppMenu extends Component<
...
           aria-label={
             isPaneOpen
               ? intl.formatMessage({ id: 'components.AppMenu.closeMenu' })
               : intl.formatMessage({ id: 'components.AppMenu.openMenu' })
           }
+          aria-roledescription="button"
           className="app-menu-icon"
           onClick={this._togglePane}
           onKeyDown={this._togglePane}
```
binh-dam-ibigroup, line 192: Won't hold this against this PR, but this traps the tab focus to the component (pressing tab or any key expands the side pane). Noting this as a new issue.

lib/components/narrative/tabbed-itineraries.js
binh-dam-ibigroup: Okay. Looks like the parent element needs a `role="tablist"` per your link above.

PR #605, opened 2022-05-11
Mapillary Tweaks and Cleanup

miles-grant-ibigroup:
The Mapillary code had an unlocalized string! That was removed.

Additionally, this PR makes use of https://github.com/opentripplanner/otp-ui/pull/401 to show a static image instead of an iframe. The iframe wasn't working with Blink.


daniel-heppner-ibigroup (approved):
It works, hope we don't need to use it!

PR #606, opened 2022-05-12
Localize overlay names

binh-dam-ibigroup:
This PR adds i18n support to overlay names (streets, P/R locations...) under the map layer dropdown menu.

Config-specific (non-i18n) overlay names are still supported using the existing `name` attribute for each overlay.

### DIFF
--- a/lib/components/map/default-map.js
+++ b/lib/components/map/default-map.js
@@ -232,7 +232,7 @@ class DefaultMap extends Component {
       location: [mapPopupLocation.lat, mapPopupLocation.lon]
     }
 
-    const baseLayersWithNames = mapConfig.baseLayers.map((baseLayer) => ({
+    const baseLayersWithNames = mapConfig.baseLayers?.map((baseLayer) => ({
       ...baseLayer,
       name: getLayerName(baseLayer, config, intl)
     }))
--- a/lib/components/map/default-map.js
+++ b/lib/components/map/default-map.js
@@ -47,6 +47,9 @@ const MapContainer = styled.div`
   }
 `
 
+/**
+ * Determines the localized name of a map layer by its type.
+ */
 function getLayerName(type, intl) {
   switch (type) {
     case 'streets':
@@ -65,6 +68,8 @@ function getLayerName(type, intl) {
       return intl.formatMessage({ id: 'components.MapLayers.park-and-ride' })
     case 'stops':
       return intl.formatMessage({ id: 'components.MapLayers.stops' })
+    default:
+      return null
   }
 }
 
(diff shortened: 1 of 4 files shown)


lib/components/map/default-map.js
```diff
@@ -59,41 +119,65 @@ class DefaultMap extends Component {
       // Determine any added/removed modes
       const oldModes = oldQuery.mode.split(',')
       const newModes = newQuery.mode.split(',')
-      const removed = oldModes.filter(m => !newModes.includes(m))
-      const added = newModes.filter(m => !oldModes.includes(m))
-      const overlayVisibility = {}
+      const removed = oldModes.filter((m) => !newModes.includes(m))
```
philip-cline, line 122: Should removing modes via the `ModeButtons` in the mode selector here: 
image
influence the added/removed modes? I don't seem to be able to change the values of `added` or `removed` and can't untoggle the park n ride layer automatically by removing the car mode for example. 
Route duplication bug

philip-cline (changes requested):
I may be confused about how to add/remove modes in the query params such that the relevant overlays are toggled, but I haven't been able to test this functionality properly. 

Otherwise it's working really well for me.

lib/components/map/default-map.js
binh-dam-ibigroup: This behavior is for the tabbed interface when you switch to/from the "Transit + Bikeshare" or "Transit + E-Scooter" travel mode. I don't think this behavior exists for the batch interface yet.

binh-dam-ibigroup: For the batch interface, the behavior does not really apply because the query parameters (almost) never include the rental modes. However, one thing we could do in the future to implement the behavior is to show the appropriate rental overlay if the corresponding mode is the only mode active in the mode selector.

binh-dam-ibigroup: I added a comment clarifying the behavior in https://github.com/opentripplanner/otp-react-redux/pull/606/commits/ba41be35d9aced13d5375c2544e525644918d480.

philip-cline (approved):
Behaviour looks good in the tabbed view. Thanks!

lib/components/map/default-map.js
```diff
@@ -44,6 +46,64 @@ const MapContainer = styled.div`
...
+        {
+          companies: getCompanyNames(overlay.companies, config, intl)
+        }
+      )
+    case 'park-and-ride':
+      return intl.formatMessage({ id: 'components.MapLayers.park-and-ride' })
+    case 'stops':
+      return intl.formatMessage({ id: 'components.MapLayers.stops' })
+    default:
+      return null
```
miles-grant-ibigroup, line 104: Do we really want the fallback to be null? Surely displaying the "raw" type is better than nothing at all?

lib/components/map/default-map.js
```diff
@@ -162,43 +258,55 @@ class DefaultMap extends Component {
...
           <ElevationPointMarker />
 
           {/* The configurable overlays */}
-          {mapConfig.overlays && mapConfig.overlays.map((overlayConfig, k) => {
+          {mapConfig.overlays?.map((overlayConfig, k) => {
+            const namedLayerProps = {
+              ...overlayConfig,
+              key: k,
+              name:
+                overlayConfig.name || getLayerName(overlayConfig, config, intl)
```
miles-grant-ibigroup, line 273: Would it be clearer what's going on if we put the `overlayConfig.name` inside `getLayerName` and replaced the `.name` default case?

miles-grant-ibigroup (approved):
Clever way to internationalize this! Just a few small nitpicks but nothing serious. Works great!

lib/components/map/default-map.js
binh-dam-ibigroup: Good one! Refactored in https://github.com/opentripplanner/otp-react-redux/pull/606/commits/1162dc3bc6d0532c5ef6ad387f9c61d193527718.

lib/components/map/default-map.js
binh-dam-ibigroup: Good point. In https://github.com/opentripplanner/otp-react-redux/pull/606/commits/1162dc3bc6d0532c5ef6ad387f9c61d193527718, it defaults to `overlay.type`. (if `overlay.name` is specified for any overlay, that still preempts the default i18n messages.

PR #607, opened 2022-05-13
Localize mode names

binh-dam-ibigroup:
This PR adds support for localizing travel mode names (transit modes, bike modes, micromobility).
Existing configs that define a `label` attribute for each travel mode continue to be supported.

Visually, the changes appear for labels in the transit mode selector (under the pinwheel button) and in the itinerary results (If configured accordingly, you can see results such as "Walk/Bike to MARTA Rail", "A pied + MARTA (métro)"... instead of the generic "Walk to Subway" in the title for each result and for the desired language).

### DIFF
--- a/lib/components/util/formatted-mode.js
+++ b/lib/components/util/formatted-mode.js
@@ -35,10 +35,9 @@ const FormattedMode = ({ mode }) => {
     case 'gondola':
       return <FormattedMessage id="common.modes.gondola" />
     case 'micromobility':
-    case 'scooter':
       return <FormattedMessage id="common.modes.micromobility" />
     case 'micromobility_rent':
-    case 'scooter_rent':
+    case 'scooter':
       return <FormattedMessage id="common.modes.micromobility_rent" />
     case 'rail':
       return <FormattedMessage id="common.modes.rail" />
--- a/lib/components/util/formatted-mode.js
+++ b/lib/components/util/formatted-mode.js
@@ -35,8 +35,10 @@ const FormattedMode = ({ mode }) => {
     case 'gondola':
       return <FormattedMessage id="common.modes.gondola" />
     case 'micromobility':
+    case 'scooter':
       return <FormattedMessage id="common.modes.micromobility" />
     case 'micromobility_rent':
+    case 'scooter_rent':
       return <FormattedMessage id="common.modes.micromobility_rent" />
     case 'rail':
       return <FormattedMessage id="common.modes.rail" />
--- a/lib/util/i18n.js
+++ b/lib/util/i18n.js
@@ -1,3 +1,4 @@
+import clone from 'clone'
 import flatten from 'flat'
 
 // deepmerge must be imported via `require`: see https://github.com/TehShrike/deepmerge#include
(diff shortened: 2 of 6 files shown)


lib/util/i18n.js
```diff
@@ -208,6 +211,62 @@ export function getFormattedMode(mode, intl) {
...
+      // TODO: support company names so this reads e.g. "Transit + BCycle"
+      return intl.formatMessage({ id: 'config.accessModes.bicycle_rent' })
+    case 'car_hail':
+      return intl.formatMessage({ id: 'config.accessModes.car_hail' })
+    case 'car_park':
+      return intl.formatMessage({ id: 'config.accessModes.car_park' })
+    case 'micromobility':
+      return intl.formatMessage({ id: 'config.accessModes.micromobility' })
+    case 'micromobility_rent':
+    case 'scooter_rent':
```
miles-grant-ibigroup, line 232:
```suggestion
    case 'scooter_rent':
    case 'scooter':
```

Still need to double check with Leonard but I believe `scooter_rent` does not exist, although it seems to have an effect when used in a query. However, `scooter` is the main scooter mode

lib/util/i18n.js
```diff
@@ -208,6 +211,62 @@ export function getFormattedMode(mode, intl) {
...
+      // This is the generic Bikeshare only.
+      // (If needed, add a separate bicycle mode with the desired label in the config for each rental company.)
+      return intl.formatMessage({ id: 'config.bicycleModes.bicycle_rent' })
+    case 'micromobility':
+    case 'scooter':
+      return intl.formatMessage({
+        id: 'config.micromobilityModes.micromobility'
+      })
+    case 'micromobility_rent':
+    case 'scooter_rent':
```
miles-grant-ibigroup, line 258: `scooter` should be moved here. Personal e-scooters do not exist in OTP2

miles-grant-ibigroup (changes requested):
I need to do some more OTP2 scooter testing, but other than that this looks great!

lib/util/i18n.js
binh-dam-ibigroup: `SCOOTER_RENT` is used in one of the configs. Should it be deleted from there?

miles-grant-ibigroup: It's been removed (as part of other scooter fixes) in https://github.com/ibi-group/configurations/pull/257

lib/components/util/formatted-mode.js
```diff
@@ -35,8 +35,10 @@ const FormattedMode = ({ mode }) => {
     case 'gondola':
       return <FormattedMessage id="common.modes.gondola" />
     case 'micromobility':
+    case 'scooter':
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Thanks for the clarification @miles-grant-ibigroup. I updated this and other instances in https://github.com/opentripplanner/otp-react-redux/pull/607/commits/2a881d96185162b4ef3c9cbb5e74469dc45b9e0e.

demory (approved):

PR #609, opened 2022-05-18
Refactor intl messages

binh-dam-ibigroup:
This PR separates some independent i18n messages that were previously combined in more complicated formatting logic, so they are easier to translate.


philip-cline (approved):
Good call, the way I had done it before was definitely not intuitive.

miles-grant-ibigroup (approved):
This is much cleaner thanks for the changes!

PR #611, opened 2022-05-25
Remove unused messages

binh-dam-ibigroup:
This PR removes unused messages (found using https://github.com/opentripplanner/otp-ui/pull/287)
and also makes sure the `LocationField` ones aren't being flagged as unused.


binh-dam-ibigroup:
> Do you think this is something we'll be doing often enough to make it worth it to add a package.json script for this?

Happens all the time. Yep, when the otp-ui scripts package is available we can start adding some of these checks to OTP-RR.

philip-cline (approved):
Cleans up the translations file really well. Thanks!

PR #612, opened 2022-05-25
fix(printable-itinerary): always print selected itinerary

miles-grant-ibigroup:
Due to the print button opening a new page, there was a chance that the `activeItinerary` could point to a new itinerary that wasn't the one the user selected.

This PR resolves this by opening the print view in the same window and adding a close button. The back button still works.


binh-dam-ibigroup (approved):
LGTM!

philip-cline (approved):
Stupendous. Working well for me.

PR #616, opened 2022-06-02
Fix from/to picker not populating location-field

miles-grant-ibigroup:
There were some bugs with the location field not being updated correctly when using the from/to picker in the stop viewer. 

I think this was due to some unnecessary reverse geocoding, which this PR removes.


binh-dam-ibigroup (approved):
LGTM

daniel-heppner-ibigroup (approved):

Issue #604, opened 2022-05-10
CallTaker can not add intermediate place


miles-grant-ibigroup:
Selecting an intermediate place currently does nothing. The trip request is not adjusted and the text box contents do not change. There seems to be some relevant FIXMEs in the code.

PR #617, opened 2022-06-02
CallTaker Regression Fixes

miles-grant-ibigroup:
Internationalization and OTP2 have created 2 bugs in the call taker UI.

The first the add intermediate stop being broken. #604
The second was the four time adjustment buttons not working.

Both have been fixed via simple patches

### DIFF
--- a/lib/actions/plan.js
+++ b/lib/actions/plan.js
@@ -1,24 +1,27 @@
-import {OTP_API_DATE_FORMAT, OTP_API_TIME_FORMAT} from '@opentripplanner/core-utils/lib/time'
-import {getTimeZoneOffset} from '@opentripplanner/core-utils/lib/itinerary'
+import { getTimeZoneOffset } from '@opentripplanner/core-utils/lib/itinerary'
+import {
+  OTP_API_DATE_FORMAT,
+  OTP_API_TIME_FORMAT
+} from '@opentripplanner/core-utils/lib/time'
 import moment from 'moment'
 
-import {getFirstStopId} from '../util/itinerary'
-import {getActiveItinerary} from '../util/state'
+import { getActiveItinerary } from '../util/state'
+import { getFirstStopId } from '../util/itinerary'
 
-import {routingQuery} from './api'
-import {setQueryParam} from './form'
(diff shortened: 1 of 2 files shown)


lib/actions/plan.js
```diff
@@ -96,17 +109,22 @@ export function planNext () {
...
-      date: moment(currentQuery.date).add('days', 1).format(OTP_API_DATE_FORMAT),
+      date: moment(currentQuery.date)
+        .add('days', 1)
+        .format(OTP_API_DATE_FORMAT),
       departArrive: 'ARRIVE',
-      startTransitStopId: getFirstStopId(itinerary),
       time: SERVICE_BREAK
     }
+    if (!state.otp?.config?.api?.v2) {
+      params.startTransitStopId = getFirstStopId(itinerary)
```
daniel-heppner-ibigroup, line 125: This if statement is repeated several times, could it just go into the `getFirstStopId` function?

daniel-heppner-ibigroup (approved):

lib/actions/plan.js
```diff
@@ -28,30 +31,32 @@ function offsetTime (itinerary, unixTime) {
...
     if (isPlanningFirst(currentQuery)) date.subtract('days', 1)
     const params = {
       date: date.format(OTP_API_DATE_FORMAT),
       departArrive: 'DEPART',
-      startTransitStopId: getFirstStopId(itinerary),
       time: SERVICE_BREAK
     }
+    if (!state.otp?.config?.api?.v2) {
+      params.startTransitStopId = getFirstStopId(itinerary)
+    }
```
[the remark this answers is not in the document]
miles-grant-ibigroup: OTP2 doesn't support `startTransitStopId` and if you provide it you get no itineraries!

lib/actions/plan.js
miles-grant-ibigroup: I tried to make this work but the problem is the field can't be set at all and even undefined creates a value!

binh-dam-ibigroup (approved):

miles-grant-ibigroup:
Assigning to @demory so you can have a look

PR #329, opened 2021-02-10
Harmonize user data across persistence strategies

binh-dam-ibigroup:
This PR exposes the logged-in user saved places and recent searches in `DefaultMainPanel` and `BatchRoutingPanel`, and in the `LocationField` drop-downs in the main panel.

Behind the scenes:
- Some new place components are introduced:
  - `<Place>` is the generic rendering component, takes a `name`, `details` and `onClick`, `onView`, `onDelete` event handlers. Variants of this component are used in the main panel and in the user's settings page.
- Only the saved/favorite locations and recent trip requests for the persistence strategy defined in `config.yml` are shown. Note: recent trip requests for an OTP Middleware user cannot be deleted.
- ~Only the first OTP web request of a batch request is logged for middleware users, to avoid repeat entries in recent searches.~ => No longer applies, a `batchId` must now be sent on every trip request to OTP after the change from https://github.com/ibi-group/otp-middleware/pull/152.
- "Starring" favorite stops from the stop viewer is disabled for logged-in middleware users.
- The user-related reducer code is moved to `create-user-reducer`.
- i18n is implemented, except for content rendered by `LocationField`.

This PR supersedes #283.

Screenshot for the `otp_middleware` persistence strategy:

image

PR #620, opened 2022-06-09
fix(actions/form): Use correct local user default query params

### LINKED ISSUE #619: UI crashes when opening the Field Trip window
The UI crashes when opening the field trip window (bug introduced after merging #329).

binh-dam-ibigroup:
Fix #619 

This PR removes the last uses of `state.otp.user` and replaces them with `state.user.localUser` introduced by #329.


binh-dam-ibigroup:
Reverting to draft (found another crash)

miles-grant-ibigroup (approved):
Glad the fix was this simple!

demory (approved):

PR #622, opened 2022-06-09
fix(transitive): fix scooter uuid with otp2

daniel-heppner-ibigroup:
Uses https://github.com/opentripplanner/otp-ui/pull/411 to fix the UUID display bug with OTP2 and micromobility.


daniel-heppner-ibigroup:
This is going to fail until OTP UI packages are updated.

daniel-heppner-ibigroup:
@binh-dam-ibigroup, I'll give this to you for first review since Miles is out.

package.json
```diff
@@ -36,13 +36,13 @@
...
     "@opentripplanner/base-map": "^2.0.0",
-    "@opentripplanner/core-utils": "^5.0.2",
+    "@opentripplanner/core-utils": "^6.0.0",
     "@opentripplanner/endpoints-overlay": "^1.4.1",
     "@opentripplanner/from-to-location-picker": "^2.1.2",
     "@opentripplanner/geocoder": "^1.2.1",
     "@opentripplanner/humanize-distance": "^1.2.0",
     "@opentripplanner/icons": "^1.2.2",
-    "@opentripplanner/itinerary-body": "^3.0.5",
+    "@opentripplanner/itinerary-body": "^4.0.0",
```
binh-dam-ibigroup, line 45: Hmm see this? Not a big deal, but for next time, try to limit breaking changes commits to just the affected packages (should have been just for core-utils, not for itinerary or transitive).

lib/util/state.js
```diff
@@ -653,19 +653,14 @@ export const getTransitiveData = createSelector(
...
-        return otpResponse
-      }
+    return {
+      hasResponse,
+      otpResponse,
+      hasItineraryResponse,
+      itineraryToRender,
+      companies,
+      disableFlexArc,
+      getTransitiveRouteLabel
```
binh-dam-ibigroup, line 663: Sort attributes

lib/components/map/connected-transitive-overlay.tsx
```diff
@@ -0,0 +1,94 @@
...
+            companies,
+            disableFlexArc,
+            getTransitiveRouteLabel,
+            intl
+          })
+        : null
+    } else if (otpResponse) {
+      transitiveData = otpResponse
+    }
+  }
```
binh-dam-ibigroup, line 59: Could this be moved to `mapStateToProps`? (e.g. by using `injectIntl` instead of `useIntl`, so that `intl` becomes an `ownProp`?) - This is to avoid repeated calculations of the transitive content on each render.

binh-dam-ibigroup: Alternately, use the old code from file `state.js` but add an `intl` argument from `injectIntl` to `getTransitiveData` that you pass to `itineraryToTransitive`.

binh-dam-ibigroup (commented):
@daniel-heppner-ibigroup What do you think about using `injectIntl` to compute the transitive content in `state.js` (see comments)?

package.json
daniel-heppner-ibigroup: Yeah, I noticed that too late... I think it was because I had to break apart the PR after the fact with cherry picks and reverts.

lib/components/map/connected-transitive-overlay.tsx
daniel-heppner-ibigroup: This is some good thinking, thank you for that! I will refactor it.

daniel-heppner-ibigroup:
Need to add an ts expect error somewhere to fix the failing test- will do tomorrow.

daniel-heppner-ibigroup:
Ready to go here, @binh-dam-ibigroup.

lib/util/state.js
```diff
@@ -644,24 +645,26 @@ export const getTransitiveData = createSelector(
...
     if (hasResponse) {
       if (hasItineraryResponse) {
         return itineraryToRender
-          ? coreUtils.map.itineraryToTransitive(
-              itineraryToRender,
-              companies,
+          ? itineraryToTransitive(itineraryToRender, {
+      companies,
+      disableFlexArc,
               getTransitiveRouteLabel,
```
binh-dam-ibigroup, line 665: Fix indent.

binh-dam-ibigroup (approved):
Seems to work with the changes.

miles-grant-ibigroup (approved):
Works good in most cases and implementation looks good too!

One bug I found, but I think it's best to address that in a new PR. Thanks for resolving this! Those uuids have been bugging me for a while.

lib/components/map/connected-transitive-overlay.tsx
```diff
@@ -18,11 +28,13 @@ const mapStateToProps = (state, ownProps) => {
     return {}
   }
 
-  return {
+  const obj = {
     labeledModes,
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: I think this isn't a conventional UUID which is why it isn't getting filtered out. In a separate PR I will expand the definition of UUID.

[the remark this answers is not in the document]
miles-grant-ibigroup: Ah that makes sense. That sounds great thanks so much!

PR #623, opened 2022-06-09
OTP custom sub-modes labeling

### LINKED ISSUE #621: Need to render a different name/content for different transit services that use the same OTP mode.
- For TriMet, "MAX" or "Portland Streetcar" should be shown in batch itinerary results instead of the generic "Tram" or equivalent for the `TRAM` OTP mode.
- Some agencies may want to display "Local bus" or "Express bus" in batch itinerary result headers depending on the dominant bus route (OTP mode: `BUS`).

binh-dam-ibigroup:
Fix #621.
Requires/blocked by #620. 

This PR adds support for a custom `TransitModes` component that renders a mode name based on an OTP itinerary leg.


philip-cline (commented):
Works well for me!

philip-cline (approved):

lib/components/util/formatted-mode.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * JSX wrapper for the i18n.getFormattedMode function (renders into a React fragment).
+ */
+const FormattedMode = ({ mode = '' }: Props): ReactElement => {
+  const intl = useIntl()
+  return <>{getFormattedMode(mode, intl)}</>
+}
+
+export default FormattedMode
```
daniel-heppner-ibigroup, line 18: What if we moved this into the file where getFormattedMode() is? Or are we keeping the components separate?

daniel-heppner-ibigroup (approved):
LGTM, just wondering if we could consolidate some files?

lib/components/util/formatted-mode.tsx
binh-dam-ibigroup: Or should we just get rid of this component?

binh-dam-ibigroup: No, that would add extra lines for adding `useIntl()`. It's better to keep components separate imo.

PR #626, opened 2022-06-10
Render times in agency local time

### LINKED ISSUE #625: Incorrect times if config.homeTimezone is different timezone than user
Querying an itinerary in a different time zone than the user's time zone results in these bugs:
- Trip time is not correctly converted in the query parameters after clicking "leave now"
- Leave/Depart at <time> text is not shown in the correct time zone
- Times in the itinerary narratives are shifted and do not correspond to the trip time in the query parameters.
- Times in the stop viewer are shifted and do not correspond to the actual schedule.

binh-dam-ibigroup:
Fix #625.

In the Stop Viewer we show an indication if the agency timezone is different than the user time zone. Should that be shown in itinerary results too?

### DIFF
--- a/lib/components/narrative/default/default-itinerary.js
+++ b/lib/components/narrative/default/default-itinerary.js
@@ -108,9 +108,7 @@ const ITINERARY_ATTRIBUTES = [
     alias: 'best',
     id: 'duration',
     order: 0,
-    render: (itinerary /*, options */) => (
-      <FormattedDuration duration={itinerary.duration} />
-    )
+    render: (itinerary) => <FormattedDuration duration={itinerary.duration} />
   },
   {
     alias: 'departureTime',
--- a/lib/components/narrative/default/default-itinerary.js
+++ b/lib/components/narrative/default/default-itinerary.js
@@ -1,8 +1,8 @@
 // This is a large file being touched by open PRs. It should be typescripted
 // in a separate PR.
-/* eslint-disable */
-import coreUtils from '@opentripplanner/core-utils'
-import React from 'react'
+/* eslint-disable react/prop-types */
+import { AccessibilityRating } from '@opentripplanner/itinerary-body'
+import { connect } from 'react-redux'
 import {
   FormattedList,
   FormattedMessage,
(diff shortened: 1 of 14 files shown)


lib/components/narrative/default/default-itinerary.js
```diff
@@ -88,20 +93,22 @@ function ItineraryDescription({ itinerary }) {
...
+  )
 }
 
 const ITINERARY_ATTRIBUTES = [
   {
     alias: 'best',
     id: 'duration',
     order: 0,
-    render: (itinerary, options) => (
+    render: (itinerary /*, options */) => (
```
philip-cline, line 111: Should we remove this rather than comment if we're not using options anymore?

philip-cline (approved):
Just one small comment, otherwise looks good!

lib/components/narrative/default/default-itinerary.js
binh-dam-ibigroup: Yeah I removed them in https://github.com/opentripplanner/otp-react-redux/pull/626/commits/93ce1d67362de7e7a11478c731c5283bba640837.

lib/components/viewers/stop-schedule-table.tsx
```diff
@@ -1,5 +1,7 @@
-import { FormattedMessage, FormattedTime } from 'react-intl'
-import moment from 'moment'
+import { FormattedMessage } from 'react-intl'
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore TYPESCRIPT TODO: wait for typescripted core-utils
```
miles-grant-ibigroup, line 3: I believe we can remove this comment, right? We've upgraded core-utils

lib/components/viewers/stop-time-cell.tsx
```diff
@@ -1,16 +1,20 @@
-import 'moment-timezone'
-import { FormattedMessage, FormattedTime } from 'react-intl'
+import { format, getTimezoneOffset, utcToZonedTime } from 'date-fns-tz'
+import { FormattedMessage } from 'react-intl'
+import addDays from 'date-fns/addDays'
 // TODO: typescript core-utils, add common types (pattern, stop, configs)
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore
```
miles-grant-ibigroup, line 6: Same as above -- I believe we can remove this line now that core-utils has been upgraded.

package.json
```diff
@@ -65,6 +65,8 @@
     "currency-formatter": "^1.4.2",
     "d3-selection": "^1.3.0",
     "d3-zoom": "^1.7.1",
+    "date-fns": "^2.23.0",
```
miles-grant-ibigroup, line 68: Annoying to have both `date-fns` and `moment` imported... Perhaps we can look at transitioning our remaining moment calls to `date-fns`? Although that doesn't have to happen in this PR.

miles-grant-ibigroup (commented):
This looks very clean and works really great thanks for putting this together! 

Looks like the percy tests are failing. I've checked the screenshots and the changes seem to make sense so I'm happy for you to approve the new screenshots (if you're unable to let me know and I can approve them/give you proper access)

Finally I agree that the banner warning on the itinerary body makes sense, but I am worried about adding clutter to an already busy interface. I think we can get away with it though, as a user using "plan now" will understand that the time zone is different. A user using "arrive at" or "leave at" might get confused, but the solution there is to provide a warning next to that input. However, since we are redesigning those fields anyway I think it makes most sense to leave that for the redesign!

package.json
binh-dam-ibigroup: We discussed that and agreed on phasing out moment.js in the technical call last week.

lib/components/viewers/stop-schedule-table.tsx
miles-grant-ibigroup: Apologies -- this has been done in the batch ui redesign, but I forgot that that hasn't been merged yet.

binh-dam-ibigroup:
> This looks very clean and works really great thanks for putting this together!
> 
> Looks like the percy tests are failing. I've checked the screenshots and the changes seem to make sense so I'm happy for you to approve the new screenshots (if you're unable to let me know and I can approve them/give you proper access)
> 
> Finally I agree that the banner warning on the itinerary body makes sense, but I am worried about adding clutter to an already busy interface. I think we can get away with it though, as a user using "plan now" will understand that the time zone is different. A user using "arrive at" or "leave at" might get confused, but the solution there is to provide a warning next to that input. However, since we are redesigning those fields anyway I think it makes most sense to leave that for the redesign!

Looking at the percy screenshots, the times in the itinerary results are earlier than the "depart at" time so I'll update the depart at time accordingly.

miles-grant-ibigroup (approved):
Looks great! Thanks for investigating that percy issue

PR #627, opened 2022-06-13
Field trip change travel date

binh-dam-ibigroup:
Requires/blocked by #620.
Fix https://github.com/ibi-group/trimet-mod-otp/issues/307
Also contains a refactor for the code that updates attributes on a field trip.

For easier review, look at the changes before the one about lint.


lib/components/admin/field-trip-details.js
```diff
@@ -73,9 +82,16 @@ class FieldTripDetails extends Component {
...
-    const convertedRequestDate = moment(evt.target.value).format(dateFormat)
-    setRequestDate(request, convertedRequestDate, intl)
+    const newDate = evt.target.value
+    // Only persist the date if it is valid (not an empty string)
+    if (newDate !== '') {
+      const { dateFormat, intl, request, setRequestDate } = this.props
+      const convertedRequestDate = moment(newDate).format(dateFormat)
+      setRequestDate(request, convertedRequestDate, intl)
+      this.setState({
+        travelDateUpdateCount: this.state.travelDateUpdateCount + 1
```
miles-grant-ibigroup, line 92: Do we use this count for anything else than updating the key? If not, would using `uuid()` not make more sense? This is what we do elsewhere when we have this problem -- the library is already in the `yarn.lock`

lib/actions/field-trip.js
```diff
@@ -122,22 +206,23 @@ export function fetchFieldTripDetails (requestId, intl) {
...
     const noteData = serialize({
-      note: {...note, userName: username},
+      note: { ...note, userName: username },
       requestId: request.id,
       sessionId
     })
-    return fetch(`${datastoreUrl}/fieldtrip/addNote`,
-      {body: noteData, method: 'POST'}
-    )
+    return fetch(`${datastoreUrl}/fieldtrip/addNote`, {
```
miles-grant-ibigroup, line 220: Is this fetch not able to use the new request method?

lib/actions/field-trip.js
```diff
@@ -149,192 +234,114 @@ export function addFieldTripNote (request, note, intl) {
...
-            { id: 'actions.fieldTrip.editSubmitterNotesError' },
-            { err: JSON.stringify(err) }
-          )
-        )
-      })
-  }
+export function deleteFieldTripNote(request, noteId, intl) {
+  return updateFieldTripRequest(
+    request,
+    'deleteNote',
```
miles-grant-ibigroup, line 274: Do you think it might be a good idea to make these an enum to avoid room for error when editing this in the future

miles-grant-ibigroup (changes requested):
Don't have much experience with the calltaker module, but changing the date seems to work and the linted code looks great! Thanks for handling that linting these huge files is so annoying.

lib/actions/field-trip.js
binh-dam-ibigroup: Handled in https://github.com/opentripplanner/otp-react-redux/pull/627/commits/42a8fb2c0864378234354ceb9827ddf380ac03bc.

lib/actions/field-trip.js
binh-dam-ibigroup: Handled in https://github.com/opentripplanner/otp-react-redux/pull/627/commits/42a8fb2c0864378234354ceb9827ddf380ac03bc.

binh-dam-ibigroup:
> 
It did when clearing one of the date fields then selecting a different field trip with the same original travel date! I fixed that in https://github.com/opentripplanner/otp-react-redux/pull/627/commits/cf6df2005fe5aebcc8324738f7e5c56ddde4976f.

miles-grant-ibigroup (approved):
Thanks for the changes! Still works great. 

Once the merge conflicts are resolved this is ready to go!

demory:
I'm getting merge conflicts too when I try to test this -- what is the status of those?

binh-dam-ibigroup:
> I'm getting merge conflicts too when I try to test this -- what is the status of those?

Github says the PR is mergeable... I'll take a look at what you have.

demory (approved):

PR #628, opened 2022-06-21
Fix calltaker trip count disparity

demory:
Resolves https://github.com/ibi-group/trimet-mod-otp/issues/308

The Calltaker UI now expresses field trips groups as consisting of separate "under 7" and "7 and over" categories, with the under-7 group traveling free of charge. Previously, this was expressed as "total students" and "free students", with the "free" group being a subset of the "total" group; the database still uses this specification and we would like to avoid changing the underlying schema if possible.

This PR addresses a client-side issue where the "numStudents" (i.e. total) field was incorrectly interpreted/displayed as the "7 and over" count. To calculate the size of 7+ (i.e. "paid") group we should instead subtract the number of "free" students from the "total".


lib/actions/field-trip.js
```diff
@@ -945,29 +946,31 @@ export function viewRequestTripItineraries(request, outbound) {
...
  * numFreeStudents, and numChaperones.
  */
-export function setRequestGroupSize (request, groupSize, intl) {
-  // rewrite to format used by database
-  groupSize = {
-    numStudents: groupSize.numPaidStudents + groupSize.numFreeStudents,
+export function setRequestGroupSize(request, groupSize, intl) {
+  // rewrite to format used by database, where numStudents is the total number of
+  // students (free and paid) and numFreeStudents is the subset that rides free
+  const rewittenGroupSize = {
```
binh-dam-ibigroup, line 952: Typo "rewritten" (mutliple instances)

binh-dam-ibigroup (approved):
See typo, and also fix the codespell error.

miles-grant-ibigroup (approved):
Thanks for figuring out this fix! Merging in dev should hopefully fix that code spell error.

PR #629, opened 2022-06-21
Fix calltaker map crash when planning scooter trips

demory:
Fixes white-screen-of-death crash when attempting to plan a scooter+transit trip in the calltaker module.


lib/components/map/default-map.js
```diff
@@ -145,7 +145,7 @@ class DefaultMap extends Component {
           const overlayCompany = oConfig.companies[0] // TODO: handle multi-company overlays
           if (added.includes(overlayMode)) {
             // Company-based mode was just selected; enable overlay iff overlay's company is active
-            if (newQuery.companies.includes(overlayCompany)) {
+            if (newQuery.companies && newQuery.companies.includes(overlayCompany)) {
```
binh-dam-ibigroup, line 148: You can also use the shorthand notation below:
```suggestion
            if (newQuery.companies?.includes(overlayCompany)) {
```

binh-dam-ibigroup (approved):
LGTM

miles-grant-ibigroup (approved):
This looks great and is to me the best way to resolve this bug. 

I agree with Binh about the shorthand notation.

demory:
@miles-grant-ibigroup / @binh-dam-ibigroup thanks! Added shorthand per comments. This ok to merge now?

miles-grant-ibigroup:
Just have to approve the percy changes and then we should be good to merge yes!

PR #630, opened 2022-06-21
Add i18n checks

binh-dam-ibigroup:
This PR adds checks for unused and untranslated messages in OTP-RR.


daniel-heppner-ibigroup (approved):
No shortage of i18n keys missing, but thanks for adding the test!

PR #634, opened 2022-07-13
Field Trip Metro UI Accomodations 

miles-grant-ibigroup:
The Metro UI introduced a few regressions in the field trip module. This PR addresses them.

Most notably, the merging of itineraries with the same legs but different departure times is now only enabled via a configuration option.

### DIFF
--- a/example-config.yml
+++ b/example-config.yml
@@ -287,8 +287,10 @@ itinerary:
   customBatchUiBackground: true
   # Whether to render itineraries below a mode description header
   groupByMode: true
-  # Whether to merge itineraries with the same origin, destination, and transit modes
+  # Whether to merge groups of itineraries with the same primary mode (Bike, Walk, etc)
   groupTransitModes: true
+  # Whether to merge itineraries with the same origin, destination, and transit modes
+  mergeItineraries: true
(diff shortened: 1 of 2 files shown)


demory:
@miles-grant-ibigroup -- I have this checked out and running locally and it seems to have addressed the issues in question, but I now am unable to click to expand planned itineraries. Clicking an itinerary fires a SET_ACTIVE_ITINERARY event but the detailed itinerary view does not appear. Are you able to recreate?
image

miles-grant-ibigroup:
This took a while to figure out -- the bug was a missing itinerary index, which in some cases failed to be assigned. I've repaired this by assigning it when it's missing.

demory:
Still seeing specific issue with a blank itinerary, but only when selecting the first itinerary for the first time after planning a request: 
image

miles-grant-ibigroup:
This problem is resolvable by setting`itinerary.showFirstResultByDefault` to `true` in the config. Field trip uses this behavior anyway, so perhaps it's just a matter of automatically setting it to true if field trip is enabled? Or do you think that's more confusing than helpful

demory (approved):

philip-cline (approved):
Works well for me after merging the required PRs!

PR #638, opened 2022-07-19
Remove velocity

daniel-heppner-ibigroup:
Velocity-react library is no longer maintained. I've replaced it with a very simple new library called react animate height that has types, hooks, and is being updated. https://www.npmjs.com/package/react-animate-height

### DIFF
--- a/lib/components/viewers/RouteRow.js
+++ b/lib/components/viewers/RouteRow.js
@@ -121,15 +121,15 @@ export class RouteRow extends PureComponent {
     }
   }
 
+  // Used to ensure details are visible until animation completes
+  setNewHeight = (newHeight) => {
+    this.setState({ newHeight })
+  }
+
   render() {
     const { intl, isActive, operator, route } = this.props
     const { newHeight } = this.state
     const { ModeIcon } = this.context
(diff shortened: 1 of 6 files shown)


daniel-heppner-ibigroup:
Percy is catching the mysterious disappearing stop icons.

lib/components/viewers/RouteRow.js
```diff
@@ -122,8 +123,14 @@ export class RouteRow extends PureComponent {
 
   render() {
     const { intl, isActive, operator, route } = this.props
+    const { newHeight } = this.state
     const { ModeIcon } = this.context
 
+    // Used to ensure details are visible until animation completes
+    const setNewHeight = (newHeight) => {
+      this.setState({ newHeight })
+    }
```
binh-dam-ibigroup, line 132: Move this to a class method.

binh-dam-ibigroup (approved):
Just one small change but otherwise looks good.

miles-grant-ibigroup (approved):
No noticeable difference and bundle size is down (a tiny bit)

Thanks for the changes!

PR #639, opened 2022-07-19
Remove eslint disable

daniel-heppner-ibigroup:
This is mostly just reordering a few huge files so that functions aren't used before they're defined, and also removed eslint-disable comments from them. I also replaced `ts-disable` with `ts-expect-error`, which doesn't require an eslint override and also will throw an error if, in the future, it is no longer needed. (Redundant `ts-expect-error` statements generate an error if there is no error)

### DIFF
--- a/lib/actions/form.js
+++ b/lib/actions/form.js
@@ -1,3 +1,4 @@
+// TODO: Typescript
 import { createAction } from 'redux-actions'
 import coreUtils from '@opentripplanner/core-utils'
 import debounce from 'lodash.debounce'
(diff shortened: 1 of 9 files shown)


miles-grant-ibigroup (approved):
Everything still works thank you

lib/actions/form.js
```diff
@@ -1,5 +1,3 @@
-// TODO: Typescript
```
philip-cline, line 1: Should the Typescript TODO stand?

philip-cline (approved):
Just one quick question, works and looks good to me.

lib/actions/form.js
daniel-heppner-ibigroup: Good catch!

PR #640, opened 2022-07-19
refactor(mailables): code split PDFKit dependency

daniel-heppner-ibigroup:
Ended up just splitting out the PDFKit dependency in this branch. This was the biggest dependency we had in our bundle, you can see it here in the analysis from after it was split (it is in a separate bundle).
image


lib/util/mailables.js
```diff
@@ -1,64 +1,66 @@
 import blobStream from 'blob-stream'
 import moment from 'moment'
 
-import PDFDocument from './PDFDocumentWithTables'
```
daniel-heppner-ibigroup, line 4: remove import

lib/util/mailables.js
```diff
@@ -69,14 +71,16 @@ function writePDF (formData, imageData, otpConfig) {
...
-  const {horizontalMargin, verticalMargin} = otpConfig
+  const { horizontalMargin, verticalMargin } = otpConfig
   const margins = {
     bottom: verticalMargin,
     left: horizontalMargin,
     right: horizontalMargin,
     top: verticalMargin
   }
-  const doc = new PDFDocument({margins})
+  const { default: PDFDocument } = await import('./PDFDocumentWithTables')
```
daniel-heppner-ibigroup, line 81: async import here

daniel-heppner-ibigroup:
This is a bit hard to test... You'll need TriMet calltaker config, but comment out the datastore section of the config so you don't get asked for a log in. The mailables button is in the sidebar nav.

miles-grant-ibigroup (approved):
Seeing a .3mb drop in final post-compression bundle size for the main js! Still well above any reasonable limit, but getting somewhere! Thanks for the changes.

daniel-heppner-ibigroup:
> Seeing a .3mb drop in final post-compression bundle size for the main js! Still well above any reasonable limit, but getting somewhere! Thanks for the changes.

Sounds about right :)

binh-dam-ibigroup (approved):
LGTM! Some conflicts to resolve as a result of merging #637.

binh-dam-ibigroup:
> Seeing a .3mb drop in final post-compression bundle size for the main js! Still well above any reasonable limit, but getting somewhere! Thanks for the changes.

Is the pdfkit package still included when building the calltaker version?

daniel-heppner-ibigroup:
> > Seeing a .3mb drop in final post-compression bundle size for the main js! Still well above any reasonable limit, but getting somewhere! Thanks for the changes.
> 
> Is the pdfkit package still included when building the calltaker version?

It is always "included" but is packaged in a separate bundle.js, which is only loaded when it's required by the app. I've tested the mailables window and the PDF generation still works with this change.

PR #643, opened 2022-07-26
Add Configurable Modal Popup Support

miles-grant-ibigroup:
This PR adds support for showing a pop-up when one of a series of trigger links is clicked. The configuration allows for many possible combinations. 

Here is a config snippet that shows off the full functionality:
```yaml
popups:
  targets:
    testTarget: 
      url: http://example.com/
    google: 
      url: http://google.com/
      appendLocale: false
  launchers:
    toolbar: google
    optionFilter: testTarget
    itineraryFooter: testTarget
    sidebarLink: google

language:
  en-US:
    config:
      popups:
        testTarget: Test Target!
        google: Google Search
```


lib/components/app/popup.tsx
```diff
@@ -0,0 +1,54 @@
...
+  }, [isMobile, compiledUrl])
+
+  if (isMobile) return null
+
+  return (
+    <Modal
+      dialogClassName="fullscreen-modal"
+      onHide={() => {
+        hideModal()
+      }}
```
binh-dam-ibigroup, line 42:
```suggestion
      onHide={hideModal}
```

lib/components/app/popup.tsx
```diff
@@ -0,0 +1,54 @@
...
+      show={shown}
+    >
+      <iframe
+        src={compiledUrl}
+        style={{ height: '100%', width: '100%' }}
```
binh-dam-ibigroup, line 47: Move the inline styles for the iframe into the CSS.

lib/reducers/create-otp-reducer.js
```diff
@@ -991,6 +992,19 @@ function createOtpReducer(config) {
             }
           }
         })
+      case 'SET_POPUP_CONTENT':
+        const target = state.config?.popups?.targets?.[action.payload]
```
binh-dam-ibigroup, line 996: Are there cases where `state.otp.config` could be null? (That's one of the first things loaded on startup)

example-config.yml
```diff
@@ -369,6 +369,27 @@ itinerary:
 #       - WALK
 #       - BICYCLE
 
+### Popup config
+### OTP-RR supports a full-screen momdal popup that can be launched from a set of
+### pre-set locations. Each button can be configured to open a different popup.
+### Popup "targets" define a URL that is opened. Their ID is used to give them names
+### via the language section below. For example, a target with the id "testTarget"
+### can be named with the i18n string "popups.testTarget"
```
binh-dam-ibigroup, line 377: Add an example of popup i18n in the `language` section below.

example-config.yml
```diff
@@ -369,6 +369,27 @@ itinerary:
...
+### Each popup button location can have a target assigned to it
+# popups:
+#   targets:
+#     testTarget: # To set a name for this target, create the i18n string "popups.testTarget"
+#       url: http://example.com/ 
+#     secondTarget: 
+#       url: http://test.com/
+#       appendLocale: false # Defaults to true, if set to false otp-rr will not append the current i18n language to the popup URL.
+#   launchers:
+#     toolbar: testTarget # This button is rendered in the top right of the toolbar
```
binh-dam-ibigroup, line 388: Complete the comment by saying that the button is only shown on desktop view.

lib/components/app/popup.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  const isMobile = coreUtils.ui.isMobile()
+  // appendLocale is true by default, so undefined is true
+  const compiledUrl = `${url}${
+    appendLocale !== false ? intl.defaultLocale : ''
+  }`
+
+  useEffect(() => {
+    if (isMobile) {
+      window.open(compiledUrl)
```
binh-dam-ibigroup, line 31: This does not work on Chrome/ium (either mobile view from devTools or on my phone). I suspect that's because these browsers block popups by default. An alternate approach for the mobile view is to pass an `href` and `target="_blank"` props to the buttons (make them react-bootstrap), or directly render `<a>` with these props and desired styling, see https://react-bootstrap-v3.netlify.app/components/buttons/#buttons-tags.

lib/components/app/popup.tsx
```diff
@@ -0,0 +1,54 @@
...
+      dialogClassName="fullscreen-modal"
+      onHide={() => {
+        hideModal()
+      }}
+      show={shown}
+    >
+      <iframe
+        src={compiledUrl}
+        style={{ height: '100%', width: '100%' }}
+        title={intl.formatMessage({ id: `popups.${id}` })}
```
binh-dam-ibigroup, line 48: Change that to `config.popups.${id}`, so it is consistent with other localizable text that is defined in config.yml.

example-config.yml
binh-dam-ibigroup: Change the nomenclature to `config.popups.testTarget`, so it is consistent with other localizable text that is defined in config.yml.

lib/components/app/app.css
```diff
@@ -84,3 +84,12 @@
 .view-switcher button.btn-link:hover {
   color: #fff;
 }
+
+/* Full screen modal styling */
+.fullscreen-modal {
+  width: 90vw;
+  height: 100vh;
```
binh-dam-ibigroup, line 91: Replace this one with the `90vh` defined below, and remove that subcontent CSS.

binh-dam-ibigroup: Then you can add the CSS for the iframe element here too.

binh-dam-ibigroup (changes requested):
For some reason the popup does not work on the mobile view, could you fix that please, along with other various comments?

lib/reducers/create-otp-reducer.js
miles-grant-ibigroup: I was being better safe than sorry but you're right `config` is in the initializer so it should be fine

example-config.yml
miles-grant-ibigroup: That's a very good idea thanks!

lib/components/app/popup.tsx
miles-grant-ibigroup: I've investigated this and the popups aren't being blocked (as they're being initiated by user action). What is going wrong is the `isMobile` check, which always returns false even on real mobile devices. I'm going to try to find a different mobile check that's more reliable.

lib/components/app/app.css
miles-grant-ibigroup: `.fullscreen-modal` is the dark background, `.fullscreen-modal .modal-content` is the actual modal. I don't see how these rules can be collapsed into one?

binh-dam-ibigroup: Oh really, then why would you want to override the size of the dark background?

miles-grant-ibigroup:
Thanks Binh, all changes are addressed.

lib/components/app/popup.tsx
miles-grant-ibigroup: Update: the issue wasn't the mobile check, but rather that `Popup` wasn't being rendered in the mobile view at all. Now that it is the regular `window.open` works, like it does in other parts of otp-rr

lib/components/app/responsive-webapp.js
```diff
@@ -194,9 +195,16 @@ class ResponsiveWebapp extends Component {
...
     const { sessionTimeoutSeconds } = this.props
     const { MainControls, MainPanel, MapWindows } = this.context
+    const { popupContent, setPopupContent } = this.props
     return (
       <div className="otp">
         <DesktopNav />
+        {popupContent && (
+          <PopupWrapper
+            content={popupContent}
+            hideModal={() => setPopupContent(null)}
```
binh-dam-ibigroup, line 205: Maybe extract `() => setPopupContent(null)` into a class method.

lib/components/app/responsive-webapp.js
```diff
@@ -194,9 +195,16 @@ class ResponsiveWebapp extends Component {
   renderDesktopView = () => {
     const { sessionTimeoutSeconds } = this.props
     const { MainControls, MainPanel, MapWindows } = this.context
+    const { popupContent, setPopupContent } = this.props
     return (
       <div className="otp">
         <DesktopNav />
+        {popupContent && (
```
binh-dam-ibigroup, line 202: Could the check for non-null `popupContent` be done inside `PopupWrapper`?

binh-dam-ibigroup (approved):
Two small optional things to consider, but the PR is still a go.

lib/components/app/responsive-webapp.js
miles-grant-ibigroup: This is much cleaner thanks!

daniel-heppner-ibigroup (changes requested):
I am assuming this worked for Binh and you so I think I'm doing something wrong, but there is no button in traveler tools for me, and the i18n strings don't seem to be working.
image

daniel-heppner-ibigroup (approved):
Great! Thank you for the update.

PR #646, opened 2022-08-04
Refactor Flex Reduction to be Optional

miles-grant-ibigroup:
Makes use of https://github.com/opentripplanner/otp-ui/pull/431 to enable the use of `FLEX_EGRESS,FLEX_ACCESS,FLEX_DIRECT` as distinct modes.

### DIFF
--- a/example-config.yml
+++ b/example-config.yml
@@ -176,7 +176,7 @@ geocoder:
 modes:
   # This is the default behavior for flex merging.
   # If set to false, each FLEX mode (FLEX_EGRESS,FLEX_ACCESS,FLEX_DIRECT) will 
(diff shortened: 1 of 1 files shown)


daniel-heppner-ibigroup (approved):
Do we not want this to default to true? (maybe should have commented this on the OTP-UI change)

philip-cline (approved):

miles-grant-ibigroup:
Sorry I forgot to reply to you earlier @daniel-heppner-ibigroup -- I think `false` as the default not only avoids a breaking change but also makes more sense in the majority of cases. Only few operators will want to distinguish between flex service for duration of an itinerary and flex service for a part of an itinerary.

PR #647, opened 2022-08-11
Add support for bugsnag error reporting

miles-grant-ibigroup:
This PR adds a new configuration option to add a bugsnag api key. If the key is added, a listener is added to the root component, and errors/crashes are reported to bugsnag. If no key is specified, nothing happens (no errors are thrown)

### DIFF
--- a/example-config.yml
+++ b/example-config.yml
@@ -4,6 +4,10 @@ api:
   port: 8001
 # v2: false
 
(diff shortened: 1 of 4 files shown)


daniel-heppner-ibigroup (approved):
LGTM

binh-dam-ibigroup (approved):

PR #648, opened 2022-08-11
refactor: Remove redundant const URL_ROOT.

binh-dam-ibigroup:
`URL_ROOT` was redundant with `window.location.origin` so this PR uses that instead.

### DIFF
--- a/lib/util/constants.js
+++ b/lib/util/constants.js
@@ -23,10 +23,6 @@ export const CREATE_TRIP_PATH = `${TRIPS_PATH}/new`
 export const TERMS_OF_SERVICE_PATH = '/terms-of-service'
 export const TERMS_OF_STORAGE_PATH = '/terms-of-storage'
 
-// Gets the root URL, e.g. https://otp-instance.example.com:8080, computed once for all.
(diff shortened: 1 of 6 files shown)


lib/util/constants.js
```diff
@@ -23,10 +23,6 @@ export const CREATE_TRIP_PATH = `${TRIPS_PATH}/new`
 export const TERMS_OF_SERVICE_PATH = '/terms-of-service'
 export const TERMS_OF_STORAGE_PATH = '/terms-of-storage'
 
-// Gets the root URL, e.g. https://otp-instance.example.com:8080, computed once for all.
-// TODO: support root URLs that involve paths or subfolders, as in https://otp-ui.example.com/path-to-ui/
```
miles-grant-ibigroup, line 27: I don't think we fixed this TODO did we?

miles-grant-ibigroup (approved):
Works great thanks for noticing this one! Just one quick question about a removed TODO

lib/util/constants.js
binh-dam-ibigroup: We have not. Noted in #652.

daniel-heppner-ibigroup (approved):

PR #649, opened 2022-08-12
fix(PopupWrapper): support not using iframe

miles-grant-ibigroup:
Sadly the original popup pr #643 failed to include support for ALWAYS avoiding the iframe. This allows us to popup popups that don't allow themselves to be in an iframe.

### DIFF
--- a/example-config.yml
+++ b/example-config.yml
@@ -386,7 +386,7 @@ itinerary:
 #   targets:
 #     testTarget: # To set a name for this target, create the i18n string "config.popups.testTarget"
 #       url: http://example.com/ 
-#       iframe: false # Whether to always launch the popup as a browser popup
+#       modal: false # Whether to always launch the popup as a browser popup
 #     secondTarget: 
 #       url: http://test.com/
 #       appendLocale: false # Defaults to true, if set to false otp-rr will not append the current i18n language to the popup URL.
--- a/example-config.yml
+++ b/example-config.yml
@@ -386,7 +386,7 @@ itinerary:
 #   targets:
 #     testTarget: # To set a name for this target, create the i18n string "config.popups.testTarget"
 #       url: http://example.com/ 
-#       modal: false # Whether to always launch the popup as a browser popup
+#       modal: false # Setting this to false always launches the popup as a browser popup
 #     secondTarget: 
 #       url: http://test.com/
 #       appendLocale: false # Defaults to true, if set to false otp-rr will not append the current i18n language to the popup URL.
--- a/example-config.yml
+++ b/example-config.yml
@@ -386,6 +386,7 @@ itinerary:
 #   targets:
 #     testTarget: # To set a name for this target, create the i18n string "config.popups.testTarget"
 #       url: http://example.com/ 
+#       iframe: false # Whether to always launch the popup as a browser popup
 #     secondTarget: 
 #       url: http://test.com/
 #       appendLocale: false # Defaults to true, if set to false otp-rr will not append the current i18n language to the popup URL.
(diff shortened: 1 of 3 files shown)


example-config.yml
```diff
@@ -386,6 +386,7 @@ itinerary:
 #   targets:
 #     testTarget: # To set a name for this target, create the i18n string "config.popups.testTarget"
 #       url: http://example.com/ 
+#       iframe: false # Whether to always launch the popup as a browser popup
```
binh-dam-ibigroup, line 389: What is a browser popup on a phone browser?

miles-grant-ibigroup: It just opens a new tab but it's up to the browser to pick an implementation

lib/components/app/popup.tsx
```diff
@@ -20,21 +25,23 @@ const isMobile = coreUtils.ui.isMobile()
...
+
   // appendLocale is true by default, so undefined is true
   const compiledUrl = `${url}${
     appendLocale !== false ? intl.defaultLocale : ''
   }`
 
   useEffect(() => {
-    if (isMobile && shown) {
+    if (!useIframe && shown) {
       window.open(compiledUrl, '_blank')
```
binh-dam-ibigroup, line 40: As implemented, the button works only once, and clicking it again does nothing. External links in new tabs should be implemented as `<a target="_blank" ...>` (react-bootstrap buttons support that).

lib/components/app/popup.tsx
```diff
@@ -4,7 +4,12 @@ import coreUtils from '@opentripplanner/core-utils'
 import React, { useEffect } from 'react'
 
 type Props = {
-  content?: { appendLocale?: boolean; id?: string; url?: string }
+  content?: {
+    appendLocale?: boolean
+    id?: string
+    iframe?: boolean
```
binh-dam-ibigroup, line 10: Maybe rename this to `embedded` or something?

binh-dam-ibigroup (changes requested):
I am going to lean towards implementing an anchor for external links. See https://github.com/opentripplanner/otp-react-redux/pull/649/files#r944525119.

lib/components/app/popup.tsx
miles-grant-ibigroup: We can't use an external anchor because not all of our targets are anchor tags and that would be a lot of needley code. I've fixed this problem inside the `Popup` component

lib/components/app/popup.tsx
miles-grant-ibigroup: Good idea! I've renamed the variable

lib/components/app/popup.tsx
binh-dam-ibigroup: We'll leave that work for another PR, probably when dealing with a11y.

binh-dam-ibigroup (approved):

example-config.yml
```diff
@@ -386,6 +386,7 @@ itinerary:
 #   targets:
 #     testTarget: # To set a name for this target, create the i18n string "config.popups.testTarget"
 #       url: http://example.com/ 
+#       modal: false # Whether to always launch the popup as a browser popup
```
daniel-heppner-ibigroup, line 389: True = the opposite of what the comment says, which I think is a little confusing right?

daniel-heppner-ibigroup (approved):
Maybe the comment could be improved? Looks good otherwise.

example-config.yml
miles-grant-ibigroup: Yes I can clarify this

PR #660, opened 2022-09-09
Itinerary Merging Improvements

miles-grant-ibigroup:
This PR allows users to click through all alternate times and see exactly which routes are involved and see if there are any minute differences.


lib/components/narrative/narrative-itineraries.js
```diff
@@ -233,17 +236,21 @@ class NarrativeItineraries extends Component {
       // If no duplicate, push full itinerary to output
       if (duplicateIndex === -1) {
         updatedItineraries.push(updatedItinerary)
-      } else {
+        // Only process itineraries less than 24 hours in the future
+      } else if (
+        updatedItinerary.startTime <
+        // 24 hours in milliseconds
+        Date.now() + 24 * 60 * 60 * 1000
```
daniel-heppner-ibigroup, line 243: This won't work well if we're using the website in a time zone other than the time zone of the agency. Do we have easy access to the agency's timezone somehow that we could use with Datefns to do this properly?

lib/components/narrative/narrative-itineraries.js
```diff
@@ -233,17 +236,21 @@ class NarrativeItineraries extends Component {
       // If no duplicate, push full itinerary to output
       if (duplicateIndex === -1) {
         updatedItineraries.push(updatedItinerary)
-      } else {
+        // Only process itineraries less than 24 hours in the future
```
daniel-heppner-ibigroup, line 239: put this comment in the if block that it pertains to

lib/components/narrative/narrative-itineraries.js
```diff
@@ -178,12 +179,14 @@ class NarrativeItineraries extends Component {
...
         active={active}
+        activeItineraryTimeIndex={activeItineraryTimeIndex}
         activeLeg={activeLeg}
         activeStep={activeStep}
         expanded={showDetails}
         index={itinerary.index}
         itinerary={itinerary}
-        key={itinerary.index}
+        // Ensure we update if either index changes
+        key={(activeItineraryTimeIndex + 1) * (itinerary.index + 1)}
```
daniel-heppner-ibigroup, line 189: Are these guaranteed to always be unique? Index a=1, b=0 and index a=0,b=1 produce the same key value with this code.

How about a template literal like `${activeItineraryTimeIndex} ${itinerary.index}`
Or you could multiply one by 100 and then add them, since I doubt either index will ever reach 100. 
Or use a binary left shift `<< 8` on one and then binary or them. IDK if that is more efficient in Javascript.

daniel-heppner-ibigroup (approved):
Approving because I don't think I need to look at this again after you resolve my comments.

However, I noticed in testing that sometimes there can be multiple itineraries that have the same transit times but different walking directions. At least, I think that's the difference between them because the map doesn't update. Could we update the map with the selected itinerary?

lib/components/narrative/narrative-itineraries.js
miles-grant-ibigroup: Good call! Luckily the leg provides a `agencyTimeZoneOffset` which we can combine with `new Date().getTimezoneOffset()` to ensure we're doing all time math in UTC!

lib/components/narrative/narrative-itineraries.js
miles-grant-ibigroup: I love the binary operator! Thanks for that suggestion works great and avoids these strange edge cases!

miles-grant-ibigroup:
Based on how the itinerary merging works it shouldn't be possible to get different walking legs... Which is why we don't store the geometry of the extra times!

Please send me some examples of differing walking legs offline if you still have them! That sounds like a bug that needs fixing (those itineraries should not be merged)

miles-grant-ibigroup:
Assigning back to myself to add batch UI support

miles-grant-ibigroup:
Assigning back to binh. Support can be achieved with configuration changes only

lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -55,8 +56,25 @@ const DepartureTimes = styled.span`
...
   width: 100%;
 
-  .first {
+  .active {
     color: #090909ee;
+    cursor: auto;
+  }
+
+  button:not(.active) {
+    text-decoration: underline;
```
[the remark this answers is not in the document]
miles-grant-ibigroup: I worry that this may make the feature less discoverable, but I do like how much cleaner it looks!

lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -55,8 +56,25 @@ const DepartureTimes = styled.span`
...
+  }
+  button:not(.active):hover {
+    color: #090909ef;
+  }
+
+  button {
+    background: none;
+    transition: all 0.05s ease-out;
```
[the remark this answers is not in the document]
miles-grant-ibigroup: It looks very janky without the transition. I've lengthened it to .1s

lib/actions/narrative.js
```diff
@@ -13,6 +15,9 @@ export function setActiveItinerary(payload) {
     const urlParams = coreUtils.query.getUrlParams()
     urlParams.ui_activeItinerary = payload.index
     dispatch(setUrlSearch(urlParams))
+
+    // If the itinerary is reset, do not select any time
+    if (payload?.index === -1) dispatch(setActiveItineraryTime(null))
```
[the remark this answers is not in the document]
miles-grant-ibigroup: It doesn't matter initially as the user won't have been able to select an alternate time yet.

lib/components/narrative/line-itin/connected-itinerary-body.js
```diff
@@ -74,6 +80,20 @@ class ConnectedItineraryBody extends Component {
...
+    if (activeItineraryTimeIndex !== undefined) {
+      const activeTime = itinerary.allStartTimes?.[activeItineraryTimeIndex]
+      if (activeTime) {
+        if (activeTime.legs) clonedItinerary.legs = activeTime.legs
+        clonedItinerary.startTime = getFirstLegStartTime(activeTime.legs)
+        clonedItinerary.endTime = getLastLegEndTime(activeTime.legs)
+      }
+      clonedItinerary.allStartTimes = null
+    }
+
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Are we expecting to use this code elsewhere? Doing this feels like adding a layer of complexity where it is not needed.

miles-grant-ibigroup:
Thanks for the review Binh, all changes are addressed.

lib/components/narrative/narrative-itineraries.js
```diff
@@ -258,9 +264,8 @@ class NarrativeItineraries extends Component {
               )
             ) {
               duplicateItin.allStartTimes.push({
-                arrival: cur.endTime,
-                realtime: firstTransitLegIsRealtime(cur),
-                time: cur.startTime
```
binh-dam-ibigroup, line 263: The deletion of the `time` field breaks the logic in the above `if` block and is causing for [this search](http://localhost:9966/#/?ui_activeSearch=8enutr3w8&ui_activeItinerary=-1&fromPlace=255%20Devin%20Place%20NE%2C%20Atlanta%2C%20GA%2C%20USA%3A%3A33.81794822741499%2C-84.37911987304688&toPlace=Plaza%20Fiesta%2C%20Buford%20Hwy%20NE%2C%20Atlanta%3A%3A33.86366%2C-84.30758&date=2022-09-21&time=13%3A42&arriveBy=false&mode=&showIntermediateStops=true&ignoreRealtimeUpdates=true&numItineraries=2&otherThanPreferredRoutesPenalty=900) a duplicate bike itinerary to be listed twice (e.g. "Leave at 1:42pm and 1:42pm"). This is not occurring on the `dev` branch.

binh-dam-ibigroup (changes requested):
I found a bug that affects merging bike itineraries...

lib/components/narrative/narrative-itineraries.js
miles-grant-ibigroup: Thank you for catching this! The `dev` merge was quite complex. Fixed in a73b83aef956dfb38929a9fd7ccbd7e4010b3c11

binh-dam-ibigroup (approved):
LGTM

PR #657, opened 2022-09-02
Move to Vector Tiles (MapLibreGL)

miles-grant-ibigroup:
Makes use of https://github.com/opentripplanner/otp-ui/pull/420 to migrate to MapLibreGL. Also includes some css refreshes to match the major changes in the otp-ui updates.

Previously blocked by https://github.com/opentripplanner/otp-ui/pull/436 waiting for all internal dependencies to be fully updated.

**Blocked by https://github.com/opentripplanner/otp-ui/pull/441 (for new gtfs-rt overlay)**. Tests will fail until this package is updated. The gtfs-rt overlay functionality will be hampered (bus emojis) until https://github.com/opentripplanner/otp-ui/pull/441 is merged.

PR #661, opened 2022-09-13
Stop Viewer Refresh

miles-grant-ibigroup:
Redesigns some panels in the stop viewer to align the design more with the metro itinerary UI. Blocked by #657, #662, https://github.com/opentripplanner/otp-ui/pull/451, https://github.com/opentripplanner/otp-ui/pull/474, https://github.com/opentripplanner/otp-ui/pull/491, https://github.com/opentripplanner/otp-ui/pull/480

#### Looking for feedback on:
The best way to transport the `hideTopBorder` info around. Rendering the top border when the background color is the same or similar to the route color looks very bad. 

The data-passing issue comes from the fact that the main stop viewer rows are being rendered differently but similarly to the related stop rows. The real solution is to rework the entire stop viewer, which is happening. The fact that most of this code will be scrapped soon anyway is why this `hideTopBorder` prop is sort of being attached in awkward places. If there is a quick solution to improve this I'd welcome it, but a larger refactoring should probably be saved for when the stop viewer is migrated to a "nearby" viewer.


lib/components/viewers/related-stops-panel.js
```diff
@@ -111,6 +113,12 @@ class RelatedStopsPanel extends Component {
                           key={id}
                           pattern={pattern}
                           route={route}
+                          routeColor={getRouteColorBasedOnSettings(
```
binh-dam-ibigroup, line 116: Is there a reason why next arrivals in this panel are separate, whereas those for that stop are "glued" together?

miles-grant-ibigroup: Aesthetics. If you glue the small ones together things look very busy if there are multiple colors going on and very messy if all lines are the same color. With the larger stop blocks we can rely on the visual hierarchy to help out more but in the small list we're only working with one horizontal row so things get muddy fast.

lib/components/viewers/stop-viewer.js
```diff
@@ -156,11 +157,23 @@ class StopViewer extends Component {
...
+    const {
+      enableFavoriteStops,
+      hideBackButton,
+      showUserSettings,
+      stopData,
+      transitOperators
+    } = this.props
+
+    // We can use the first route, as this operator will only be used if there is only one operator
+    const stationOperator = transitOperators.find(
```
binh-dam-ibigroup, line 170: Would love to see a mode icon in the header the same way it is shown in related stops!

miles-grant-ibigroup: I think we agreed to save this for the nearby view. I like it though -- great idea!

lib/components/viewers/stop-viewer.js
```diff
@@ -156,11 +157,23 @@ class StopViewer extends Component {
...
+      (o) =>
+        o.agencyId === stopData?.agencyId || stopData?.routes?.[0]?.agencyId
+    )
+
     return (
-      <div className="stop-viewer-header">
+      <div className="stop-viewer-header base-color-bg">
```
binh-dam-ibigroup, line 176: The font color should be set where the ambient background color is set (multiple instances).

lib/components/viewers/stop-viewer.js
```diff
@@ -156,11 +157,23 @@ class StopViewer extends Component {
...
+      hideBackButton,
+      showUserSettings,
+      stopData,
+      transitOperators
+    } = this.props
+
+    // We can use the first route, as this operator will only be used if there is only one operator
+    const stationOperator = transitOperators.find(
+      (o) =>
+        o.agencyId === stopData?.agencyId || stopData?.routes?.[0]?.agencyId
```
binh-dam-ibigroup, line 172: Surround the `||` statement with parentheses (try with this stop http://localhost:9966/#/stop/2:841 to see the difference).

lib/components/viewers/stop-viewer.js
```diff
@@ -175,7 +188,10 @@ class StopViewer extends Component {
         {/* Header Text */}
         <div className="header-text">
           {stopData ? (
-            <span>{stopData.name}</span>
+            <>
+              {agencyCount <= 1 && <OperatorLogo operator={stationOperator} />}
```
binh-dam-ibigroup, line 192: Nit: could that go inside the span?

lib/components/viewers/viewers.css
```diff
@@ -43,6 +42,7 @@
 }
 
 .otp .stop-viewer-body input[type="date"] {
+  background: inherit;
```
binh-dam-ibigroup, line 45: Might need to set the color of the user agent calendar widget too.

lib/components/viewers/pattern-row.tsx
```diff
@@ -68,113 +72,63 @@ class PatternRow extends Component<Props, State> {
...
+            color: getMostReadableTextColor(routeColor, route?.textColor)
+          }}
+        >
           {/* route name */}
           <div className="route-name">
-            <strong>
-              <RouteRenderer leg={generateFakeLegForRouteRenderer(route)} />
+            <strong
+              style={
+                routeName && routeName?.length >= 4 ? { fontSize: '3vb' } : {}
```
binh-dam-ibigroup, line 91: I don't see that applied to the `RAPID` bus (try with http://localhost:9966/#/stop/2:841)

binh-dam-ibigroup: Is this going to accommodate long long `routeShortName` that some agencies on the West Coast use?

lib/components/viewers/pattern-row.tsx
```diff
@@ -68,113 +72,63 @@ class PatternRow extends Component<Props, State> {
...
+                }}
+              >
+                {showOperatorLogos && (
+                  <OperatorLogo operator={route?.operator} />
+                )}
+                <RouteRenderer
+                  // All GTFS bg colors look strange with the top border
+                  hideTopBorder={route?.operator?.colorMode?.includes('gtfs')}
+                  leg={generateFakeLegForRouteRenderer(route, true)}
+                />
```
binh-dam-ibigroup, line 109: Sounds like you need another RouteRenderer component.

lib/components/viewers/viewers.css
```diff
@@ -95,36 +95,46 @@
 /* stop viewer styles */
 
 .otp .stop-viewer .route-row {
-  border: 1px solid #1113;
-  border-top: none;
+}
+.otp .stop-viewer .route-row-container,
+.related-panel-container {
+  box-shadow: 2px 2px 5px 1px rgba(0, 0, 0, 0.1);
+  border-radius: 10px;
```
binh-dam-ibigroup, line 102: Sort CSS (multiple instances)

lib/util/viewer.js
```diff
@@ -406,6 +387,30 @@ export function getColorAndNameFromRoute(operator, route) {
...
 
+/**
+ * Helper method to determine a route color based on config color settings.
+ *
+ * Colors can be one of the following configured values:
+ * - gtfs (default): colors are taken directly from gtfs/operator config
+ * - gtfs-softened: colors are taken from gtfs/operator config and "softened" to be less intense
+ * - disabled: color is always a generic gray
+ */
+export function getRouteColorBasedOnSettings(operator = {}, route = {}) {
```
binh-dam-ibigroup, line 398: I would much rather get-a-`RouteNameRenderer`-based-on-settings, and each route renderer can decide which colors/shapes/etc to use based on the mode and other route attributes.

lib/util/viewer.js
```diff
@@ -425,10 +430,12 @@ export function stopIsFlex(stopData) {
  * The RouteRenderer requires a leg. To get it to render a Route object,
  * we use this method
  */
-export function generateFakeLegForRouteRenderer(route) {
+export function generateFakeLegForRouteRenderer(route, hideTopBorder = false) {
   // RouteRenderer requires a "leg"
   return {
     agencyName: route?.agencyName || route.agency?.name,
+    // Don't render top border when colors are mismatched
+    hideTopBorder,
```
binh-dam-ibigroup, line 438: If you create different kinds of route renderers, each can decide whether to render a border or not.

lib/components/viewers/stop-time-cell.tsx
```diff
@@ -89,42 +97,51 @@ const StopTimeCell = ({
...
-    <div>
-      <div className="pull-left">
-        <StyledIconWrapperTextAlign
-          style={{ color: '#888', fontSize: '0.8em', marginRight: 2 }}
-        >
-          {stopTime.realtimeState === 'UPDATED' ? <Rss /> : <Clock />}
-        </StyledIconWrapperTextAlign>
-      </div>
+    <>
+      <StyledIconWrapperTextAlign
```
binh-dam-ibigroup, line 103: Add a `title` and `aria-label` to this element that say the real-time status.

lib/components/viewers/stop-time-cell.tsx
```diff
@@ -89,42 +97,51 @@ const StopTimeCell = ({
...
+        style={{
+          fontSize: '0.6em',
+          margin: 0,
+          marginRight: 2,
+          transform: `scaleX(${realtime ? '-1' : '1'})`
```
binh-dam-ibigroup, line 108: Move the `realtime` portion to the `PulsingRss` component and move the other styles to a styled wrapper?

lib/components/viewers/related-stops-panel.js
binh-dam-ibigroup: I still think "gluing" the arrivals by stop would provide visual separation between nearby stops. The headings don't outline groups of arrivals as nicely.

miles-grant-ibigroup: I think the real solution is to move to the nearby view we've been discussing for some time. The entire idea of the "nearby stops" view below the stop has never sat right with me.

lib/components/narrative/metro/default-route-renderer.tsx
```diff
@@ -4,22 +4,25 @@ import styled from 'styled-components'
...
 `
 
 type RouteRendererProps = {
-  leg: Leg
+  leg: Leg & { hideTopBorder?: boolean }
 }
 
 const DefaultRouteRenderer = ({ leg }: RouteRendererProps): JSX.Element => (
-  <Block color={leg.routeColor || '333333'}>
+  <Block color={leg.routeColor || '333333'} hideTopBorder={leg.hideTopBorder}>
```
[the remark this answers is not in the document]
miles-grant-ibigroup: It's definitely a hack... The issue comes from the fact that the `RouteRenderer` component doesn't really exist if that makes sense. Since it's pulled in from context, we can't change the props. I could try to find the exact detailed reason why this was done but it was a while ago and I remember I spent a good amount of time trying to figure out how to avoid it.

lib/components/viewers/pattern-row.tsx
```diff
@@ -68,113 +72,63 @@ class PatternRow extends Component<Props, State> {
...
+          style={{
+            backgroundColor: routeColor,
+            color: getMostReadableTextColor(routeColor, route?.textColor)
+          }}
+        >
           {/* route name */}
           <div className="route-name">
-            <strong>
-              <RouteRenderer leg={generateFakeLegForRouteRenderer(route)} />
+            <strong
```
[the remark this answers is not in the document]
miles-grant-ibigroup: We replace it here because the default one looks far too busy

lib/components/viewers/stop-schedule-table.tsx
```diff
@@ -113,17 +112,17 @@ class StopScheduleTable extends Component<{
         <thead>
           <tr>
             {showBlockIds && (
-              <th>
+              <th className="base-color-bg">
```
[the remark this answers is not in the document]
miles-grant-ibigroup: I originally had this, but it didn't color enough of the elements.

lib/components/viewers/stop-viewer.js
miles-grant-ibigroup: I tried to do this, but there were too many instances of the wrong elements being colored.

lib/components/viewers/pattern-row.tsx
miles-grant-ibigroup: I'm not sure I understand what you mean. This RouteRenderer will be whichever one is defined in the config

lib/util/viewer.js
miles-grant-ibigroup: I'm not sure if I understand what you mean. This is what's happening now. The reason that some of the syntax of this is so strange is precisely because the custom route renderer can generate any HTML it would like.

lib/util/viewer.js
miles-grant-ibigroup: I think I see what you mean now. But wouldn't this wrapper route renderer still wind up passing the same prop to the underlying route renderer? I agree this prop is poorly named and should instead be used to send context information to the route renderer. Maybe something like `isBeingRenderedInCompactSpace` or similar.

binh-dam-ibigroup: In my mind, each combination of visual styles (borders, how color is applied, shape...) would be a different renderer.

Regarding adjusting route names to the available space, one technique would be to render text as SVG, then the SVG would be fitted in the space available. See for instance https://github.com/opentripplanner/otp-ui/blob/47302686e8bb53256d3214c1cf1d54db7ebdf754/packages/transit-vehicle-overlay/src/VehicleIcon.tsx#L30 in https://github.com/opentripplanner/otp-ui/pull/470.

binh-dam-ibigroup: So if the config is so that route background colors should be used in `PatternRow`, then use one kind of renderer. If the config sets a say white background color to `PatternRow`, use another kind of renderer.

miles-grant-ibigroup: Also for the SVG trick -- are we sure that this doesn't have a11y implications?

binh-dam-ibigroup: Correct, an 1`aria-label` is needed if using SVG.

binh-dam-ibigroup: Going back to route renderers, to me, the sample below features four different route name renderers.
image

lib/components/viewers/next-arrival-for-pattern.tsx
```diff
@@ -64,11 +74,21 @@ function NextArrivalForPattern(props: Props) {
   const title = `${routeName} ${toHeadsign}`
 
   return (
-    <div className="next-arrival-row">
+    <div
+      className="next-arrival-row"
+      style={{
+        backgroundColor: routeColor,
+        color: getMostReadableTextColor(routeColor, route?.textColor)
```
binh-dam-ibigroup, line 81: I am seeing inconsistent colors with this stop: http://localhost:9966/#/stop/5:21097, where the text for route 140 can be both black and white on the same screen.

lib/components/util/operator-logo.tsx
```diff
@@ -0,0 +1,29 @@
...
+}
+
+const OperatorLogo = ({ operator }: Props): JSX.Element | null => {
+  const intl = useIntl()
+
+  if (!operator?.logo) return null
+
+  return (
+    <OperatorImg
+      alt={intl.formatMessage(
```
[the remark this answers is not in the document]
miles-grant-ibigroup: I've changed the translation string to "logo for {operatorName}". This feels like a cleaner solution that hiding elements for screen readers?

lib/components/viewers/amenities-panel.tsx
```diff
@@ -58,29 +74,35 @@ class AmenitiesPanel extends Component {
           </div>
         </li>
       ))
-      : <li className='related-item'>
-        <div className='item-label'>
-          {parkAndRideMarker}<FormattedMessage id='components.AmenitiesPanel.noPRLotsFound' />
+    ) : (
+      <li className="related-item">
+        <div className="item-label">
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Those styles weren't even doing anything!

lib/components/viewers/next-arrival-for-pattern.tsx
miles-grant-ibigroup: This is because the related stops endpoint includes `textColor` while the stops endpoint doesn't. Repairing this involves a lot of redux plumbing, but it should be fixed once we source all the stop info from one endpoint in the future.

lib/components/viewers/pattern-row.tsx
```diff
@@ -68,113 +72,63 @@ class PatternRow extends Component<Props, State> {
...
             </strong>
-            <FormattedMessage
-              id="components.PatternRow.routeName"
-              values={{
-                headsign: pattern.headsign,
-                routeName: '',
-                strong: Strong
-              }}
-            />
+            <span>{pattern.headsign}</span>
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Screen Shot 2022-11-14 at 9 26 52 AM
It sure is

lib/components/viewers/pattern-row.tsx
```diff
@@ -68,113 +72,63 @@ class PatternRow extends Component<Props, State> {
...
           {/* route name */}
           <div className="route-name">
-            <strong>
-              <RouteRenderer leg={generateFakeLegForRouteRenderer(route)} />
+            <strong
+              style={
+                routeName && routeName?.length >= 4 ? { fontSize: '3vb' } : {}
+              }
+            >
+              <div
```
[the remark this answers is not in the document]
miles-grant-ibigroup: If we did this the flex box would affect the headsign as well. We could accomplish the div removal by using grid but that would add a lot of complexity to what is otherwise a simple component (style-wise).

lib/components/viewers/stop-time-cell.tsx
```diff
@@ -89,42 +97,51 @@ const StopTimeCell = ({
...
   return (
-    <div>
-      <div className="pull-left">
-        <StyledIconWrapperTextAlign
-          style={{ color: '#888', fontSize: '0.8em', marginRight: 2 }}
-        >
-          {stopTime.realtimeState === 'UPDATED' ? <Rss /> : <Clock />}
-        </StyledIconWrapperTextAlign>
-      </div>
+    <>
```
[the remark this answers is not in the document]
miles-grant-ibigroup: This makes sense in theory, except it's now an `li` in one place and a `div` in another. Not sure if there's a non-hack way to do that well...

lib/components/viewers/stop-viewer.js
miles-grant-ibigroup: Very good catch thank you!

lib/components/viewers/stop-viewer.js
miles-grant-ibigroup: No, the padding looked awkward.

lib/components/viewers/pattern-row.tsx
miles-grant-ibigroup: Yes if you have a look at septa which has long `routeShortName`s on it's rail lines they fit nicely.

lib/components/viewers/stop-time-cell.tsx
miles-grant-ibigroup: I think this would make it more difficult to a reader to understand what's going on

miles-grant-ibigroup:
Thanks for the careful review @binh-dam-ibigroup! All changes are addressed or responded to. I couldn't find any regressions based on the changes, but I would appreciate a double-check, especially the CSS had quite a few adjustments.

lib/components/viewers/stop-time-cell.tsx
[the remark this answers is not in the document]
binh-dam-ibigroup: You can place spans inside list items and divs.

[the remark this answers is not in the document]
miles-grant-ibigroup: But wouldn't this add extra nesting where it's not needed

lib/components/viewers/stop-time-cell.tsx
```diff
@@ -19,6 +21,10 @@ const { getUserTimezone } = coreUtils.time
 const ONE_HOUR_IN_SECONDS = 3600
 const ONE_DAY_IN_SECONDS = 86400
 
+const PulsingRss = styled(Rss)`
+  animation: pulse-opacity 2s ease-in-out infinite;
+`
```
binh-dam-ibigroup, line 26: Following up on https://github.com/opentripplanner/otp-react-redux/pull/661/files#r1018446380, insert the following and remove the corresponding inline style from the containing span (that style attribute actually has no effect, so maybe a transform is not needed at all):

```css
          transform: scaleX(-1);
```

lib/components/viewers/stop-time-cell.tsx
binh-dam-ibigroup: See follow-up comment in `PulsingRss`, and move the non-transform style into a `styled(StyledIconWrapperTextAlign)`.

lib/components/viewers/amenities-panel.tsx
```diff
@@ -114,13 +138,13 @@ class AmenitiesPanel extends Component {
               <>
                 <div>
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Can use `<br />` tags instead of `<divs>`, with the advantage that you can improve the text indentation using:
```css
.related-item {
    text-indent: -27px; /* 22px image width + 5px margin */
    padding-left: 27px;
}
```

binh-dam-ibigroup (changes requested):
Some more comments, but the blocking item is regarding fitting not-so-long route names such as "RAPID" into the available in `PatternRow`.

i18n/es.yml
```diff
@@ -492,6 +492,8 @@ components:
     destination: A
     route: 'Ruta:'
   StopTimeCell:
+    realtime: Based on realtime data
+    scheduled: Based on schedule data
```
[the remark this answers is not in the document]
miles-grant-ibigroup: I had to add the strings because of the i18n messages check lint step!

lib/components/narrative/metro/default-route-renderer.tsx
```diff
@@ -4,22 +4,28 @@ import styled from 'styled-components'
...
   border-radius: 5px;
-  text-align: center;
+  border-top: 5px solid #${(props) => props.color};
   display: inline-block;
   max-width: 12vw;
+  padding: 3px 7px;
+  text-align: center;
+  white-space: break-spaces;
+
+  ${(props) => props.onColoredBackground && 'border-top: none'}
```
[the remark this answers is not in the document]
miles-grant-ibigroup: I agree that these extra props are messy, but I don't think either of us have an ideal solution here.

I think having the route renderer accept a `leg` may have been a bit ambitious. A future refactor that might be good is to have the route renderer accept only route name and color. This should be enough to still do the custom styling we're currently doing, and it would make this `onColoredBackground` prop less messy.

lib/components/util/operator-logo.tsx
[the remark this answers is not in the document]
miles-grant-ibigroup: That's a good aria label!

lib/components/viewers/RouteRow.js
```diff
@@ -154,18 +150,8 @@ export class RouteRow extends PureComponent {
...
-                alt={intl.formatMessage(
-                  {
-                    id: 'components.RouteRow.operatorLogoAltText'
-                  },
-                  { operatorName: operator.name }
-                )}
-                src={operator.logo}
-              />
-            )}
+          <RouteRowElement style={{ paddingRight: 8 }}>
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Slept right through that one, thanks!

lib/components/viewers/live-stop-times.tsx
```diff
@@ -146,24 +153,33 @@ class LiveStopTimes extends Component<Props, State> {
...
-              return routeIsValid(route, getRouteIdForPattern(pattern)) ? (
+              return (
                 <PatternRow
                   homeTimezone={homeTimezone}
                   key={id}
                   pattern={pattern}
-                  route={route}
+                  route={{
+                    ...route,
+                    operator: transitOperators.find(
```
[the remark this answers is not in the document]
miles-grant-ibigroup: It could be, but we'd then have to assemble it back together within the file for `next-arrival-for-pattern`. We can't ship operator as its own prop there since the related stops panel deals with a list of routes. Matching those arrays up would get messy.

lib/components/viewers/live-stop-times.tsx
```diff
@@ -146,24 +153,33 @@ class LiveStopTimes extends Component<Props, State> {
...
+                      (o: TransitOperator) => o.agencyId === route.agencyId
+                    )
+                  }}
+                  showOperatorLogos={showOperatorLogos}
```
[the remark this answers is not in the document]
miles-grant-ibigroup: sure!

lib/components/viewers/next-arrival-for-pattern.tsx
```diff
@@ -17,18 +18,25 @@ type Props = {
   homeTimezone?: string
   intl: IntlShape
   pattern: Pattern
-  route: Route
+  // Not the true operator type, but the one that's used here
+  route: Route & { operator?: { colorMode?: string } }
```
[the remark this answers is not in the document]
miles-grant-ibigroup: See previous explanation -- I originally had it like this but found this solution to be cleaner than doing weird array length matching. I'll add a comment

lib/components/viewers/pattern-row.tsx
binh-dam-ibigroup: This item still needs to be resolved.

binh-dam-ibigroup (changes requested):
https://github.com/opentripplanner/otp-react-redux/pull/661#discussion_r1024492832 still needs to be resolved, I don't think that was updated in recent commits.

miles-grant-ibigroup:
Thanks for catching that -- seems to be a blink issue. Will look into a polyfill. Screen Shot 2022-11-21 at 11 12 13 AM

lib/components/viewers/amenities-panel.tsx
```diff
@@ -92,20 +114,22 @@ class AmenitiesPanel extends Component {
...
+                count: company.stations.length
+              }
+            )
         return (
-          <li className='related-item' key={key}>
-            <div className='item-label'>
-              <div className='overflow-ellipsis' title={label}>
+          <li className="related-item" key={key}>
+            <div className="item-label">
+              <div className="overflow-ellipsis" title={label}>
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Pretty soon, we will start deeming `<div>`s inside `<li>`s to be invalid markup.

lib/components/viewers/pattern-row.tsx
```diff
@@ -88,7 +88,7 @@ class PatternRow extends Component<Props, State> {
           <div className="route-name">
             <strong
               style={
-                routeName && routeName?.length >= 4 ? { fontSize: '3vb' } : {}
+                routeName && routeName?.length >= 4 ? { fontSize: '50%' } : {}
```
binh-dam-ibigroup, line 91: This or similar logic also needs to be applied to the related stops rows (see screenshot)

Screenshot 2022-11-21 at 3 29 39 PM

binh-dam-ibigroup (changes requested):
Almost... one more change is needed regarding the route font size for related stops.

lib/components/viewers/amenities-panel.tsx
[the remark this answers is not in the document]
miles-grant-ibigroup: This change was made in another pr based on pr review...

lib/components/viewers/pattern-row.tsx
miles-grant-ibigroup: Spent a long time trying to get the grid alignment right while also doing dynamic text sizing, in the end decided to just un-grid it. 31f58ccb70b6b8e55470e28c9eb85f09a8139dc3

lib/components/viewers/viewers.css
```diff
@@ -219,24 +219,39 @@
...
+.otp .stop-viewer .stop-viewer-body .next-arrival-row:last-of-type {
+  margin-bottom: 0;
+}
 
 .otp .stop-viewer .stop-viewer-body .next-arrival-row .next-arrival-label {
-  display: table-cell;
-  max-width: 200px;
+  align-items: center;
+  display: flex;
+  max-width: 250px;
```
binh-dam-ibigroup, line 238: max-width causes unnecessary clipping:

```suggestion
  flex-grow: 1;
  margin-right: 4px; /* or insert a non-shrinkable fixed-size element after. */
```

lib/components/viewers/viewers.css
```diff
@@ -219,24 +219,39 @@
...
   font-weight: 700;
-  display: inline-block;
-  min-width: 5ch;
+}
+.otp .stop-viewer .stop-viewer-body .next-arrival-row .route-name section {
+  border-top: none;
 }
 
 .otp .stop-viewer .stop-viewer-body .next-arrival-row .next-arrival-time {
   display: table-cell;
```
binh-dam-ibigroup, line 253: Inside a `flex` element you should use flex attributes (although in some future life this might have to become a table again)
```suggestion
  flex-shrink: 0;
```

binh-dam-ibigroup (approved):
LGTM. Thanks for the changes! Just a few visual tweaks.

lib/components/viewers/viewers.css
miles-grant-ibigroup: I tried this originally but ran into the following issue on long destinations. Have tried all kinds of combos of flex shrinking and growing, and even grid but couldn't come up with something as clean as the max width that behaved appropriately. Screen Shot 2022-11-28 at 8 54 10 AM

lib/components/viewers/viewers.css
miles-grant-ibigroup: This one works, thanks!

miles-grant-ibigroup:
Tests are failing pending `core-utils` updates being published and used in this PR

miles-grant-ibigroup:
@daniel-heppner-ibigroup all dependencies are now working out of the box and this is ready for your review

daniel-heppner-ibigroup:
I am seeing an issue where the stop times table headers are overlapping with the table contents. 
image

binh-dam-ibigroup:
> I am seeing an issue where the stop times table headers are overlapping with the table contents. image

I thought I addressed that with #695 + the appropriate CSS file.

miles-grant-ibigroup:
I think the specific fix got lost in a merge... I will attempt to fix now

miles-grant-ibigroup:
> > I am seeing an issue where the stop times table headers are overlapping with the table contents. image
> 
> I thought I addressed that with #695 + the appropriate CSS file.

Yes you are right the appropriate CSS file being key to this fix working. @daniel-heppner-ibigroup the CSS your config is using might need to be updated.

daniel-heppner-ibigroup (approved):
Thank you for the great work on this! It looks really great! Had a few comments and suggestions, I think I can still approve though.

lib/components/viewers/viewers.css
binh-dam-ibigroup: For this comment, I still think replacing `max-width` with `margin-right` renders better on wide screens.

miles-grant-ibigroup: I agree, but this specific reflow issue to me is unacceptable

lib/components/viewers/amenities-panel.tsx
```diff
@@ -74,8 +90,9 @@ class AmenitiesPanel extends Component {
     const { intl, stopData } = this.props
     if (!stopData || !stopData.bikeRental) return null
     const { stations } = stopData.bikeRental
-    const stationCounts = {}
-    stations.forEach((station) => {
+    // TODO: Rewrite this in a way that the types make sense
```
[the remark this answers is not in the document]
miles-grant-ibigroup: This is really great thanks so much!

lib/components/narrative/metro/default-route-renderer.tsx
```diff
@@ -2,29 +2,40 @@ import { Leg } from '@opentripplanner/types'
 import React from 'react'
 import styled from 'styled-components'
 
-const Block = styled.span<{ color: string }>`
+// TODO: This should not be a section, but using div/span
+// doesn't allow us to style it from within the RouteBlock
+const Block = styled.span<{ color: string; onColoredBackground?: boolean }>`
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Well spotted thanks

lib/components/narrative/metro/default-route-renderer.tsx
```diff
@@ -2,29 +2,40 @@ import { Leg } from '@opentripplanner/types'
...
 type RouteRendererProps = {
-  leg: Leg
+  leg: Leg & { onColoredBackground?: boolean }
 }
 
 const DefaultRouteRenderer = ({ leg }: RouteRendererProps): JSX.Element => (
-  <Block color={leg.routeColor || '333333'}>
+  <Block
+    color={leg.routeColor || '333333'}
+    onColoredBackground={leg.onColoredBackground}
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Good idea changed!

lib/components/viewers/amenities-panel.tsx
```diff
@@ -1,8 +1,10 @@
-// Disabled for now because this file should be removed when switching to the nearby view
 /* eslint-disable react/prop-types */
-import { connect } from 'react-redux'
-import { FormattedMessage, injectIntl } from 'react-intl'
+import { ConfiguredCompany, Station, Stop } from '@opentripplanner/types'
+import { FormattedMessage, injectIntl, IntlShape } from 'react-intl'
+// @ts-expect-error not typescripted yet
 import { getCompanyIcon } from '@opentripplanner/icons/lib/companies'
+// Disabled for now because this file should be removed when switching to the nearby view
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Good catch. Gone!

miles-grant-ibigroup:
Thanks @daniel-heppner-ibigroup for those notes! This is now ready to merge once the Percy fix #730 is merged

PR #662, opened 2022-09-13
Add CodeQL Analysis

miles-grant-ibigroup:
This PR adds GitHub codeql analysis


daniel-heppner-ibigroup (approved):

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,72 @@
+# For most projects, this workflow file will not need changing; you simply need
```
binh-dam-ibigroup, line 1: Add a comment saying the file was initially generated by github.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,72 @@
...
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "dev", master ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "dev" ]
+  schedule:
+    - cron: '23 11 * * 4'
```
binh-dam-ibigroup, line 21: Is the intent to run checks periodically in addition to PR and PR updates to `dev`?

binh-dam-ibigroup (approved):
LGTM, only wondering about whether periodicity of the checks is absolutely needed.

.github/workflows/codeql-analysis.yml
miles-grant-ibigroup: It seems like it! Which makes sense -- even if we don't make changes new vulnerabilities might be discovered

binh-dam-ibigroup: Good point!

PR #664, opened 2022-09-21
Address two react console warnings.

binh-dam-ibigroup:
This PR resolves React console warnings regarding keys for the locale selector and a prop for the app menu.


daniel-heppner-ibigroup (approved):
Wow, thank you for the fix!

miles-grant-ibigroup (approved):
Thanks! This is much cleaner. 

Would it be possible to merge https://github.com/opentripplanner/otp-react-redux/pull/407 into this? They seem related.

binh-dam-ibigroup:
> Thanks! This is much cleaner.
> 
> Would it be possible to merge #407 into this? They seem related.

The change from #407 is already included in the current dev branch.

PR #667, opened 2022-09-28
Improve point popup

binh-dam-ibigroup:
Requires/blocked by #657, https://github.com/opentripplanner/otp-ui/pull/471.

This PR adds to the generic location popup a button for the user to zoom to that location. The button replaces behavior that was previously supposed to be triggered when the popup appears, leaving users control on when the map moves.


lib/components/map/point-popup.tsx
```diff
@@ -1,45 +1,68 @@
...
-  width: 240px;
+/**
+ * Adds a box shadow and tweaks border radius to make popups easier to read.
+ */
+const Popup = styled(MapGlPopup)`
+  & > .maplibregl-popup-content,
+  & > .mapboxgl-popup-content {
+    border-radius: 6px;
+    box-shadow: 0 3px 14px rgb(0 0 0 / 40%);
+  }
```
miles-grant-ibigroup, line 22: Screen Shot 2022-10-20 at 2 02 42 PM
I love these changes, but would it be possible to extract this into core-utils? That way the popups can be a bit more consistent.

lib/components/map/point-popup.tsx
```diff
@@ -56,22 +79,31 @@ function MapPopup({
...
-        <div>
-          <FromToLocationPicker
-            label
-            location={{ ...mapPopupLocation, name: popupName }}
-            setLocation={onSetLocationFromPopup}
-          />
-        </div>
-      </PopupContainer>
+      <ZoomButton
+        className="pull-right"
```
miles-grant-ibigroup, line 87: I wouldn't use float for this. We can `flex` the container instead. That way we can `align-items: center/baseline` as well

miles-grant-ibigroup (changes requested):
Really love these changes! Great experience and the shadows look clean as well. Just a few cleanup suggestions to make sure things are consistent and gridded.

lib/components/map/point-popup.tsx
binh-dam-ibigroup: Did you intend the magnifier to center vertically when the place name occupies multiple lines? My intuition was rather to leave it at a fixed place relative to the top-right corner of the popup, similar with the "close" button elsewhere.

miles-grant-ibigroup: Off the top of my head not sure if it would look better to align it with the first text line or the entire text block but I definitely think it should be tied to the text in some way -- the magnifying glass has nothing to do with the popup, but rather the title. The close button should be semantically tied to the popup as the close button affects the popup and only has to do with the popup. The zoom button's context is firmly tied to the active stop though.

lib/components/map/point-popup.tsx
binh-dam-ibigroup: They would be in the map package, but I get the idea and agree. Should that be a `feat.` or a `fix` change?

miles-grant-ibigroup: fix I think since it's mostly cosmetic

binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/667/commits/3e590c94392954fd6870e1895e5205d82b72a9f1.

lib/components/map/point-popup.tsx
binh-dam-ibigroup: I switched to `flex` and improved the alignment of the magnifier with the top line of the place text with https://github.com/opentripplanner/otp-react-redux/pull/667/commits/3e590c94392954fd6870e1895e5205d82b72a9f1.

lib/components/map/point-popup.tsx
```diff
@@ -56,22 +79,31 @@ function MapPopup({
       closeOnClick
       latitude={mapPopupLocation.lat}
       longitude={mapPopupLocation.lon}
-      onClose={() => setMapPopupLocation({ location: null })}
+      onClose={clearMapPopupLocation}
+      // Override inline style supplied by react-map-gl to accommodate long "plan a trip" translations.
+      style={{ maxWidth: '260px', width: '260px' }}
     >
-      <PopupContainer>
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Partially changed with https://github.com/opentripplanner/otp-react-redux/pull/667#discussion_r1012207290.

miles-grant-ibigroup (approved):
Looks great thanks for the changes!

lib/components/map/point-popup.tsx
```diff
@@ -56,22 +77,33 @@ function MapPopup({
...
-        <div>
-          <FromToLocationPicker
-            label
-            location={{ ...mapPopupLocation, name: popupName }}
-            setLocation={onSetLocationFromPopup}
-          />
-        </div>
-      </PopupContainer>
+        <ZoomButton
+          // className="pull-right"
```
daniel-heppner-ibigroup, line 91: Is this comment still needed?

daniel-heppner-ibigroup (approved):
Looks a lot nicer, the shadow is a huge help. And of course the functionality is great too. Thank you!

lib/components/map/point-popup.tsx
binh-dam-ibigroup: No, good catch, removed in https://github.com/opentripplanner/otp-react-redux/pull/667/commits/9c07af0c666936e4a65cb7328f9a6bca456d842e.

PR #670, opened 2022-10-10
Reduce Bundle Size pt 1

daniel-heppner-ibigroup:
This addresses the following bundle size problems:
- Fix styled-icons code splitting. ~~(Blocked by https://github.com/opentripplanner/otp-ui/pull/456)~~
- Fix unneeded import in mailables for pdfkit
- Removes stylized map completely, along with d3.


package.json
```diff
@@ -68,8 +68,6 @@
     "copy-to-clipboard": "^3.0.8",
     "country-flag-icons": "^1.4.19",
```
miles-grant-ibigroup, line 69: While we're removing things, I don't think this is used anymore

package.json
```diff
@@ -68,8 +68,6 @@
...
     "currency-formatter": "^1.4.2",
```
miles-grant-ibigroup, line 70: Same for this

miles-grant-ibigroup (approved):
Seems like the Percy tests are facing the am/pm bug again. I think they really are affected by the time they're run... I'll try re-running them tomorrow morning and we can see if that changes thigns

package.json
daniel-heppner-ibigroup: It's used in tnc-leg. It seems like we might not be using tnc-leg anymore though. Can it be removed?

package.json
daniel-heppner-ibigroup: Shouldn't affect bundle size, but it's been removed.

PR #671, opened 2022-10-11
Split findRoutes() into OTP1/2-specific implementations

binh-dam-ibigroup:
This (minimal) PR splits `findRoutes` from the API actions into OTP1 and OTP2 versions and fetch route and agency data while taking advantage of OTP2 graphql endpoint.

### DIFF
--- a/lib/actions/apiV2.js
+++ b/lib/actions/apiV2.js
@@ -521,9 +521,10 @@ const getVehiclePositionsForRoute = (routeId) =>
   }
 
 /**
- * Helper function to extract the agency id string from an OTP2-generated agency id.
+ * Helper function to extract the agency id string from an OTP2-generated agency id,
+ * so it matches the (simple) agency id defined in config.
  */
-function getAgencyId(agency /* Agency */) {
+function getAgencyIdMatchingConfig(agency /* Agency */) {
   const agencyParts = agency.id.split(':')
   // agencyId is the last portion of agencyParts (the first one is the feed id).
   return agencyParts[agencyParts.length - 1]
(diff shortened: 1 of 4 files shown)


binh-dam-ibigroup:
Reverting to draft as there is a regression in the route viewer.

lib/actions/apiV2.js
```diff
@@ -518,6 +520,16 @@ const getVehiclePositionsForRoute = (routeId) =>
...
+/**
+ * Helper function to extract the agency id string from an OTP2-generated agency id,
+ * so it matches the (simple) agency id defined in config.
+ */
+function getAgencyIdMatchingConfig(agency /* Agency */) {
+  const agencyParts = agency.id.split(':')
+  // agencyId is the last portion of agencyParts (the first one is the feed id).
+  return agencyParts[agencyParts.length - 1]
+}
+
```
miles-grant-ibigroup, line 532: Can this be moved to a separate utils file?

lib/actions/apiV2.js
```diff
@@ -601,6 +616,7 @@ export const findRoute = (params) =>
             })
             newRoute.patterns = routePatterns
             // TODO: avoid explicit behavior shift like this
+            newRoute.agency.id = getAgencyIdMatchingConfig(newRoute.agency)
```
miles-grant-ibigroup, line 619: This is a clever solution to this problem, but there is a problem: the transit operator agency id is also used to display operator logos in the itinerary body and stop viewer. The agency ids need to be of the same format there! Either that or the config needs to support defining multiple agency ids to match against

miles-grant-ibigroup (changes requested):
A few issues to resolve before I can fully test this.

lib/actions/apiV2.js
binh-dam-ibigroup: Okay, so it sounds like the full agency id returned from OTP should be used and my assumptions about nomenclature were incorrect.

miles-grant-ibigroup (changes requested):
I'm seeing agency id information in the routes object, but transit operator matching still isn't working -- I think it's because for OTP2 the relevant methods expect the agency id to be `route.agency.id`. 

Not sure what the better fix is -- we can either rewrite the response to include `route.agency.id` or make the methods check for `agencyId` no matter the OTP version

miles-grant-ibigroup (approved):
Great fix thanks so much!

daniel-heppner-ibigroup (approved):
Works well for me! Hopefully we can get rid of some of the old reliance on v1 data structures.

PR #673, opened 2022-10-13
Fix mobile map

binh-dam-ibigroup:
Fixes initializing of the mobile map and resizing of the map when toggling between split results and full map view.


miles-grant-ibigroup (approved):
Thanks for fixing this! Works great

daniel-heppner-ibigroup (approved):
Thanks for the fix!

PR #674, opened 2022-10-14
fix(components/narriative/utils): resolve a11y routing fa regression

miles-grant-ibigroup:
Everyone missed a pretty glaring error in #651! This prevented the accessibility routing from working correctly. This PR resolves this and fixes accessibility labels.

### DIFF
--- a/lib/components/narrative/utils.tsx
+++ b/lib/components/narrative/utils.tsx
@@ -19,7 +19,7 @@ export function localizeGradationMap(
   gradationMap?: Record<string, any>
 ): Record<string, any> {
   if (!gradationMap) return {}
-  const newGradationMap: Record<string, any> = {}
+  const newGradationMap: Record<string, any> = { ...gradationMap }
 
   Object.keys(newGradationMap).forEach((key) => {
     // As these localization keys are in the config, rather than


lib/components/narrative/utils.tsx
```diff
@@ -19,7 +19,7 @@ export function localizeGradationMap(
   gradationMap?: Record<string, any>
 ): Record<string, any> {
   if (!gradationMap) return {}
-  const newGradationMap: Record<string, any> = {}
+  const newGradationMap: Record<string, any> = { ...gradationMap }
 
   Object.keys(newGradationMap).forEach((key) => {
```
daniel-heppner-ibigroup, line 24: I would probably just change `newGradationMap` to `gradationMap` instead of copying `gradationMap` into the new object.

daniel-heppner-ibigroup (approved):
Suggestion of another way to do this, but your way seems fine too. Thanks for catching my error!

lib/components/narrative/utils.tsx
miles-grant-ibigroup: I think that would modify the original object and I want to double check if we want to do that (although since you wrote this if that is what we want to do please let me know and I'll make the change)

philip-cline (approved):

PR #675, opened 2022-10-14
Reduce API calls from Route Viewer.

binh-dam-ibigroup:
Requires #671.

This PR changes the route viewer and the api plumbing, so that only one request to fetch all routes is sent and only one request to fetch route details is sent per route. Route details are now fetched on clicking a route in the viewer instead of on mouse over.

Bugs/inefficiencies are addressed when opening route viewer in the following scenarios:
- from the initial view
- from the `/route` url, no route selected
- from the `/route/1:1234` urls to show details expanded (url, patterns) and its shape on the map
- from the `/route/1:1234/pattern/1:1234:0:01` urls to show a route pattern.

### DIFF
--- a/lib/actions/api.js
+++ b/lib/actions/api.js
@@ -738,7 +738,7 @@ export function findRouteIfNeeded(params) {
     if (!routeId) return
 
     // If route details were already requested or fetched, don't fetch them again.
-    const route = (getState().otp.transitIndex.routes || {})[routeId]
+    const route = getState().otp.transitIndex.routes?.[routeId]
     if (route?.patterns || route?.pending) return
 
     dispatch(findingRoute(routeId))
--- a/lib/actions/api.js
+++ b/lib/actions/api.js
@@ -8,6 +8,7 @@ import coreUtils from '@opentripplanner/core-utils'
 import qs from 'qs'
 
 import { convertToPlace, getPersistenceMode } from '../util/user'
+import { FETCH_STATUS } from '../util/constants'
 import { getSecureFetchOptions } from '../util/middleware'
 import { getStopViewerConfig, queryIsValid } from '../util/state'
 
@@ -701,11 +702,19 @@ export function findStopTimesForStop(params) {
 
 // Routes lookup query
 
+export const findingRoutes = createAction('FINDING_ROUTES')
 export const findRoutesResponse = createAction('FIND_ROUTES_RESPONSE')
 export const findRoutesError = createAction('FIND_ROUTES_ERROR')
 
-export function findRoutes(params) {
-  return executeOTPAction('findRoutes', params)
+export function findRoutesIfNeeded(params) {
+  return function (dispatch, getState) {
+    if (
+      getState().otp.transitIndex.routesFetchStatus === FETCH_STATUS.UNFETCHED
+    ) {
+      dispatch(findingRoutes())
+      dispatch(executeOTPAction('findRoutes', params))
+    }
+  }
 }
 
 // Patterns for Route lookup query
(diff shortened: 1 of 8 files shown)


lib/components/viewers/RouteRow.js
```diff
@@ -119,7 +119,8 @@ export class RouteRow extends PureComponent {
   }
 
   _onClick = () => {
-    const { isActive, route, setViewedRoute } = this.props
+    const { findRouteIfNeeded, isActive, route, setViewedRoute } = this.props
+    findRouteIfNeeded({ routeId: route.id })
```
miles-grant-ibigroup, line 123: Going from hover to onClick is causing a bit of a negative user experience, especially on slow connections. The geometry is sometimes not loaded in time to display before the animation fires. Even if it does arrive in time, the delay means that all stops briefly flash unpleasantly.

miles-grant-ibigroup (changes requested):
Thanks for fixing this! This has been on my list to improve for a long time. Things are definitely working smoothly on both OTP1 and OTP2 which is delightful! 

The only issue I'm having for now is moving from hover to click to get route geometry. Once we move back to hover though I think this will be a great change!

lib/components/viewers/RouteRow.js
binh-dam-ibigroup: I'm thinking hover can be problematic for a11y, and users with certain assistive technologies may not be able to trigger the behavior otherwise. Maybe implementing a "focus" thing such as in a regular list could help resolve that?

miles-grant-ibigroup: I'd be ok with firing the onHover handler on focus

binh-dam-ibigroup: Cool. Updated in https://github.com/opentripplanner/otp-react-redux/pull/675/commits/0d581bae5c9cf4b964631312bd70e388bcc88e33.

lib/components/viewers/RouteRow.js
```diff
@@ -154,7 +150,8 @@ export class RouteRow extends PureComponent {
           className="clear-button-formatting"
           isActive={isActive}
           onClick={this._onClick}
-          onMouseEnter={this._onMouseEnter}
+          onFocus={this._onFocusOrMouseEnter}
```
miles-grant-ibigroup, line 153: Realizing now we need touch support as well! Do you think it makes sense to include a `onTouchStart` handler as well?

binh-dam-ibigroup: I added a `touchStart` handler in https://github.com/opentripplanner/otp-react-redux/pull/675/commits/13adba9ee416c7cb0a9db96b7b16f754c8a9632d, but I am really wondering what the expanded `RouteRow` is really trying to achieve.

miles-grant-ibigroup: The geometry data is quite massive... But agreed we should think about figuring out how to load it in at clever times

lib/actions/api.js
```diff
@@ -719,11 +728,22 @@ export const findPatternsForRouteError = createAction(
...
 
-export function findRoute(params) {
-  return executeOTPAction('findRoute', params)
+export function findRouteIfNeeded(params) {
+  return function (dispatch, getState) {
+    const { routeId } = params
+    if (!routeId) return
+
+    // If route details were already requested or fetched, don't fetch them again.
+    const route = (getState().otp.transitIndex.routes || {})[routeId]
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Not really actually, as `transitIndex` and `transitIndex.routes` are always initialized, so I simplified in https://github.com/opentripplanner/otp-react-redux/pull/675/commits/13adba9ee416c7cb0a9db96b7b16f754c8a9632d and https://github.com/opentripplanner/otp-react-redux/pull/675/commits/9c7bceafa3c1db7cc5826ca381b0289c84287336.

daniel-heppner-ibigroup (approved):
Great changes! I love how it preloads based on where your cursor moves. Feels very fancy.

PR #676, opened 2022-10-17
Remove tab view

daniel-heppner-ibigroup:


lib/index.js
```diff
@@ -90,8 +79,6 @@ export {
 
   // mobile components
   MobileMain,
-  MobileResultsScreen,
```
miles-grant-ibigroup, line 93: I'm getting an error when trying to run call taker. it seems to want to import this for its mobile view. This can probably be replaced with something else but I think I'd prefer to not have to update all the mobile configs. I think the solution here is a "fake" export in the right place. 

If you like we can discuss more offline

miles-grant-ibigroup (changes requested):
This has less of an impact on things than I expected! Works fairly well for me, just a few i18n keys I'd switch around and a small bug with launching the call taker

lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -293,6 +293,18 @@ class MetroItinerary extends NarrativeItinerary {
       return routeBlocks
     }
 
+    const onClick = () => {
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: that's better!

i18n/en-US.yml
```diff
@@ -144,6 +144,7 @@ components:
     destination: destination
     origin: origin
     planTripTooltip: Plan trip
+    settings: Settings
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: How is "Trip Settings"?

i18n/en-US.yml
```diff
@@ -245,9 +246,11 @@ components:
         one {# issue}
         other {# issues}
       }"
+    changeSortDir: "Change sort direction"
     # Note to translator: text below is width-constrained
     # (about half-pane width, or half-screen width in mobile views).
     searching: Finding your options...
+    selectTitle: Sort By
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: Changed key to "sortBy"

lib/index.js
daniel-heppner-ibigroup: Replacing this with fake exports seems to work ok, although it seems to crash when I try to actually plan a trip. (it crashes the tab, so no error message.)

miles-grant-ibigroup: I wonder what the problem is? Do you think it might be that the call taker mobile views are somehow broken?

lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -339,7 +347,7 @@ class MetroItinerary extends NarrativeItinerary {
             )}
             {!mini && (
               <ItineraryGrid className="itin-grid">
-                <Routes enableDot={enableDot}>
+                <Routes aria-hidden enableDot={enableDot}>
```
miles-grant-ibigroup, line 350: I would at least add a comment here that we need to add an aria-label to the parent

miles-grant-ibigroup (approved):
Happy to see all these reductions and happy to see bundle size is down! Thanks for all this work.

philip-cline (approved):

PR #679, opened 2022-10-20
fix(components/narriative): support hiding calories

miles-grant-ibigroup:
Makes use of https://github.com/opentripplanner/otp-ui/pull/466 to make showing calories in the trip details component optional.


daniel-heppner-ibigroup (approved):
Works for me

amy-corson-ibigroup (approved):

PR #682, opened 2022-10-24
Fix regression with field trip date display

demory:
Fixes bug where clicking on a field trip entry in the FT request list (see screenshot) caused browser crash due to date library issue. Should now work when selecting requests both with date set (e.g. requests 35-37) and not set (e.g. request 34).

image

### DIFF
--- a/lib/components/admin/field-trip-details.js
+++ b/lib/components/admin/field-trip-details.js
@@ -166,11 +166,8 @@ class FieldTripDetails extends Component {
               }}
               type="date"
             />
-            {travelDateValue && (
-              <span>
-                ({formatDistanceToNow(travelDate, { addSuffix: true })})
-              </span>
-            )}
+            {travelDateValue &&
+              formatDistanceToNow(travelDate, { addSuffix: true })}
           </small>
         </div>
       </WindowHeader>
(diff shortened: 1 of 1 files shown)


lib/components/admin/field-trip-details.js
```diff
@@ -167,7 +166,11 @@ class FieldTripDetails extends Component {
               }}
               type="date"
             />
-            ({formatDistanceToNow(travelDate, { addSuffix: true })})
+            {travelDateValue && (
+              <span>
```
binh-dam-ibigroup, line 170: Do we need this extra `<span>`?

binh-dam-ibigroup (approved):
One small nitpick, but I'm good with merging as is.

lib/components/admin/field-trip-details.js
```diff
@@ -8,6 +8,7 @@ import { format, formatDistanceToNow, parse } from 'date-fns'
 import {
   getDateFormat,
   // yyyy-MM-dd, which is same format required for date input controls.
+  OTP_API_DATE_FORMAT,
   OTP_API_DATE_FORMAT_DATE_FNS
```
amy-corson-ibigroup, line 12: Out of curiosity, does `OTP_API_DATE_FORMAT_DATE_FNS` still need to be imported if it's been removed below?

amy-corson-ibigroup (approved):

lib/components/admin/field-trip-details.js
binh-dam-ibigroup: Good spot @amy-corson-ibigroup, thanks, that constant doesn't exist anymore so it can be removed.

PR #684, opened 2022-10-27
Ignore non-english i18n files.

binh-dam-ibigroup:
This PR changes the codespell exclusion list to ignore all files in the `i18n` folder except for the English one(s).

PR #685, opened 2022-10-27
Attempt again to exclude non-english i18n files.

binh-dam-ibigroup:
#684 had the incorrect syntax (codespell only ignored files not starting with `e` or `n`!).
This PR ~tries to fix that~ now just excludes the `i18n` folder from spellcheck.


binh-dam-ibigroup:
Closing in favor of #686.

PR #686, opened 2022-10-28
Exclude entire i18n folder from codespell

binh-dam-ibigroup:
#684 had the incorrect syntax (codespell only ignored files not starting with `e` or `n`!).
This PR now excludes the entire i18n folder from spellcheck. I tried multiple patterns to try to include the English-US file, but none seem to work (see #685).


miles-grant-ibigroup (approved):
I think this is reasonable. Codespell is mostly for code spelling

amy-corson-ibigroup (approved):

PR #687, opened 2022-10-28
Support OTP2 Tiles

miles-grant-ibigroup:
Makes use of https://github.com/opentripplanner/otp-ui/pull/472 to support native OTP2 vector tile layers. 

The new layer is configured slightly differently than previous ones. The most basic use (showing stops and rental vehicles) is the following:

```yaml
- type: otp2
      layers:
      - type: stops
      - type: rentalVehicles
      - type: rentalStations
```

Network-based layers are also supported.

These layers will only work if the OTP server has vector tiles enabled! http://docs.opentripplanner.org/en/latest/sandbox/MapboxVectorTilesApi/

Layer names in the configuration must match what's in the front end config


lib/components/map/enhanced-stop-marker.js
```diff
@@ -101,6 +101,8 @@ class EnhancedStopMarker extends Component {
     const { activeStopId, highlight, languageConfig, modeColors, stop } =
       this.props
     const { code, id, lat, lon, name } = stop
+    if (!id) return null
+
     const [, stopId] = id.split(':')
     const stopCode = code || stopId
     if (!stopCode) return null
```
amy-corson-ibigroup, line 108: Can line 104 & 108 be consolidated?

lib/actions/api.js
```diff
@@ -355,7 +355,7 @@ export function routingQuery(searchId = null, updateSearchInReducer = false) {
  * @param  {Object} options  The options object for the API request
  * @return {string}          The URL to use for making the http request
  */
-function makeApiUrl(config, endpoint, options) {
+export function makeApiUrl(config, endpoint, options) {
```
amy-corson-ibigroup, line 358: probably a question more for me, but why does this get exported?

amy-corson-ibigroup (approved):
LGTM! Just a few clarifying questions

lib/components/map/enhanced-stop-marker.js
miles-grant-ibigroup: If only! `stopCode` relies on `id` exisitng

lib/actions/api.js
miles-grant-ibigroup: It's so we can use it in `default-map.tsx`!

amy-corson-ibigroup: Great, thanks!

binh-dam-ibigroup (commented):
Some other comments for now.

lib/components/map/default-map.tsx
```diff
@@ -5,17 +5,22 @@ import { connect } from 'react-redux'
...
 import {
   bikeRentalQuery,
   carRentalQuery,
+  makeApiUrl,
   vehicleRentalQuery
 } from '../../actions/api'
 import { ComponentContext } from '../../util/contexts'
 import { getActiveItinerary, getActiveSearch } from '../../util/state'
-import { setMapPopupLocationAndGeocode } from '../../actions/map'
+import {
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Good catch thanks! Not sure how that slipped in

lib/components/map/default-map.tsx
```diff
@@ -380,6 +408,7 @@ const mapDispatchToProps = {
   bikeRentalQuery,
   carRentalQuery,
   setMapPopupLocationAndGeocode,
+  setViewedStop,
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Good catch thanks!

lib/components/map/default-map.tsx
```diff
@@ -238,13 +249,16 @@ class DefaultMap extends Component {
...
+      setLocation,
+      setViewedStop,
       vehicleRentalQuery,
       vehicleRentalStations
     } = this.props
     const { getCustomMapOverlays, getTransitiveRouteLabel, ModeIcon } =
       this.context
     const { baseLayers, maxZoom, overlays } = mapConfig || {}
     const { lat, lon, zoom } = this.state
+    const vectorTilesEndpoint = makeApiUrl(config, 'vectorTiles', {})
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Just double checked and can confirm `base-map` only makes the request once the first layer is selected.

binh-dam-ibigroup (approved):
Approving on the assumption that the expected tile layer ids are correct in this repo (one config I saw had different layer ids for bike rentals and bike stations).

PR #691, opened 2022-11-07
fix(call-taker): support otp2 walk reluctance

miles-grant-ibigroup:
Calltaker previously had some OTP1-specific functionality hardcoded into it. This PR takes the max walk selector and replaces it with a slider, only when the config api is marked as `v2`.

This PR adds a new language key `components.AdvancedOptions.walkReluctance`. I tried to re-use `otpUi.queryParameters.walkReluctance` (which works for this purpose) but the i18n key check script complained that it couldn't find the key. If anyone has a solution for this please let me know.


demory (approved):

lib/components/form/call-taker/advanced-options.js
```diff
@@ -186,21 +193,33 @@ class AdvancedOptions extends Component {
...
           }}
         >
-          <label style={{ fontWeight: '400' }}>
-            <FormattedMessage
-              id="components.AdvancedOptions.maxWalk"
-              values={{ unitsString }}
+          {usingOtp2 ? (
+            <SliderSelector
+              label={
+                <FormattedMessage id="components.AdvancedOptions.walkReluctance" />
```
binh-dam-ibigroup, line 199: Need to use `intl.formatMessage` so that the `aria-label` is a real string instead of a react element.

binh-dam-ibigroup (approved):
LGTM, please address one minor comment regarding the `aria-label` of the walk reluctance selector.

lib/components/form/call-taker/advanced-options.js
miles-grant-ibigroup: Very good catch thanks!

PR #700, opened 2022-11-17
New mode selector

daniel-heppner-ibigroup:
Implements the new mode selector and GraphQL plan queries.

Will require a weblate update afterwards, en-US merged properly but the rest didn't


daniel-heppner-ibigroup:
@miles-grant-ibigroup discovered a very strange crash when accessing [this](http://localhost:9966/#/?ui_activeSearch=1yyhdmszs&ui_activeItinerary=-1&fromPlace=33.78325%2C+-84.39406%3A%3A33.783249048627454%2C-84.39406179237416&toPlace=33.8956%2C+-84.09812%3A%3A33.89560378231087%2C-84.09811757850667&date=2023-02-17&time=17%3A14&arriveBy=false&mode=WALK%2CBUS%2CSUBWAY%2CTRAM%2CFLEX_EGRESS%2CFLEX_ACCESS%2CFLEX_DIRECT&showIntermediateStops=true&walkSpeed=1.34&ignoreRealtimeUpdates=true&wheelchair=false&numItineraries=3&otherThanPreferredRoutesPenalty=900&modeButtons=transit_walk) and clicking on any button. I wasn't able to repeat this with any other url, and I am not sure what it is about this specific instance that causes a crash. I fixed it by adding a null check in itineraries header component.

miles-grant-ibigroup:
Another bug: the accessible routing switch doesn't seem to be working properly!

lib/components/form/batch-settings.tsx
```diff
@@ -71,57 +75,43 @@ type Props = {
...
+  const [expanded, setExpanded] = useState<null | string>(null)
+  const { ModeIcon } = useContext(ComponentContext)
+
+  const modeButtonsWithIcons: ModeButtonDefinition[] =
+    config.modes.modeButtons.map((button: ModeButtonDefinition) => ({
+      ...button,
+      Icon: React.memo(function ModeButtonIcon() {
+        return <ModeIcon mode={button.iconName} />
+      })
+    }))
```
binh-dam-ibigroup, line 102: Can these be moved to `mapStateToProps`? I think redux applies an memo or some optimization that prevents too many frequent changes.

lib/components/form/batch-settings.tsx
```diff
@@ -143,65 +133,36 @@ class BatchSettings extends Component<Props, State> {
...
-            </StyledIconWrapper>
-          </SettingsPreview>
-          <StyledDateTimePreview
-            // as='button'
-            expanded={expanded === 'DATE_TIME'}
-            hideButton
-            onClick={this._toggleDateTime}
+  const _updateExpanded = (type: string) => (expanded === type ? null : type)
+
+  const _toggleDateTime = () => setExpanded(_updateExpanded('DATE_TIME'))
```
binh-dam-ibigroup, line 144: These methods seem superfluous and should be removed. Use the setter from `useState` directly.

lib/components/form/batch-settings.tsx
```diff
@@ -143,65 +133,36 @@ class BatchSettings extends Component<Props, State> {
...
+  return (
+    <>
+      {/* <ModeButtonsFullWidthContainer className="hidden-lg">
+      </ModeButtonsFullWidthContainer> */}
```
binh-dam-ibigroup, line 148: Delete comment.

lib/components/form/batch-settings.tsx
```diff
@@ -216,20 +177,20 @@ class BatchSettings extends Component<Props, State> {
...
+      {expanded === 'DATE_TIME' && (
+        <DateTimeModalContainer>
+          <DateTimeModal />
+        </DateTimeModalContainer>
+      )}
+      {expanded === 'SETTINGS' && (
+        <BatchPreferencesContainer>
+          <BatchPreferences />
+        </BatchPreferencesContainer>
+      )}
```
binh-dam-ibigroup, line 191: Remove. This no longer applies.

lib/components/form/batch-preferences.tsx
```diff
@@ -21,44 +20,9 @@ interface Props {
...
-  (newQueryParamsMode: string) =>
-  (combination: Combination): Combination => {
-    // Split out walk so it's not duplicated
-    const newMode = (newQueryParamsMode || 'WALK,TRANSIT').split('WALK,')?.[1]
-    // Replace TRANSIT with the newly selected parameters
-    const mode = combination.mode.replace('TRANSIT', newMode)
-    return { ...combination, mode }
-  }
-
 class BatchPreferences extends Component<Props> {
```
binh-dam-ibigroup, line 23: This component can be removed now, do you think?

binh-dam-ibigroup (changes requested):
First round of comments. Mostly cleanup until the mode selector API solidifies.

lib/components/form/batch-settings.tsx
```diff
@@ -71,57 +75,43 @@ type Props = {
...
-    })
-  }
+function BatchSettings({
+  activeSearch,
+  config,
+  currentQuery,
+  intl,
+  routingQuery
+}: Props) {
+  const [expanded, setExpanded] = useState<null | string>(null)
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: This variable is keeping track of both the settings element and the datetime element, which is why it's a string.

lib/components/form/batch-settings.tsx
daniel-heppner-ibigroup: oh, of course...

lib/components/form/batch-settings.tsx
daniel-heppner-ibigroup: I don't think we can do that because of the use of ModeIcon from context, which is only available in a component. I believe I looked into putting this in the `mapStateToProps` when I wrote it.

binh-dam-ibigroup:
I think the main thing missing now is that the mode selector selections aren't really synced with the URL params. One example is that when I set `wheelchair=true` in the URL, the accessible routing box doesn't come out checked as it should.

lib/actions/apiV2.js
```diff
@@ -701,12 +742,183 @@ export const findPatternsForRoute = (params) =>
...
+      state.otp.config.modes.modeButtons,
+      [
+        ...defaultModeSettings,
+        ...(state.otp.config.modes.modeSettingDefinitions || [])
+      ],
+      state.otp.config.modes.initialState
+    )
+    // Get the raw query param strings to save for the rider's search history
+    const urlSearchParams = new URLSearchParams(state.router.location.search)
+    const rawModeButtonQP = urlSearchParams.get('modeButtons')
```
binh-dam-ibigroup, line 813: I'm wondering whether the `modeButtons` query param is needed. It mostly mirrors the `mode` query param, right?

binh-dam-ibigroup:
> moved away from Redux-URL State sync

I think being to paste a link and get the mode selected according to that in the UI is an awesome feature.

lib/actions/apiV2.js
daniel-heppner-ibigroup: I'd like to remove the `mode` queryParam since that is for the old system. Mode buttons do not necessarily map one to one with modes either since a button could activate multiple modes.

daniel-heppner-ibigroup:
> I think the main thing missing now is that the mode selector selections aren't really synced with the URL params. One example is that when I set `wheelchair=true` in the URL, the accessible routing box doesn't come out checked as it should.

It is synced, but the mode settings are all grouped together under `modeSettings` query param. There are not specific query parameters for individual settings, since they're all customizable from the config. If we want to add some hard coded mappings for backwards compatibility, we could check those on application startup for backwards compatibility, but I think that would be quite messy.

lib/actions/api.js
```diff
@@ -204,6 +211,15 @@ export function updateOtpUrlParams(state, searchId) {
       .forEach((key) => {
         params[key] = urlParams[key]
       })
+
+    // Populate the URL parameters for the mode buttons if those are specified in the currentQuery
+    if (currentQuery.modeButtons) {
+      params.modeButtons = currentQuery.modeButtons
+    }
+    if (currentQuery.modeSettings) {
+      params.modeSettings = currentQuery.modeSettings
```
binh-dam-ibigroup, line 220: Remove query parameters from unactivated modes to shorten and clean up the URL. Remove query parameters that are set on their default value.

binh-dam-ibigroup: Can we use `:` as separator? We use `::` as separator for query from/to locations.

binh-dam-ibigroup (changes requested):
Remove the unnecesary file changes, and implement removal of mode settings variables for modes that aren't active and variables set to their default value.

lib/components/form/plan-trip-button.tsx
```diff
@@ -0,0 +1,43 @@
...
+    if (typeof onClick === 'function') onClick()
+    if (!coreUtils.ui.isMobile()) setMainPanelContent(null)
+  }, [onClick, setMainPanelContent])
+
+  const locationMissing = false // TODO: get from query params
+  const grayedOut = locationMissing || disabled
+  return (
+    <Button
+      className="plan-trip-button"
+      disabled={grayedOut}
```
binh-dam-ibigroup, line 33: A11Y: Do not disable the plan trip button, instead, let users click it, and if something is missing, tell them at that point. See https://css-tricks.com/making-disabled-buttons-more-inclusive/, https://www.smashingmagazine.com/2021/08/frustrating-design-patterns-disabled-buttons/

binh-dam-ibigroup (changes requested):
There is one more big blocking issue regarding the mode buttons resetting when clicking the "Plan Trip" button.

lib/components/util/styledIcon.tsx
```diff
@@ -51,19 +51,21 @@ export const StyledIconWrapper = styled.span<Props>`
   animation: ${(props) => (props.spin ? rotateAnimation : 'none')} 1s linear
     infinite;
   display: ${(props) => (props.spin ? 'inline-block' : 'initial')};
-  ${StyledIconBase} {
+  svg {
     font-size: ${(props) => getFontSize(props.size)};
     height: 1em;
+    height: 1em;
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: bad merge

lib/components/form/batch-settings.tsx
```diff
@@ -134,99 +94,54 @@ class BatchSettings extends Component<Props, State> {
...
-          >
-            {coreUtils.query.isNotDefaultQuery(currentQuery, config) && (
-              <Dot className="dot" />
+  return (
+    <div role="group">
+      <MainSettingsRow>
+        <StyledDateTimePreviewContainer
+          aria-controls="date-time-modal"
+          aria-expanded={dateTimeExpanded}
+          aria-label="Date/Time settings"
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: Are you asking me to make a change besides merging dev?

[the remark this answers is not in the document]
binh-dam-ibigroup: Merging with dev should be enough.

lib/actions/api.js
```diff
@@ -204,6 +211,15 @@ export function updateOtpUrlParams(state, searchId) {
       .forEach((key) => {
         params[key] = urlParams[key]
       })
+
+    // Populate the URL parameters for the mode buttons if those are specified in the currentQuery
+    if (currentQuery.modeButtons) {
+      params.modeButtons = currentQuery.modeButtons
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: This was actually brought up a while ago, I think. The problem with this is that `modes` are different from `modeButtons`, since `modeButtons` refers to the selected buttons, each of which contains various modes in it that get activated or deactivated with the button. 

I've had to re-add support for directly specifying the modes in the currentQuery object for call taker and field trip.

lib/components/form/batch-settings.tsx
```diff
@@ -1,118 +1,78 @@
...
+      config.modes.initialState,
+      [...defaultModeSettings, ...(config.modes.modeSettingDefinitions || [])],
+      {
+        queryParamState: true
+      }
+    )
 
-  _planTrip = () => {
-    const { currentQuery, intl, routingQuery } = this.props
+  const _planTrip = () => {
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: The fix for this involves breaking the saved searches functionality. I think this is something we should chat about, I need another mind to think about a solution.

lib/components/form/batch-settings.tsx
```diff
@@ -1,118 +1,124 @@
...
   routingQuery: any
   setQueryParam: (queryParam: any) => void
+  setUrlSearch: (evt: any) => void
+  urlSearchParams: URLSearchParams
 }
 
-type State = {
-  expanded?: string | null
-  selectedModes: string[]
+function pipe<T>(...fns: Array<(arg: T) => T>) {
```
binh-dam-ibigroup, line 58: Add a brief comment about this function.

lib/components/form/batch-settings.tsx
```diff
@@ -71,57 +75,43 @@ type Props = {
...
-      combinations: newCombinations,
-      disabledModes,
-      enabledModes: newModes
-    })
-  }
+function BatchSettings({
+  activeSearch,
+  config,
+  currentQuery,
+  intl,
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Still needs to be addressed.

lib/components/form/batch-settings.tsx
```diff
@@ -134,114 +140,108 @@ class BatchSettings extends Component<Props, State> {
...
+          aria-label="Date/Time settings"
+          expanded={dateTimeExpanded}
+          onClick={() => setDateTimeExpanded(!dateTimeExpanded)}
+        >
+          <StyledDateTimePreview hideButton />
+        </StyledDateTimePreviewContainer>
+        <MetroModeSelector
+          fillModeIcons={fillModeIcons}
+          modeButtons={processedModeButtons}
+          onSettingsUpdate={(evt) => setUrlSearch(evt)}
```
binh-dam-ibigroup, line 179:
```suggestion
          onSettingsUpdate={setUrlSearch}
```

lib/components/form/batch-styled.ts
```diff
@@ -94,13 +106,19 @@ export const BatchPreferencesContainer = styled.div`
...
-
 export const MainSettingsRow = styled.div`
   align-items: top;
   display: flex;
   flex-direction: row;
-  justify-content: flex-start;
+  justify-content: space-between;
   margin-bottom: 5px;
+
+  @media (max-width: 355px), (min-width: 991px) and (max-width: 1110px) {
```
binh-dam-ibigroup, line 116: Can you use the bootstrap thresholds? Otherwise, add a comment to explain the choice of thresholds.

lib/components/form/batch-styled.ts
```diff
@@ -94,13 +106,19 @@ export const BatchPreferencesContainer = styled.div`
...
+    display: grid;
```
binh-dam-ibigroup, line 117: This produces the behavior in the screenshot below. Is that desired?

image

lib/components/form/batch-styled.ts
```diff
@@ -94,13 +106,19 @@ export const BatchPreferencesContainer = styled.div`
...
+    grid-template-columns: repeat(4, 20%);
+    grid-auto-rows: 1fr;
```
binh-dam-ibigroup, line 119: Sort CSS.

lib/components/util/styledIcon.tsx
```diff
@@ -51,19 +51,19 @@ export const StyledIconWrapper = styled.span<Props>`
...
-  ${StyledIconBase} {
+  svg {
     font-size: ${(props) => getFontSize(props.size)};
     height: 1em;
     transform: ${(props) => `
-      ${props.flipHorizontal ? 'scale(-1,1) ' : ''}
-      ${props.rotate90 ? 'rotate(90deg)' : ''}
-      `};
+      ${props.flipHorizontal ? 'scale(-1,1) ' : ''} 
+      ${props.rotate90 ? 'rotate(90deg)' : ''}
```
binh-dam-ibigroup, line 59: Remove trailing whitespace.

binh-dam-ibigroup (commented):
Reminder about some items still to be resolved, and a few other things.

lib/actions/api.js
binh-dam-ibigroup: > Remove query parameters from unactivated modes to shorten and clean up the URL. Remove query parameters that are set on their default value.

I think this still needs to be addressed.

lib/components/form/batch-settings.tsx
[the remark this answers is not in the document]
binh-dam-ibigroup: This has been fixed by https://github.com/opentripplanner/otp-react-redux/pull/700/commits/5d881fd3f9b467c153455695d831c502126a4030. Nice job.

lib/actions/apiV2.js
```diff
@@ -19,12 +34,46 @@ import {
...
+    query: getTripOptionsFromQuery(
+      { ...state.otp.currentQuery, queryParamData },
+      true
+    ),
+    timestamp: new Date().getTime()
+  }
+}
+
+function isStoredPlace(place) {
+  return ['home', 'work', 'suggested', 'stop'].indexOf(place.type) !== -1
```
miles-grant-ibigroup, line 74: This might need some work...

daniel-heppner-ibigroup: This code doesn't look familiar to me at all

binh-dam-ibigroup: What does this have to do with the mode selector?

daniel-heppner-ibigroup: Actually, since the functions are no longer used anywhere else, I just deleted them from the api.js file. I think at this point any further refactoring can be left to the future OTPv1 removal PR, which will change all of this.

miles-grant-ibigroup:
Percy is still flagging 2 issues: accessibility scores are still missing and trip data isn't coming back either

lib/actions/apiV2.js
```diff
@@ -706,12 +756,255 @@ export const findPatternsForRoute = (params) =>
...
+        ? convertModeSettingValue(setting, fromUrl)
+        : config?.modes?.initialState?.modeSettingValues?.[setting.key] ||
+          setting.default
+      return acc
+    }, {})
+    const modeSettings = modeSettingDefinitions?.map(
+      populateSettingWithValue(modeSettingValues)
+    )
+
+    // Get the raw query param strings to save for the rider's search history
```
binh-dam-ibigroup, line 876: Some params aren't being parsed correctly. For instance, removing the bus option still results in itineraries with a bus leg being returned (see screenshot).

image

It looks like boolean params are converted to strings in the OTP-UI package in use, which probably resolve to `true` (see debugger screenshot).
image

binh-dam-ibigroup (changes requested):
Setting state for checkboxes seems to be parsed incorrectly and needs to be fixed. See also other minor items.

lib/components/narrative/metro/metro-error-renderer.tsx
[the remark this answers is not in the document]
miles-grant-ibigroup: It was, it was then merged into this one

lib/actions/apiV2.js
daniel-heppner-ibigroup: Yep, great catch. That was happening because I added a new type, "SUBMODE" which behaves just like "CHECKBOX", but I forgot to add that to the utility function which converts the true/false string into an actual boolean, so it was staying as a string.

lib/components/app/call-taker-panel.js
```diff
@@ -120,8 +122,17 @@ class CallTakerPanel extends Component {
     }
   }
 
+  _handleUpdateModes = (modes) => {
+    this.setState({ enabledModes: modes })
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: No. We could. Is that needed for call taker?

lib/actions/apiV2.js
daniel-heppner-ibigroup: Fixed in OTP UI, new alpha package version added to this PR.

lib/components/util/styledIcon.tsx
daniel-heppner-ibigroup: There's no trailing whitespace on my local copy, so there's nothing here for me to remove.

lib/actions/apiV2.js
```diff
@@ -701,12 +742,183 @@ export const findPatternsForRoute = (params) =>
...
+          (response) => {
+            const dispatchedRoutingResponse = routingResponse(response)
+            // If tracking is enabled, store locations and search after successful
+            // search is completed.
+            if (
+              persistenceMode.isLocalStorage &&
+              state.user?.localUser?.storeTripHistory
+            ) {
+              const { from, to } = currentQuery
+              if (!isStoredPlace(from)) {
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: Moving them outside the loop increases the function's complexity too high. Maybe I can simplify it somewhere else. I don't want to make a new dispatched action for this since that would require a ton of argument passing that would be messy.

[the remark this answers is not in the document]
daniel-heppner-ibigroup: I couldn't find a nice way of doing this. Moving these evaluations outside of the loop increases the function's complexity too much. I think this really should just all be refactored in a future PR.

lib/components/narrative/narrative-itineraries.js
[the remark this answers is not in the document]
daniel-heppner-ibigroup: It was a PR into this one along with the other error stuff

lib/components/narrative/metro/metro-error-renderer.tsx
```diff
@@ -0,0 +1,87 @@
...
+    <List>
+      {Object.keys(errors).map((error: string) => {
+        const localizedInputFieldList = Array.from(errors[error])?.map(
+          (inputField) =>
+            intl.formatMessage({
+              id: `components.OTP2ErrorRenderer.inputFields.${inputField}`
+            })
+        )
+
+        // The search window is set by the client and can't be changed by the user.
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: @miles-grant-ibigroup

i18n/fr.yml
```diff
@@ -744,9 +748,9 @@ config:
     both: Plus de détails au bas de l'itinéraire
     call-ahead: Appelez pour réserver
     continuous-dropoff: Demandez l'arrêt au conducteur
+  micromobilityModes:
     flex-service: Service Flex
     flex-service-colon: "Service Flex :"
-  micromobilityModes:
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: Oh shoot! Bad merge

miles-grant-ibigroup:
I've addressed everything assigned to me

lib/components/form/batch-settings.tsx
```diff
@@ -124,7 +123,7 @@ function BatchSettings({
     )
   )
 
-  const _planTrip = () => {
+  const _planTrip = useCallback(() => {
```
binh-dam-ibigroup, line 126: Just thought of this... Changes from #861 need to be merged into this PR.

lib/components/form/batch-styled.ts
```diff
@@ -73,10 +82,31 @@ export const PlanTripButton = styled(Button)`
   }
 `
 
+export const ModeSelectorContainer = styled.div<{ squashed?: boolean }>`
```
binh-dam-ibigroup, line 85: Nice-to-have: The top margin of the `<input>` elements, set by bootstrap, should probably be removed in the popups.

lib/components/form/batch-settings.tsx
```diff
@@ -136,81 +146,60 @@ class BatchSettings extends Component<Props, State> {
...
+  )
+
+  // We can rely on this existing, as there is a default
+  const baseColor = getComputedStyle(document.documentElement).getPropertyValue(
+    '--main-base-color'
+  )
+  const accentColor = tinycolor(baseColor).darken(10)
+
+  return (
+    <div role="group">
```
binh-dam-ibigroup, line 180: Remove the "group" role per client request, so that the `<fieldset>` element can be correctly announced by assistive technology.

binh-dam-ibigroup (changes requested):
Need features from #861 and another a11y tweak. Also, the checkboxes still don't have the desired effect.

lib/components/form/batch-settings.tsx
daniel-heppner-ibigroup: Right, this had been mentioned to me, thanks for the reminder. I replaced the div with a fragment.

lib/components/form/batch-styled.ts
daniel-heppner-ibigroup: That makes the alignment much better! Thank you!

binh-dam-ibigroup (commented):
Much better now, seeing the light! The basic functionality works now. There is some polishing needed but that can be done in a subsequent PR. I'll approve once the non-alpha version is released.

daniel-heppner-ibigroup:
Current plan is to let @philip-cline review this while we wait for all the alpha builds to be merged.

package.json
```diff
@@ -38,28 +38,28 @@
...
-    "@opentripplanner/route-viewer-overlay": "^2.0.9",
-    "@opentripplanner/stop-viewer-overlay": "^2.0.4",
-    "@opentripplanner/stops-overlay": "^5.0.7",
-    "@opentripplanner/transit-vehicle-overlay": "^4.0.2",
-    "@opentripplanner/transitive-overlay": "^3.0.11",
-    "@opentripplanner/trip-form": "^3.1.0-alpha.25",
-    "@opentripplanner/trip-details": "^4.0.0",
-    "@opentripplanner/trip-viewer-overlay": "^2.0.4",
-    "@opentripplanner/vehicle-rental-overlay": "^2.1.0",
+    "@opentripplanner/trip-form": "^3.0.6",
```
binh-dam-ibigroup: Sort dependencies. Is this the correct version number?

lib/actions/api.js
binh-dam-ibigroup, line 1: Could you address the "defined but never used" warnings for this file?

binh-dam-ibigroup (commented):
One more thing about passing custom mode names.

binh-dam-ibigroup:
@daniel-heppner-ibigroup Please do the cleanup and advise regarding versions and passing custom mode names.

lib/components/narrative/default/itinerary-description.tsx
```diff
@@ -3,6 +3,7 @@ import { Itinerary } from '@opentripplanner/types'
 import coreUtils from '@opentripplanner/core-utils'
 import React, { useContext } from 'react'
 
+import { isRideshareLeg } from '@opentripplanner/core-utils/lib/itinerary'
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Addressed in 0382c541cb579e3cf41f2545bc060a7e1aa85a9e

package.json
binh-dam-ibigroup: This PR should use `@opentripplanner/trip-form@^3.1.0`.

lib/components/form/batch-settings.tsx
```diff
@@ -1,119 +1,129 @@
...
+  const populateSettingWithIcon = useCallback(
+    (msd: ModeSetting) => ({
+      ...msd,
+      icon: <ModeIcon mode={msd.iconName} width={16} />
+    }),
+    [ModeIcon]
+  )
+
+  const processedModeSettings = modeSettingDefinitions.map(
+    pipe(populateSettingWithIcon, populateSettingWithValue(modeSettingValues))
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: I think I had in my head that is handled by i18n stuff, which I realize is not right. I can add another step to this to pull in those strings.

daniel-heppner-ibigroup:
I added functionality for custom mode names using the old `getFormattedMode` function. However, this was always possible thanks to the i18n keys.

binh-dam-ibigroup:
🚨 @daniel-heppner-ibigroup  and @miles-grant-ibigroup The type check is failing.

lib/actions/apiV2.js
```diff
@@ -1,6 +1,22 @@
+import {
+  aggregateModes,
+  filterModeDefitionsByKey,
+  populateSettingWithValue
+} from '@opentripplanner/trip-form/esm/MetroModeSelector/utils'
+import { decodeQueryParams, DelimitedArrayParam } from 'use-query-params'
+import { randId } from '@opentripplanner/core-utils/lib/storage'
```
binh-dam-ibigroup, line 7: Use `const { randId } = `coreUtils.storage`.

lib/actions/apiV2.js
```diff
@@ -706,12 +757,254 @@ export const findPatternsForRoute = (params) =>
...
+    case 'SCHEDULED':
+      return 'scheduled'
+    case 'NONE':
+      return 'none'
+    default:
+      return null
+  }
+}
+
+const processLeg = (leg) => ({
```
binh-dam-ibigroup, line 784: Add JSDoc

lib/actions/ui.js
```diff
@@ -308,11 +308,7 @@ export function setItineraryView(value) {
     // set the desired ui query param, or remove it if same as default,
     // and store the current view as previousItineraryView.
     if (value !== urlParams.ui_itineraryView) {
-      if (value !== ItineraryView.LIST) {
-        urlParams.ui_itineraryView = value
-      } else if (urlParams.ui_itineraryView) {
-        delete urlParams.ui_itineraryView
-      }
+      urlParams.ui_itineraryView = value
```
binh-dam-ibigroup, line 311: Also delete the "remove" phrase from the comment above.

lib/components/form/batch-settings.tsx
```diff
@@ -136,81 +163,60 @@ class BatchSettings extends Component<Props, State> {
...
+    (buttonId: string, newState: boolean) => {
+      let newButtons
+      if (newState) {
+        newButtons = [...enabledModeButtons, buttonId]
+      } else {
+        newButtons = enabledModeButtons.filter((c) => c !== buttonId)
+      }
+
+      // encodeQueryParams serializes the mode buttons for the URL
+      // It uses encodeQueryParams to get nice looking URL params and consistency
```
binh-dam-ibigroup, line 182: Deduplicate comment.

lib/components/form/batch-settings.tsx
```diff
@@ -136,81 +163,60 @@ class BatchSettings extends Component<Props, State> {
...
+  )
+  const accentColor = tinycolor(baseColor).darken(10)
 
+  return (
+    <>
+      <MainSettingsRow>
+        <StyledDateTimePreviewContainer
+          aria-controls="date-time-modal"
+          aria-expanded={dateTimeExpanded}
+          aria-label="Date/Time settings"
```
binh-dam-ibigroup, line 202: Use `intl.formatMessage` here.

lib/components/form/batch-styled.ts
```diff
@@ -39,7 +46,9 @@ export const StyledDateTimePreviewContainer = styled(Button)<{
   margin-right: 5px;
   padding: 0;
   position: relative;
-  width: 120px;
+  min-width: 90px;
```
binh-dam-ibigroup, line 49: Sort CSS.

lib/components/narrative/narrative-itineraries.js
```diff
@@ -464,12 +475,28 @@ class NarrativeItineraries extends Component {
...
+          {this._renderLoadingSpinner()}
         </div>
       </S.NarrativeItinerariesContainer>
     )
   }
 }
 
+const reduceErrorsFromResponse = (acc, cur) => {
+  const { routingErrors } = cur?.plan
+  // code and inputfield
```
binh-dam-ibigroup, line 487: Remove or update this comment.

lib/components/narrative/metro/metro-error-renderer.tsx
```diff
@@ -0,0 +1,85 @@
...
+        if (error === 'NO_TRANSIT_CONNECTION_IN_SEARCH_WINDOW') {
+          return null
+        }
+
+        return (
+          <Container key={error}>
+            <Icon Icon={ExclamationCircle} size="3x" />
+            <h3>
+              <FormattedMessage
+                id={`components.OTP2ErrorRenderer.${error}.header`}
```
binh-dam-ibigroup, line 65: Note for another PR: After pasting a URL in the browser's search bar, the error message might appear, but can remain even if results were updated.

image

lib/components/form/call-taker/mode-dropdown.tsx
```diff
@@ -0,0 +1,58 @@
...
+      const newModes = modeDropdownOptions.find(
+        (mdo) => mdo.label === e.target.value
+      )?.combination
+      onChangeModes(newModes || [])
+    },
+    []
+  )
+
+  return (
+    <select onBlur={onChange} onChange={onChange} value={selectedMode}>
```
binh-dam-ibigroup, line 40: Note for another PR: not sure if it is a config issue, but the "Rail only" option seems to be too strict, returning no itineraries whereas the "transit" option produces a Walk-MAX itinerary.

binh-dam-ibigroup (approved):
Nice work (that took too long)! Please see the minor tweaks suggested and fix the TypeScript errors. Include #921 to fix the broken date/time selector styles.

lib/components/form/call-taker/mode-dropdown.tsx
daniel-heppner-ibigroup: Probably just bad config.

lib/components/form/call-taker/advanced-options.js
```diff
@@ -104,7 +104,16 @@ class AdvancedOptions extends Component {
   _onSubModeChange = (changedMode) => {
     // Get previous transit modes from state and all modes from query.
     const transitModes = [...this.state.transitModes]
-    const allModes = this.props.currentQuery.mode.split(',')
+    const allModes = this.props.currentQuery?.modes
+      // todo: @daniel-heppner-ibigroup clean this up
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: This is my first time seeing this... this code is pretty wild. Hopefully we can move calltaker out of OTP-RR soon.

philip-cline (approved):
I tried to test this pretty rigorously in the following areas and couldn't find anything wrong: 
- call-taker
- trip saving
- slow network speeds
- various different configurations

So it's looking pretty good for me. Fixed that one whitespace issue.

Great work !

PR #705, opened 2022-11-18
Do not reset UI upon acquiring location if viewer is active

### LINKED ISSUE #702: Mobile view: UI resets when acquiring device location
In mobile view, the UI resets to the landing page (`/#/` path) after the location of the device is acquired.

Steps to reproduce on Chrome Desktop with mobile emulation:
1. Load `http://localhost:9966/#/` and enable location access (reload the page after enabling that setting)
2. Observe that the device location is not immediately available
3. Go to the route viewer
4. Observe that the UI "returns" to the root path and that the "from" location is set to your device location.

binh-dam-ibigroup:
Fix #702.

This PR removes the resetting of the UI after device location is acquired.
Test using for instance http://localhost:9966/#/route, http://localhost:9966/#/route/<route_id>, http://localhost:9966/#/stop/<stop_id> where the viewed route or stop should remain on screen after device location is acquired.

### DIFF
--- a/lib/actions/form.js
+++ b/lib/actions/form.js
@@ -206,9 +206,6 @@ export function formChanged(oldQuery, newQuery) {
         if (ui.mobileScreen !== MobileScreens.WELCOME_SCREEN) {
           dispatch(setMobileScreen(MobileScreens.SEARCH_FORM))
         }
-        if (ui.mainPanelContent > 0) {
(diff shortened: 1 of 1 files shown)


daniel-heppner-ibigroup (approved):
I am having trouble reproducing the original bug in my browser, but the change looks fine to me and didn't cause any obvious bugs.

PR #708, opened 2022-11-22
fix(actions/apiV2): Add stop code field to stops graphql query.

binh-dam-ibigroup:
This PR adds the stop code field to stops graphql queries, so the stop code is shown instead of the GTFS stop id in the stop viewer and related stops.


miles-grant-ibigroup (approved):
Good catch thanks! I feel this is already part of another PR but no worries it should merge cleanly.

binh-dam-ibigroup:
> Good catch thanks! I feel this is already part of another PR but no worries it should merge cleanly.

Do you have the PR number? I thought it would have been in #661 but it is not.

amy-corson-ibigroup (approved):

PR #710, opened 2022-11-23
Restore Route Preview Overlay

miles-grant-ibigroup:
The move to vector tiles removed the route preview overlay. This PR restores the overlay.


lib/components/map/route-preview-overlay.tsx
```diff
@@ -0,0 +1,100 @@
...
+  const geometries = getActiveSearch(state)?.response.flatMap(
+    (serverResponse: { plan?: { itineraries?: Itinerary[] } }) =>
+      serverResponse?.plan?.itineraries?.flatMap((itinerary) => {
+        return itinerary.legs?.map((leg) => leg.legGeometry.points)
+      })
+  )
+
+  return {
+    geometries,
+    visible:
```
binh-dam-ibigroup, line 87: (non-blocking) Could it be shown always while the list of results is shown?

binh-dam-ibigroup (approved):
Some light clean up needed, but this does the job.

lib/components/map/route-preview-overlay.tsx
```diff
@@ -0,0 +1,100 @@
...
+          return {
+            geometry: polyline.toGeoJSON(segment),
+            properties: [],
+            type: 'Feature'
+          }
+        }),
+      type: 'FeatureCollection'
+    }
+    return (
+      <>
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Nope! Good catch thanks

lib/components/map/route-preview-overlay.tsx
miles-grant-ibigroup: It looks very weird when overlapping with an active itinerary

PR #711, opened 2022-11-29
Update map overlays

binh-dam-ibigroup:
This PR updates map overlays with OTP-UI packages recently released:
- core-utils (fix null route color)
- stops overlay (suppress console warnings)
- transitive overlay (place and route labels)
- transit vehicle overlay (new symbols, and showing popups)
- route viewer overlay fitting flex routes.


daniel-heppner-ibigroup (approved):
Seems to work fine for me, although I did just notice an issue with the transitive labels. It seems like there's some condition where the label can appear but the background doesn't.  It seems pretty rare and only happens and certain zoom levels in Safari. 
I'll approve since it's not a big deal but maybe this is something to revisit in OTP-UI in the future.
image

binh-dam-ibigroup:
> It seems like there's some condition where the label can appear but the background doesn't.

That is correct. The background has a larger size than the label, so occasionally maplibre will skip the background if it thinks space is constrained.

lib/components/map/connected-transit-vehicle-overlay.js
```diff
@@ -110,6 +117,8 @@ const mapStateToProps = (state) => {
   }
   return {
```
miles-grant-ibigroup, line 118: We should style the transit vehicle overlay to have the arrows attached to the border. I think we should also restore the color hover effect.

miles-grant-ibigroup (changes requested):
Works and looks great thanks for all the updates! However, I think we need to adjust the transit vehicle overlay styling

lib/components/map/connected-transit-vehicle-overlay.js
binh-dam-ibigroup: Good point, fixed in https://github.com/opentripplanner/otp-react-redux/pull/711/commits/89b0e2e3cf8e55d418d0cb0870f36101cc938cdd.

miles-grant-ibigroup (approved):
Looks good now! Could use some final styling tweaks but looks good for now

PR #712, opened 2022-12-01
Route viewer visual tweaks

binh-dam-ibigroup:
This PR makes improves the route viewer layout, so route numbers are not shrunk on narrow screens and long route numbers are properly clipped.

### DIFF
--- a/lib/components/narrative/metro/default-route-renderer.tsx
+++ b/lib/components/narrative/metro/default-route-renderer.tsx
@@ -7,8 +7,13 @@ const Block = styled.span<{ color: string }>`
   border-radius: 5px;
   border-top: 5px solid #${(props) => props.color};
   display: inline-block;
-  max-width: 150px;
   padding: 3px 7px;
+
+  /* Below is for route names that are too long: cut-off and show ellipsis. */
+  max-width: 150px;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
 `
 
 type RouteRendererProps = {
--- a/lib/components/narrative/metro/default-route-renderer.tsx
+++ b/lib/components/narrative/metro/default-route-renderer.tsx
@@ -13,6 +13,8 @@ const Block = styled.span<{ color: string }>`
   max-width: 150px;
   overflow: hidden;
   text-overflow: ellipsis;
+  /* (Change the vertical alignment when changing the overflow attribute.) */
+  vertical-align: middle;
   white-space: nowrap;
 `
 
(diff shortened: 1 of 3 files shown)


amy-corson-ibigroup (approved):
This works really well! Just had a comment regarding margins and spacing, but it's solving the clipping problem beautifully on smaller screens.

lib/components/viewers/RouteRow.js
```diff
@@ -24,41 +24,34 @@ export const RouteRowButton = styled(Button)`
...
-const RouteNameElement = styled(Label)`
-  background-color: transparent;
-  color: inherit;
-  flex: 0 1 auto;
+const RouteNameElement = styled.span`
+  flex-shrink: 0;
   font-size: 16px;
   font-weight: 400;
   margin-left: 8px;
-  margin-top: 2px;
```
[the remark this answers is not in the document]
binh-dam-ibigroup: I like to have the route number close to the mode icon, but increasing the margin between the route number and route long name (e.g. to 16px or 20px) makes sense.

lib/components/narrative/metro/default-route-renderer.tsx
```diff
@@ -2,16 +2,21 @@ import { Leg } from '@opentripplanner/types'
...
-  text-align: center;
+  border-top: 5px solid #${(props) => props.color};
   display: inline-block;
-  max-width: 12vw;
+  margin-top: -2px;
+  padding: 3px 7px;
+
+  /* Below is for route names that are too long: cut-off and show ellipsis. */
+  max-width: 150px;
+  overflow: hidden;
```
miles-grant-ibigroup, line 15: Not sure if it's this line, but there is some trouble in the itinerary body that wasn't there previously.
Screenshot 2022-12-06 at 9 58 50 AM

miles-grant-ibigroup (approved):
Looking good -- one minor bug but nothing blocking. Thanks for the cleanup!

lib/components/narrative/metro/default-route-renderer.tsx
binh-dam-ibigroup: Is this on Safari? This also appears in the dev branch.

miles-grant-ibigroup: Might've been a cache issue when I switched back to dev to check. If it's not a regression then we can leave as is

binh-dam-ibigroup: That probably comes from this line: https://github.com/opentripplanner/otp-react-redux/blob/20ca3fa4531a28a9a7d0c0fb2101f0db9a1c03a4/lib/components/narrative/metro/route-block.tsx#L30

PR #713, opened 2022-12-01
Render itinerary bodies using ambient font

binh-dam-ibigroup:
This overrides the `ItineraryBody` component's internal font so that the ambient font is used throughout the UI. (This helps support fonts that support Vietnamese letters, among others).

### DIFF
--- a/lib/components/narrative/line-itin/connected-itinerary-body.js
+++ b/lib/components/narrative/line-itin/connected-itinerary-body.js
@@ -37,6 +37,10 @@ const ItineraryBodyContainer = styled.div`
 `
 
 const StyledItineraryBody = styled(ItineraryBody)`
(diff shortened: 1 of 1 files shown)


amy-corson-ibigroup (approved):

miles-grant-ibigroup (commented):
Do you think this might be better in the global stylesheet?

binh-dam-ibigroup:
> Do you think this might be better in the global stylesheet?

It's going to be difficult because
1. `ItineraryBody` does not have a set class name, and
2. Even with that, `ItineraryBody`'s own styles would override the stylesheet.

miles-grant-ibigroup (approved):
That makes sense. I prefer having the same style multiple places over `!important`! Thanks for thinking it through with me

PR #717, opened 2022-12-05
Make first itinerary leg clickable

binh-dam-ibigroup:
Fixes a bug that prevents fitting the shape of the first leg of an itinerary on the map when clicking that in the narrative.


daniel-heppner-ibigroup (approved):
Awesome! I was dealing with some related bugs the other day in my own branch related to this, I think.

PR #719, opened 2022-12-06
Rebuild Calltaker date time selector

miles-grant-ibigroup:
The previous date time selector was built before react-intl and therefore had some serious design flaws. The new one is simpler and hopefully more bug-free.

I would appreciate more extensive testing! I tried to break it in every way I could think of but there may be more ways...


lib/components/form/call-taker/date-time-picker.tsx
```diff
@@ -0,0 +1,223 @@
...
+  'h:mmaaaaa',
+  'hmmaaaaa',
+  'haaaaa',
+  'Hmm',
+  'Hm',
+  'H:mm',
+  'H:m',
+  'H',
+  'HH:mm'
+]
```
amy-corson-ibigroup, line 44: Is there a reason why you're only using 'a' and 'p' here as supported formats? It's definitely faster to type "5p" instead of needing to type "5pm", but it does feel odd for '5pm' to throw an error

amy-corson-ibigroup:
timeloop

Seeing a little bug here where the time input refuses to be empty

lib/components/form/call-taker/date-time-picker.tsx
```diff
@@ -0,0 +1,223 @@
...
+          onKeyDown={onKeyDown}
+          style={{
+            fontSize: 'inherit',
+            lineHeight: '.8em',
+            marginLeft: '3px',
+            padding: '0px',
+            width: '50px'
+          }}
+          // Don't use intl.formatTime, so that users can enter time in 12hr or 24hr format at their leisure.
+          value={time || format(dateTime, 'H:mm', { timeZone: homeTimezone })}
```
amy-corson-ibigroup, line 189: This is causing a tiny bug where the input refuses to be empty
timeloop

lib/components/form/call-taker/date-time-picker.tsx
```diff
@@ -0,0 +1,223 @@
+import { connect } from 'react-redux'
+import { format, toDate } from 'date-fns-tz'
+import { getCurrentTime } from '@opentripplanner/core-utils/lib/time'
+import { IntlShape, useIntl } from 'react-intl'
+import { isMatch, parse } from 'date-fns'
+import { OverlayTrigger, Tooltip } from 'react-bootstrap'
+import coreUtils from '@opentripplanner/core-utils'
+import React, { useEffect, useState } from 'react'
```
amy-corson-ibigroup, line 8: Super nitpicky, but do you need to import `React` here?

amy-corson-ibigroup (changes requested):
Overall looks really good! I was able to find a couple small bugs and unfortunately it does look like emptying the time and date inputs crashes the site. 

Also, I wasn't able to see the map tiles but that could be a config error in my setup!

https://user-images.githubusercontent.com/115499534/207439655-2b8db3b8-c15c-404c-b296-9525d9476703.mov

lib/components/form/call-taker/date-time-picker.tsx
miles-grant-ibigroup: Good catch thanks! Totally just forgot it. In there now

lib/components/form/call-taker/date-time-picker.tsx
miles-grant-ibigroup: I think our new compiler requires it unfortunately. I agree it's really annoying

amy-corson-ibigroup: Ah, gotcha!! Good to know for the future too!

miles-grant-ibigroup:
I've made some adjustments and tested a little, but I'd appreciate you'd taking another closer look! I feel it's easy to miss things with this

lib/components/form/call-taker/date-time-picker.tsx
```diff
@@ -0,0 +1,235 @@
...
+            {text}
+          </option>
+        ))}
+      </select>
+      <OverlayTrigger
+        overlay={
+          <Tooltip id="time-tooltip">
+            {safeFormat(dateTime, timeFormat, {
+              timeZone: homeTimezone
+            })}
```
amy-corson-ibigroup, line 175: Would it be helpful to have a little error msg get rendered here instead of an empty tooltip when the time format doesn't work? Something like 
   ```
        <Tooltip id="time-tooltip">
            {dateTime ? safeFormat(dateTime, timeFormat, {
              timeZone: homeTimezone
            }) : "Invalid Time"}
        </Tooltip>
```
I'll defer to you, I think it works okay as is!

File: lib/components/form/call-taker/date-time-picker.tsx [hunk 5]
```diff
@@ -0,0 +1,235 @@
...
+        onKeyDown={onKeyDown}
+        style={{
+          fontSize: '14px',
+          lineHeight: '1em',
+          outline: 'none',
+          width: '109px'
+        }}
+        type="date"
+        value={safeFormat(dateTime, OTP_API_DATE_FORMAT, {
+          timeZone: homeTimezone
```
amy-corson-ibigroup, line 219: Very small issue but this line causes the date to zero out if the time input is invalid: 
https://user-images.githubusercontent.com/115499534/208951708-5c17ac68-6be7-438e-88ea-4fd6faccb800.mov

amy-corson-ibigroup (approved):
Looks great! I noticed a few small additional things but no serious issues!

Only other thing I was wondering is it does feel weird to be able to input past dates/times, but maybe that's outside the scope of this PR?

lib/components/form/call-taker/date-time-picker.tsx
miles-grant-ibigroup: What's the expected behavior here? The trip can't be planned if the date is wrong. Maybe it should be grayed out let me try that

lib/components/form/call-taker/date-time-picker.tsx
miles-grant-ibigroup: This is a huge ux improvement thanks so much! Implemented

lib/components/form/call-taker/date-time-picker.tsx
amy-corson-ibigroup: I think it would be enough just for the date field to stay consistent, even if there's an error in the time field! But honestly I don't think it's too distracting as is!

demory:
Thanks, this looks good! 

One thing I'm noticing is that the time field "blanks out" after going to some other part of the app (e.g. the stop / trip viewer) and then back to the search form.

Should that be fixed as part of this PR or is that a separate issue?

miles-grant-ibigroup:
Could you possibly upload a screen recording of this happening? I can't replicate on any browser! Both fields are still populated when I return to the trip planner.

miles-grant-ibigroup:
Have just merged in `dev` and am still unable to replicate.

demory:
@miles-grant-ibigroup I'm finally getting back to this. The blanking issue I described earlier seems to have resolved, that must have been due to other merged work, etc. 

But, I am seeing that when I type in a time the date field occasionally changes to "mm/yy/dddd" rather that the active date. For example when typing "2:30pm" this happens as I type the ":" and then again after typing the "m" (and stays in that state). The application seems to still correctly parse this date, but the tooltip for the time field still says "invalid date":

image

Can you reproduce? I am running this branch (`ct-rework-time-date-selector`) on its own and with the latest `trimet-call-taker-otp2` config.

miles-grant-ibigroup:
Good catch thanks! Have added more supported time formats, which should fix this. Let me know if it's fixed for you! 

Failing tests will be fixed once https://github.com/opentripplanner/otp-react-redux/pull/739 is merged

demory (approved):

PR #720, opened 2022-12-06
Fix route viewer fitting pt2

binh-dam-ibigroup:
Requires/blocked by https://github.com/opentripplanner/otp-ui/pull/507.

This PR optimizes rendering in the route viewer by preventing the stops overlay from displaying all transit stops when selecting one route after the other in the route viewer.


lib/components/map/connected-stops-overlay.js
```diff
@@ -1,64 +1,62 @@
...
-          prev.push(cur)
-        }
-        return prev
-      }, [])
-      // Override the minimum zoom so that the stops appear even if zoomed out
-      minZoom = 2
-    }
+      return prev
+    }, [])
+    minZoom = 2
```
miles-grant-ibigroup, line 43: The comment about this got lost!

miles-grant-ibigroup (approved):
Will wait for the upcoming behavior change to fully test functionality but looks a lot cleaner for now! Some comments seem to have gotten lost but I suspect there will be large changes here soon anyway.

lib/components/map/connected-stops-overlay.js
binh-dam-ibigroup: Oops, I added the comment back in https://github.com/opentripplanner/otp-react-redux/pull/720/commits/6e6a9726464f4793b915a7c01c1bcd3c7c1d44cc.

daniel-heppner-ibigroup (approved):
This works well for me and there's not noticeable delay, so happy to see it!  Thank you!

PR #721, opened 2022-12-06
chore(deps): Update map, transitive, route, and trip overlays.

binh-dam-ibigroup:
This PR updates map-related packages so that transitive overlay better autofits bounds, especially in print view.


miles-grant-ibigroup (changes requested):
Tests seem to be failing based on these package updates. I've fixed this in https://github.com/opentripplanner/otp-react-redux/pull/722. Not sure if you'd rather close this PR in favor of #722 or want to replicate those fixes here.

binh-dam-ibigroup:
> Tests seem to be failing based on these package updates. I've fixed this in #722. Not sure if you'd rather close this PR in favor of #722 or want to replicate those fixes here.

Let's use #722 instead.

PR #728, opened 2022-12-14
Don't show stops when all route patterns are previewed.

binh-dam-ibigroup:
This PR changes the route viewer so that, when a route is previewed (when the user clicks the route number) and all patterns are shown on the map, no stops are shown. Stops for a pattern continue to be displayed when the user selects a pattern from the destination drop-down selector.

### DIFF
--- a/lib/components/map/connected-stops-overlay.js
+++ b/lib/components/map/connected-stops-overlay.js
@@ -21,25 +21,29 @@ const mapStateToProps = (state) => {
     patterns
   ) {
     // Display stops for a pattern or all patterns for a route.
-    let viewedPattern = patternId
 
-    // If a flex route is being shown but the pattern viewer is not active, then the
-    // stops of the first pattern of the route should be shown
-    // This will ensure that the flex zone stops are shown.
-
-    // Preferably, the flex stops would be rendered in a separate layer.
+    // Preferably, the flex zones would be rendered in a separate layer.
     // However, without changes to GraphQL, getting this data is very expensive
-    if (!viewedPattern && v2) {
-      viewedPattern = Object.keys(patterns)?.[0]
+    if (v2) {
+      // If a flex route is being shown, show flex zones from all patterns
+      // using the stops overlay, whether the pattern viewer is active or not.
+      Object.values(patterns).forEach((p) => {
+        stops = stops.concat(
+          p?.stops?.filter((s) => s.geometries?.geoJson?.type === 'Polygon')
+        )
+      })
     }
 
-    // Discard duplicates from pattern stops.
-    stops = (patterns?.[viewedPattern]?.stops || []).reduce((prev, cur) => {
-      if (!prev.find((stop) => stop.id === cur.id)) {
-        prev.push(cur)
-      }
-      return prev
-    }, [])
+    // Discard duplicates from pattern stops and flex zones.
+    stops = stops
+      .concat(patterns?.[patternId]?.stops || [])
+      .reduce((prev, cur) => {
+        if (!prev.find((stop) => stop.id === cur.id)) {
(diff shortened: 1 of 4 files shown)


amy-corson-ibigroup (approved):
Looks great!! Makes the Route Viewer look very clean

lib/components/map/connected-stops-overlay.js
```diff
@@ -21,25 +21,29 @@ const mapStateToProps = (state) => {
...
-    // Discard duplicates from pattern stops.
-    stops = (patterns?.[viewedPattern]?.stops || []).reduce((prev, cur) => {
-      if (!prev.find((stop) => stop.id === cur.id)) {
-        prev.push(cur)
-      }
-      return prev
-    }, [])
+    // Discard duplicates from pattern stops and flex zones.
+    stops = stops
+      .concat(patterns?.[patternId]?.stops || [])
```
miles-grant-ibigroup, line 39: What does this `concat` line do?

miles-grant-ibigroup (changes requested):
Getting some annoying flashing when I switch from the pattern viewer back to the route viewer. It was happening before but is now more noticeable because we slowed down the render method. Curious if you have any immediate ideas. Otherwise I will investigate further. If I can't find a simple fix I'll approve and we can push this into a new issue.

lib/components/map/connected-stops-overlay.js
binh-dam-ibigroup: It was supposed to include flex stops in the duplicate stop removal, but good point, the two things are independent so concatenating was not necessary. The new logic for removing duplicate stops is in https://github.com/opentripplanner/otp-react-redux/pull/728/commits/f3b720e8128b7dcbd2a3716ab26e29153f4597da.

binh-dam-ibigroup:
> Getting some annoying flashing when I switch from the pattern viewer back to the route viewer. It was happening before but is now more noticeable because we slowed down the render method.

It might be a side effect of #720. Let me take a look.

binh-dam-ibigroup:
> > Getting some annoying flashing when I switch from the pattern viewer back to the route viewer. It was happening before but is now more noticeable because we slowed down the render method.
> 
> It might be a side effect of #720. Let me take a look.

Okay this one was a little crazy. The URL scheme for the Pattern Viewer within the Route Viewer was not listed in the list of routes, so switching to/from the pattern viewer triggered a re-render of the ENTIRE webapp, hence the flickering you saw. See https://github.com/opentripplanner/otp-react-redux/pull/728/commits/a7b3009eb8c485e080378c1ba60cb9403501823f for the fix.

miles-grant-ibigroup (approved):
This works so well now thanks for figuring out that URL thing! Only regression I noticed is that after clicking a stop from the pattern viewer the back button now skips the pattern viewer and goes all the way back to the route viewer. However, this is minor enough I'm willing to open a new issue and mark this PR as done!

binh-dam-ibigroup:
> Only regression I noticed is that after clicking a stop from the pattern viewer the back button now skips the pattern viewer and goes all the way back to the route viewer. However, this is minor enough I'm willing to open a new issue and mark this PR as done!

Is this the Back button of the stop viewer? Yeah we'll need a discussion of what "Back" means, compared to using the browser's Back button.

miles-grant-ibigroup:
> > Only regression I noticed is that after clicking a stop from the pattern viewer the back button now skips the pattern viewer and goes all the way back to the route viewer. However, this is minor enough I'm willing to open a new issue and mark this PR as done!
> 
> Is this the Back button of the stop viewer? Yeah we'll need a discussion of what "Back" means, compared to using the browser's Back button.

Yeah the stop viewer back button although I think the browser back has a similar issue

PR #729, opened 2022-12-15
Make App Menu and Popup Link Icons Configurable

binh-dam-ibigroup:
This PR brings back configurability of icons for the app menu and popup links.

### DIFF
--- a/example-config.yml
+++ b/example-config.yml
@@ -460,6 +460,7 @@ itinerary:
 #           0.9: 'Mostly Accessible'
 #       popups:
 #         testTarget: Test Target!
+#         testTarget-narrow: Target! # used for narrow UI elements such as link buttons under trip details.
 #       menuItems:
 #         demo-item: Demo Item
 #     errorMessages:
--- a/example-config.yml
+++ b/example-config.yml
@@ -460,7 +460,8 @@ itinerary:
 #           0.9: 'Mostly Accessible'
 #       popups:
 #         testTarget: Test Target!
-#         testTarget-narrow: Target! # used for narrow UI elements such as link buttons under trip details.
+#         # Optional string used for narrow UI elements such as link buttons under trip details.
+#         testTarget-narrow: Target!
 #       menuItems:
 #         demo-item: Demo Item
 #     errorMessages:
--- a/lib/components/app/app-menu.tsx
+++ b/lib/components/app/app-menu.tsx
@@ -11,7 +11,6 @@ import { MenuItem } from 'react-bootstrap'
 import { Undo } from '@styled-icons/fa-solid/Undo'
 import { withRouter } from 'react-router'
 import AnimateHeight from 'react-animate-height'
-import qs from 'qs'
 import React, { Component, Fragment, useContext } from 'react'
 import SlidingPane from 'react-sliding-pane'
 import type { RouteComponentProps } from 'react-router'
@@ -61,6 +60,8 @@ class AppMenu extends Component<
   AppMenuProps & WrappedComponentProps & RouteComponentProps,
   AppMenuState
 > {
+  static contextType = ComponentContext
+
   state = {
     expandedSubmenus: {} as Record<string, boolean>,
     isPaneOpen: false
@@ -174,6 +175,7 @@ class AppMenu extends Component<
     } = this.props
 
     const { isPaneOpen } = this.state
+    const { SvgIcon } = this.context
     return (
       <>
         <div
@@ -219,7 +221,7 @@ class AppMenu extends Component<
             {popupTarget && (
               <MenuItem className="menu-item" onClick={this._triggerPopup}>
                 <StyledIconWrapper>
-                  <ExternalLinkSquareAlt />
+                  <SvgIcon iconName={popupTarget} />
                 </StyledIconWrapper>
                 <FormattedMessage id={`config.popups.${popupTarget}`} />
               </MenuItem>
--- a/lib/components/app/app-menu.tsx
+++ b/lib/components/app/app-menu.tsx
@@ -319,14 +319,10 @@ const IconAndLabel = ({
           )}
           src={iconUrl}
         />
+      ) : iconType ? (
+        <SvgIcon iconName={iconType} />
       ) : (
-        <StyledIconWrapper>
-          {iconType ? (
-            <SvgIcon iconName={iconType} />
-          ) : (
-            <ExternalLinkSquareAlt />
-          )}
-        </StyledIconWrapper>
+        <ExternalLinkSquareAlt />
       )}
       {label}
     </span>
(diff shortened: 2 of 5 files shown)


lib/components/narrative/trip-tools.js
```diff
@@ -169,7 +171,20 @@ LinkButton.propTypes = {
   url: PropTypes.string
 }
 
+const makeIcon = memoize((SvgIcon, iconType) => {
```
daniel-heppner-ibigroup, line 174: By default, lodash's memoize only caches based on the first argument provided to the function. Might that be a problem here? 
> By default, the first argument provided to the memoized function is used as the map cache key.

https://lodash.com/docs/#memoize

daniel-heppner-ibigroup (approved):
Concern about the lodash memo function, but it doesn't seem to be actively causing issues, so I approve.

lib/components/narrative/trip-tools.js
binh-dam-ibigroup: Good catch! Also tricky. If I pass an object such as `{ icon: SvgIcon, type: popupTarget }`, the value that determines the hash changes at every render and makes memoizing moot. Interestingly, if only SvgIcon gets memoized, the console warning regarding creating a dynamic component still appears only once, which is the desired outcome.

lib/components/narrative/trip-tools.js
```diff
@@ -169,7 +171,20 @@ LinkButton.propTypes = {
...
+  const IconComponent = (props) => {
+    return (
+      <StyledIconBase>
+        <SvgIcon iconName={iconType} {...props} />
```
miles-grant-ibigroup, line 178: Is this the approach we want to take? This would require shipping every icon we want in the `config.js`. Could we not use dynamic imports instead to import directly from fa?

miles-grant-ibigroup (changes requested):
Works well but I'm not sure if this is the best approach? Pulling the icons from `config.js` is a nice feature but feels a bit cumbersome maybe?

lib/components/narrative/trip-tools.js
binh-dam-ibigroup: Shipping every icon in config makes sense, as the label for the popup button is defined in config.

miles-grant-ibigroup: But if you wanted to use a `fa` import wouldn't you have to define that/import it during the build step in the config instead of just importing it dynamically?

binh-dam-ibigroup: A dynamic import assumes we have a predetermined set of known icons like the ones from `SvgIcon`. Do we want to do that?

binh-dam-ibigroup: Also, defining an icon for the popup link/button remains optional, the default is still the "external link" icon.

miles-grant-ibigroup: I just don't like having to define all the icons as svg in the config and the pulling them out. Would prefer something less cumbersome

binh-dam-ibigroup: Turns out that, from the config file, you can import any icon from any applicable package referenced by `package.json`, so extra SVG files are not required in these cases.

binh-dam-ibigroup: And no dynamic imports are required that way.

lib/components/app/app-menu.tsx
```diff
@@ -319,14 +321,10 @@ const IconAndLabel = ({
           )}
           src={iconUrl}
         />
+      ) : iconType ? (
+        <SvgIcon iconName={iconType} />
```
miles-grant-ibigroup, line 325: This seems to have been lost already in the font awesome pr, but there used to be support for an icon being a URL that pointed to an image. Do you think we could restore that functionality? (By checking if the `iconType` is a URL and if so rendering an `img`

lib/components/narrative/trip-tools.js
```diff
@@ -169,91 +170,102 @@ LinkButton.propTypes = {
...
+    switch (type) {
+      case 'COPY_URL':
+        buttonComponents.push(<CopyUrlButton />)
+        break
+      case 'PRINT':
+        buttonComponents.push(<PrintButton />)
+        break
+      case 'REPORT_ISSUE':
+        if (!reportConfig || !reportConfig.mailto) break
+        buttonComponents.push(<ReportIssueButton {...reportConfig} />)
```
miles-grant-ibigroup, line 205: Could the extra break be removed if we flipped the if condition?

lib/components/narrative/trip-tools.js
```diff
@@ -169,91 +170,102 @@ LinkButton.propTypes = {
...
-              text={<FormattedMessage id="common.forms.startOver" />}
-              url={startOverUrl}
+              Icon={PopupIcon}
+              onClick={() => setPopupContent(popupTarget)}
+              text={
+                <FormattedMessage
+                  defaultMessage={intl.formatMessage({
+                    id: `config.popups.${popupTarget}`
+                  })}
+                  id={`config.popups.${popupTarget}-narrow`}
```
miles-grant-ibigroup, line 232: Clever way to make this optional...

example-config.yml
```diff
@@ -460,6 +460,7 @@ itinerary:
 #           0.9: 'Mostly Accessible'
 #       popups:
 #         testTarget: Test Target!
+#         testTarget-narrow: Target! # used for narrow UI elements such as link buttons under trip details.
```
miles-grant-ibigroup, line 463: We should probably specify that this is optional!

miles-grant-ibigroup (approved):
Looking much better! Just a few small nitpicky fixes

lib/components/app/app-menu.tsx
miles-grant-ibigroup: Sorry it is the `iconUrl` property we are ignoring!

binh-dam-ibigroup: From the changes in #651, I am not sure that a check for whether an image loaded successfully was there before.

miles-grant-ibigroup: Yeah I think my memory is playing tricks on me. I believe we just rendered the image if `iconUrl` was present, otherwise the svg tag

binh-dam-ibigroup: If your question was regarding the fallback if `iconType` is unknown, then yes, `SvgIcon` falls back to the external link icon.

miles-grant-ibigroup: No we used to support `iconUrl`, and `iconType` was the fallback

lib/components/narrative/trip-tools.js
binh-dam-ibigroup: Fixed in https://github.com/opentripplanner/otp-react-redux/pull/729/commits/866fdc2bfbe16ae5c0c97c0868f3bd3a6f37e541.

example-config.yml
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/729/commits/866fdc2bfbe16ae5c0c97c0868f3bd3a6f37e541.

PR #730, opened 2022-12-20
Update Percy Tests for Dec. 2022 GTFS Changes

miles-grant-ibigroup:
Because we make use of a real OTP server to make the Percy tests more realistic, we run into the problem of having to change the tests as MARTA schedules change.


binh-dam-ibigroup (approved):
Thanks for figuring this out!

daniel-heppner-ibigroup (approved):
perfect, thank you for the update!

PR #731, opened 2022-12-20
co2 Tweaks

miles-grant-ibigroup:
This PR does a tiny bit of cleanup on the co2 display to add final polish.

* removes period at the end of the statement
* the number is blurred until all results are no longer pending -- this ensures that the number "jump" when a driving leg arrives is less jarring.

### DIFF
--- a/lib/components/narrative/metro/metro-itinerary.tsx
+++ b/lib/components/narrative/metro/metro-itinerary.tsx
@@ -7,8 +7,9 @@ import {
   IntlShape
 } from 'react-intl'
 import { Itinerary, Leg } from '@opentripplanner/types'
+import { Leaf } from '@styled-icons/fa-solid/Leaf'
 import React from 'react'
-import styled from 'styled-components'
+import styled, { keyframes } from 'styled-components'
 
 import * as narrativeActions from '../../../actions/narrative'
 import * as uiActions from '../../../actions/ui'
@@ -20,7 +21,6 @@ import {
 } from '../../../util/accessibility-routing'
 import { getActiveSearch, getFare } from '../../../util/state'
 import { ItineraryDescription } from '../default/itinerary-description'
-import { Leaf } from '@styled-icons/fa-solid/Leaf'
 import { localizeGradationMap } from '../utils'
 import FormattedDuration, {
   formatDuration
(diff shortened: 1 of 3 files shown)


lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -197,6 +197,19 @@ const ItineraryGridSmall = styled.div`
...
+const blurAnimation = keyframes`
+ 0% { filter: blur(${BLUR_AMOUNT + 1}px); }
+ 50% { filter: blur(${BLUR_AMOUNT}px) }
+`
+
+const LoadingBlurred = styled.span<{ loading: boolean }>`
+  ${(props) => props.loading && `filter: blur(${BLUR_AMOUNT}px)`};
+  animation-name: ${(props) => (props.loading ? blurAnimation : '')};
+  animation-duration: 1s;
+  animation-iteration-count: infinite;
```
daniel-heppner-ibigroup, line 210: Is it possible to add a transition to the filter so that the blur sort of fades out instead of being instant? Otherwise I think it looks good. I don't really notice the animation that much, but that's fine.

daniel-heppner-ibigroup (approved):
This looks good to me, thanks for the change. I have a suggestion about animating the fade out of the blur, if possible. 

Unrelatedly, I noticed it can be kind of hard to see which itinerary the CO2 note corresponds to. I wonder if we could add a top border to it to make it easier to differentiate. Maybe could be added to this PR. 
image

lib/components/narrative/metro/metro-itinerary.tsx
miles-grant-ibigroup: Good catch! Looks much better thanks!

lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -479,6 +496,9 @@ const mapStateToProps = (state: any, ownProps: Props) => {
     // @ts-expect-error state is not yet typed
     activeSearch && activeSearch.activeItineraryTimeIndex
 
+  // @ts-expect-error TODO: type activeSearch
+  const pending = activeSearch ? Boolean(activeSearch.pending) : false
```
binh-dam-ibigroup, line 500: This variable could be inlined.

binh-dam-ibigroup (approved):
Does the job.

lib/components/narrative/metro/metro-itinerary.tsx
miles-grant-ibigroup: Good call!

PR #733, opened 2022-12-22
fix(forms): Adjust icon layout and remove .fa references.

binh-dam-ibigroup:
This PR adjust icon positioning within their buttons in the trip form, and removes references to unused fontawesome element classes.

### DIFF
--- a/index.css
+++ b/index.css
@@ -119,12 +119,6 @@ input[type="text"]::-ms-clear {
   }
 }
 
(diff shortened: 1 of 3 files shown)


daniel-heppner-ibigroup (approved):
LGTM

miles-grant-ibigroup (approved):
Looks like this was all safe to remove! Thanks for the cleanup

PR #734, opened 2022-12-23
refactor: prioritize config transit operators over gtfs agencies

amy-corson-ibigroup:
This PR prioritizes using the Transit Operator names in the client's config file and will fall back on using the GTFS agency if no Transit operators are specified in the config file.

OTP-UI: https://github.com/ibi-group/otp-ui/tree/prioritize-config-operators
Config: https://github.com/ibi-group/configurations/tree/atl-temp-flex-feeds

### DIFF
--- a/lib/components/viewers/route-details.js
+++ b/lib/components/viewers/route-details.js
@@ -146,7 +146,7 @@ class RouteDetails extends Component {
               <FormattedMessage
                 id="components.RouteDetails.operatedBy"
                 values={{
-                  agencyName: agency.name
+                  agencyName: operator.name ? operator.name : agency.name
                 }}
               />
             )}
--- a/lib/util/state.js
+++ b/lib/util/state.js
@@ -581,17 +581,25 @@ function getItineraryToRender(state) {
 const routeSelector = (state) => Object.values(state.otp.transitIndex.routes)
 const routeViewerFilterSelector = (state) => state.otp.ui.routeViewer.filter
 
+const getAgencyIdForFilter = (name, transitOperators) => {
+  const operator = transitOperators.find((operator) => operator.name === name)
+  return operator.agencyId
+}
+
 /**
  * Returns all routes that match the route viewer filters
  */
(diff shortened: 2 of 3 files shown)


demory (approved):

lib/components/viewers/route-details.js
```diff
@@ -147,7 +147,7 @@ class RouteDetails extends Component {
                 <FormattedMessage
                   id="components.RouteDetails.operatedBy"
                   values={{
-                    agencyName: agency.name
+                    agencyName: operator.name ? operator.name : agency.name
```
miles-grant-ibigroup, line 150:
```suggestion
                    agencyName: operator?.name || agency.name
```

Does this work?

lib/util/state.js
```diff
@@ -581,17 +581,28 @@ function getItineraryToRender(state) {
 const routeSelector = (state) => Object.values(state.otp.transitIndex.routes)
 const routeViewerFilterSelector = (state) => state.otp.ui.routeViewer.filter
 
+const getAgencyIdForFilter = (name, transitOperators) => {
+  const operators = transitOperators.filter(
+    (operator) => operator.name === name
+  )
+  return operators.map((x) => x.agencyId)
```
miles-grant-ibigroup, line 588: Probably cleaner to do this all as one call! Then we don't need to assign the variable

miles-grant-ibigroup: Also the name doesn't make it clear that this will return an array!

miles-grant-ibigroup (changes requested):
Very clean way of fixing this! Just a few style nits. Also might want to block this until https://github.com/opentripplanner/otp-ui/pull/519 is merged and in this PR.

lib/util/state.js
amy-corson-ibigroup: Good call! Let me know if changing it to `getAgencyIdsForFilter` works okay or if it could be more descriptive!

PR #738, opened 2022-12-28
Update map tiles URL in example config

otrichet:
Currently, doing `yarn start` with the example configuration leads to a blank map, because the configuration is still referencing raster tile server URL.

### DIFF
--- a/example-config.yml
+++ b/example-config.yml
@@ -93,11 +93,10 @@ map:
   initLon: -122.682
   baseLayers:
     - name: Streets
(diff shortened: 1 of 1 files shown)


miles-grant-ibigroup (approved):
Very good catch! Thanks for the change!

philip-cline (approved):

miles-grant-ibigroup:
Tests failing is fine, as it doesn't have access to our secrets!

PR #739, opened 2022-12-29
Various Small Layout Adjustments

miles-grant-ibigroup:
This PR brings many small layout adjustments and minor accessibility improvements. 

Changes include:
- correction of various aria attributes 
- avoiding text clashing into navbar elements by moving navbar elements to app menu
- removes back buttons which replicate browser back button behavior while increasing UI noise
- adds typescript to various files
- repair various small behavioral bugs, avoiding blank screens and blank UI errors


lib/components/mobile/mobile.css
```diff
@@ -32,6 +32,13 @@
   display: inline-block;
 }
 
+@media (max-width: 800px) {
+  .otp .navbar .mobile-header-text {
+    font-size: 3vw;
```
miles-grant-ibigroup, line 37: Really not sure about this one...  It seems to work but doesn't feel very stable.

lib/components/narrative/default/itinerary-summary.js
```diff
@@ -47,6 +47,7 @@ const Block = connect(mapStateToProps)(styled.div`
 
   span {
     display: ${(props) => (props.renderRouteNamesAndColors ? 'block' : 'none')};
+    margin-top: -3px;
```
miles-grant-ibigroup, line 50: This fixes things, but I don't know what it's accounting for... feels risky

lib/components/mobile/mobile.css
miles-grant-ibigroup: Will experiment with some javascript based solutions...

lib/components/app/app.css
```diff
@@ -32,11 +32,13 @@
 }
 
 /* Don't show route viewer link in the app menu on desktop as it is in the navbar */
```
binh-dam-ibigroup, line 34: Update this comment for trip planning.

lib/components/app/app.css
```diff
@@ -32,11 +32,13 @@
...
-.app-menu-route-viewer-link {
+.app-menu-route-viewer-link,
+.app-menu-trip-planner-link {
   display: none;
 }
 @media (max-width: 800px) {
-  .app-menu-route-viewer-link {
+  .app-menu-route-viewer-link,
+  .app-menu-trip-planner-link {
     display: block;
```
binh-dam-ibigroup, line 42: Should probably be `display: inherit;`

lib/components/app/app-menu.tsx
```diff
@@ -200,7 +259,23 @@ class AppMenu extends Component<
...
-          <div className="app-menu">
+          <nav
+            aria-label={intl.formatMessage({
+              id: 'components.AppMenu.appMenu'
+            })}
+            className="app-menu"
+          >
+            {/* This item is duplicated by the view-switcher, but only shown on mobile
+            when the view switcher isn't shown (using css) */}
+            <MenuItem
```
binh-dam-ibigroup, line 270: `<MenuItem>` resolves to `<li>`, so surround the list items with `<ul>`.

lib/components/app/app-menu.tsx
```diff
@@ -269,25 +347,31 @@ class AppMenu extends Component<
...
   resetAndToggleCallHistory: callTakerActions.resetAndToggleCallHistory,
   resetAndToggleFieldTrips: fieldTripActions.resetAndToggleFieldTrips,
+  setLocale: uiActions.setLocale,
   setMainPanelContent,
   setPopupContent: uiActions.setPopupContent,
   toggleMailables: callTakerActions.toggleMailables
 }
 
 export default injectIntl(
+  // @ts-expect-error TODO: type setMainPanelContent
```
binh-dam-ibigroup, line 374: How wasn't this an error before?

lib/components/app/app-menu.tsx
```diff
@@ -163,17 +184,55 @@ class AppMenu extends Component<
...
+                loggedInUser,
+                createOrUpdateUser,
+                setLocale,
+                intl
+              ),
+            subMenuDivider: false
+          })),
+        iconType: <GlobeAmericas />,
+        id: 'app-menu-locale-selector',
+        label: <FormattedMessage id="components.SubNav.languageSelector" />,
```
binh-dam-ibigroup, line 231: Tricky a11y: To avoid setting up radio button elements or roles in the submenu items, maybe add a mention about the current language here so the label reads, with a screen reader: "Select language - current language: English". ("Current language..." can be visually hidden)

lib/components/app/app-menu.tsx
```diff
@@ -163,17 +184,55 @@ class AppMenu extends Component<
...
+    const languageMenuItems: menuItem[] | undefined = configLanguages && [
+      {
+        children: Object.keys(configLanguages)
+          .filter((locale) => locale !== 'allLanguages')
+          .map((locale) => ({
+            iconType: <></>,
+            id: configLanguages[locale].name,
+            label:
+              activeLocale === locale ? (
+                <strong>{configLanguages[locale].name}</strong>
```
binh-dam-ibigroup, line 212: An alternative to saying the current language in the heading would be to add invisible text saying that that is the selected language.

lib/components/app/app-menu.tsx
```diff
@@ -317,12 +401,16 @@ const IconAndLabel = ({
             {
               id: 'components.AppMenu.menuItemIconAlt'
             },
-            { label }
+            { label: typeof label === 'string' ? label : '' }
```
binh-dam-ibigroup, line 404: a11y: The icons are purely decorative here, so `alt` text can be set to an empty string (see https://www.w3.org/WAI/tutorials/images/decorative/). Delete the corresponding entry from the language YML files.

lib/util/locale.ts
```diff
@@ -0,0 +1,28 @@
...
+import { MouseEvent } from 'react'
+import { User } from '@auth0/auth0-react'
+
+export const handleLocaleSelection = (
+  e: MouseEvent<Element>,
+  locale: string,
+  currentLocale: string,
+  loggedInUser: User,
+  createOrUpdateUser: (user: User, silent: boolean, intl: IntlShape) => void,
+  setLocale: (locale: string) => void,
```
binh-dam-ibigroup, line 11: Portions of the code for changing the locale should be a redux action (`actions/ui`), so it can take advantage of the redux state so that we don't have to pass all of these arguments around.

lib/components/app/app-menu.tsx
```diff
@@ -269,25 +347,31 @@ class AppMenu extends Component<
...
 const mapStateToProps = (state: Record<string, any>) => {
-  const { extraMenuItems } = state.otp.config
+  const { extraMenuItems, language } = state.otp.config
   return {
+    activeLocale: state.otp.ui.locale,
     callTakerEnabled: isModuleEnabled(state, Modules.CALL_TAKER),
+    configLanguages: language,
     extraMenuItems,
     fieldTripEnabled: isModuleEnabled(state, Modules.FIELD_TRIP),
+    loggedInUser: state.user.loggedInUser,
```
binh-dam-ibigroup, line 357: See comment about making changing the language an action.

binh-dam-ibigroup (changes requested):
First round of comments. Lots going on in this PR, I think the route viewer improvements that are not a11y, and itinerary changes could have been in a separate PR, but that's not blocking compared to some or the more basic changes that I have noted.

lib/components/app/app-menu.tsx
```diff
@@ -256,8 +331,11 @@ class AppMenu extends Component<
                 <FormattedMessage id="components.AppMenu.mailables" />
               </MenuItem>
             )}
-            {this._addExtraMenuItems(extraMenuItems)}
-          </div>
+            {this._addExtraMenuItems([
+              ...(extraMenuItems || []),
+              ...(languageMenuItems || [])
```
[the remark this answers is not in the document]
miles-grant-ibigroup: I see the gap you're referring to, but I'm not seeing the `overflow: hidden` you're asking me to remove?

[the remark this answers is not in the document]
miles-grant-ibigroup: Looks like it's added by the library... oh boy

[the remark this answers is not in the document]
miles-grant-ibigroup: Have added a css rule that includes `!important` which doesn't feel great but I can't think of another way to override the inline style

lib/components/app/app.css
miles-grant-ibigroup: Sure!

lib/components/app/app-menu.tsx
miles-grant-ibigroup: Surprised we didn't ready have this!

lib/components/app/app-menu.tsx
miles-grant-ibigroup: Good question

lib/util/locale.ts
miles-grant-ibigroup: Good call! Turns out most of the wrapper stuff wasn't even needed! Things are much cleaner now get excited

lib/components/app/app-menu.tsx
```diff
@@ -163,17 +184,55 @@ class AppMenu extends Component<
...
+      loggedInUser,
       mailablesEnabled,
       popupTarget,
       resetAndToggleCallHistory,
       resetAndToggleFieldTrips,
+      setLocale,
       toggleMailables
     } = this.props
 
+    const languageMenuItems: menuItem[] | undefined = configLanguages && [
```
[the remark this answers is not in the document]
binh-dam-ibigroup: What is not possible?

lib/components/app/app-menu.tsx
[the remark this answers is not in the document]
binh-dam-ibigroup: Add a comment in the CSS regarding that.

lib/components/viewers/view-trip-button.tsx
```diff
@@ -0,0 +1,64 @@
...
+  }): void
+  text: JSX.Element
+  toIndex: number
+  tripId: string
+}
+
+class ViewTripButton extends Component<Props> {
+  _onClick = () => {
+    this.props.setMainPanelContent(MainPanelContent.TRIP_VIEWER)
+    this.props.setViewedTrip({
```
binh-dam-ibigroup, line 32: Can `setViewedTrip` also set the main panel content?

lib/components/viewers/view-trip-button.tsx
```diff
@@ -0,0 +1,64 @@
...
+          <FormattedMessage id="components.TripViewer.header" />
+        )}
+      </Button>
+    )
+  }
+}
+
+const mapStateToProps = () => {
+  return {}
+}
```
binh-dam-ibigroup, line 56: Delete this var. The first argument of `connect` can be `null`.

binh-dam-ibigroup (changes requested):
See my other comments. Itinerary narrative summaries should be lists, each containing a button with the itinerary summary.

lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -350,7 +350,6 @@ class MetroItinerary extends NarrativeItinerary {
         onMouseLeave={this._onMouseLeave}
         role="presentation"
       >
-        {/* Semantically this is incorrect, but this helps a11y tests pass. Buttons may not contain html. TODO FIX? */}
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Looks like that will be done in #744, so you can discard comments regarding itinerary result lists.

lib/components/mobile/mobile.css
miles-grant-ibigroup: Sometimes the header doesn't fit in the available space.

miles-grant-ibigroup: This change ensures it always fits, but the text is sometimes a bit small

lib/actions/user.js
```diff
@@ -323,7 +323,7 @@ export function createOrUpdateUser(userData, silentOnSuccess = false, intl) {
 
     // TODO: improve the UI feedback messages for this.
     if (status === 'success' && returnedUser) {
-      if (!silentOnSuccess) {
+      if (!silentOnSuccess && intl) {
```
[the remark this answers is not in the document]
miles-grant-ibigroup: When we call this method from within redux we don't have access to an `intl` object and it's therefore null. Since these user updates are never based on user input anyway this is no big deal (`silentOnSuccess` is true)

lib/components/app/view-switcher.tsx
```diff
@@ -0,0 +1,118 @@
...
+ */
+class ViewSwitcher extends Component<Props> {
+  static propTypes = {
+    activePanel: PropTypes.number,
+    setMainPanelContent: PropTypes.func,
+    sticky: PropTypes.bool
+  }
+
+  _showRouteViewer = () => {
+    this.props.setMainPanelContent(null)
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Some redux actions are funky, and going from pattern viewer to route viewer sometimes doesn't trigger the url path to update

lib/components/app/view-switcher.tsx
```diff
@@ -0,0 +1,118 @@
...
+    const routeViewerActive = activePanel === MainPanelContent.ROUTE_VIEWER
+
+    return (
+      <div
+        aria-label={intl.formatMessage({
+          id: 'components.ViewSwitcher.switcher'
+        })}
+        className="view-switcher"
+        id="view-switcher"
+        role="tablist"
```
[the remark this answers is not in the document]
miles-grant-ibigroup: What else do you suggest? I think the tabs are better than buttons

lib/components/mobile/mobile.css
```diff
@@ -32,6 +32,13 @@
   display: inline-block;
 }
 
+@media (max-width: 800px) {
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Yes. Wide phones can render the regular text size fine.

lib/components/mobile/options-screen.tsx
```diff
@@ -1,33 +1,35 @@
...
   _planTripClicked = () => {
     this.props.setMobileScreen(MobileScreens.RESULTS_SUMMARY)
   }
 
-  render () {
+  render() {
     return (
       <MobileContainer>
         <MobileNavigationBar
           backScreen={MobileScreens.SEARCH_FORM}
```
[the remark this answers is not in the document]
miles-grant-ibigroup: See previous comments about potentially restoring back button behavior in the future

lib/components/narrative/line-itin/connected-itinerary-body.js
```diff
@@ -135,7 +140,10 @@ class ConnectedItineraryBody extends Component {
           RouteDescription={RouteDescriptionOverride || RouteDescription}
           setActiveLeg={setActiveLeg}
           setLegDiagram={setLegDiagram}
-          setViewedTrip={setViewedTrip}
+          setViewedTrip={(tripId) => {
+            setViewedTrip(tripId)
+            setMainPanelContent(MainPanelContent.TRIP_VIEWER)
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Are we sure that every call to `setViewedTrip` intends to change the view? I didn't want to introduce new edge cases

lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -360,7 +359,9 @@ class MetroItinerary extends NarrativeItinerary {
           onKeyDown={() => {}}
           // TODO: use _onHeaderClick for tap only -- this will require disabling
           // this onClick handler after a touchStart
-          role="presentation"
+          role="menuitem"
+          // TODO test this with a screen reader
+          tabIndex={expanded ? 1 : 0}
```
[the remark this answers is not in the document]
miles-grant-ibigroup: `expanded` is true when there is only one itinerary being shown. Will add a comment

lib/components/user/nav-login-button-auth0.tsx
```diff
@@ -11,10 +11,10 @@ type AccountLink = {
 }
 
 interface NavLoginButtonAuth0Props {
-  className?: string
+  className?: string | undefined
   id: string
   links: Array<AccountLink>
-  style?: string
+  style?: CSSProperties | undefined
 }
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Beautiful thank yoU!

lib/components/user/nav-login-button.css
```diff
@@ -36,3 +41,16 @@
   color: #ddd;
   text-decoration: none;
 }
+
+#app-menu-locale-selector {
+  display: none;
+}
+
+@media (max-width: 800px) {
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Updated

lib/components/viewers/route-viewer.js
```diff
@@ -163,16 +163,6 @@ class RouteViewer extends Component {
             className="route-viewer-header"
             style={{ backgroundColor: routeColor }}
           >
-            {/* Always show back button, as we don't write a route anymore */}
-            <div className="back-button-container">
-              <Button bsSize="small" onClick={this._backClicked}>
```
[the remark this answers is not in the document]
miles-grant-ibigroup: See previous comments about potentially bringing this back

lib/components/viewers/view-trip-button.tsx
miles-grant-ibigroup: See previous comments about avoiding regressions. I've mentioned elsewhere that I think our entire routing system needs a major overhaul.

lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -360,7 +359,9 @@ class MetroItinerary extends NarrativeItinerary {
           onKeyDown={() => {}}
           // TODO: use _onHeaderClick for tap only -- this will require disabling
           // this onClick handler after a touchStart
-          role="presentation"
+          role="menuitem"
```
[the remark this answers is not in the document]
miles-grant-ibigroup: The problem is to do this the `div` would need to become a `ul` which is not allowed to have event listeners.

lib/components/narrative/narrative-itineraries.js
```diff
@@ -389,6 +389,8 @@ class NarrativeItineraries extends Component {
         <div
           // FIXME: Change to a ul with li children?
           className="list"
+          id="itinerary-menu"
+          role="menu"
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Listitems can't be interactive: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-noninteractive-element-interactions.md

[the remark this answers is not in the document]
binh-dam-ibigroup: In our case, each list item would contain a `<button>` for showing itinerary details.

lib/components/narrative/metro/metro-itinerary.tsx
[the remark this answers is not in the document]
binh-dam-ibigroup: This `<div>` would have to become an `<li>` and include a `<button>` that handles the mouse and keyboard events.

lib/components/narrative/metro/metro-itinerary.tsx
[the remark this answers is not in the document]
binh-dam-ibigroup: After speaking with @amy-corson-ibigroup, #744 does not include semantic changes for this element, so this PR should include them IMO.

lib/components/app/view-switcher.tsx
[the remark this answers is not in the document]
binh-dam-ibigroup: The other thing with tabs is that we would have to implement keyboard actions (arrows) too.

lib/components/mobile/mobile.css
binh-dam-ibigroup: Yes it is, and it kinda defeats the requirement for making the text legible. I think the bigger question is how do we handle text that is too long for the space available, regardless of the width. This is important for i18n. One workaround would be to show the title over two lines instead of one.

lib/components/narrative/metro/metro-itinerary.tsx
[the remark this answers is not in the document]
amy-corson-ibigroup: Actually, https://github.com/opentripplanner/otp-react-redux/pull/744 does change itinerary results into lists and list items. Should that be consolidated into this PR?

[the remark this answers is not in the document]
binh-dam-ibigroup: Oh cool, then @miles-grant-ibigroup leave these changes out of this PR.

binh-dam-ibigroup:
Things to be addressed later (undo changes in this PR):
- The font stuff in the mobile title bar
- The view switcher role

demory:
@miles-grant-ibigroup testing this this with the new st-harmonize config branch.

For the list of optional menu items, I'm noticing the list rendering in screenreader view as a combination of ordered and unordered items?
image

Is this because the `<li>` for the language selector is missing a `presentation` role tag?
image

demory:
Also, the language selector options look OK when it it in the navbar, but there is an issue when it moves to the app menu:
image

image

miles-grant-ibigroup:
> @miles-grant-ibigroup testing this this with the new st-harmonize config branch.
> 
> For the list of optional menu items, I'm noticing the list rendering in screenreader view as a combination of ordered and unordered items? image
> 
> Is this because the `<li>` for the language selector is missing a `presentation` role tag? image

Sadly we have no control over this... These roles are added by bootstrap! Not sure how we want to handle this. Could be a sign we're not using it correctly?

miles-grant-ibigroup:
> Also, the language selector options look OK when it it in the navbar, but there is an issue when it moves to the app menu: image
> 
> image

Good catch thanks! Resolved

lib/components/app/app-menu.tsx
```diff
@@ -204,13 +203,11 @@ class AppMenu extends Component<
...
                 <>
-                  <strong
-                    aria-label={intl.formatMessage({
-                      id: 'components.SubNav.activeLanguage'
-                    })}
-                  >
-                    {configLanguages[locale].name}
-                  </strong>
+                  {/* // Text is screen-reader only */}
+                  <span style={{ display: 'none' }}>
```
binh-dam-ibigroup, line 207: Revert this change in favor of #751.

lib/components/app/view-switcher.tsx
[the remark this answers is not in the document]
binh-dam-ibigroup: This still needs to be resolved. I recommend just using `<div role="group">`. I don't think tabs apply in our situation because not one tab will be selected at all times. If you still want to use `tablist`, then implement the keyboard interactions per https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tablist_role#keyboard_interactions.

lib/components/mobile/mobile.css
binh-dam-ibigroup: @miles-grant-ibigroup Have you tried fitting the header text into two lines as we discussed last time?

binh-dam-ibigroup (changes requested):
Some items I flagged for follow-up that we agreed should be addressed still need to be addressed.

lib/components/app/app-menu.tsx
miles-grant-ibigroup: There is no room for the language selector on mobile. It should stay in the app menu

lib/components/mobile/mobile.css
miles-grant-ibigroup: Do you remember what that fix was? I have a fix now using flex box but I'm not sure if that's the fix we found together.

lib/components/app/app-menu.tsx
binh-dam-ibigroup: Okay, thanks for the reminder; however `display: none` also hides contents from screen readers, so either hide the text visually using small dimensions, or use `aria-selected`, or `aria-label`.

miles-grant-ibigroup:
Wow that was a complex merge. Fingers crossed everything still works

lib/components/app/app-menu.tsx
```diff
@@ -123,46 +120,23 @@ class AppMenu extends Component<
...
-          )
-        }
-
         return (
           <AppMenuItem
             className={subMenuDivider ? 'app-menu-divider' : undefined}
             href={href}
-            icon={<Icon iconType={iconType} iconUrl={iconUrl} />}
+            icon={
+              typeof iconType === 'string' ? (
```
binh-dam-ibigroup, line 128: This did not seem to work with some entries in the ATL config.

binh-dam-ibigroup: See screenshot: 
image

lib/components/app/app.css
```diff
@@ -31,15 +31,17 @@
...
+.app-menu .app-menu-route-viewer-link,
+.app-menu .app-menu-trip-planner-link,
+.app-menu #app-menu-locale-selector {
   display: none;
 }
-@media (max-width: 768px) {
-  .app-menu-route-viewer-link,
-  .app-menu-trip-planner-link {
-    display: inherit;
+@media (max-width: 800px) {
```
binh-dam-ibigroup, line 40: I think the threshold width that has been used in this PR is 768px.

lib/components/app/app.css
```diff
@@ -31,15 +31,17 @@
   color: #ddd;
 }
 
-/* Don't show route viewer or trip plan link in the app menu on desktop as it is in the navbar */
-.app-menu-route-viewer-link,
-.app-menu-trip-planner-link {
+/* Don't show route viewer link in the app menu on desktop as it is in the navbar */
+.app-menu .app-menu-route-viewer-link,
+.app-menu .app-menu-trip-planner-link,
+.app-menu #app-menu-locale-selector {
```
binh-dam-ibigroup, line 37: Why is the last one an id whereas the others are class names?

binh-dam-ibigroup (changes requested):
Thanks for figuring out the merging!... There is one resulting big glitch with some icons missing, see https://github.com/opentripplanner/otp-react-redux/pull/739/files/1eb2213eb17534a1e20314fe080146fce0cd1501..dbfffa0d4f8c474f67f781b29f3725f8b5a6b1af#r1081926370.

lib/components/app/app.css
miles-grant-ibigroup: Because the locale selector is generated whereas the others are fixed.

miles-grant-ibigroup:
Thanks for catching that! So many potential regressions with all this merging...

lib/components/app/app-menu.tsx
binh-dam-ibigroup: This still needs to be resolved.

binh-dam-ibigroup: Ok @miles-grant-ibigroup  and I agreed to solve the selected language indication in a separate PR.

miles-grant-ibigroup: I'd like to use `aria-selected` if possible and that will likely require some syntax changes

lib/components/app/desktop-nav.tsx
```diff
@@ -68,16 +77,14 @@ const DesktopNav = ({ otpConfig, popupTarget, setPopupContent }: Props) => {
 
           {brandingOrTitle}
         </Navbar.Brand>
-      </Navbar.Header>
-      {/* @ts-expect-error typescript is having some trouble with complex type merges here */}
-      <ViewSwitcher sticky />
+        {/* @ts-expect-error typescript is having some trouble with complex type merges here */}
+        <ViewSwitcher sticky />
 
-      <Navbar.Collapse>
```
binh-dam-ibigroup: Removing `Navbar.Collapse` is causing the nav bar to slightly expand vertically at widths less than 768px

Screen Recording 2023-01-20 at 1 01 47 PM

miles-grant-ibigroup: Yep I noticed this.... I chalked it up to something that's better worth fixing by replacing the entire navbar? Which we want to do anyway

binh-dam-ibigroup: For the time being, add `display: none` at widths less than 768px to the underlying `<ul>` that has the top-right menus, please.

binh-dam-ibigroup: Bug seems to have been introduced prior to this PR, so that can be fixed separately.

binh-dam-ibigroup (approved):

lib/components/user/stacked-pane-display.tsx
```diff
@@ -59,7 +59,8 @@ const StackedPaneDisplay = ({
         okayButton={{
           disabled: buttonClicked === 'okay',
```
binh-dam-ibigroup, line 60: Also, disabling the button prevents the form from being submitted.

miles-grant-ibigroup: Yep this was fixed

lib/components/narrative/default/itinerary-summary.js
amy-corson-ibigroup: `align-self: center` works!!

amy-corson-ibigroup (approved):
I couldn't find any major regressions! I did get into the weeds with some css/alignment that I think will make things a bit cleaner, but let me know if you disagree!

lib/components/app/view-switcher.tsx
```diff
@@ -57,19 +61,17 @@ const ViewSwitcher = ({
       }
     >
       <Button
+        aria-controls="view-switcher"
         bsStyle="link"
-        className={activePanel === null && !accountsActive ? 'active' : ''}
+        className={`${tripPlannerActive && 'active'}`}
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Good catch thank you! The entire point of the template string is to avoid that haha

lib/components/app/app.css
```diff
@@ -74,6 +78,7 @@
 }
 .view-switcher button.btn-link.active {
   background: rgba(0, 0, 0, 0.15);
+  font-weight: 600;
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Yeah I had the same qualm when I first did this. I tried it but then the effect is that it looks like the bubble grows which I liked less than the bold one sort of scooting a bit

lib/components/mobile/mobile.css
```diff
@@ -18,18 +18,30 @@
...
+  align-items: center;
+  display: flex;
+  height: 60px;
+  justify-content: center;
   left: 50px;
+  max-width: 70%;
+  position: fixed;
   right: 50px;
   text-align: center;
+  top: -11px;
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Perfect fix thank you!

lib/components/mobile/mobile.css
```diff
@@ -18,18 +18,30 @@
...
 
 .otp .navbar .mobile-header {
-  position: fixed;
-  top: 5px;
+  align-items: center;
+  display: flex;
+  height: 60px;
+  justify-content: center;
   left: 50px;
+  max-width: 70%;
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Ignore my previous comment -- this is excellent thanks very much!

PR #743, opened 2023-01-03
Lazy-load expensive react-select

miles-grant-ibigroup:
`react-select` is an expensive load, but luckily we can easily lazy load it! This PR gives us a quick 20kb bundle size win.


daniel-heppner-ibigroup:
common miles W right here

daniel-heppner-ibigroup (approved):
works great for me!

philip-cline (approved):
Amazing... almost makes you wonder if this could be used anywhere else 🤔

PR #744, opened 2023-01-04
Itinerary Narrative Semantic HTML

amy-corson-ibigroup:
- Changes headers across the itinerary narrative and results views to `h1`
- Changes results to `ul`
- Adds headers and groupings to trip tools

### DIFF
--- a/lib/components/narrative/metro/metro-itinerary.tsx
+++ b/lib/components/narrative/metro/metro-itinerary.tsx
@@ -358,13 +358,10 @@ class MetroItinerary extends NarrativeItinerary {
 
     // Use first leg's agency as a fallback
     return (
-      <div
+      <li
         className={`option metro-itin${active ? ' active' : ''}${
           expanded ? ' expanded' : ''
         }`}
-        onMouseEnter={this._onMouseEnter}
-        onMouseLeave={this._onMouseLeave}
-        role="presentation"
       >
         {/* Semantically this is incorrect, but this helps a11y tests pass. Buttons may not contain html. TODO FIX? */}
         <div
@@ -374,6 +371,8 @@ class MetroItinerary extends NarrativeItinerary {
           // first time?
           // eslint-disable-next-line @typescript-eslint/no-empty-function
           onKeyDown={() => {}}
+          onMouseEnter={this._onMouseEnter}
+          onMouseLeave={this._onMouseLeave}
           // TODO: use _onHeaderClick for tap only -- this will require disabling
           // this onClick handler after a touchStart
           role="presentation"
@@ -483,7 +482,7 @@ class MetroItinerary extends NarrativeItinerary {
             />
           </>
         )}
-      </div>
+      </li>
     )
   }
 }
--- a/lib/components/narrative/metro/metro-itinerary.tsx
+++ b/lib/components/narrative/metro/metro-itinerary.tsx
@@ -358,7 +358,7 @@ class MetroItinerary extends NarrativeItinerary {
 
     // Use first leg's agency as a fallback
     return (
-      <li
+      <div
         className={`option metro-itin${active ? ' active' : ''}${
           expanded ? ' expanded' : ''
         }`}
@@ -482,7 +482,7 @@ class MetroItinerary extends NarrativeItinerary {
             />
           </>
         )}
-      </li>
+      </div>
     )
   }
 }
--- a/lib/components/narrative/narrative-itineraries-header.tsx
+++ b/lib/components/narrative/narrative-itineraries-header.tsx
@@ -1,6 +1,7 @@
 import { ArrowLeft } from '@styled-icons/fa-solid/ArrowLeft'
 import { ExclamationTriangle } from '@styled-icons/fa-solid/ExclamationTriangle'
 import { FormattedMessage, useIntl } from 'react-intl'
+import { Itinerary } from '@opentripplanner/types'
 import { SortAmountDown } from '@styled-icons/fa-solid/SortAmountDown'
 import { SortAmountUp } from '@styled-icons/fa-solid/SortAmountUp'
 import React from 'react'
@@ -8,6 +9,7 @@ import styled from 'styled-components'
 
 import { IconWithText, StyledIconWrapper } from '../util/styledIcon'
 
+import { ItineraryDescription } from './default/itinerary-description'
 import PlanFirstLastButtons from './plan-first-last-buttons'
 import SaveTripButton from './save-trip-button'
 
@@ -24,6 +26,7 @@ export default function NarrativeItinerariesHeader({
   customBatchUiBackground,
   errors,
   itineraries,
+  itinerary,
   itineraryIsExpanded,
   onSortChange,
   onSortDirChange,
@@ -39,6 +42,7 @@ export default function NarrativeItinerariesHeader({
   customBatchUiBackground?: boolean
   errors: unknown[]
   itineraries: unknown[]
+  itinerary: Itinerary
   itineraryIsExpanded: boolean
   onSortChange: () => void
   onSortDirChange: () => void
@@ -84,6 +88,9 @@ export default function NarrativeItinerariesHeader({
               <FormattedMessage id="components.NarrativeItinerariesHeader.viewAll" />
             </IconWithText>
           </button>
+          <h1 style={{ display: 'none' }}>
+            <ItineraryDescription itinerary={itinerary} />
+          </h1>
           {itineraryIsExpanded && (
             // marginLeft: auto is a way of making something "float right"
             // within a flex container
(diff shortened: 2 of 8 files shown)


lib/components/narrative/narrative-itineraries.js
```diff
@@ -404,30 +406,37 @@ class NarrativeItineraries extends Component {
...
-                      this._renderItineraryRow(itin, true)
-                    )
-                  )}
-              </S.SingleModeRowContainer>
+              {groupItineraries && !itineraryIsExpanded && (
+                <li style={{ padding: '0 1em' }}>
+                  <h2>
+                    {intl.formatMessage({ id: 'otpUi.LocationField.other' })}
+                  </h2>
+                  <S.SingleModeRowContainer aria-label="Non-Transit Options">
```
binh-dam-ibigroup, line 428: aria-label is redundant with heading above.

lib/components/narrative/narrative-itineraries.js
```diff
@@ -404,30 +406,37 @@ class NarrativeItineraries extends Component {
               {groupItineraries && !itineraryIsExpanded
                 ? Object.keys(groupedMergedItineraries.multi).map((mode) => {
                     return (
-                      <div key={mode} style={{ padding: '0 1em' }}>
+                      <li key={mode} style={{ padding: '0 1em' }}>
```
binh-dam-ibigroup, line 409: I don't think the list is needed here, as headings already follow (2 instances).

lib/components/narrative/narrative-itineraries.js
```diff
@@ -386,12 +388,12 @@ class NarrativeItineraries extends Component {
           showingErrors={showingErrors}
           sort={sort}
         />
-        <div
-          // FIXME: Change to a ul with li children?
-          className="list"
+        <ul
```
binh-dam-ibigroup, line 391: I think this no longer needs to be a list (there are lists already under each mode heading).

lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -358,13 +358,10 @@ class MetroItinerary extends NarrativeItinerary {
 
     // Use first leg's agency as a fallback
     return (
-      <div
+      <li
```
binh-dam-ibigroup, line 361: When viewing one itinerary after a search, the viewed itinerary is still presented as a list with a single item. Instead, there should be no list. Therefore, `<li>` should move out of this component (retain `<div>` for now).

lib/components/narrative/narrative-itineraries.js
```diff
@@ -404,30 +406,37 @@ class NarrativeItineraries extends Component {
...
-                      <div key={mode} style={{ padding: '0 1em' }}>
+                      <li key={mode} style={{ padding: '0 1em' }}>
                         <h2>{mode}</h2>
-                        {groupedMergedItineraries.multi[mode].map((itin) =>
-                          this._renderItineraryRow(itin)
-                        )}
-                      </div>
+                        <ul style={{ listStyle: 'none', padding: '0' }}>
+                          {groupedMergedItineraries.multi[mode].map((itin) =>
+                            this._renderItineraryRow(itin)
```
binh-dam-ibigroup, line 413: Surround with `<li>` (assuming moving `<li>` out of `MetroItinerary`).

lib/components/narrative/narrative-itineraries.js
```diff
@@ -404,30 +406,37 @@ class NarrativeItineraries extends Component {
...
-                {groupItineraries &&
-                  !itineraryIsExpanded &&
-                  Object.keys(groupedMergedItineraries.single).map((mode) =>
-                    groupedMergedItineraries.single[mode].map((itin) =>
-                      this._renderItineraryRow(itin, true)
-                    )
-                  )}
-              </S.SingleModeRowContainer>
+              {groupItineraries && !itineraryIsExpanded && (
+                <li style={{ padding: '0 1em' }}>
```
binh-dam-ibigroup, line 424: This list item is not needed if removing the parent `<ul>`.

lib/components/narrative/narrative-itineraries.js
```diff
@@ -404,30 +406,37 @@ class NarrativeItineraries extends Component {
...
+                  <h2>
+                    {intl.formatMessage({ id: 'otpUi.LocationField.other' })}
+                  </h2>
+                  <S.SingleModeRowContainer aria-label="Non-Transit Options">
+                    {Object.keys(groupedMergedItineraries.single).map((mode) =>
+                      groupedMergedItineraries.single[mode].map((itin) =>
+                        this._renderItineraryRow(itin, true)
```
binh-dam-ibigroup, line 431: Surround with `<li>` (assuming moving `<li>` out of `MetroItinerary`).

binh-dam-ibigroup (changes requested):
Move `<li>` out of `MetroItinerary` and remove the list when rendering mode headings. Looks good otherwise.

lib/components/narrative/narrative-itineraries.js
amy-corson-ibigroup: Unfortunately I'm realizing the issue with turning this one back into a div is that in the batch-itinerary view, the results do not show up as a list

amy-corson-ibigroup: Which probably means if this stays as a <ul>, the sub-lists of modes in the metro-itinerary should probably also remain?

miles-grant-ibigroup (approved):
One small nitpick but with that resolved this is good to go! Thanks for doing this finnicky work

lib/components/narrative/narrative-itineraries-header.tsx
```diff
@@ -84,6 +88,9 @@ export default function NarrativeItinerariesHeader({
               <FormattedMessage id="components.NarrativeItinerariesHeader.viewAll" />
             </IconWithText>
           </button>
+          <h1 style={{ display: 'none' }}>
```
binh-dam-ibigroup, line 91: For use with screen readers only, use CSS `visibility: hidden` and set a zero-width or zero-height (see other comment regarding `display: none`).

lib/components/narrative/narrative-itineraries.js
```diff
@@ -256,6 +260,10 @@ class NarrativeItineraries extends Component {
 
     if (!activeSearch) return null
 
+    // render lists become divs if itinerary is expanded, to avoid rendering a list with list item
+    const ListContainer = itineraryIsExpanded ? 'div' : 'ul'
+    const ListItem = itineraryIsExpanded ? 'div' : 'li'
```
binh-dam-ibigroup, line 265: Not needed per https://github.com/opentripplanner/otp-react-redux/pull/744#discussion_r1063525664.

lib/components/narrative/narrative.css
```diff
@@ -44,6 +44,20 @@
 }
 
 /* TRIP TOOLS ROW */
+
+.otp .trip-tools-container {
+  margin: 15px 5px;
+  padding: 10px 16px;
+  background-color: #eee;
+  border-radius: 6px;
```
binh-dam-ibigroup, line 52: Sort attributes.

binh-dam-ibigroup (changes requested):
A few more touches needed, but we're almost there!

lib/components/narrative/narrative-itineraries.js
```diff
@@ -197,6 +197,7 @@ class NarrativeItineraries extends Component {
       visibleItinerary
     } = this.props
     const { ItineraryBody, LegIcon } = this.context
+    const ListItem = itineraryIsExpanded ? 'div' : 'li'
```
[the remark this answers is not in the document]
amy-corson-ibigroup: Have to give some credit to @miles-grant-ibigroup for that one 😁

lib/components/narrative/narrative-itineraries.js
amy-corson-ibigroup: I think the same issue we pointed out in the above thread still applies - removing the outer <ul> means the batch-itin doesn't render as a list. Let me know if I'm misunderstanding!

binh-dam-ibigroup: To preserve that, how about moving the `ListContainer` at line 399 to the block around line 429 (make it surround `itinerariesWithCo2.map(...)`.

amy-corson-ibigroup: That works great, thanks so much Binh!

lib/components/narrative/narrative-itineraries-header.tsx
```diff
@@ -138,6 +151,7 @@ export default function NarrativeItinerariesHeader({
               marginLeft: showHeaderText ? 'inherit' : 'auto'
             }}
           >
+            <h1 style={{ display: 'none' }}>{itinerariesFound}</h1>
```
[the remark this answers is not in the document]
binh-dam-ibigroup: I am sorry, [`visibility: hidden` also hides content from screen readers](https://developer.mozilla.org/en-US/docs/Web/CSS/visibility#accessibility_concerns), so just keep the zero-size for these items.

lib/components/narrative/narrative-itineraries-header.tsx
binh-dam-ibigroup: Sorry, don't render this `<h1>` at all. It was not going to be shown to screen readers (see https://developer.mozilla.org/en-US/docs/Web/CSS/visibility#accessibility_concerns). Rather, set a zero-size to the `<h1>` above this one as needed.

binh-dam-ibigroup (changes requested):
My apologies... I gave you wrong advice regarding `visibility: hidden`.

lib/components/narrative/narrative-itineraries-header.tsx
amy-corson-ibigroup: Thank you so much for catching this!! I think removing this h1 leaves the itinerary narrative without a header, but I can fix both with `height: 0; width: 0; overflow: hidden;`

PR #747, opened 2023-01-06
CallTaker Clean Up

miles-grant-ibigroup:
Throughout all our recent merges, some regressions have slipped into the calltaker. This PR resolves some of these.


demory (approved):

lib/actions/api.js
```diff
@@ -443,6 +443,13 @@ export function createQueryAction(
   return async function (dispatch, getState) {
     const state = getState()
     const { config } = state.otp
+
+    if (options.appendBounds) {
+      const bounds = config?.geocoder?.boundary?.rect
+      const stringifiedBounds = '&' + qs.stringify(bounds)
```
binh-dam-ibigroup, line 449: Should there be a check for null `bounds`?

binh-dam-ibigroup (approved):

lib/actions/api.js
miles-grant-ibigroup: Good catch thanks! Addressed in 4ddf2c1901d354aea38d19c23a38edb8b0de2eab

PR #749, opened 2023-01-12
App Menu Keyboard Navigation

binh-dam-ibigroup:
This PR improves keyboard navigation around the app menu.

### DIFF
--- a/index.css
+++ b/index.css
@@ -72,11 +72,6 @@ input[type="text"]::-ms-clear {
   margin: 52px auto 0 0;
 }
 
-.slide-pane__header,
-.slide-pane__close {
-  display: none;
-}
-
 .slide-pane__content {
   padding: 6px 0;
 }
@@ -87,29 +82,35 @@ input[type="text"]::-ms-clear {
 
 .app-menu {
   margin: 0;
-  padding: 0;
+  padding: 0.5rem 0;
 }
 
-.app-menu li {
-  list-style-type: none;
-  margin: 1rem 2rem;
-  cursor: pointer;
-}
-
-.app-menu li img,
-.app-menu li svg {
+.app-menu img,
+.app-menu svg {
   max-height: 1em;
   margin-right: 2rem;
   vertical-align: middle;
   width: 1em;
 }
 
-.app-menu li a {
+.app-menu a,
+.app-menu button {
+  background: none;
+  border: none;
   color: inherit;
+  cursor: pointer;
+  display: flex;
   font-size: 20px;
+  padding: 0.5rem 2rem;
   text-decoration: none;
-  outline: none;
+  width: 100%;
 }
+
+.app-menu a:hover,
+.app-menu button:hover {
+  color: #4c97f5;
+}
+
 
 .view-switcher {
   display: none;
@@ -119,11 +120,6 @@ input[type="text"]::-ms-clear {
   .view-switcher {
     display: flex;
   }
-}
-
-.menu-item:hover *,
-.app-menu li a span *:hover {
-  color: #4c97f5;
 }
 
 .expand-menu-chevron {
@@ -137,34 +133,22 @@ input[type="text"]::-ms-clear {
   white-space: nowrap;
 }
 
-.expand-submenu-button a {
-  display: flex;
-  width: 100%;
+.expand-submenu-button {
   justify-content: space-between;
-  background: transparent;
-  border: none;
-  outline: none;
 }
 
 .sub-menu-container {
   border-top: 1px solid #cccccc;
+  margin-top: 0.5rem;
 }
 
-.sub-menu-container li a {
+.sub-menu-container > * {
   margin-left: 2rem;
 }
 
-.sub-menu-container li a span {
-  color: inherit;
-}
-
-.app-menu-divider::after {
-  content: "";
-  display: block;
-  height: 1px;
-  background: #ccc;
-  margin: 1rem;
-  width: 200%; /* ensures the line extends to the edge of the pane */
+.app-menu .app-menu-divider {
+  border-bottom: 1px solid #ccc;
+  padding: 1rem 2rem;
 }
 
 /* Buttons */
--- a/index.css
+++ b/index.css
@@ -23,43 +23,45 @@ input[type="text"]::-ms-clear {
 
 /* New app menu */
 .app-menu-icon {
-  width: 21px;
-  height: 15px;
+  background: none;
+  border: none;
+  cursor: pointer;
   display: flex;
   flex-direction: column;
+  height: 15px;
   justify-content: space-between;
-  cursor: pointer;
+  padding: 0;
   position: absolute;
-  z-index: 10;
-  transition: all 1s ease;
   top: 16px;
+  transition: all 1s ease;
+  width: 21px;
+  z-index: 10;
 }
 
 @media only screen and (max-width: 768px) {
   .app-menu-icon {
-    margin-left: 15px;
+    left: 15px;
   }
 }
 
-.menu-top-line,
-.menu-middle-line,
-.menu-bottom-line,
-.menu-left-x,
-.menu-right-x {
+.app-menu-icon > span {
   border-bottom: 3px solid #ffffff;
+  display: block;
+  position: relative;
   transition: all 0.5s ease;
+  width: 100%;
 }
 
-.menu-left-x {
+.app-menu-icon.open span:first-child {
   transform: rotate(45deg);
   top: 7px;
-  position: relative;
 }
-
-.menu-right-x {
+.app-menu-icon.open span:last-child {
   transform: rotate(-45deg);
   bottom: 5px;
-  position: relative;
+}
+.app-menu-icon.open span:nth-child(2) {
+  display: none;
 }
 
 .slide-pane {
--- a/index.css
+++ b/index.css
@@ -137,10 +137,7 @@ input[type="text"]::-ms-clear {
 .sub-menu-container {
   border-top: 1px solid #cccccc;
   margin-top: 0.5rem;
-}
-
-.sub-menu-container > * {
-  margin-left: 2rem;
+  padding-left: 2rem;
 }
 
 .app-menu .app-menu-divider {
--- a/index.css
+++ b/index.css
@@ -88,7 +88,7 @@ input[type="text"]::-ms-clear {
 .app-menu img,
 .app-menu svg {
   max-height: 1em;
-  margin-right: 2rem;
+  margin: 0 2rem;
   vertical-align: middle;
   width: 1em;
 }
@@ -101,7 +101,7 @@ input[type="text"]::-ms-clear {
   cursor: pointer;
   display: flex;
   font-size: 20px;
-  padding: 0.5rem 2rem;
+  padding: 0.5rem 0;
   text-decoration: none;
   width: 100%;
 }
@@ -123,7 +123,8 @@ input[type="text"]::-ms-clear {
 }
 
 .expand-menu-chevron {
-  margin-right: -2rem;
+  flex-grow: 1;
+  text-align: end;
 }
 
 .dropdown-header {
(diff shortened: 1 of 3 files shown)


index.css
```diff
@@ -23,43 +23,45 @@ input[type="text"]::-ms-clear {
...
+.app-menu-icon > span {
   border-bottom: 3px solid #ffffff;
+  display: block;
+  position: relative;
   transition: all 0.5s ease;
+  width: 100%;
 }
 
-.menu-left-x {
+.app-menu-icon.open span:first-child {
```
miles-grant-ibigroup, line 55: I think I preferred the 3 classes to `nth-child` which can sometimes get messy, but I'll leave it to the second reviewer to decide if this is necessary to change.

lib/components/app/app-menu-item.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * Helper method to find the element within the app menu at the given offset
+ * (e.g. previous or next) relative to the specified element.
+ * The query is limited to the app menu so that arrow navigation is contained within
+ * (tab navigation is not restricted).
+ */
+function getEntryRelativeTo(element: EventTarget, offset: 1 | -1): HTMLElement {
+  const entries = Array.from(
+    document.querySelectorAll('.app-menu a, .app-menu button')
```
miles-grant-ibigroup, line 27: This isn't the best but given the fact that we don't have time to rebuild the menu from scratch happy to let this one pass. We should maybe add a TODO to rebuild the entire menu? I feel it's very bloated -- I don't think we need the library it uses and some of the styling is overkill as well I feel.

miles-grant-ibigroup, line 27: This isn't the best but given the fact that we don't have time to rebuild the menu from scratch happy to let this one pass. We should maybe add a TODO to rebuild the entire menu? I feel it's very bloated -- I don't think we need the library it uses and some of the styling is overkill as well I feel.

lib/components/app/app-menu.tsx
```diff
@@ -176,85 +138,80 @@ class AppMenu extends Component<
...
-          role="button"
-          tabIndex={0}
         >
-          <span className={isPaneOpen ? 'menu-left-x' : 'menu-top-line'} />
-          <span className={isPaneOpen ? '' : 'menu-middle-line'} />
-          <span className={isPaneOpen ? 'menu-right-x' : 'menu-bottom-line'} />
-        </div>
+          <span />
+          <span />
+          <span />
```
miles-grant-ibigroup, line 156: Can we add a comment here saying what this is? It looks really funky like this. I think my preferred solution would be to make all those line styles into a styled component and then having a `Hamburger` component or something

miles-grant-ibigroup, line 156: Can we add a comment here saying what this is? It looks really funky like this. I think my preferred solution would be to make all those line styles into a styled component and then having a `Hamburger` component or something

lib/components/app/app-menu.tsx
```diff
@@ -176,85 +138,80 @@ class AppMenu extends Component<
...
+        </button>
+        {/* Use a button for skipping navigation. A regular <a> element would modify the URL,
```
miles-grant-ibigroup, line 158: Is there no way to use `preventDefault` instead? That feels less bad than using a button...

lib/components/app/app-menu.tsx
```diff
@@ -176,85 +138,80 @@ class AppMenu extends Component<
...
+          shouldCloseOnEsc
           width="320px"
         >
-          <div className="app-menu">
+          <div className="app-menu" id="app-menu">
             {/* This item is duplicated by the view-switcher, but only shown on mobile
             when the view switcher isn't shown (using css) */}
-            <MenuItem
+            <AppMenuItem
               className="app-menu-route-viewer-link"
```
miles-grant-ibigroup, line 178: I don't think this className is getting properly applied -- `app-menu-route-viewer-link` should disappear on >768px screen sizes but the rule doesn't seem to be applying.

lib/components/app/app-menu.tsx
```diff
@@ -294,39 +251,21 @@ export default injectIntl(
...
-        />
-      ) : iconType ? (
-        <SvgIcon iconName={iconType} />
-      ) : (
-        <ExternalLinkSquareAlt />
-      )}
-      {label}
-    </span>
+  return iconUrl ? (
+    <img alt="" src={iconUrl} />
```
miles-grant-ibigroup, line 265: Are you sure about the blank alt? I get how the UX might be better but if we make it invisible to screen readers shouldn't it have a `role="presentation"`?

lib/components/app/responsive-webapp.js
```diff
@@ -212,10 +212,10 @@ class ResponsiveWebapp extends Component {
             <Col className="sidebar" md={4} sm={6}>
               {/*
                 Note: the main tag provides a way for users of screen readers
-                to skip to the primary page content.
+                to skip to the primary page content (tabinde = -1 needed for programmatic navigation skip).
```
miles-grant-ibigroup, line 215: Typo `tabIndex`

miles-grant-ibigroup, line 215: Typo `tabIndex`

miles-grant-ibigroup, line 215: Typo `tabIndex`

lib/components/app/responsive-webapp.js
```diff
@@ -235,8 +235,9 @@ class ResponsiveWebapp extends Component {
     const { popupContent } = this.props
 
     return (
-      // <main> Needed for accessibility checks. TODO: Find a better place.
-      <main>
+      // <main> Needed for accessibility checks.
+      // Tabindes of -1 required for programmatic navigation skip.
```
miles-grant-ibigroup, line 239: Typo `tabIndex`

miles-grant-ibigroup (changes requested):
Based on the timeframe for this I'll let that `document.querySelector` go even if ideally we use refs instead.

A few nitpicky comments -- the only blocking one is the view routes not disappearing properly (my guess is it's the class not applying)

lib/components/app/app-menu-item.tsx
binh-dam-ibigroup: Which library? `AnimateHeight`?

lib/components/app/app-menu.tsx
binh-dam-ibigroup: Benefits of `alt=""` are explained here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-alt

binh-dam-ibigroup: `alt=""` is used by screen readers and also by visual browsers (e.g. when the `src` is broken.)

lib/components/app/app-menu-item.tsx
binh-dam-ibigroup: This function supports navigation using the ArrowUp/ArrowDown keys, which is a nice-to-have for keyboard navigation.

lib/components/app/app-menu.tsx
miles-grant-ibigroup: Ok I trust you on this one

lib/components/app/app-menu-item.tsx
miles-grant-ibigroup: `react-sliding-pane`

miles-grant-ibigroup: A properly built menu wouldn't require javascript hacks for it to work properly, but that's outside the scope of this PR.

lib/components/app/responsive-webapp.js
binh-dam-ibigroup: Thanks, typos fixed in https://github.com/opentripplanner/otp-react-redux/pull/749/commits/d8aa9b7ec5ba8ff7bae24bdbcd2ebac08c4ec5b2.

index.css
binh-dam-ibigroup: I replaced the ordinal positioning CSS with the top/middle/bottom nomenclature in  https://github.com/opentripplanner/otp-react-redux/pull/749/commits/3523d0d99f2b12c7b43aadbde6ee57d915966e86 so it is closer from how it was before (and improved).

lib/components/app/app-menu.tsx
binh-dam-ibigroup: See https://github.com/opentripplanner/otp-react-redux/pull/749#discussion_r1070096755.

lib/components/app/app-menu.tsx
binh-dam-ibigroup: Fixed in https://github.com/opentripplanner/otp-react-redux/pull/749/commits/2434a1c3b347761d8228160badf161dced0dbc18. The `display` attribute I added to the buttons conflicted with the one that sets visibility for the route viewer menu.

lib/components/app/app-menu.tsx
binh-dam-ibigroup: The anchor + href implementation is more complicated unfortunately, the main complicated things being generating the URL (e.g. `#content/?ui_activeSearch=...`) and removing the `#content` part of the URL once the link is clicked.

miles-grant-ibigroup (approved):
It's not perfect, but given how much better the accessible experience is I'm happy with it! Couldn't spot any regressions.

amy-corson-ibigroup (changes requested):
This looks and works so much better!! Just a few things to address below.

lib/components/app/app-menu.tsx
```diff
@@ -176,85 +138,81 @@ class AppMenu extends Component<
...
+          <Bar className="menu-line bottom" />
+        </button>
+        {/* Use a button for skipping navigation. A regular <a href=...> element would modify the URL,
+            and such change would be captured by the router without changing the focused element. */}
+        <button
+          className="skip-nav-button"
+          onClick={this._handleSkipNavigation}
+        >
+          <FormattedMessage id="components.AppMenu.skipNavigation" />
+        </button>
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Addressed in https://github.com/opentripplanner/otp-react-redux/pull/749/commits/52f6eb57dc71ea974788d4d63f13142402c8e62d.

index.css
```diff
@@ -85,28 +82,45 @@ input[type="text"]::-ms-clear {
...
+}
+
+.app-menu a:focus,
+.app-menu button:focus {
+  background-color: #ddd;
+  outline: 2px solid #4c97f5;
+  outline-offset: -2px;
+}
+
+.skip-nav-button {
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Good catch! Fixed in https://github.com/opentripplanner/otp-react-redux/pull/749/commits/52f6eb57dc71ea974788d4d63f13142402c8e62d.

lib/components/app/app.css
```diff
@@ -32,11 +32,11 @@
...
 
 /* Don't show route viewer link in the app menu on desktop as it is in the navbar */
-.app-menu-route-viewer-link {
+.app-menu .app-menu-route-viewer-link {
   display: none;
 }
 @media (max-width: 800px) {
-  .app-menu-route-viewer-link {
+  .app-menu .app-menu-route-viewer-link {
     display: block;
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Thanks... fixed in https://github.com/opentripplanner/otp-react-redux/pull/749/commits/52f6eb57dc71ea974788d4d63f13142402c8e62d.

index.css
```diff
@@ -85,28 +82,49 @@ input[type="text"]::-ms-clear {
...
+  color: initial;
+  position: fixed;
+  top: -30px;
+}
+.skip-nav-button:focus {
+  top: 0;
 }
```
amy-corson-ibigroup, line 128: What about this, to make the button a little more cohesive with the nav bar? Currently the skip button is covered with the nav button, so the z-index feels the most important out of the below changes

```suggestion
.skip-nav-button:focus {
  padding: 7px 24px;
  top: 7px;
  z-index: 100
}
```

amy-corson-ibigroup (approved):
LGTM, just a small suggestion re: button styling!

index.css
binh-dam-ibigroup: Sure, thanks!

PR #752, opened 2023-01-13
Support mutliple base layers

miles-grant-ibigroup:
Vector tiles dropped support for alternate base layers. This PR brings that support back. Configuration is the same as before.


lib/components/map/default-map.tsx
```diff
@@ -263,12 +263,16 @@ class DefaultMap extends Component {
...
+    const baseLayerUrls = baseLayersWithNames?.map((bl) => bl.url)
+    const baseLayerNames = baseLayersWithNames?.map((bl) => bl.name)
 
     return (
       <MapContainer>
         <BaseMap
-          // Only 1 base layer is supported at the moment
-          baseLayer={baseLayersWithNames?.[0]?.url}
+          baseLayer={
+            baseLayerUrls?.length > 1 ? baseLayerUrls : baseLayerUrls?.[0]
```
daniel-heppner-ibigroup, line 273: Why can't you just pass in an array of 1 and handle it on the OTP UI side? Instead of having this accept two different types.

daniel-heppner-ibigroup (approved):
Looks good, just one thought about general structure to make the types nicer.

binh-dam-ibigroup (approved):

PR #753, opened 2023-01-13
Grouping, Labels, and Headings in Itineraries

amy-corson-ibigroup:
Metro Itinerary:
- invisible header for screen readers
- group itin details in a list

Default/Batch Itinerary:
- make "click to expand details" button consistent for screen reader usage
- invisible header for screen readers
- remove empty list item when no co2 config present


miles-grant-ibigroup (changes requested):
Thank you so much for doing a lot of finicky work to get this in this much much better state! 

I have a few mostly nitpicks to get this over the finish line but overall thanks so much! Very exciting stuff

lib/components/narrative/narrative-itinerary.js
```diff
@@ -30,6 +31,12 @@ export default class NarrativeItinerary extends Component {
     }
   }
 
+  onDirectClick = () => {
```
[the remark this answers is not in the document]
amy-corson-ibigroup: `_onHeaderClick` requires two clicks on the result to expand the batch-itinerary (clicking once highlights the result and shows the "click to see details" hint, clicking again expands the itin)

`onDirectClick` opens the itinerary in one click, and it's essentially so screen readers can open the itinerary directly without having to select the option twice!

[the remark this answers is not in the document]
miles-grant-ibigroup: This makes sense! I think we discussed a while ago doing something similar for touch screens so thank you for doing this! We can probably make good use of this in the future. Do you think there's room for having the 2 methods have a shared util method they both depend on? I can't immediately tell if it'll make things better by looking at the code, so I trust you to decide if it's a good idea

lib/components/narrative/default/default-itinerary.js
```diff
@@ -263,68 +287,81 @@ class DefaultItinerary extends NarrativeItinerary {
         onMouseLeave={this._onMouseLeave}
         role="presentation"
       >
-        <button
+        <ItinerarySummaryWrapper
```
AdrianaCeric, line 290: We could add an `aria-label` here?

lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -316,6 +333,20 @@ class MetroItinerary extends NarrativeItinerary {
...
+          leg.mode
+        }`
+        if (index === routeLegs.length) {
+          return legHeader
+        }
+        return ` ${legHeader}`
+      })
+      return routeHeader.toString()
+    }
+
```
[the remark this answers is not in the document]
amy-corson-ibigroup: Exactly what I was looking for!! I didn't wind up creating a new translation key, so let me know if this isn't what you had in mind!

lib/components/narrative/narrative-itinerary.js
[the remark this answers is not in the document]
amy-corson-ibigroup: Added a tiny util, it may be more trouble than it's worth? Let me know

lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -75,22 +86,32 @@ const DepartureTimes = styled.span`
   }
 `
 
-const PrimaryInfo = styled.span`
+const ItineraryDetails = styled.ul`
+  display: flex;
+  flex-direction: column;
+  grid-column-start: -1;
```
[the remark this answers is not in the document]
amy-corson-ibigroup: nope! good catch!

lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -400,54 +431,64 @@ class MetroItinerary extends NarrativeItinerary {
...
-                    values={{
-                      time: <FormattedDuration duration={itinerary.walkTime} />
-                    }}
-                  />
-                </SecondaryInfo>
+                  </SecondaryInfo>
+                </ItineraryDetails>
                 <DepartureTimes>
                   <FormattedMessage id="components.MetroUI.leaveAt" />{' '}
                   <DepartureTimesList
```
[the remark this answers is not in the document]
amy-corson-ibigroup: Ahh I agree this feels messy in the expanded itin... unfortunately moving this button (sometimes list of buttons) into the `ItineraryDetails` list works great for screen readers but really otherwise breaks the UI 😅 Maybe there's a different way to separate it other than adding to the `ItineraryDetails` list? or maybe i'm missing something?

lib/components/narrative/default/default-itinerary.js
```diff
@@ -263,68 +287,81 @@ class DefaultItinerary extends NarrativeItinerary {
...
           className="header"
-          // _onHeaderClick comes from super component (NarrativeItinerary).
```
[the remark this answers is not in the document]
amy-corson-ibigroup: Good catch, thank you!!

lib/components/narrative/default/default-itinerary.js
amy-corson-ibigroup: I definitely see where you're coming from here, however I think we may be okay without it? This element does have an unmarked interaction, but it also has an accompanying button for screen readers so they can access the same interaction in a more straightforward way! 
image

lib/components/narrative/narrative-itinerary.js
```diff
@@ -20,23 +20,28 @@ export default class NarrativeItinerary extends Component {
     toggleDetailedItinerary: PropTypes.func
   }
 
+  _setItin = () => {
```
miles-grant-ibigroup, line 23: I would rename this to `_setItinToActiveIndex`

lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -432,7 +423,10 @@ class MetroItinerary extends NarrativeItinerary {
               <ItineraryGrid className="itin-grid">
                 {/* TODO: a11y: add aria-label to parent element */}
                 <InvisibleHeader as={expanded && 'h2'}>
-                  {renderAccessibilityHeader(itinerary.legs)}
+                  <FormattedList
+                    type="conjunction"
+                    value={routeAndModeStrings}
+                  />
```
miles-grant-ibigroup, line 429: Yess love it

lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -332,26 +332,17 @@ class MetroItinerary extends NarrativeItinerary {
...
-
-    const renderRouteBlocks = (legs: Leg[], firstOnly = false) => {
-      const routeBlocks = legs
-        .filter(removeInsignifigantWalkLegs)
+    const routeLegs = itinerary.legs.filter(removeInsignifigantWalkLegs)
+    const routeAndModeStrings = routeLegs.map((leg: Leg, index: number) => {
+      return `${
+        leg.routeShortName ? leg.routeShortName : ''
+      } ${getFormattedMode(leg.mode, intl)}`
+    })
```
miles-grant-ibigroup, line 340: Much cleaner! Sadly though I think you'll have to use a new i18n key here -- so we can support RTL languages.

react-intl also supports most of the logic you're doing here! If you peek around you can see some examples of changing the string based on if variables are present

miles-grant-ibigroup (changes requested):
Looking really good! Just a few small nitpicks about the strings

lib/components/narrative/metro/metro-itinerary.tsx
amy-corson-ibigroup: Unfortunately the closest thing I found was the [Select](https://formatjs.io/docs/core-concepts/icu-syntax/#select-format) format, but I didn't have much luck filtering undefined variables. In the interest of time and i18n I went ahead and paired this header down to just the `FormattedList` of modes, although that may not be descriptive enough?

lib/components/narrative/metro/metro-itinerary.tsx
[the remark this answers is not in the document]
amy-corson-ibigroup: Great idea, thank you!

lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -52,7 +55,18 @@ const ItineraryWrapper = styled.div.attrs((props) => {
   padding: 0;
 `
 
+// invisible header rendered for screen readers and a11y technologies
+const InvisibleHeader = styled.h3`
+  width: 0;
+  height: 0;
+  overflow: hidden;
+  //place it in an unused grid cell so it doesn't add a row
```
miles-grant-ibigroup, line 63: Not sure how I feel about this method of hiding the element -- will leave it up to the second reviewer to decide

miles-grant-ibigroup (approved):
Haven't spotted any regressions and looks clean to me!

lib/components/narrative/default/default-itinerary.js
```diff
@@ -263,68 +287,82 @@ class DefaultItinerary extends NarrativeItinerary {
...
+              if (bSelected) return 1
+              return a.order - b.order
+            })
+              .filter(
+                (x) =>
+                  x.render(
+                    itinerary,
+                    itineraryAttributeOptions,
+                    defaultFareKey
+                  ) !== undefined
```
binh-dam-ibigroup, line 333: Can we reuse these rendering results on line 346?

lib/components/narrative/default/default-itinerary.js
```diff
@@ -52,11 +52,28 @@ const LegIconWrapper = styled.div`
...
+  border: none;
+  background: transparent;
+  position: absolute;
+`
 const DetailsHint = styled.div`
   clear: both;
   color: #685c5c;
   font-size: small;
   text-align: center;
+  cursor: pointer;
```
binh-dam-ibigroup, line 76: Sort CSS

lib/components/narrative/default/default-itinerary.js
```diff
@@ -52,11 +52,28 @@ const LegIconWrapper = styled.div`
...
 `
 
+const DetailsHintButton = styled.button`
+  &.visible {
+     display: contents
+  }
+
+  // the width/height/overflow trick still renders a tiny button, 
+  so make it invisible and out of the way, but still visible to 
+  screen readers
```
binh-dam-ibigroup, line 62: Use `/*` for multi-line comments

lib/components/narrative/default/default-itinerary.js
```diff
@@ -52,11 +52,28 @@ const LegIconWrapper = styled.div`
...
+
+  width: 0;
+  height: 0;
```
binh-dam-ibigroup, line 65: Sort CSS.

lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -398,56 +419,66 @@ class MetroItinerary extends NarrativeItinerary {
...
-                      textOnly
-                    />
-                  ) : (
-                    firstTransitStop && (
-                      <FormattedMessage
-                        id="components.MetroUI.fromStop"
-                        values={{ stop: firstTransitStop }}
+                <ItineraryDetails
+                  aria-label={intl.formatMessage({
+                    id: 'components.RouteDetails.moreDetails'
```
binh-dam-ibigroup, line 432: This one should say "Itinerary details" as in the batch UI.

lib/components/narrative/metro/metro-itinerary.tsx
binh-dam-ibigroup: This is fine. Eventually we'll need to combine all the "visually hidden" CSS into a single place.

binh-dam-ibigroup (changes requested):
I think we need a "Search Results" `<h1>` (invisible) for both the "batch" and "metro" UIs to introduce the results. After that the rest will be good to go.

lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -52,7 +55,18 @@ const ItineraryWrapper = styled.div.attrs((props) => {
   padding: 0;
 `
 
+// invisible header rendered for screen readers and a11y technologies
+const InvisibleHeader = styled.h3`
+  width: 0;
+  height: 0;
+  overflow: hidden;
+  //place it in an unused grid cell so it doesn't add a row
+  grid-row: 2;
```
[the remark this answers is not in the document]
AdrianaCeric: Found a quick hack to sort CSS attributes in the entire file:
1. Install "PostCSS Sorting"

2. Add this to your `settings.json`

```
"postcssSorting.config": {
  "properties-order": "alphabetical"
}
```

3. `(cmd+shift+p)` and choose `PostCSS Sorting: Run`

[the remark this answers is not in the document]
AdrianaCeric: ^ This hack doesn't work for styled divs, however 😞

amy-corson-ibigroup:
> I think we need a "Search Results" `<h1>` (invisible) for both the "batch" and "metro" UIs to introduce the results.

@binh-dam-ibigroup Right now, both the batch and metro uis have h1 elements along the lines of "# of itineraries found" (Batch UI has a visible h1, Metro is invisible). Are you thinking these should be changed to just "search results"?

lib/components/narrative/default/default-itinerary.js
amy-corson-ibigroup: Good idea! Added in c388b25ddda47344657e8b72a354a43e1b69a9c2 thanks!!

PR #754, opened 2023-01-18
add a11y labeling to Trip Viewer

amy-corson-ibigroup:
- Adds trip description
- Adds aria labels for stops related to trip
- Re-orders trip viewer list items for screen readers


lib/components/viewers/trip-viewer.js
```diff
@@ -159,30 +187,33 @@ class TripViewer extends Component {
...
-                    <div className="strip-map-container">
-                      {highlightClass && <div className={highlightClass} />}
-                      <div className={stripMapLineClass} />
-                      <div className="strip-map-icon">
-                        <StyledIconWrapper>
-                          <Circle />
-                        </StyledIconWrapper>
-                      </div>
-                    </div>
+                  <Stop aria-label={stopLabel} key={i} title={stopLabel}>
```
miles-grant-ibigroup, line 190: I don't think we need the title as well!

lib/components/viewers/trip-viewer.js
```diff
@@ -38,6 +40,16 @@ const HeaderText = styled.h1`
...
 
+const TripDescription = styled.div`
+  border-left: 4px solid #13c1c1;
+  font-size: 14px;
+  margin-bottom: 1.5em;
+  padding-left: 1em;
+`
+const FlexWrapper = styled.div`
+  display: flex;
+  width: ${(props) => props.width && '80%'};
```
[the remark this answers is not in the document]
amy-corson-ibigroup: I'm not proud of this one

lib/components/viewers/trip-viewer.js
```diff
@@ -38,6 +40,16 @@ const HeaderText = styled.h1`
   margin: 2px 0 0 0;
 `
 
+const TripDescription = styled.div`
+  border-left: 4px solid #13c1c1;
```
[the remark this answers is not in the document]
amy-corson-ibigroup: Hmm I tried a different approach, without the border, tightening the text block and adding a subtle background color. I'm not sure it's quite there, but maybe it's in the right direction? Let me know!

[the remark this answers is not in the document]
miles-grant-ibigroup: This is it! I love it. My only 2 comments on it at this point are to match the color with the highlights on the blobs (I'd adjust the blob highlight color -- I like yours more)

and maybe adding
```css
border-radius: 5px;
box-shadow: 2px 2px 5px 1px rgba(0, 0, 0, 0.02);
```
to the box? What do you think

lib/components/viewers/trip-viewer.js
```diff
@@ -33,22 +33,23 @@ const Stop = styled.li`
...
 
 const TripDescription = styled.div`
-  border-left: 4px solid #13c1c1;
-  font-size: 14px;
-  margin-bottom: 1.5em;
-  padding-left: 1em;
+  font-size: 12px;
+  margin-bottom: 1em;
+  background-color: rgba(19, 193, 193, 0.2);
+  // border-left: 4px solid rgb(19, 193, 193);
```
miles-grant-ibigroup, line 47: We can probably fully remove this!

lib/components/viewers/trip-viewer.js
```diff
@@ -187,8 +195,8 @@ class TripViewer extends Component {
                 }
 
                 return (
-                  <Stop aria-label={stopLabel} key={i} title={stopLabel}>
-                    <FlexWrapper width>
+                  <Stop aria-label={stopLabel} key={i}>
+                    <FlexWrapper width="80%">
```
miles-grant-ibigroup, line 199: I think I'd almost prefer an inline style here rather than this extra prop, but will leave it to the second reviewer to decide!

miles-grant-ibigroup (approved):
Looks so good! Thanks for your work on this. Curious for your thoughts on the proposed css, otherwise approved pending 2 tiny nitpicks!

lib/components/viewers/trip-viewer.js
amy-corson-ibigroup: whoops

lib/components/viewers/trip-viewer.js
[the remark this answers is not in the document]
amy-corson-ibigroup: LOVE this! I added these, and a TINY bit of padding, and I think it's looking pretty good!

Changing the highlight color on the blobs to match the description is tricky because it's unfortunately barely visible. The description background is decoration, but the highlight is at least in part trying to communicate information? I split the difference and made the opacity `0.5` but it may just look inconsistent, so I'm very open to feedback!

lib/components/viewers/trip-viewer.js
```diff
@@ -106,6 +120,17 @@ class TripViewer extends Component {
                 )}
               </RouteName>
 
+              <TripDescription>
+                <FormattedMessage
+                  id="components.TripViewer.tripDescription"
+                  values={{
+                    boardAtStop: tripData.stops[viewedTrip.fromIndex].name,
+                    disembarkAtStop: tripData.stops[viewedTrip.toIndex].name,
```
binh-dam-ibigroup, line 128: Good idea!

You might be able to surround the contents with `<strong>` directly instead of passing a `<strong>` tag. 

Add a TODO to add links to the entry in the list of stops, so that when people navigate, they can see/hear the time they will reach the boarding and arriving stops, and also access the stop viewer links without having to go through all the stops not on the trip.

lib/components/viewers/trip-viewer.js
```diff
@@ -159,30 +196,33 @@ class TripViewer extends Component {
...
-                    <div className="strip-map-container">
-                      {highlightClass && <div className={highlightClass} />}
-                      <div className={stripMapLineClass} />
-                      <div className="strip-map-icon">
-                        <StyledIconWrapper>
-                          <Circle />
-                        </StyledIconWrapper>
-                      </div>
-                    </div>
+                  <Stop aria-label={stopLabel} key={i}>
```
binh-dam-ibigroup, line 199: `aria-label` does not seem to be picked up on a regular `<li>`, probably because it is not interactive. I think it is better to insert a `<span>` before or after the stop name that is not visible to sighted users (e.g. use zero-width/height and/or clip).

lib/components/viewers/trip-viewer.js
```diff
@@ -25,19 +25,33 @@ import ViewStopButton from './view-stop-button'
...
 const RouteName = styled.h2`
-  margin-bottom: 20px;
   font-size: inherit;
+  margin: 0 0 1em 0;
 `
 const HeaderText = styled.h1`
   margin: 2px 0 0 0;
 `
 
+const TripDescription = styled.div`
```
binh-dam-ibigroup, line 43: Or use `<Alert bsStyle="info">` (as long as we are using bootstrap)

lib/components/viewers/viewers.css
```diff
@@ -452,17 +452,18 @@
 
 .otp .trip-viewer .stop-name {
   font-size: 14px;
-  margin-left: 80px;
-  margin-right: 55px;
-  padding-top: 6px;
+  margin-left: 10px;
+  padding-top: 8px;
   white-space: nowrap;
   overflow: hidden;
```
binh-dam-ibigroup, line 458: While you are at it... sort CSS props.

binh-dam-ibigroup (changes requested):
The labels for "Trip begins/ends here" are not showing in ARIA dev tools for some reason, please fix that (see comment).

lib/components/viewers/trip-viewer.js
amy-corson-ibigroup: Agree it looks very similar to Bootstrap, are we trying to move away from Bootstrap long term?

miles-grant-ibigroup: Agreed I think we should avoid adding any more bootstrap

miles-grant-ibigroup: But perhaps making this `TripDescription` thing into a custom `Alert` component and putting in utils might be a good middle ground!

lib/components/viewers/trip-viewer.js
amy-corson-ibigroup: Smart, thank you!

lib/components/viewers/trip-viewer.js
amy-corson-ibigroup: added a component to utils in fb0e0f4b1006719b8ae4fbf1f1e3e3f80d95c2e7 !

i18n/en-US.yml
```diff
@@ -554,7 +554,7 @@ components:
     bicyclesAllowed: Allowed
     header: Trip Viewer
     listOfRouteStops: List of stops on this route
-    tripDescription: Board at <strong>{boardAtStop}</strong> and disembark at <strong>{disembarkAtStop}</strong>
+    tripDescription: Board at {boardAtStop} and disembark at {disembarkAtStop}
```
binh-dam-ibigroup, line 557: ❤️

lib/components/util/alert.tsx
```diff
@@ -0,0 +1,19 @@
...
+    background-color: rgba(19, 193, 193, 0.2);
+    border-radius: 5px;
+    box-shadow: 2px 2px 5px 1px rgba(0, 0, 0, 0.02);
+    font-size: 12px;
+    margin-bottom: 1em;
+    padding: 0.75em 1em;
+  `
+
+  return <TripDescription>{children}</TripDescription>
+}
```
binh-dam-ibigroup, line 19:
```suggestion
const Alert = styled.div`
  background-color: rgba(19, 193, 193, 0.2);
  border-radius: 5px;
  box-shadow: 2px 2px 5px 1px rgba(0, 0, 0, 0.02);
  font-size: 12px;
  margin-bottom: 1em;
  padding: 0.75em 1em;
`

export default Alert
```

lib/components/viewers/trip-viewer.js
```diff
@@ -39,19 +40,15 @@ const RouteName = styled.h2`
...
-  padding: 0.75em 1em;
-`
 const FlexWrapper = styled.div`
   display: flex;
   width: ${(props) => props.width};
 `
+const InvisibleAccessibilityLabel = styled.span`
+  height: 0;
+  overflow: 0;
+  width: 0;
```
binh-dam-ibigroup, line 50: For spans, you need `display: block` or `display: inline-block` for the text to be clipped.

lib/components/viewers/trip-viewer.js
```diff
@@ -175,14 +181,10 @@ class TripViewer extends Component {
...
-                  stopLabel = intl.formatMessage({
-                    id: 'components.tripviewer.startOfTrip'
-                  })
+                  stopLabel = 'components.TripViewer.startOfTrip'
                 }
                 if (viewedTrip.toIndex === i) {
-                  stopLabel = intl.formatMessage({
-                    id: 'components.tripviewer.endOfTrip'
-                  })
+                  stopLabel = 'components.TripViewer.endOfTrip'
```
binh-dam-ibigroup, line 187: You want to keep the id value together with `<FormattedMessage>` (or with intl.formatMessage), so that the formatjs CLI tool can pick up the string to translate. (In other words, don't use dynamic ids - the exception is for using ids defined in config.)

lib/components/viewers/trip-viewer.js
binh-dam-ibigroup: I favor inlining the `80%` right into the CSS.

lib/components/viewers/trip-viewer.js
```diff
@@ -196,10 +198,15 @@ class TripViewer extends Component {
...
 
                 return (
-                  <Stop aria-label={stopLabel} key={i}>
+                  <Stop key={i}>
                     <FlexWrapper width="80%">
                       <FlexWrapper>
                         {/* the departure time */}
+                        {stopLabel && (
+                          <InvisibleAccessibilityLabel>
+                            <FormattedMessage id={stopLabel} />
```
binh-dam-ibigroup, line 207: Keep `<FormattedMessage>` and the value for `id` together.

binh-dam-ibigroup (changes requested):
Some more cleanup needed, please.

lib/components/viewers/trip-viewer.js
amy-corson-ibigroup: The overflow property is wrong too, what a mess, thank you!!

lib/components/viewers/trip-viewer.js
amy-corson-ibigroup: Good to know! Thanks! Fixed in 67caf50f05cfa00f575984c8559b788eb79f36e8

lib/components/util/alert.tsx
```diff
@@ -1,19 +1,11 @@
...
-}
+const Alert = styled.div`
+  background-color: rgba(19, 193, 193, 0.2);
+  border-radius: 5px;
+  box-shadow: 2px 2px 5px 1px rgba(0, 0, 0, 0.02);
+  font-size: 12px;
+  margin-bottom: 1em;
+  padding: 0.75em 1em;
+`
+export default Alert
```
binh-dam-ibigroup, line 11: Rename to `alert.ts` now that this file no longer has markup.

binh-dam-ibigroup (approved):
Thanks for the changes!

PR #750, opened 2023-01-12
Improve User Feedback in Accounts Section

miles-grant-ibigroup:
This PR resolves some UX no-nos that were present in the accounts section. 

Loading spinners show up in the icons when things are loading and the custom locations save panel works again!

PR #755, opened 2023-01-18
(Fix): Work and Home location saving bug

AdrianaCeric:
Adding onto @miles-grant-ibigroup's [PR](https://github.com/opentripplanner/otp-react-redux/pull/750) with help from Binh.

Fixes problem where Work or Home lack of `name` doesn't pass Formik name validation checks before saving. Rather than removing name validation method, assign Home or Work location a name with its type.

UI:
image

Don't mind fork PR, I don't have permissions to push/commit yet 👍


miles-grant-ibigroup (approved):
Works great! Thank you for the clean fix. Will merge this into my PR and then revert my change.

PR #761, opened 2023-01-23
fix(route-block): repair slanted rendering

miles-grant-ibigroup:
At some point we introduced a regression breaking the slanted metro UI rendering. Luckily the fix appears to be quite simple, although I'm worried some regressions may have been introduced based on how simple the fix is.


AdrianaCeric (approved):

daniel-heppner-ibigroup (approved):
Seems to be working

PR #762, opened 2023-01-23
Stop Viewer Screen Reader Tweaks

miles-grant-ibigroup:
This PR tweaks the stop viewer to be more compatible with screen readers. I'm not sure if it's quite there yet, looking for feedback. It's more structured and usable, but I'm not sure if it follows all best practices.


__tests__/components/viewers/__snapshots__/stop-viewer.js.snap
```diff
@@ -606,56 +607,46 @@ exports[`components > viewers > stop viewer should render countdown times after
...
-                                className=""
+                              <LocationIcon
                                 size="0.9em"
-                                title="From Location Icon"
+                                type="from"
                               >
-                                <DotCircle
-                                  className="styled__FromIcon-n5xcvc-0 bnFcGf"
+                                <styled__FromIcon
+                                  className=""
```
AdrianaCeric, line 639: Why the empty `className`s for `styled_FromIcon`/`styled_ToIcon`?

__tests__/components/viewers/__snapshots__/stop-viewer.js.snap
```diff
@@ -3199,56 +3209,46 @@ exports[`components > viewers > stop viewer should render countdown times for st
                       </IconWithText>
                     </button>
                   </div>
-                  <FromToLocationPicker
-                    label={true}
-                    onFromClick={[Function]}
-                    onToClick={[Function]}
+                  <span
+                    role="group"
```
AdrianaCeric, line 3213: Why wrap this with a span? Could we add `role="group"` to `FromToLocationPicker` itself?

__tests__/components/viewers/__snapshots__/stop-viewer.js.snap
miles-grant-ibigroup: I'm not sure! These files are auto-generated so who knows why it decided to do that haha

__tests__/components/viewers/__snapshots__/stop-viewer.js.snap
miles-grant-ibigroup: If we did that then we'd either have to make all `FromToLocationPicker`s groups or add a new prop to turn it on and off which I'm not crazy about. This isn't perfect but I feel it's a happy middle ground

lib/components/viewers/amenities-panel.tsx
```diff
@@ -133,7 +135,7 @@ class AmenitiesPanel extends Component<Props, State> {
           <li className="related-item" key={key}>
             <div className="item-label">
               <div className="overflow-ellipsis" title={label}>
-                {company.icon}
+                <span aria-hidden>{company.icon}</span>
```
AdrianaCeric, line 138: This is a good addition. I would also recommend adding it to the micro mobility icon (line 228)

lib/components/util/dropdown.tsx
```diff
@@ -161,7 +161,7 @@ const Dropdown = ({
       id={`${id}-wrapper`}
       onKeyDown={_handleKeyDown}
       ref={containerRef}
-      role="group"
+      role="listitem"
```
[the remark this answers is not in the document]
miles-grant-ibigroup: If you look at the a11y structure you'll see the container is inside a list of containers!

lib/components/util/dropdown.tsx
```diff
@@ -161,7 +161,7 @@ const Dropdown = ({
       id={`${id}-wrapper`}
```
AdrianaCeric, line 161: I learned about `aria-live` while looking over the accessibility guidelines. Maybe adding an `aria-live="polite"` would help the screen reader get notified when the language changes languages in the dropdown:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions

lib/components/viewers/stop-viewer.js
```diff
@@ -190,12 +190,12 @@ class StopViewer extends Component {
           {stopData ? (
             <>
               {agencyCount <= 1 && <OperatorLogo operator={stationOperator} />}
-              <span style={{ paddingLeft: '0.5ch' }}>{stopData.name}</span>
+              <h1 style={{ paddingLeft: '0.5ch' }}>{stopData.name}</h1>
```
AdrianaCeric, line 193: :+1:

lib/components/viewers/live-stop-times.tsx
```diff
@@ -197,6 +197,9 @@ class LiveStopTimes extends Component<Props, State> {
             <FormattedMessage id="components.LiveStopTimes.autoRefresh" />
           </label>
           <button
+            // Functionality is provided by auto-refresh, this only adds confusion
+            // in a screen reader context
```
AdrianaCeric, line 201: Do you think `aria-live` could help here?

AdrianaCeric (changes requested):
Overall the additions are really great. I left a couple questions and comments!

lib/components/viewers/amenities-panel.tsx
miles-grant-ibigroup: Good idea thanks!

lib/components/util/dropdown.tsx
miles-grant-ibigroup: I'm not sure if this is what that is meant for? Although I might be confused. I see this more as being about an autocomplete suggestions dropdown rather than this which really could be considered more of a submenu? Not sure what others think

miles-grant-ibigroup:
Thanks for the good catches! Not sure about the `aria-live` stuff, I'll defer to the second reviewer

AdrianaCeric (approved):

lib/components/viewers/related-stops-panel.js
```diff
@@ -100,7 +100,7 @@ class RelatedStopsPanel extends Component {
                     }
                   />
                 </div>
-                <div className="pattern-list">
+                <div className="pattern-list" role="list">
```
amy-corson-ibigroup, line 103: Looks like using `<ul>` [may be recommended](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles#:~:text=available%20and%20supported.-,Avoid%20using%3A,-application) over using `role="list"`

lib/components/viewers/stop-viewer.js
```diff
@@ -303,11 +307,13 @@ class StopViewer extends Component {
             />
           )}
         </div>
-        <FromToLocationPicker
-          label
-          onFromClick={this._onClickPlanFrom}
-          onToClick={this._onClickPlanTo}
-        />
+        <span role="group">
```
amy-corson-ibigroup, line 310: Thoughts on incorporating this group role into the `FromToLocationPicker` itself?

lib/components/util/dropdown.tsx
amy-corson-ibigroup, line 164: Using role='listitem' is also, unfortunately, [not ideal](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles#:~:text=available%20and%20supported.-,Avoid%20using%3A,-application)

amy-corson-ibigroup (changes requested):
Looking really good so far, thank you so much for this!! A few best practice things below.

I'm also noticing there's no way for screenreaders to see whether the times listed are scheduled or based on realtime data, and that feels like a pretty key purpose of the stop viewer? I'm wondering if you'd be open to adding invisible labels that announce those statuses, or removing the `aria-hidden` from the scheduled/realtime icons.

lib/components/util/dropdown.tsx
miles-grant-ibigroup: I think the checkmark next to English is the selected indication right?

lib/components/viewers/stop-viewer.js
miles-grant-ibigroup: I was talking about this with Adriana too -- we should probably do it if both of you came up with the same idea independently.

lib/components/viewers/related-stops-panel.js
miles-grant-ibigroup: Good catch thanks!

lib/components/util/dropdown.tsx
miles-grant-ibigroup: It's an `li` anyway! Thanks for the catch

lib/components/viewers/stop-viewer.js
AdrianaCeric: Amy and I have the same mind haha. But yeah up to you

miles-grant-ibigroup:
Yeah the realtime readout is a bit tricky since you risk the word realtime being read out every line which could be a bit annoying right?

lib/components/util/dropdown.tsx
amy-corson-ibigroup: Yeah, I just tried this on screenreader and it reads the selected language as "clickable, selected"!

amy-corson-ibigroup: Tried it yesterday and for some reason i didn't catch the "selected" which is why I brought up the potential invisible labels. I think this is good the way it is!

amy-corson-ibigroup (approved):
Looks good, nothing blocking! Thanks for the changes.

I see what you mean about the "realtime" indicators potentially being a lot, but it feels like info we're giving to sighted users and not to screenreader users. And I think knowing if the bus is actually on the way vs just scheduled is a key distinction to make. Regardless, happy to make it a bigger decision with the group, it doesn't have to be dependent on this PR!

PR #763, opened 2023-01-23
refactor(ViewSwitcher): Clean up types

binh-dam-ibigroup:
PR to fix types on #750, plus a status tweak.


miles-grant-ibigroup (approved):
Makes me a bit nervous but looks like it works!

PR #764, opened 2023-01-24
Route Mode Override

AdrianaCeric:
Adds support for overriding a route mode in the config. Helpful if an agency's GTFS doesn't correctly define route modes


lib/reducers/create-otp-reducer.js
```diff
@@ -306,6 +307,23 @@ function createOtpReducer(config) {
...
+        response.plan.itineraries = response.plan.itineraries.map(
+          (itinerary) => {
+            itinerary.legs = itinerary.legs.map((leg) => {
+              if (leg.routeId) {
+                leg.mode = checkForRouteModeOverride(
+                  {
+                    id: leg.routeId,
+                    mode: leg.mode
+                  },
+                  state?.config?.routeModeOverrides
```
binh-dam-ibigroup, line 319: `state` is defined, so the `?` operator is not needed after `state`.

binh-dam-ibigroup (approved):
LGTM

lib/reducers/create-otp-reducer.js
miles-grant-ibigroup: Good catch thanks!

philip-cline (approved):

PR #765, opened 2023-01-24
App Menu Selected Language a11y

binh-dam-ibigroup:
This PR extracts shared language selector logic and adds `aria-selected`, `lang`, and `role` attributes to the language selector under the app menu.


lib/components/app/app-menu.tsx
```diff
@@ -50,7 +51,10 @@ type menuItem = {
   iconType: string | JSX.Element
   iconUrl?: string
   id: string
+  isRadio?: boolean
```
miles-grant-ibigroup, line 54: Can't `isRadio` be inferred by if `isSelected` is true/false? If it's undefined it's not radio, right?

miles-grant-ibigroup (changes requested):
Good fix, thanks. One quick question and then I'm happy to merge into my PR

lib/components/app/app-menu.tsx
binh-dam-ibigroup: No, because all of the language choices need to have the `option` role, so that the user knows what the different choices are (see lines 178-179).

miles-grant-ibigroup: Yes but can't we use the `true/false` of `isSelected` to detect if `isRadio` is true? basically my question is is `isSelected` ever explicitly true or false (instead of undefined) if `isRadio` is false?

binh-dam-ibigroup: `isSelected = true` implies `isRadio` indeed, but that does not apply other way around. The unselected languages still need to be `isRadio` so that `role="option"` is applied to them too.

miles-grant-ibigroup: Yes but won't `isSelected=false` always imply `isRadio`? a non-radio option would have `isSelected=undefined`, right?

binh-dam-ibigroup: That is true indeed.

binh-dam-ibigroup: Refactored in https://github.com/opentripplanner/otp-react-redux/pull/765/commits/1ec68f3a43f1467cc89ba1bfcff501b60931de6c.

PR #767, opened 2023-01-31
Stop and Route Viewer a11y Tweaks

amy-corson-ibigroup:
- Stops route viewer "towards" select from clipping into the edge of the sidebar
- Adds focus states and improves focus order for stop viewer controls
- Makes timetable container focusable (and therefore scrollable for keyboard navigation)
- Changes `StopLink` from `a` to `button` (and therefore focusable)

Follow up PRs necessary for:
- OTP-UI keyboard focusable stops on map
- OTP-RR stop-viewer updated departures layout

### DIFF
--- a/lib/components/viewers/stop-viewer.js
+++ b/lib/components/viewers/stop-viewer.js
@@ -404,6 +404,8 @@ class StopViewer extends Component {
         {stopData && (
           <div className="stop-viewer-body">
             {this._renderControls()}
+            {/* scrollable list of scheduled stops requires tabIndex 
+            for keyboard navigation */}
             <Scrollable tabIndex={0}>{contents}</Scrollable>
             {/* Future: add stop details */}
           </div>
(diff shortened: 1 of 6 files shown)


lib/components/viewers/stop-viewer.js
```diff
@@ -404,7 +404,7 @@ class StopViewer extends Component {
         {stopData && (
           <div className="stop-viewer-body">
             {this._renderControls()}
-            <Scrollable>{contents}</Scrollable>
+            <Scrollable tabIndex={0}>{contents}</Scrollable>
```
binh-dam-ibigroup, line 407: Add a comment for why keyboard access should be handled here.

binh-dam-ibigroup (approved):
Looks good

miles-grant-ibigroup (approved):
Works great thanks for all your changes!

PR #772, opened 2023-02-03
Adjust Itinerary View Scroll Behavior

miles-grant-ibigroup:
At the moment, the itineraries scroll in their own box while the batch planner stays static. This PR modifies this so the _entire column_ scrolls. This can potentially be changed so that the entire column scrolls only when the container is small (due to something like high zoom).

As part of this work the layout of the batch routing panel is modified. The previous one didn't allow itself to be scrolled. 

The new behavior only kicks in at high zoom levels.


lib/components/form/form.css
```diff
@@ -414,3 +414,26 @@
 .otp .date-time-modal .date-time-selector.with-desktop-touchscreen input::-webkit-datetime-edit {
   margin-inline-end: 8px;
 }
+
+.batch-routing-panel-location-fields {
+  display: grid;
+  grid-template-rows: 1fr 1fr;
+  grid-template-columns: 1fr 40px 10px;
```
binh-dam-ibigroup, line 421: The grid stuff that follows is not needed if you do for this CSS class:
```suggestion
  display: block;
  position: relative;
```

Note that the switch button container's existing CSS should handle the positioning correctly already with this suggestion.

lib/components/app/batch-routing-panel.js
```diff
@@ -68,8 +71,7 @@ class BatchRoutingPanel extends Component {
         <div
           className="desktop-narrative-container"
           style={{
-            flexGrow: 1,
-            overflowY: 'hidden'
+            flexGrow: 1
           }}
```
binh-dam-ibigroup, line 75: Remove all of it. (allowing scrolling removes any bounds that would apply for a flex layout)

binh-dam-ibigroup: Also, around Line 33, remove `display: 'flex'` and `flexDirection` from the `ViewerContainer` style.

binh-dam-ibigroup (changes requested):
The layout can be simplified (the grid for location fields is a bit overkill) and styles cleaned up.

lib/components/form/form.css
miles-grant-ibigroup: Screenshot 2023-02-08 at 1 21 18 PM
Not sure if I missed something but this doesn't exactly work

lib/components/app/batch-routing-panel.js
miles-grant-ibigroup: Removing this causes the background of the pane to not extend to the full window height.

lib/components/app/responsive-webapp.js
```diff
@@ -212,7 +212,7 @@ class ResponsiveWebapp extends Component {
             <Col className="sidebar" md={4} sm={6}>
               {/* Note: the main tag provides a way for users of screen readers to skip to the
                   primary page content (tabindex = -1 needed for programmatic navigation skip). */}
-              <main tabIndex={-1}>
+              <main style={{ overflow: 'scroll' }} tabIndex={-1}>
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Good call thank you!

lib/components/form/form.css
binh-dam-ibigroup: For the switch button container, apply `position: absolute; right: ..., top: ...` and for its parent element, apply `position: relative`.

miles-grant-ibigroup: I tried this and couldn't get it to work, since the container moves around so much (and forget about mobile). Maybe you'll have more luck

binh-dam-ibigroup:
Note that the change to scroll the location fields creates this side effect where the location field popup list gets clipped by the `<main>` element (see screenshot).

image

miles-grant-ibigroup:
@binh-dam-ibigroup rules are refactored to only apply at high zoom

lib/components/app/app.css
```diff
@@ -102,3 +102,23 @@
...
 }
+
+/** These changes kick in when the map is hidden. These rules ensure that the entire "sidebar"
+    (which at this point fills the entire screen) scrolls as a single unit.
+    These rules can only apply at this high zoom level because the location field dropdown is
+    otherwise "stuck" inside the sidebar and can't overlap the map. */
+@media (max-width: 768px) {
+  .container-fluid > .navbar-header {
+    margin-bottom: -7px;
+  }
```
binh-dam-ibigroup, line 113: Revert this change, it causes the mobile title to be not vertically centered.

binh-dam-ibigroup (approved):
One thing to revert on the mobile title bar, please, but the scrolling threshold is clever.

daniel-heppner-ibigroup (approved):

PR #774, opened 2023-02-03
Fix notification settings

binh-dam-ibigroup:
This PR fixes:
- undesired scrolling up the page when changing a user setting
- displaying focus cues on the notification settings when navigating using the keyboard - this is achieved by separating out (rewriting) the options group components.
- phone number editor crashes when typing letters

Behind the scenes, the phone number editor is extracted to its own TypeScript file,
and the notifications preferences pane is converted to TypeScript.


lib/components/user/user-account-screen.js
```diff
@@ -122,34 +127,26 @@ class UserAccountScreen extends Component {
             // and to its own blur/change/submit event handlers that automate the state.
             // We pass the Formik props below to the components rendered so that individual controls
             // can be wired to be managed by Formik.
-            (formikProps) => {
-              const DisplayComponent = isCreating
-                ? lazy(() => import('./new-account-wizard'))
```
miles-grant-ibigroup, line 127: Why was this change made? This increases bundle size by about 0.2mb, which is significant. Would it be possible to maybe lazy-load the entire user-account-screen instead?

lib/components/user/phone-number-editor.tsx
```diff
@@ -0,0 +1,332 @@
...
+    const { validationCode } = values
+
+    if (!errors.validationCode) {
+      onSubmitCode(validationCode)
+    }
+  }
+
+  _isPhoneNumberPending = () => {
+    const { initialPhoneNumber, initialPhoneNumberVerified } = this.props
+    return !isBlank(initialPhoneNumber) && !initialPhoneNumberVerified
```
miles-grant-ibigroup, line 172: I'm noticing that the component briefly flashes validated before switching to pending. I'm not sure if this was introduced by these changes, but it seems to be related to this check

miles-grant-ibigroup (changes requested):
Works well and is much much cleaner thanks for this work! Just one worry about bundle size and one nit

lib/components/user/user-account-screen.js
binh-dam-ibigroup: Maybe. This lazy loading was causing an entire re-render of the account settings page when you edited a value.

binh-dam-ibigroup: `UserAccountScreen` is already being lazy-loaded from `webapp-routes`, so something else is going on.

binh-dam-ibigroup: `UserAccountScreen`, `NewAccountWizard` and `ExistingAccountDisplay` appear to be in the same bundle `6.bundle.js`, size 33kb.

lib/components/user/phone-number-editor.tsx
binh-dam-ibigroup: Correct, that is an old issue.

binh-dam-ibigroup: This is fixed in https://github.com/opentripplanner/otp-react-redux/pull/774/commits/e2936f327849821f02bbe10751910ac0808a1483.

lib/components/user/user-account-screen.js
miles-grant-ibigroup: That's strange... might have been a caching issue on my end. Glad it's still being lazy loaded! Thanks for looking into it

miles-grant-ibigroup (approved):
Thanks for all the fixes! Works great

amy-corson-ibigroup (changes requested):
Biggest issue I'm seeing is that 1) entering a non-number into the phone number input or 2) emptying the value of the phone number input both crash the site. This looks like it's a live issue (not something that's being added in this PR) but seems like this would be the correct place to fix it?

binh-dam-ibigroup:
> 939c48f

Good catch @amy-corson-ibigroup, thanks, I fixed that with https://github.com/opentripplanner/otp-react-redux/pull/774/commits/939c48f761acfaf951ded20a60ddf473e067380c.

amy-corson-ibigroup (approved):
Great fix! I wasn't able to find any other issues. Thanks!

PR #775, opened 2023-02-03
No More Endless Accounts Spinning

miles-grant-ibigroup:
This PR resolves an issue where an invalid form state would cause the loading spinner to never disappear. Fix is to reset the state when form data updates. Forms that were not updated by this PR weren't affected by the bug.

Blocked by https://github.com/opentripplanner/otp-react-redux/pull/774


binh-dam-ibigroup (approved):
LGTM

amy-corson-ibigroup (approved):
There is still a spinner getting rendered when nothing's loading and I think looks a little confusing, but this does fix the more immediate issue so I'm happy to table for now! Thanks so much for fixing this


ezgif com-video-to-gif

miles-grant-ibigroup:
How can I reproduce this @amy-corson-ibigroup? This looks quite bad

amy-corson-ibigroup:
In Chrome, I'm getting it when i'm signed in, save a trip, and then click the Save Preferences button without adding a trip name. Let me know if you can't reproduce!

miles-grant-ibigroup:
I'm unable to reproduce after merging in dev! I'm going to merge this for now and if we continue to see the issue we can always open an issue!

PR #779, opened 2023-02-06
Repair Various a11y regressions

miles-grant-ibigroup:
All the recent a11y work has caused a few small styling regressions and crashes. This PR resolves some of them.

### DIFF
--- a/lib/components/narrative/metro/metro-itinerary.tsx
+++ b/lib/components/narrative/metro/metro-itinerary.tsx
@@ -1,14 +1,11 @@
 import { AccessibilityRating } from '@opentripplanner/itinerary-body'
 import { connect } from 'react-redux'
 import {
-  FormattedList,
   FormattedMessage,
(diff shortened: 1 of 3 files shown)


lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -23,6 +20,7 @@ import {
   itineraryHasAccessibilityScores
 } from '../../../util/accessibility-routing'
 import { getActiveSearch, getFare } from '../../../util/state'
+import { getFormattedMode } from '../../../util/i18n'
```
amy-corson-ibigroup, line 23: actually, I think this is only used by `modeStrings` on line 324, which is also not getting used. You may be able to take them both out!

amy-corson-ibigroup (approved):
Thank you for fixing these!

daniel-heppner-ibigroup (approved):
LGTM

PR #780, opened 2023-02-07
Stop Viewer Departure Layout

amy-corson-ibigroup:
Separates upcoming stop times by day and tweaks flexbox/grid elements to avoid clipping and overflow at smaller screen sizes.

Definitely looking for feedback around how the times are being sorted by day (it feels like there may be a more efficient way to render each day!)

Before                                 |  After
:-------------------------:|:-------------------------:
image  |  
image

### DIFF
--- a/lib/components/viewers/live-stop-times.tsx
+++ b/lib/components/viewers/live-stop-times.tsx
@@ -231,7 +231,7 @@ class LiveStopTimes extends Component<Props, State> {
           )}
           {tomorrowTimes.length > 0 && (
             <div className="list-container">
-              <p>
+              <p className="day-label">
                 <FormattedDayOfWeek
                   // 'iiii' returns the long ISO day of the week (independent of browser locale).
                   // See https://date-fns.org/v2.28.0/docs/format
@@ -265,7 +265,7 @@ class LiveStopTimes extends Component<Props, State> {
           )}
           {dayAfterTomorrowTimes.length > 0 && (
             <div className="list-container">
-              <p>
+              <p className="day-label">
                 <FormattedDayOfWeek
                   day={format(dayAfterTomorrow, 'iiii', {
                     timeZone: homeTimezone
--- a/lib/components/viewers/live-stop-times.tsx
+++ b/lib/components/viewers/live-stop-times.tsx
@@ -1,7 +1,12 @@
+import { format, utcToZonedTime } from 'date-fns-tz'
 import { FormattedMessage, FormattedTime } from 'react-intl'
 import { Redo } from '@styled-icons/fa-solid/Redo'
 import { TransitOperator } from '@opentripplanner/types'
+import addDays from 'date-fns/addDays'
 import coreUtils from '@opentripplanner/core-utils'
+
+import FormattedDayOfWeek from '../util/formatted-day-of-week'
+import isSameDay from 'date-fns/isSameDay'
 import React, { Component } from 'react'
 
 import {
@@ -11,6 +16,7 @@ import {
 } from '../../util/viewer'
 import { IconWithText } from '../util/styledIcon'
 import SpanWithSpace from '../util/span-with-space'
+import type { Time } from '../util/types'
 
 import AmenitiesPanel from './amenities-panel'
 import PatternRow from './pattern-row'
@@ -23,10 +29,12 @@ const defaultState = {
   timer: null
 }
 
+const ONE_DAY_IN_SECONDS = 86400
+
 type Props = {
   autoRefreshStopTimes: boolean
   findStopTimesForStop: ({ stopId }: { stopId: string }) => void
-  homeTimezone?: string
+  homeTimezone: string
   nearbyStops: any // TODO: shared types
   setHoveredStop: (stopId: string) => void
   showNearbyStops: boolean
@@ -151,35 +159,143 @@ class LiveStopTimes extends Component<Props, State> {
       return patternHeadsignComparator(patternA, patternB)
     }
 
+    const routeTimes = Object.values(stopTimesByPattern)
+      .sort(patternComparator)
+      .filter(({ pattern, route }) =>
+        routeIsValid(route, getRouteIdForPattern(pattern))
+      )
+
+    // Determine if arrival occurs on different day, making sure to account for
+    // any extra days added to the service day if it arrives after midnight. Note:
+    // this can handle the rare (and non-existent?) case where an arrival occurs
+    // 48:00 hours (or more) from the start of the service day.
+
+    const now = utcToZonedTime(Date.now(), homeTimezone)
+    const tomorrow = addDays(now, 1)
+    const dayAfterTomorrow = addDays(now, 2)
+
+    const findTimes = (times: any, date: any): any[] => {
+      return times
+        .map((route: any) => {
+          return {
+            ...route,
+            times: route.times.filter((time: Time) => {
+              const departureTime = time.realtimeDeparture
+              const serviceDay = utcToZonedTime(
+                new Date(time.serviceDay * 1000),
+                homeTimezone
+              )
+              const departureTimeRemainder = departureTime % ONE_DAY_IN_SECONDS
+              const daysAfterServiceDay =
+                (departureTime - departureTimeRemainder) / ONE_DAY_IN_SECONDS
+              const departureDay = addDays(serviceDay, daysAfterServiceDay)
+              return isSameDay(date, departureDay)
+            })
+          }
+        })
+        .filter((route: any) => route.times.length > 0)
+    }
+
+    const todayTimes = findTimes(routeTimes, now)
+    const tomorrowTimes = findTimes(routeTimes, tomorrow)
+    const dayAfterTomorrowTimes = findTimes(routeTimes, dayAfterTomorrow)
+
     return (
       <>
-        <ul className="route-row-container">
-          {Object.values(stopTimesByPattern)
-            .sort(patternComparator)
-            .filter(({ pattern, route }) =>
-              routeIsValid(route, getRouteIdForPattern(pattern))
-            )
-            .map(({ id, pattern, route, times }) => {
-              // Only add pattern if route info is returned by OTP.
-              return (
-                <PatternRow
-                  homeTimezone={homeTimezone}
-                  key={id}
-                  pattern={pattern}
-                  route={{
-                    ...route,
-                    operator: transitOperators.find(
-                      (o: TransitOperator) => o.agencyId === route.agencyId
-                    )
-                  }}
-                  showOperatorLogo={showOperatorLogo}
-                  stopTimes={times}
-                  stopViewerArriving={stopViewerArriving}
-                  stopViewerConfig={stopViewerConfig}
+        <div>
+          {todayTimes.length > 0 && (
+            <div className="list-container">
+              <ul className="route-row-container">
+                {todayTimes.map(({ id, pattern, route, times }) => {
+                  // Only add pattern if route info is returned by OTP.
+                  return (
+                    <PatternRow
+                      homeTimezone={homeTimezone}
+                      key={id}
+                      pattern={pattern}
+                      route={{
+                        ...route,
+                        operator: transitOperators.find(
+                          (o: TransitOperator) => o.agencyId === route.agencyId
+                        )
+                      }}
+                      showOperatorLogo={showOperatorLogo}
+                      stopTimes={times}
+                      stopViewerArriving={stopViewerArriving}
+                      stopViewerConfig={stopViewerConfig}
+                    />
+                  )
+                })}
+              </ul>
+            </div>
+          )}
+          {tomorrowTimes.length > 0 && (
+            <div className="list-container">
+              <p>
+                <FormattedDayOfWeek
+                  // 'iiii' returns the long ISO day of the week (independent of browser locale).
+                  // See https://date-fns.org/v2.28.0/docs/format
+                  day={format(tomorrow, 'iiii', {
+                    timeZone: homeTimezone
+                  }).toLowerCase()}
                 />
-              )
-            })}
-        </ul>
+              </p>
+              <ul className="route-row-container">
+                {tomorrowTimes.map(({ id, pattern, route, times }) => {
+                  return (
+                    <PatternRow
+                      homeTimezone={homeTimezone}
+                      key={id}
+                      pattern={pattern}
+                      route={{
+                        ...route,
+                        operator: transitOperators.find(
+                          (o: TransitOperator) => o.agencyId === route.agencyId
+                        )
+                      }}
+                      showOperatorLogo={showOperatorLogo}
+                      stopTimes={times}
+                      stopViewerArriving={stopViewerArriving}
+                      stopViewerConfig={stopViewerConfig}
+                    />
+                  )
+                })}
+              </ul>
+            </div>
+          )}
+          {dayAfterTomorrowTimes.length > 0 && (
+            <div className="list-container">
+              <p>
+                <FormattedDayOfWeek
+                  day={format(dayAfterTomorrow, 'iiii', {
+                    timeZone: homeTimezone
+                  }).toLowerCase()}
+                />
+              </p>
+              <ul className="route-row-container">
+                {dayAfterTomorrowTimes.map(({ id, pattern, route, times }) => {
+                  return (
+                    <PatternRow
+                      homeTimezone={homeTimezone}
+                      key={id}
+                      pattern={pattern}
+                      route={{
+                        ...route,
+                        operator: transitOperators.find(
+                          (o: TransitOperator) => o.agencyId === route.agencyId
+                        )
+                      }}
+                      showOperatorLogo={showOperatorLogo}
+                      stopTimes={times}
+                      stopViewerArriving={stopViewerArriving}
+                      stopViewerConfig={stopViewerConfig}
+                    />
+                  )
+                })}
+              </ul>
+            </div>
+          )}
+        </div>
 
         {/* Auto update controls for realtime arrivals */}
         <div style={{ marginTop: '20px' }}>
--- a/lib/components/viewers/pattern-row.tsx
+++ b/lib/components/viewers/pattern-row.tsx
@@ -73,12 +73,21 @@ class PatternRow extends Component<Props, State> {
 
     const routeName = route.shortName ? route.shortName : route.longName
     const routeColor = getRouteColorBasedOnSettings(route.operator, route)
+    console.log(stopTimes)
+    let style
+    if (routeName && routeName?.length >= 4 && routeName?.length <= 6) {
+      style = { fontSize: '20px' }
+    } else if (routeName && routeName?.length > 7) {
+      style = { fontSize: '16px' }
+    } else {
+      style = { fontSize: '32px' }
+    }
 
     return (
       <li className="route-row">
         {/* header row */}
         <div
-          className="header"
+          className="header stop-view"
           style={{
             backgroundColor: routeColor,
             color: getMostReadableTextColor(routeColor, route?.textColor)
@@ -86,11 +95,7 @@ class PatternRow extends Component<Props, State> {
         >
           {/* route name */}
           <div className="route-name">
-            <strong
-              style={
-                routeName && routeName?.length >= 4 ? { fontSize: '50%' } : {}
-              }
-            >
+            <strong style={style}>
               <div
                 style={{
                   alignContent: 'center',
@@ -98,6 +103,7 @@ class PatternRow extends Component<Props, State> {
                   justifyContent: 'center',
                   whiteSpace: 'nowrap'
                 }}
+                title={routeName}
               >
                 {showOperatorLogo && (
                   <OperatorLogo operator={route?.operator} />
@@ -111,7 +117,7 @@ class PatternRow extends Component<Props, State> {
                 />
               </div>
             </strong>
-            <span>{pattern.headsign}</span>
+            <span title={pattern.headsign}>{pattern.headsign}</span>
           </div>
           {/* next departure preview */}
           {hasStopTimes && (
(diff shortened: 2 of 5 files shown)


lib/components/viewers/live-stop-times.tsx
```diff
@@ -151,35 +159,143 @@ class LiveStopTimes extends Component<Props, State> {
...
-                    ...route,
-                    operator: transitOperators.find(
-                      (o: TransitOperator) => o.agencyId === route.agencyId
-                    )
-                  }}
-                  showOperatorLogo={showOperatorLogo}
-                  stopTimes={times}
-                  stopViewerArriving={stopViewerArriving}
-                  stopViewerConfig={stopViewerConfig}
+        <div>
```
miles-grant-ibigroup, line 205: Do we need this outermost div?

lib/components/viewers/pattern-row.tsx
```diff
@@ -73,31 +73,37 @@ class PatternRow extends Component<Props, State> {
 
     const routeName = route.shortName ? route.shortName : route.longName
     const routeColor = getRouteColorBasedOnSettings(route.operator, route)
+    console.log(stopTimes)
```
miles-grant-ibigroup, line 76: Whoops I think this slipped in!

lib/components/viewers/pattern-row.tsx
```diff
@@ -73,31 +73,37 @@ class PatternRow extends Component<Props, State> {
...
+    let style
+    if (routeName && routeName?.length >= 4 && routeName?.length <= 6) {
+      style = { fontSize: '20px' }
```
miles-grant-ibigroup, line 79: Instead of setting the entire style object, we should instead set a font size variable we insert later

lib/components/viewers/pattern-row.tsx
```diff
@@ -73,31 +73,37 @@ class PatternRow extends Component<Props, State> {
 
     const routeName = route.shortName ? route.shortName : route.longName
     const routeColor = getRouteColorBasedOnSettings(route.operator, route)
+    console.log(stopTimes)
+    let style
+    if (routeName && routeName?.length >= 4 && routeName?.length <= 6) {
```
miles-grant-ibigroup, line 78: I don't think you need the first routeName check! Since you have the question mark

lib/components/viewers/viewers.css
```diff
@@ -102,48 +102,71 @@
...
+  overflow: hidden;
   width: 100%;
 }
 .otp .stop-viewer .route-row .header .route-name {
   align-items: center;
   display: grid;
   font-size: 42px;
   gap: 5px;
-  grid-template-columns: 4ch 2fr;
+  grid-template-columns: fit-content(8ch) fit-content(8ch);
```
miles-grant-ibigroup, line 145: This causes the grid to sometimes flow strangely, with destinations being unaligned. I think we need a better solution here

lib/components/viewers/viewers.css
```diff
@@ -102,48 +102,71 @@
...
 }
 .otp .stop-viewer .route-row .header .route-name span {
```
miles-grant-ibigroup, line 147: I'm not a fan of the webkit specific rules... What's going on here?

lib/components/viewers/viewers.css
```diff
@@ -246,7 +286,7 @@
 .otp .stop-viewer .stop-viewer-body .next-arrival-row .next-arrival-label {
   align-items: center;
   display: flex;
-  max-width: 250px;
+  width: 75%;
```
miles-grant-ibigroup, line 289: This might also be what's causing the grid misalignment?

lib/components/viewers/live-stop-times.tsx
```diff
@@ -151,35 +159,143 @@ class LiveStopTimes extends Component<Props, State> {
...
+          {todayTimes.length > 0 && (
+            <div className="list-container">
+              <ul className="route-row-container">
+                {todayTimes.map(({ id, pattern, route, times }) => {
+                  // Only add pattern if route info is returned by OTP.
+                  return (
+                    <PatternRow
```
miles-grant-ibigroup, line 212: I don't like repeating the same code 3 times. Is there any way to replace the 3 arrays with a new field inside of a new single array that's date or something, we can then check for that date changing inside of the map and render a heading accordingly. Let me know if you'd like to discuss this more offline, I'm not sure if that was super clear

miles-grant-ibigroup (changes requested):
The overall vision is good! Just a few specifics that I think we need to figure out

lib/components/viewers/live-stop-times.tsx
```diff
@@ -144,36 +150,88 @@ class LiveStopTimes extends Component<Props, State> {
...
+          ...route,
+          times: route.times.concat()?.sort(stopTimeComparator)
+        }
+      })
+      .filter(({ pattern, route }) =>
+        routeIsValid(route, getRouteIdForPattern(pattern))
+      )
+
+    const routesByDay: any = []
+    const firstDayOfService = routeTimes[0].times[0].serviceDay
```
miles-grant-ibigroup, line 167:
```suggestion
    const firstDayOfService = routeTimes?.[0].times?.[0].serviceDay
```

We can't rely on this existing!
image

lib/components/viewers/live-stop-times.tsx
```diff
@@ -144,36 +150,88 @@ class LiveStopTimes extends Component<Props, State> {
...
+    routeTimes.forEach((route) => {
+      const { serviceDay } = route.times[0]
+      const daysPastFirstService =
+        (serviceDay - firstDayOfService) / ONE_DAY_IN_SECONDS
+
+      // Create a separate array for each service day
+      routesByDay[daysPastFirstService]
+        ? routesByDay[daysPastFirstService].push(route)
+        : routesByDay.push([route])
+    })
```
miles-grant-ibigroup, line 179: I'm not sure I'm crazy about `routesByDay` wrapper array. Would it not be possible to just add a `day` field to the route time object?

miles-grant-ibigroup (changes requested):
I think we're getting there! Just a few small architecture changes -- I think we can get it cleaner

lib/components/viewers/live-stop-times.tsx
amy-corson-ibigroup: Added in [e6abad9](https://github.com/opentripplanner/otp-react-redux/pull/780/commits/e6abad93592a98933c323342b67150019b12ff32)!

lib/components/viewers/live-stop-times.tsx
```diff
@@ -144,36 +148,81 @@ class LiveStopTimes extends Component<Props, State> {
...
+      )
+
+    // Next 3 service days at this stop
+    const daysOfService = [
+      ...Array.from(new Set(routeTimes.map((route) => route.day)))
+    ]
+      .sort(function (a, b) {
+        return a - b
+      })
+      .slice(0, 3)
```
miles-grant-ibigroup, line 173: I think we can get away with not needing this! See #806

lib/components/viewers/pattern-row.tsx
```diff
@@ -73,31 +70,36 @@ class PatternRow extends Component<Props, State> {
...
     const routeName = route.shortName ? route.shortName : route.longName
     const routeColor = getRouteColorBasedOnSettings(route.operator, route)
+    let fontSize = '32px'
+    if (routeName) {
+      if (routeName?.length >= 4 && routeName?.length <= 6) {
+        fontSize = '20px'
+      } else if (routeName?.length > 7) {
+        fontSize = '16px'
+      }
+    }
```
miles-grant-ibigroup, line 80: Can this be replaced with `routeNameFontSize`?

miles-grant-ibigroup (changes requested):
Getting there! Just a little bit more cleanup

lib/components/viewers/pattern-row.tsx
amy-corson-ibigroup: Yes good catch!!

lib/components/viewers/viewers.css
```diff
@@ -102,48 +102,86 @@
...
+  display: flex;
+  gap: 10px;
 }
 .otp .stop-viewer .route-row .header .route-name span {
+  display: -webkit-box;
   font-size: 15px;
+  line-clamp: 3;
+  overflow: hidden;
+  -webkit-box-orient: vertical;
+  -webkit-line-clamp: 3;
```
miles-grant-ibigroup, line 165: I'm not sure we should use webkit specific rules. We have a post processor that's supposed to automatically apply these if we use the generic rules! What is this doing anyway? When we originally built this we moved away from `line-clamp` because it was so poorly supported

lib/components/viewers/viewers.css
```diff
@@ -102,48 +102,86 @@
...
 }
 .otp .stop-viewer .route-row .header .route-name strong span {
```
miles-grant-ibigroup, line 167: Some of these rules could probably be replaced by a stronger grid for the entire row, but this isn't blocking.

lib/components/viewers/viewers.css
```diff
@@ -102,48 +102,86 @@
 
 /* stop viewer styles */
 
-.otp .stop-viewer .route-row {
```
miles-grant-ibigroup, line 105: Any way we can set some form of min-height on this? To prevent this issue: 
image

lib/components/viewers/live-stop-times.tsx
```diff
@@ -144,36 +169,72 @@ class LiveStopTimes extends Component<Props, State> {
...
+      .filter(({ times }) => times.length !== 0)
+      .sort(patternComparator)
+      .map((route) => {
+        const sortedTimes = route.times
+          .concat()
+          ?.sort(stopTimeComparator)
+          // Only show times within 24 hours of next arrival time
+          .filter((time: any, i: number, array: Array<any>) => {
+            const firstDepartureTime =
+              array[0].serviceDay + array[0].realtimeDeparture
```
miles-grant-ibigroup, line 182:
```suggestion
              array?.[0]?.serviceDay + array?.[0]?.realtimeDeparture
```

We can't rely on any of this data!

lib/components/viewers/live-stop-times.tsx
```diff
@@ -144,36 +169,72 @@ class LiveStopTimes extends Component<Props, State> {
...
+            const departureTime = time.serviceDay + time.realtimeDeparture
+            return i === 0 || (departureTime - firstDepartureTime) / 3600 < 24
+          })
+        const { serviceDay } = sortedTimes[0]
```
miles-grant-ibigroup, line 186: I see this causing crashes too. Maybe we can add some sort of null check?

lib/components/viewers/live-stop-times.tsx
```diff
@@ -144,36 +169,72 @@ class LiveStopTimes extends Component<Props, State> {
...
+        return {
+          ...route,
+          day: serviceDay,
+          times: sortedTimes
+        }
+      })
+      .filter(({ pattern, route }) =>
+        routeIsValid(route, getRouteIdForPattern(pattern))
```
miles-grant-ibigroup, line 194: Any way we can move that previous filter into this filter step? Just so we don't loop through the array twice

lib/components/viewers/live-stop-times.tsx
```diff
@@ -144,36 +169,72 @@ class LiveStopTimes extends Component<Props, State> {
...
+      )
+      .filter((route) => {
+        /* If the route's first departure time falls on the 2nd available service day,
+      only show it if it's within 6 hours of now. */
+        if (!isSameDay(utcToZonedTime(route.day * 1000, homeTimezone), now)) {
+          const departureTime = route.times[0].realtimeDeparture + route.day
+          return isBefore(departureTime * 1000, addHours(now, 6))
+        }
+        return route
```
miles-grant-ibigroup, line 203: And ideally combine this one as well. There's some shared logic here as well so I think we can get clever with it

lib/components/viewers/viewers.css
```diff
@@ -102,48 +102,86 @@
...
+  -webkit-line-clamp: 3;
 }
 .otp .stop-viewer .route-row .header .route-name strong span {
-  font-size: 32px;
+  display: block;
+  font-size: inherit;
+  overflow: hidden;
+  padding-left: 15px;
+  text-align: center;
+  text-overflow: ellipsis;
```
miles-grant-ibigroup, line 173: This sometimes causes strange results... 
Screenshot 2023-03-06 at 11 12 17 AM

miles-grant-ibigroup (changes requested):
This PR is such a tricky one but I think we're getting there. Let me know if you'd like to chat about this offline a bit there's some open questions in the air

lib/components/viewers/live-stop-times.tsx
amy-corson-ibigroup: I can move the `routeIsValid` filter into the `times !== 0` filter no problem! The 2nd service day filter below is a bit tricker, because it relies on the `sort(stopTimeComparator)` which  in turn needs the `times !== 0` filter to avoid throwing an error. There could definitely be a better way to do this that I'm missing!

lib/components/viewers/viewers.css
```diff
@@ -102,48 +102,85 @@
...
   list-style-type: none;
   padding-left: 0;
 }
+.otp .stop-viewer .departures {
+  box-shadow: 2px 2px 5px 1px rgb(0 0 0 / 10%);
+}
+
+.otp .stop-viewer .day-container .route-row-container li .header.stop-view {
+  border-radius: 0;
+  height: fit-content;
```
miles-grant-ibigroup, line 116: This is still creating varied row heights for me!

lib/components/viewers/viewers.css
```diff
@@ -246,7 +301,7 @@
 .otp .stop-viewer .stop-viewer-body .next-arrival-row .next-arrival-label {
   align-items: center;
   display: flex;
-  max-width: 250px;
+  max-width: 75%;
```
miles-grant-ibigroup, line 304: Percentages on text can be wonky especially on strange screen widths. Can we do this with grid instead?

lib/components/viewers/pattern-row.tsx
```diff
@@ -111,7 +97,7 @@ class PatternRow extends Component<Props, State> {
                 />
               </div>
             </strong>
-            <span>{pattern.headsign}</span>
+            <span title={pattern.headsign}>{pattern.headsign}</span>
```
miles-grant-ibigroup, line 100: I forget if we discussed this already but I don't think it's good practice to have the `title` contain the exact same content as the element itself!

miles-grant-ibigroup (changes requested):
Looking much much cleaner! However still seeing some potential css issues, plus we're getting a lot of useless divs in the markup.

There are also merge conflicts

lib/components/viewers/pattern-row.tsx
amy-corson-ibigroup: In this case I think it's useful when often we're seeing gtfs where the shortname or headsign significantly overflows the container! A stopgap so users still have access to the info

lib/components/viewers/viewers.css
```diff
@@ -102,48 +102,76 @@
...
-.otp .stop-viewer .route-row {
-}
 .otp .stop-viewer .route-row-container,
 .related-panel-container {
-  border-radius: 10px;
-  box-shadow: 2px 2px 5px 1px rgba(0, 0, 0, 0.1);
   list-style-type: none;
   padding-left: 0;
 }
+.otp .stop-viewer .departures {
```
miles-grant-ibigroup, line 110: Can we add this as a comma rule for the related-panel-container

miles-grant-ibigroup (approved):
Approved pending the changes we discussed. good luck with the merge conflicts!

miles-grant-ibigroup:
A little worried about that snapshot update commit -- it's updating a lot of things that this PR wasn't supposed to touch?

amy-corson-ibigroup:
I saw that, I poked around for a while last night and couldn't figure out why! I figured I'd wait to investigate in case the next round of review requires another snapshot update, but I can revert the commit if that would be better.

binh-dam-ibigroup (changes requested):
Implement the change we agreed upon during the team discussion to reuse/repurpose the config item for how long in advance to display upcoming trips. Otherwise, looks solid!

lib/components/viewers/live-stop-times.tsx
```diff
@@ -155,36 +180,75 @@ class LiveStopTimes extends Component<Props, State> {
...
-        <ul className="route-row-container">
-          {Object.values(stopTimesByPattern)
-            .filter(({ times }) => times.length !== 0)
-            .sort(patternComparator)
-            .filter(({ pattern, route }) =>
-              routeIsValid(route, getRouteIdForPattern(pattern))
-            )
-            .map(({ id, pattern, route, times }) => {
-              // Only add pattern if route info is returned by OTP.
+        <div className="departures">
```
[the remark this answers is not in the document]
amy-corson-ibigroup: Good catch thank you!

lib/components/viewers/live-stop-times.tsx
```diff
@@ -155,36 +180,75 @@ class LiveStopTimes extends Component<Props, State> {
...
+        const { serviceDay } = sortedTimes[0]
+        return {
+          ...route,
+          day: serviceDay || null,
+          times: sortedTimes
+        }
+      })
+      .filter((route) => {
+        /* If the route's first departure time falls on the 2nd available service day,
+      only show it if it's within 6 hours of now. */
```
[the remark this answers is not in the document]
amy-corson-ibigroup: Yes, however the changes we discussed on the tech call removed this filter & comment entirely!

lib/components/viewers/live-stop-times.tsx
```diff
@@ -124,6 +128,26 @@ class LiveStopTimes extends Component<Props, State> {
       this._startAutoRefresh()
   }
 
+  renderDay = (homeTimezone: string, day: number, now: Date): JSX.Element => {
+    const formattedDay = utcToZonedTime(day * 1000, homeTimezone)
+    return (
+      <React.Fragment key={day}>
+        {/* If the service day is not today, add a label */}
```
binh-dam-ibigroup, line 135: A very small thing... inserting a label visually "breaks" the grouping (the edges of the arrivals it sit between are not rounded).

lib/components/viewers/live-stop-times.tsx
```diff
@@ -151,25 +176,55 @@ class LiveStopTimes extends Component<Props, State> {
       return stopTimeComparator(stopTimesA[0], stopTimesB[0])
     }
 
+    const configDepartureWindow = stopViewerConfig?.timeRange / 86400
```
binh-dam-ibigroup, line 179: Per how the redux state is initialized, `stopViewerConfig.timeRange` should always be defined.

lib/components/viewers/live-stop-times.tsx
```diff
@@ -151,25 +176,55 @@ class LiveStopTimes extends Component<Props, State> {
...
+          times &&
+          times.length !== 0 &&
+          routeIsValid(route, getRouteIdForPattern(pattern))
+      )
+      .sort(patternComparator)
+      .map((route) => {
+        const sortedTimes = route.times
+          .concat()
+          ?.sort(stopTimeComparator)
+          // Only show times within the next 2 days, unless otherwise indicated by the config.
```
binh-dam-ibigroup, line 197: Update the redux initial state too. The default there is still 4 days, see https://github.com/opentripplanner/otp-react-redux/blob/46c32ead25dc421d7baba0c4df9a2097ea2c6961/lib/reducers/create-otp-reducer.js#L87.

lib/components/viewers/live-stop-times.tsx
```diff
@@ -151,25 +176,55 @@ class LiveStopTimes extends Component<Props, State> {
...
+          .filter((time: any, i: number, array: Array<any>) => {
+            const departureWindow = configDepartureWindow || 4
```
binh-dam-ibigroup, line 199: Isn't this variable redundant now? (`configDepartureWindow` always defined -- assuming it is probably non-zero?)

lib/components/viewers/live-stop-times.tsx
```diff
@@ -151,25 +176,55 @@ class LiveStopTimes extends Component<Props, State> {
...
+          {routeTimes.map((time: any, index: number) => {
+            const { id, pattern, route, times } = time
+            return (
+              <React.Fragment key={id}>
+                {((index > 0 &&
+                  !isSameDay(
+                    time.day * 1000,
+                    routeTimes[index - 1]?.day * 1000
+                  )) ||
+                  // TODO: is new Date() the right approach?
```
binh-dam-ibigroup, line 223: Delete this comment.

binh-dam-ibigroup (approved):
There is one missing change to how the default time range is initialized in the redux state, please fix that, otherwise changes look good.

lib/components/viewers/live-stop-times.tsx
amy-corson-ibigroup: This is actually intentional! The idea was for the entire departure container to be rounded, so the middle rows should stay square. Original wireframe for reference: 

image

amy-corson-ibigroup:
The a11y tests were failing because technically the `renderDay` label is a `<p>` sitting within a `<ul>`. To work around this, I added `aria-hidden` to the `p` and added an `InvisibleA11yLabel` to each time that does not fall on the current day.

| UI    | Screenreader |
| -------- | ------- |
|image|image|

Tried using aria-role, but it kept failing! Let me know if this doesn't seem like the best solution.

amy-corson-ibigroup:
A poorly executed merge from dev ([2aa1792](https://github.com/opentripplanner/otp-react-redux/pull/780/commits/2aa179263ddf8605729baee390258640da1bdc3c)) could cause potentially hundreds of merge conflicts in weblate, so closing in favor of #837!

PR #781, opened 2023-02-08
Invisible Metro Itinerary Header

amy-corson-ibigroup:
Creates an invisible header in the metro itinerary that renders as an h2 when the itinerary is expanded, ~~and an h3 when it is displayed as a result~~. 

The a11y/CI tests will not accept the dynamic `InvisibleHeader` that was introduced in #753, so this PR instead renders the itinerary result as a `span` when it appears as a result `li`, and as an `h2` when the result is expanded.


binh-dam-ibigroup (commented):
This is in contrast with the itinerary summary that visually displays the modes and the route numbers. Do we want to follow that instead?

amy-corson-ibigroup:
@binh-dam-ibigroup I agree the best thing would be for the header to match the route summary as closely as possible. Feel free to assign back to me and I can work on adding the route numbers today!

amy-corson-ibigroup:
Changing the `routeBlocks` to an `h3` didn't quite work the way we were hoping - I couldn't see a recommended way to replace the blocks themselves with alt-text in screenreader view. I kept the InvisibleHeader and added the route names as discussed! Let me know if this falls short, or if you have any thoughts.

lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -413,6 +406,15 @@ class MetroItinerary extends NarrativeItinerary {
             {!mini && (
               <ItineraryGrid className="itin-grid" role="group">
                 {/* TODO: a11y: add aria-label to parent element */}
+                <InvisibleHeader as={expanded && 'h2'}>
```
binh-dam-ibigroup, line 409: When expanded, the `<h2>` is not needed IMO. Indeed, just above it is an `<h1>` with the main mode summary (e.g. "Walk + Transit"). Alternately, this element could replace the `<h1>` above, what do you think?

amy-corson-ibigroup: Good point, I went ahead and removed the previous `h1` and replaced it with this header, since it was more descriptive

lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -406,14 +407,24 @@ class MetroItinerary extends NarrativeItinerary {
...
+                      values={{
+                        routes: transitRoutes,
+                        time: formatDuration(itinerary.duration, intl)
+                      }}
+                    />
+                  ) : (
+                    <FormattedMessage
+                      id="components.MetroUI.singleModeItineraryDescription"
+                      values={{
+                        mode: itinerary.legs[0].mode,
```
binh-dam-ibigroup, line 423: Mode has to be translated. As written, it is the OTP mode enum value. For bike and walk, use for instance `common.modes.bike/walk` or `otpUi.ItineraryBody.travelByMode.bike/walk`.

binh-dam-ibigroup (changes requested):
Please internationalize the first leg mode for single-mode itineraries.

lib/components/narrative/metro/metro-itinerary.tsx
amy-corson-ibigroup: Good catch, sorry for missing that! Fixed in cbd2965

binh-dam-ibigroup (approved):
Looks better, thanks for changes!

lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -204,6 +203,15 @@ const ItineraryGridSmall = styled.div`
   }
 `
 
+// invisible header rendered for screen readers and a11y technologies
+const InvisibleHeader = styled.span`
+  //place it in an unused grid cell so it doesn't add a row
```
miles-grant-ibigroup, line 208: Could we adjust this comment to make it clear that this is a bit of a hack? I don't mind it for now but I'd like to have it be a TODO to find a cleaner solution for this

miles-grant-ibigroup (approved):
Works great! Thanks for these finicky fixes

PR #783, opened 2023-02-08
improvement(BatchRoutingPanel): Cleanup CSS.

binh-dam-ibigroup:
Some suggestions to simplify the code. HTML and CSS handle stacking things on top of each other very well.

### DIFF
--- a/lib/components/app/batch-routing-panel.js
+++ b/lib/components/app/batch-routing-panel.js
@@ -28,7 +28,14 @@ class BatchRoutingPanel extends Component {
         })
 
     return (
-      <ViewerContainer className="batch-routing-panel">
+      <ViewerContainer
+        className="batch-routing-panel"
+        style={{
+          display: 'flex',
+          flexDirection: 'column',
+          height: '100%'
+        }}
+      >
         <form
           className="form"
           onSubmit={this.handleSubmit}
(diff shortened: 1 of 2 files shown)


lib/components/app/batch-routing-panel.js
```diff
@@ -28,14 +28,7 @@ class BatchRoutingPanel extends Component {
         })
 
     return (
-      <ViewerContainer
-        className="batch-routing-panel"
-        style={{
-          display: 'flex',
```
miles-grant-ibigroup, line 34: The removal of this causes the white block to appear at the bottom of an incomplete load as you can see here.

https://user-images.githubusercontent.com/86619099/217650871-6f6e8905-b1fc-4893-bf66-fbf280a196ad.mov

miles-grant-ibigroup (changes requested):
Much cleaner thank you! Just one small issue to fix

lib/components/app/batch-routing-panel.js
binh-dam-ibigroup: Okay. It's back in https://github.com/opentripplanner/otp-react-redux/pull/783/commits/6832c0465b2b33e95266419e2ab2d7c6ed17f413.

PR #785, opened 2023-02-09
Sort Stop Viewer Routes by Arrival/Departure Time

AdrianaCeric:
Using `StopTimeComparator` to sort routes in the Stop Viewer.
Co-authored by Binh!
image

### DIFF
--- a/lib/components/viewers/live-stop-times.tsx
+++ b/lib/components/viewers/live-stop-times.tsx
@@ -7,7 +7,8 @@ import React, { Component } from 'react'
 import {
   getRouteIdForPattern,
   getStopTimesByPattern,
-  routeIsValid
+  routeIsValid,
+  stopTimeComparator
 } from '../../util/viewer'
 import { IconWithText } from '../util/styledIcon'
 import SpanWithSpace from '../util/span-with-space'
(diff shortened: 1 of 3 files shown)


lib/util/viewer.js
```diff
@@ -211,8 +211,8 @@ export const getModeFromStop = (stop) => {
  * (in chronological order - 9:13am, 9:15am, etc.)
  */
 export function stopTimeComparator(a, b) {
-  const aTime = a.serviceDay + a.scheduledDeparture
-  const bTime = b.serviceDay + b.scheduledDeparture
+  const aTime = a.serviceDay + a.realtimeDeparture
```
miles-grant-ibigroup, line 214: Do we know that `realtimeDeparture` is always present?

lib/components/viewers/live-stop-times.tsx
```diff
@@ -134,27 +135,20 @@ class LiveStopTimes extends Component<Props, State> {
...
 
     // TODO: Shared types
     const patternComparator = (patternA: any, patternB: any) => {
-      // first sort by routes
-      const routeCompareValue = routeComparator(patternA.route, patternB.route)
-      if (routeCompareValue !== 0) return routeCompareValue
-
-      // if same route, sort by headsign
-      return patternHeadsignComparator(patternA, patternB)
+      const arrayA = [].concat(patternA.times).sort(stopTimeComparator)
```
miles-grant-ibigroup, line 141: Why are we concatenating an empty array instead of `map`ping `pattern.times`?

lib/components/viewers/live-stop-times.tsx
```diff
@@ -134,27 +135,20 @@ class LiveStopTimes extends Component<Props, State> {
...
+      const arrayB = [].concat(patternB.times).sort(stopTimeComparator)
```
miles-grant-ibigroup, line 142: Would it be possible to rename these arrays?
```suggestion
      const stopTimesB = [].concat(patternB.times).sort(stopTimeComparator)
```

miles-grant-ibigroup (changes requested):
Works great thanks for the changes! Just a few small tweaks

lib/components/viewers/live-stop-times.tsx
AdrianaCeric: I remember the decision to do this came down to State best practices in React. Do you think mapping is a better way?

miles-grant-ibigroup: Sounds good to me

lib/util/viewer.js
AdrianaCeric: Good catch

miles-grant-ibigroup (approved):
Looks good! Just need to update snapshots and then this is good to go

amy-corson-ibigroup (approved):
Looks good to me!

PR #786, opened 2023-02-10
Add OTP-rr PR Template

AdrianaCeric:
**Description:**
This PR is a PR template (super meta)!

**PR Checklist:**
    - [x] N/A
    - [x] N/A
    - [x] N/A

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,15 @@
+<!--Please provide a brief description of what this PR accomplishes, whether it should be considered/merged with any parallel PR(s) in another repository-->
+**Description:**
+
+
+<!--Ensure you have the following recommendations met before requesting a review, if applicable:-->
+**PR Checklist:**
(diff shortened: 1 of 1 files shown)


daniel-heppner-ibigroup (approved):
Looks good! Thank you for this improvement!

amy-corson-ibigroup (approved):
Looks great thanks for doing this!! Super nitpicky, but I'm wondering how you feel about the N/A checkboxes, versus just asking folks to remove the items in the checklist that aren't applicable (which I think we do in datatools-ui). To me these nested checkboxes are a little confusing on first glance, but I will defer to you!!

AdrianaCeric:
@amy-corson-ibigroup you have a really good point and it would make it less cluttered. I think I initially wanted to distinguish not having completed a check box and it being N/A but if we assume that once someone opens a PR, it should be at a level where all checkboxes should be checked off, otherwise blank if N/A, then it's safe to remove the N/A imo too! I'll make the change

PR #789, opened 2023-02-13
Additional Itinerary A11y Details

amy-corson-ibigroup:
- Changes screenreader itinerary buttons to reflect departure time instead of arrival time.


amy-corson-ibigroup:
@demory Very odd... can you share a search that's returning these? I just tried and wasn't able to reproduce!

demory (approved):

demory:
FYI, merged the latest dev and am no longer seeing the above issue

i18n/fr.yml
```diff
@@ -233,7 +233,6 @@ components:
     stops: Arrêts et stations
     streets: Plan des rues
   MetroUI:
-    arriveAtTime: "Arrivée à {time}"
```
AdrianaCeric, line 236: Why not also remove this key in other languages (ko, vi..)?

AdrianaCeric (approved):
Looks and works great, thanks for the changes! Just a localization comment

lib/components/narrative/metro/departure-times-list.tsx
```diff
@@ -27,10 +24,7 @@ export const DepartureTimesList = (props: DepartureTimesProps): JSX.Element => {
     return (
       <button
         className={isRealTime ? 'realtime active' : 'active'}
-        title={`${intl.formatMessage(
-          { id: 'components.MetroUI.arriveAtTime' },
```
miles-grant-ibigroup, line 31: Why was this removed?

miles-grant-ibigroup (changes requested):
Quickly blocking this -- think there may be some regressions

lib/components/narrative/metro/departure-times-list.tsx
amy-corson-ibigroup: Because it replaces the departure time, creating inconsistencies for screenreaders: 
image

miles-grant-ibigroup: Doesn't aria-label override `title`? I don't like that you hover over the text and the title contains the exact same text

amy-corson-ibigroup: The title should actually contain the time and also whether or not the departure is realtime, so in some cases it will be different! 

I can add an aria label as well! I think it should probably have both, you're right.

i18n/fr.yml
amy-corson-ibigroup: Looks like someone already merged a commit into dev that took care of the other languages!

lib/components/narrative/metro/departure-times-list.tsx
amy-corson-ibigroup: I went ahead and added the "based on schedule data" string to scheduled trips as well, so that the title is never the same text as the button!

miles-grant-ibigroup: I like this more! Would it be possible to re-add the arrives by info? That is a helpful piece of information

amy-corson-ibigroup: I agree the arrives by info is helpful, but I don't think putting it as the title of this button feels like a super intuitive place for it. May be worth having a larger team discussion for the best place to present that info within the itinerary?

miles-grant-ibigroup: This decision came out of a team discussion on this issue a while ago. I agree it's not perfect but would prefer the info in the title to it not being there at all!

amy-corson-ibigroup:
The keyboard navigation for non-transit options has been split into https://github.com/opentripplanner/otp-react-redux/pull/815

lib/components/narrative/metro/departure-times-list.tsx
```diff
@@ -23,18 +20,17 @@ export const DepartureTimesList = (props: DepartureTimesProps): JSX.Element => {
   const { activeItineraryTimeIndex, itinerary, setItineraryTimeIndex } = props
   const intl = useIntl()
   const isRealTime = firstTransitLegIsRealtime(itinerary)
+  const departureLabelText = `${intl.formatTime(itinerary.startTime)} ${
+    isRealTime
+      ? intl.formatMessage({ id: 'components.StopTimeCell.realtime' })
+      : intl.formatMessage({ id: 'components.StopTimeCell.scheduled' })
```
miles-grant-ibigroup, line 26: I believe @binh-dam-ibigroup has had notes before on doing string concatenation like this, but I forget what the note is

lib/components/narrative/metro/departure-times-list.tsx
```diff
@@ -58,10 +54,7 @@ export const DepartureTimesList = (props: DepartureTimesProps): JSX.Element => {
             className={classNames.join(' ')}
             key={getFirstLegStartTime(time.legs)}
             onClick={() => setItineraryTimeIndex(index)}
-            title={`${intl.formatMessage(
-              { id: 'components.MetroUI.arriveAtTime' },
-              { time: intl.formatTime(getLastLegEndTime(time.legs)) }
-            )} ${
+            title={`${intl.formatTime(getFirstLegStartTime(time.legs))} ${
```
miles-grant-ibigroup, line 57: Based on our discussion I would remove this title as well. It only shows the time for me and a space.

miles-grant-ibigroup (changes requested):
After discussion ok to remove the arrive by title, but I'm not sure about the string concatenation (especially for potential future RTL languages)

lib/components/narrative/metro/departure-times-list.tsx
binh-dam-ibigroup: It is probably ok to do telegraph style.

lib/components/narrative/metro/departure-times-list.tsx
```diff
@@ -23,18 +20,17 @@ export const DepartureTimesList = (props: DepartureTimesProps): JSX.Element => {
...
         className={isRealTime ? 'realtime active' : 'active'}
-        title={`${intl.formatMessage(
-          { id: 'components.MetroUI.arriveAtTime' },
-          { time: intl.formatTime(itinerary.endTime) }
-        )} ${
-          isRealTime
-            ? intl.formatMessage({ id: 'components.StopTimeCell.realtime' })
-            : ''
-        }`}
+        title={departureLabelText}
```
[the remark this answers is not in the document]
binh-dam-ibigroup: If removing the title, you can also remove `aria-label` and instead place the real-time or schedule hint next to the time, as visually-hidden.

lib/components/narrative/metro/departure-times-list.tsx
binh-dam-ibigroup: I would put parentheses around the real-time/scheduled hints. That way, at least for French (and Spanish?), you could say "10:30 (real-time data)", "10:45 (theoretical schedule)", and we wouldn't have to worry about conjunctions.

lib/components/viewers/realtime-status-label.tsx
```diff
@@ -96,7 +96,7 @@ const RealtimeStatusLabel = ({
     // and display the updated time underneath.
     renderedTime = isEarlyOrLate ? (
       <TimeBlock>
-        <TimeStruck>
+        <TimeStruck aria-hidden>
```
binh-dam-ibigroup, line 99: IMO, we still need to say what the original time was.

amy-corson-ibigroup: This actually feeds into the work we're doing on https://github.com/opentripplanner/otp-ui/pull/549. If we go with the below description, how would we incorporate the old time? 

`6:44 pm, from <location>, ride 4 stops to <next location>`

Something like?

`6:44 pm (4 minutes late, previously 6:40PM), from <location>, ride 4 stops to <next location>`

binh-dam-ibigroup: Indeed. Maybe "originally" instead of "previously".

amy-corson-ibigroup:
I kept the button titles, because I do think they're helpful for sighted users who may not be familiar with how trip planners typically indicate realtime departure data, but @binh-dam-ibigroup @miles-grant-ibigroup if you both feel strongly they should be removed in favor or the `InvisibleAccessibilityLabel` for screen readers, happy to do so!

lib/components/narrative/metro/departure-times-list.tsx
```diff
@@ -20,17 +20,18 @@ export const DepartureTimesList = (props: DepartureTimesProps): JSX.Element => {
   const { activeItineraryTimeIndex, itinerary, setItineraryTimeIndex } = props
   const intl = useIntl()
   const isRealTime = firstTransitLegIsRealtime(itinerary)
-  const departureLabelText = `${intl.formatTime(itinerary.startTime)} ${
-    isRealTime
-      ? intl.formatMessage({ id: 'components.StopTimeCell.realtime' })
-      : intl.formatMessage({ id: 'components.StopTimeCell.scheduled' })
-  }`
+  const departureLabelText = (time: any, realTime: boolean) =>
```
binh-dam-ibigroup, line 23: Move this new function outside of the component.

lib/components/narrative/metro/departure-times-list.tsx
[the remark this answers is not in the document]
AdrianaCeric: Good point Amy. Instead "based on realtime/scheduled data" we can just have a title that says "Realtime" to indicate what the green wifi symbol means for non-power users. I don't know if we need a title that says this is from scheduled data at all (seems implied) but let me know what you think
image

binh-dam-ibigroup:
> I kept (and improved - no more random spaces) the button titles, because I do think they're helpful for sighted users who may not be familiar with how trip planners typically indicate realtime departure data...

I'm fine with keeping the title.

lib/components/viewers/realtime-status-label.tsx
```diff
@@ -82,6 +83,9 @@ const RealtimeStatusLabel = ({
     delay,
     onTimeThresholdSeconds
   )
+  const formattedOriginalTime = (
+    <FormattedTime timeStyle="short" value={originalTime} />
+  )
```
miles-grant-ibigroup, line 88: You can inline this!

miles-grant-ibigroup (approved):
One small refactor, but looking really great otherwise!

PR #790, opened 2023-02-14
add labeling to stop viewer

amy-corson-ibigroup:
- Additional aria-labeling around stop viewer schedule inputs and refresh button.
- Fix "spin" animation


demory:
@amy-corson-ibigroup The refresh button is showing up with no label/content in screen reader view:
image

The button also appears to be tagged as `aria-hidden=true`. I don't think interactive content should ever be tagged as aria hidden?

image

demory (approved):

demory:
@amy-corson-ibigroup right after approving I noticed an issue that requires a (hopefully) quick change. The changes to IconWithText now cause the date/time preview to look like this:

image

I think it's because these icon buttons are now wrapped with divs which combined the the <br> in the date/time preview are causing that gap. Should probably also check for any other uses of IconWithText that might be affected.

amy-corson-ibigroup:
Thanks for catching that @demory!! I added a prop to `IconWithText` so that the changes will only affect this stop times refresh button. See [6c3cc0d](https://github.com/opentripplanner/otp-react-redux/pull/790/commits/6c3cc0d0a87d85a31d07f98000c6ac4eb1af0e29).

lib/components/util/styledIcon.tsx
```diff
@@ -72,18 +73,26 @@ export const StyledIconWrapperTextAlign = styled(StyledIconWrapper)<Props>`
   }
 `
 
+const IconButton = styled.div`
+  display: flex;
+  gap: 5px;
+`
```
miles-grant-ibigroup, line 79: Can we replace the `flex` prop with one that lets you specify a component to use instead of `React.Fragment`? Then we can move this outside of this component and keep it a bit more abstract

miles-grant-ibigroup (changes requested):
Looks good! Just one small cleanup suggestion

lib/components/viewers/live-stop-times.tsx
```diff
@@ -191,14 +202,18 @@ class LiveStopTimes extends Component<Props, State> {
...
           </label>
           <button
-            // Functionality is provided by auto-refresh, this only adds confusion
-            // in a screen reader context
-            aria-hidden
+            aria-label={refreshButtonText}
             className="link-button pull-right percy-hide"
             onClick={this._refreshStopTimes}
             style={{ fontSize: 'small' }}
+            title={refreshButtonText}
```
miles-grant-ibigroup, line 209: Do we need both title and aria label?

miles-grant-ibigroup (approved):
Working great for me! Just a few potential cleanups

lib/components/util/styledIcon.tsx
```diff
@@ -47,7 +49,7 @@ const rotateAnimation = keyframes`
 `
 
 export const StyledIconWrapper = styled.span<Props>`
-  animation: ${(props) => (props.spin ? rotateAnimation : 'none')} 1.2s linear
+  animation: ${(props) => (props.spin ? rotateAnimation : 'none')} 1s linear
```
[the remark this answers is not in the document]
amy-corson-ibigroup: at `1.2s` the icon doesn't quite make the full rotation, and jumps back to its original position. `1s` just looks cleaner.

lib/components/viewers/live-stop-times.tsx
miles-grant-ibigroup: Oh I thought `title` was automatically read out by screen readers. Thanks for clearing that up!

PR #792, opened 2023-02-14
Fix dropdown menus

binh-dam-ibigroup:
This PR fixes broken menu behavior for the language selector and the user account menu:
- the dropdown menu is now dismissed when selecting a language,
- the dropdown menu is now hidden when the keyboard focus reaches another element not in the menu.
- the logout command works again.

The PR also cleans up code and types and moves event handling from `<li>`s to `<button>`s.

### DIFF
--- a/index.css
+++ b/index.css
@@ -122,7 +122,7 @@ input[type="text"]::-ms-clear {
 }
 
 .app-menu button[aria-selected="true"],
-#locale-selector li[aria-selected="true"] {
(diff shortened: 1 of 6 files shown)


AdrianaCeric (approved):
Everything works as expected! 👍 I don't see anything wrong with the implementation as of now.

miles-grant-ibigroup (approved):
Working well for me and much cleaner! Thanks so much for this cleanup

PR #793, opened 2023-02-14
deps: update trip-details

daniel-heppner-ibigroup:
Updates trip details to support the new OTP2 fare table.


lib/components/narrative/line-itin/connected-itinerary-body.js
```diff
@@ -37,6 +37,11 @@ const noop = () => {}
 
 const ItineraryBodyContainer = styled.div`
   padding: 0px 10px;
+
+  /* TODO: Find a better way of doing this */
+  table {
+    font-size: 14px;
```
miles-grant-ibigroup, line 43: I don't think this is as much of a hack as you think it is!

miles-grant-ibigroup (approved):
Works and looks great!

lib/components/narrative/line-itin/connected-itinerary-body.js
daniel-heppner-ibigroup: Should probably add a class name to the table someday

miles-grant-ibigroup: I think since this is styled components you don't need to!

demory (approved):

PR #806, opened 2023-03-01
Stop viewer refactor departures refactor

miles-grant-ibigroup:
**Description:**
This PR cleans up the main loop of `live-stop-times`.

**PR Checklist:**

### DIFF
--- a/lib/components/viewers/live-stop-times.tsx
+++ b/lib/components/viewers/live-stop-times.tsx
@@ -3,8 +3,6 @@ import { FormattedMessage, FormattedTime } from 'react-intl'
 import { Redo } from '@styled-icons/fa-solid/Redo'
 import { TransitOperator } from '@opentripplanner/types'
 import coreUtils from '@opentripplanner/core-utils'
-
-import FormattedDayOfWeek from '../util/formatted-day-of-week'
 import isSameDay from 'date-fns/isSameDay'
 import React, { Component } from 'react'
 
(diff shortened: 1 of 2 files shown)


lib/components/viewers/live-stop-times.tsx
```diff
@@ -163,65 +183,35 @@ class LiveStopTimes extends Component<Props, State> {
...
-                            stopViewerConfig={stopViewerConfig}
-                          />
-                        )
-                      })}
-                  </ul>
-                </div>
+                <React.Fragment key={id}>
+                  {index > 0 &&
+                    time.day !== routeTimes[index - 1].day &&
+                    this.renderDay(homeTimezone, time.day)}
```
amy-corson-ibigroup, line 196: You may need to get rid of this `index > 0`, in case the first available day needs a label!

```suggestion
                  {time.day !== routeTimes[index - 1]?.day &&
                    this.renderDay(homeTimezone, time.day)}
```

amy-corson-ibigroup (approved):
Thank you so much for figuring this out, it looks great!! Just one quick note

lib/components/viewers/live-stop-times.tsx
amy-corson-ibigroup: Maybe we could move the `isSameDay` check out of `this.renderDay` and put it here instead? And then just 
`{index === 0 && !isSameDay(today, time.day) && this.renderDay}`

PR #807, opened 2023-03-01
Fix transit overlay highlight color

AdrianaCeric:
**Description:**
When hovered over, transit overlay vehicles show up as a low-contrast white color on a light grey background. This PR unblocks `otp-ui` route color highlight functionality w/ contrast. 

**PR Checklist:**

| Before | After |
|--------|-------|
| image | image |

### DIFF
--- a/lib/components/map/connected-transit-vehicle-overlay.js
+++ b/lib/components/map/connected-transit-vehicle-overlay.js
@@ -124,7 +124,10 @@ const mapStateToProps = (state) => {
   if (viewedRoute?.routeId) {
     vehicleList = route?.vehicles?.map((vehicle) => {
       vehicle.routeType = route?.mode
-      vehicle.routeColor = route?.color ? '#' + route.color : null
+      vehicle.routeColor =
+        route.color && !route.color.includes('#')
+          ? '#' + route.color
+          : route?.color || '#5A5A5A'
       // Try to populate this attribute, which is required for the vehicle popup to appear.
       vehicle.routeShortName = vehicle.routeShortName || route?.shortName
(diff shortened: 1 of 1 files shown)


lib/components/map/connected-transit-vehicle-overlay.js
```diff
@@ -124,7 +124,7 @@ const mapStateToProps = (state) => {
   if (viewedRoute?.routeId) {
     vehicleList = route?.vehicles?.map((vehicle) => {
       vehicle.routeType = route?.mode
-      vehicle.routeColor = route?.color
+      vehicle.routeColor = route?.color ? '#' + route.color : null
```
miles-grant-ibigroup, line 127:
```suggestion
      vehicle.routeColor = !route?.color.includes("#") ? '#' + route.color : route?.color
```

This covers more cases while (hopefully) still fixing the issue

miles-grant-ibigroup (changes requested):
Good fix thank you! I'm just nervous about mandating the lack of `#`. I think we can support both cases

lib/components/map/connected-transit-vehicle-overlay.js
AdrianaCeric: Good points! Added the change to https://github.com/opentripplanner/otp-react-redux/pull/807/commits/efea50e67e2786c99d8057c566fb081feb9a6b07. Also handled the case where the color is undefined, null or "" by defaulting it to a dark grey.
Pasted Graphic 66

miles-grant-ibigroup (approved):
Great work!

PR #812, opened 2023-03-02
Remove caret if bearing is null

AdrianaCeric:
**Description:**
Some feeds do not include vehicle bearing, making transit overlay carets default to pointing upwards. This PR removes the caret if vehicle.heading is null. Let me know if you think there's a better way to go about this; I had to break up the original function a bit. The `waterloo` instance is a good example of a feed without bearings (shown below).

**PR Checklist:**

| Before | After |
|--------|-------|
| Pasted Graphic 63 | Pasted Graphic 60 |

### DIFF
--- a/lib/components/map/connected-transit-vehicle-overlay.js
+++ b/lib/components/map/connected-transit-vehicle-overlay.js
@@ -102,15 +102,28 @@ function VehicleTooltip(props) {
   )
 }
 
-// Round vehicle symbol with arrow/caret on the border,
-// and showing the route color with a transparent effect on hover.
-const IconContainer = withRouteColorBackground(
-  withCaret(Circle, { height: 5, offset: 1.5, width: 10 }),
-  {
+/**
+ *
+ * @param {Object} props vehicle object (includes heading)
+ * @returns circlular mode icon overlay that includes a caret only if bearing is not null
+ */
+const IconContainer = (props) => {
+  const { vehicle } = props
+  const caretHeight = vehicle?.heading === null ? -1 : 5
+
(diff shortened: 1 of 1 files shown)


lib/components/map/connected-transit-vehicle-overlay.js
```diff
@@ -102,15 +102,28 @@ function VehicleTooltip(props) {
...
+/**
+ *
+ * @param {Object} props vehicle object (includes heading)
+ * @returns circular mode icon overlay that includes a caret only if bearing is not null
+ */
+const IconContainer = (props) => {
+  const { vehicle } = props
+  const caretHeight = vehicle?.heading === null ? -1 : 5
+
+  const OverlayIcon = withCaret(Circle, {
```
binh-dam-ibigroup, line 114: This sounds like a patch is needed in the transit vehicle overlay package instead. For performance reasons, we don't want to be generating component classes in the rendering code of another component.

AdrianaCeric: Yeah agreed the function is clunky. Do you mean changing the TransitVehicleOverlay function directly? Right now it looks like it's in my node_modules from @opentripplnner, would this need to be patched in another repo?

AdrianaCeric: Update: I will take a stab at patching the `TransitVehicleOverlay` in `otp-ui`
PR will be linked here after it is made

AdrianaCeric:
Closed because change is being made in `otp-ui` instead: https://github.com/opentripplanner/otp-ui/pull/556

PR #827, opened 2023-03-16
Fix: Dismissed App Menu does not re-open 

amy-corson-ibigroup:
**Description:**
Fixes a small bug where `AppMenu` could be opened and closed by clicking anywhere on the page if you clicked before `react-sliding-pane`'s transition ended.

To reproduce bug:
- Open app menu
- Click outside app menu to dismiss menu
- Immediately click again (within 500ms)
- App menu should re-open

https://user-images.githubusercontent.com/115499534/225717982-37e06c76-d499-4129-b2ce-ec4067bef522.mov

**PR Checklist:**

### DIFF
--- a/lib/components/app/app-menu.tsx
+++ b/lib/components/app/app-menu.tsx
@@ -216,7 +216,7 @@ class AppMenu extends Component<
           from="left"
           hideHeader
           isOpen={isPaneOpen}
-          onRequestClose={this._togglePane}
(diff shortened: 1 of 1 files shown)


miles-grant-ibigroup (approved):
Wow that's a clean fix thanks for making that! Can't believe we had it the other way.

AdrianaCeric (approved):
Great catch and implementation!

PR #829, opened 2023-03-17
Route View to Pattern View Button

amy-corson-ibigroup:
**Description:**
This PR removes the "route details" with headsign select from the active route in the route viewer. Instead, an active route will have an arrow button that brings the user directly to the pattern viewer.

The pattern view button will only show up when route is active for sighted users, similarly to how the route details were only displayed when the route was active. However, the button is available to all screen readers.

I'm worried the arrow button being displayed for active routes looks strange when you keyboard navigate further down the page. Maybe the button should disappear when the parent `li` element loses focus?

**PR Checklist:**

| Before | After |
|--------|-------|
|image|image| -->

### DIFF
--- a/lib/components/viewers/RouteRow.js
+++ b/lib/components/viewers/RouteRow.js
@@ -238,13 +238,13 @@ export class RouteRow extends PureComponent {
                   getModeFromRoute(route).toLowerCase(),
                   intl
                 )}
-                height={22}
+                height={28}
                 leg={{
                   routeLongName: route?.longName,
                   routeShortName: route?.shortName
                 }}
                 mode={getModeFromRoute(route)}
-                width={22}
+                width={28}
               />
             </ModeIconElement>
           )}
--- a/lib/components/viewers/RouteRow.js
+++ b/lib/components/viewers/RouteRow.js
@@ -52,8 +52,9 @@ export const OperatorImg = styled.img`
 `
 
 export const ModeIconElement = styled.span`
+  height: 28px;
   padding-left: inherit;
-  padding-top: 3px;
+  width: 28px;
 `
 
 const RouteNameElement = styled.span`
--- a/lib/components/viewers/RouteRow.js
+++ b/lib/components/viewers/RouteRow.js
@@ -3,9 +3,9 @@
 import { ArrowRight } from '@styled-icons/fa-solid'
 import { Button } from 'react-bootstrap'
 
+import { Icon } from '../util/styledIcon'
 import React, { PureComponent } from 'react'
 import styled from 'styled-components'
-import { Icon } from '../util/styledIcon'
 
 import { ComponentContext } from '../../util/contexts'
 import {
--- a/lib/components/viewers/RouteRow.js
+++ b/lib/components/viewers/RouteRow.js
@@ -33,7 +33,7 @@ export const RouteRowButton = styled(Button)`
   left: 0;
   position: absolute;
   top: 0;
-  transition: all ease-out 0.17s;
+  transition: all ease-out 0.1s;
   width: 100%;
 
   &:before {
@@ -55,7 +55,7 @@ export const RouteRowButton = styled(Button)`
   &:active:before,
   &.btn-default:active:focus:before {
     background-color: #eee;
-    transition: all ease-out 0.17s;
+    transition: all ease-out 0.1s;
   }
 `
 
@@ -111,12 +111,12 @@ const PatternButton = styled.button`
   border: none;
   border-radius: 50%;
   color: #004f96;
-  opacity: ${(props) => (props.display ? 100 : 0)};
   height: 40px;
   margin-right: 8px;
-  position: relative;
-  transition: all ease-out 0.17s;
   min-width: 40px;
+  opacity: ${(props) => (props.display ? 100 : 0)};
+  position: relative;
+  transition: all ease-out 0.1s;
   z-index: ${(props) => (props.display ? 2 : -2)};
 
   svg {
@@ -125,9 +125,14 @@ const PatternButton = styled.button`
     width: 22px !important;
   }
 
+  :active {
+    opacity: 80%;
+    transition: all ease-out 0.1s;
+  }
+
   &:hover {
     background: #eee;
-    transition: all ease-out 0.17s;
+    transition: all ease-out 0.1s;
   }
 `
 
--- a/lib/components/viewers/RouteRow.js
+++ b/lib/components/viewers/RouteRow.js
@@ -1,7 +1,9 @@
 // TODO: Typescript, which doesn't make sense to do in this file until common types are established
 /* eslint-disable react/prop-types */
+import { ArrowRight } from '@styled-icons/fa-solid'
 import { Button } from 'react-bootstrap'
-import AnimateHeight from 'react-animate-height'
+
+import { Icon } from '../util/styledIcon'
 import React, { PureComponent } from 'react'
 import styled from 'styled-components'
 
@@ -14,20 +16,31 @@ import { getFormattedMode } from '../../util/i18n'
 import DefaultRouteRenderer from '../narrative/metro/default-route-renderer'
 import OperatorLogo from '../util/operator-logo'
 
-import RouteDetails from './route-details'
-
 export const StyledRouteRow = styled.li`
-  background-color: white;
-  margin: 10px 0;
-  list-style: none;
-`
-
-export const RouteRowButton = styled(Button)`
   align-items: center;
+  background-color: white;
   display: flex;
-  padding: 6px;
+  gap: 7px;
+  justify-content: space-between;
+  list-style: none;
+  margin: 5px 0;
+  padding: 5px;
+  position: relative;
+`
+// Route Row Button sits invisble on top of the route name and info.
+export const RouteRowButton = styled(Button)`
+  height: 100%;
+  left: 0;
+  position: absolute;
+  top: 0;
   transition: all ease-in-out 0.1s;
   width: 100%;
+
+  &:active,
+  &.btn-default:active:focus {
+    background-color: rgba(212, 212, 212, 0.2);
+    transition: all ease-in-out 0.1s;
+  }
 `
 
 export const RouteRowElement = styled.span`
@@ -39,8 +52,6 @@ export const OperatorImg = styled.img`
 `
 
 export const ModeIconElement = styled.span`
-  height: 28px;
-  width: 28px;
   padding-left: inherit;
   padding-top: 3px;
 `
@@ -51,6 +62,12 @@ const RouteNameElement = styled.span`
   font-weight: 400;
   margin-left: 8px;
 `
+const RouteDetailsContainer = styled.div`
+  align-items: center;
+  display: grid;
+  grid-template-columns: repeat(4, auto);
+  justify-items: center;
+`
 const RouteLongNameElement = styled.span`
   font-size: 16px;
   font-weight: 500;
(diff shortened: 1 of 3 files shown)


miles-grant-ibigroup (changes requested):
This is a really good start! Overall I love the new interaction method it feels much smoother. 

Some immediate slight things to adjust, also the Percy tests are failing! I'm guessing they haven't been told yet how to select a pattern direction now which is why they're failing to finish.

Also there are some possible performance improvements, but that can wait for now.

lib/components/viewers/RouteRow.js
```diff
@@ -68,6 +90,31 @@ const SimpleRouteRenderer = styled.span`
   padding: 0;
 `
 
+const PatternButton = styled.button`
+  background: transparent;
+  border: none;
+  border-radius: 50%;
+  color: #004f96;
+  opacity: ${(props) => (props.display ? 100 : 0)};
```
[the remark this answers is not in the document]
amy-corson-ibigroup: Yes, when there isn't an active route the button is hidden and un-focusable, but it remains in the dom tree for screen readers!

lib/components/viewers/RouteRow.js
```diff
@@ -14,20 +16,31 @@ import { getFormattedMode } from '../../util/i18n'
...
+  height: 100%;
+  left: 0;
+  position: absolute;
+  top: 0;
   transition: all ease-in-out 0.1s;
   width: 100%;
+
+  &:active,
+  &.btn-default:active:focus {
+    background-color: rgba(212, 212, 212, 0.2);
```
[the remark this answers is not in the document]
amy-corson-ibigroup: In Bootstrap there's a default `active:focus` state that makes the background color grey, but since now the button is an overlay, it's turning everything grey! I tried to just lower the opacity, but I agree it looks weird, I'm just going to take the whole thing out!

lib/components/viewers/RouteRow.js
```diff
@@ -68,6 +89,31 @@ const SimpleRouteRenderer = styled.span`
...
+  transition: all ease-in-out 0.2s;
+  width: 40px;
+  z-index: ${(props) => (props.display ? 2 : -2)};
+
+  svg {
+    margin-top: -3px;
+    height: 22px !important;
+    width: 22px !important;
+  }
+
```
miles-grant-ibigroup, line 110: There's no focus effect! This lack of feedback is confusing

lib/components/viewers/RouteRow.js
```diff
@@ -138,26 +183,29 @@ export class RouteRow extends PureComponent {
     }
   }
 
-  // Used to ensure details are visible until animation completes
-  setNewHeight = (newHeight) => {
-    this.setState({ newHeight })
+  _patternButtonClicked = () => {
+    const { route, setViewedRoute } = this.props
+    const { id, patterns } = route
+    const firstPattern = route && patterns && Object.values(patterns)[0]?.id
```
miles-grant-ibigroup, line 189:
```suggestion
    const firstPattern = route && patterns && Object.values(patterns)?.[0]?.id
```

Can we guarantee that patterns will be a non-empty object?

lib/components/viewers/RouteRow.js
```diff
@@ -179,16 +227,28 @@ export class RouteRow extends PureComponent {
...
+          onClick={this._onClick}
+          onFocus={this._onFocusOrEnter}
+          onMouseEnter={this._onFocusOrEnter}
+          onTouchStart={this._onFocusOrEnter}
+          title={routeMapToggleText}
+        />
+        <PatternButton
+          aria-label={patternViewerButtonText}
+          display={isActive}
+          id={`pattern-button-${route.shortName || route.longName}`}
```
miles-grant-ibigroup, line 244:
```suggestion
          id={`open-route-button-${route.shortName || route.longName}`}
```

Can we rename this?

miles-grant-ibigroup (changes requested):
Mostly a few nitpicks! Overall very clean thanks for this

miles-grant-ibigroup:
Another issue -- I'm seeing weird behavior when selecting a flex route!

amy-corson-ibigroup:
> Another issue -- I'm seeing weird behavior when selecting a flex route!

@miles-grant-ibigroup is there a route in particular? I looked at a few of the RideNoCo flex routes and I'm not seeing anything odd!

lib/components/viewers/RouteRow.js
amy-corson-ibigroup: Safari does not handle focus on buttons by default, users adjust their keyboard settings to be able to use keyboard navigation: https://www.a11yproject.com/posts/macos-browser-keyboard-navigation/

miles-grant-ibigroup: This makes sense for keyboard navigation, but there is also no styling when I press down on a row item with my mouse!

amy-corson-ibigroup: I can add an `:active` state across all browsers! Would we want to add a state for whether the route is selected as well? This would be in addition to the blue outline in other browsers.

miles-grant-ibigroup: Absolutely!

lib/components/viewers/RouteRow.js
```diff
@@ -68,6 +89,31 @@ const SimpleRouteRenderer = styled.span`
...
+  width: 40px;
+  z-index: ${(props) => (props.display ? 2 : -2)};
+
+  svg {
+    margin-top: -3px;
+    height: 22px !important;
+    width: 22px !important;
+  }
+
+  &:hover {
```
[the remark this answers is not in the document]
amy-corson-ibigroup: Hmm you're not seeing this?
image

lib/components/viewers/RouteRow.js
amy-corson-ibigroup: Okay I overpromised undelivered on this one. After testing a few things, I decided to leave out the `isActive` state, because everything I tried looked very messy on browsers that do have the focus outlines. The `:active` is back though!

lib/components/viewers/RouteRow.js
amy-corson-ibigroup: Good call!

lib/components/viewers/RouteRow.js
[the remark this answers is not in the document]
miles-grant-ibigroup: I see that if I click the arrow button, but not if I click the entire row!

lib/components/viewers/RouteRow.js
miles-grant-ibigroup: I think if we provide our own outline style we can remove the browser-built-in ones!

lib/components/viewers/RouteRow.js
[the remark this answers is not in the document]
miles-grant-ibigroup: All working now thanks

lib/components/viewers/RouteRow.js
```diff
@@ -68,6 +106,31 @@ const SimpleRouteRenderer = styled.span`
...
+    background: #eee;
+    transition: all ease-in-out 0.3s;
```
miles-grant-ibigroup, line 130: Can we speed this up? The delay is annoying when I deselect a row and have to wait for the button to disappear. Also, could we change the function to `ease-out`? I worry that the delayed feedback gives the perception of lag

lib/components/viewers/RouteRow.js
```diff
@@ -14,20 +16,47 @@ import { getFormattedMode } from '../../util/i18n'
...
+
+  &:active,
+  &.btn-default:active:focus {
+    background-color: transparent !important;
+  }
+
+  &:active:before,
+  &.btn-default:active:focus:before {
+    background-color: #eee;
+    transition: all ease-in-out 0.1s;
```
miles-grant-ibigroup, line 58: Isn't this transition inherited?

lib/components/viewers/RouteRow.js
```diff
@@ -68,6 +106,31 @@ const SimpleRouteRenderer = styled.span`
...
+  margin-right: 8px;
+  position: relative;
+  transition: all ease-in-out 0.2s;
+  width: 40px;
+  z-index: ${(props) => (props.display ? 2 : -2)};
+
+  svg {
+    margin-top: -3px;
+    height: 22px !important;
+    width: 22px !important;
```
miles-grant-ibigroup, line 125: this is a massive nitpick, but can we set the width and height on the container instead? I'm getting a weird non-circle highlight
Screenshot 2023-03-29 at 10 56 59 AM

miles-grant-ibigroup (changes requested):
Functionality is perfect! Just a few nitpicks about specific interaction that I want to nail down

lib/components/viewers/RouteRow.js
amy-corson-ibigroup: Ooh good catch. So moving the height and width out of the svg makes it so small, but I fixed this by adding a `min-width` to the button! Let me know if you're still seeing it.

lib/components/viewers/RouteRow.js
```diff
@@ -14,20 +16,47 @@ import { getFormattedMode } from '../../util/i18n'
...
 export const RouteRowButton = styled(Button)`
-  align-items: center;
-  display: flex;
-  padding: 6px;
-  transition: all ease-in-out 0.1s;
+  height: 100%;
+  left: 0;
+  position: absolute;
+  top: 0;
+  transition: all ease-out 0.17s;
```
miles-grant-ibigroup, line 36: Can we bump this all the way up to .1? .17 still feels a bit slow

lib/components/viewers/RouteRow.js
```diff
@@ -68,6 +106,31 @@ const SimpleRouteRenderer = styled.span`
...
+  min-width: 40px;
+  z-index: ${(props) => (props.display ? 2 : -2)};
+
+  svg {
+    margin-top: -3px;
+    height: 22px !important;
+    width: 22px !important;
+  }
+
+  &:hover {
```
miles-grant-ibigroup, line 128: Can we add an `:active` state for this button as well? Maybe `#ddd` or something or actually `#eeeeeeaa` is usually a classic trick

miles-grant-ibigroup: maybe just opacity: 0.8

lib/components/viewers/RouteRow.js
```diff
@@ -138,26 +200,29 @@ export class RouteRow extends PureComponent {
...
   }
 
-  // Used to ensure details are visible until animation completes
-  setNewHeight = (newHeight) => {
-    this.setState({ newHeight })
+  _patternButtonClicked = () => {
+    const { route, setViewedRoute } = this.props
+    const { id, patterns } = route
+    const firstPattern = route && patterns && Object.values(patterns)?.[0]?.id
+    setViewedRoute({ patternId: firstPattern, routeId: id })
```
miles-grant-ibigroup, line 207: Does `setViewedRoute` require a pattern id? It has all the pattern geometry, shouldn't it be able to render all of the patterns at once?

miles-grant-ibigroup (changes requested):
Just a few more small UI nitpicks!

lib/components/viewers/RouteRow.js
amy-corson-ibigroup: Yep! Passing the pattern id is what takes it to the pattern viewer. The `_onClick` for the `RouteRowButton` renders all the patterns.

miles-grant-ibigroup: Oh sorry! Slept right through that one!

miles-grant-ibigroup (approved):
It's ready to go! Looking beautiful

AdrianaCeric (commented):
I actually have a question:
Do we want this arrow in mobile view and why is it now an orange border rather than blue when in a focus state (SEPTA)
image

amy-corson-ibigroup:
I think it makes sense to keep the arrow in mobile! I don't think the button is small enough to be an issue on touchscreens, be we can always adjust the padding down the line if that's not the case. And the yellow focus ring is actually specific to chrome, it's a browser setting. We could override it but that could get tricky across browsers and the yellow doesn't bother me! Let me know if you feel differently.

AdrianaCeric:
@amy-corson-ibigroup you're right on the mobile interaction. Yellow doesn't really bother me either. Thanks for explanation!

PR #835, opened 2023-03-24
Fix icon styling

AdrianaCeric:
**Description:**
- Customize padding and whether highlight on hover occurs -- now configurable in config settings! This is useful for custom/inverted icons that have different behaviour or need custom styling
- Set a max height for route viewer and nearby amenities icons for consistency and to fix large icon size problem

**PR Checklist:**


example-config.yml
```diff
@@ -558,6 +558,10 @@ itinerary:
 #   # Whether to render routes within flex zones of a route's patterns. If set to true,
 #   # routes will not be rendered within flex zones.
 #   hideRouteShapesWithinFlexZones: true
+#   # disable vehicle highlight if necessary (e.g. inverted icons)
+#   noVehicleIconHighlight: true
+#   # customize vehicle icon padding (the default iconPadding is 2 in otp-ui)
```
miles-grant-ibigroup, line 563:
```suggestion
#   # customize vehicle icon padding (the default iconPadding is 2px in otp-ui)
```

I'd make the true size clear!

miles-grant-ibigroup (changes requested):
Just a few nitpicks! Overall everything works great thanks for the changes.

lib/components/map/connected-transit-vehicle-overlay.js
```diff
@@ -102,22 +102,31 @@ function VehicleTooltip(props) {
...
+  display: 'onhover'
+})
 
 // connect to the redux store
 
 const mapStateToProps = (state) => {
   const viewedRoute = state.otp.ui.viewedRoute
   const route = state.otp.transitIndex?.routes?.[viewedRoute?.routeId]
 
+  const PassedIconContainer = state.otp.config?.routeViewer
```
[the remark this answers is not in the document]
AdrianaCeric: I like configured better too

miles-grant-ibigroup (changes requested):
Not seeing the config item flip but otherwise ready to go!

miles-grant-ibigroup (approved):
Works great thanks for the changes

lib/components/map/connected-transit-vehicle-overlay.js
```diff
@@ -102,22 +102,31 @@ function VehicleTooltip(props) {
...
+})
 
 // connect to the redux store
 
 const mapStateToProps = (state) => {
   const viewedRoute = state.otp.ui.viewedRoute
   const route = state.otp.transitIndex?.routes?.[viewedRoute?.routeId]
 
+  const ConfiguredIconContainer =
+    state.otp.config?.routeViewer?.vehicleIconHighlight === false
```
amy-corson-ibigroup, line 126: I think you can replace this with `!state.otp.config?.routeViewer?.vehicleIconHighlight`

amy-corson-ibigroup (commented):
Looks good!

amy-corson-ibigroup (approved):

lib/components/map/connected-transit-vehicle-overlay.js
miles-grant-ibigroup: Careful! That will treat undefined differently than this does

amy-corson-ibigroup: ok! Feel free to disregard!

AdrianaCeric: Good catch, but yeah this is super confusing and special ✨ Since we're setting `vehicleIconHighlight` as `true` or `false` in the config, we want to check for false specifically, and distinguish the other falsy values like `""`, `0`, `null`,  and `undefined` as *not* false

PR #838, opened 2023-03-27
Repair a11y Itinerary Regressions

miles-grant-ibigroup:
**Description:**

The "other" category always showed up, whether or not single-mode itineraries were present. This PR repairs this, as well as adjusts the string used.

**PR Checklist:**
- [X] Does the code follow accessibility standards (WCAG 2.1 AA Compliant)?
- [X] Are all languages supported (Internationalization/Localization)?
- [X] Are appropriate Typescript types implemented?

### DIFF
--- a/i18n/en-US.yml
+++ b/i18n/en-US.yml
@@ -265,7 +265,7 @@ components:
   DefaultItinerary:
     clickDetails: Click to view details
     multiModeSummary: "{accessMode} + {transitMode}"
-    nonTransit: Without Transit
+    nonTransit: Alternative options
   DeleteUser:
     deleteMyAccount: Delete my account
   EnhancedStopMarker:
--- a/i18n/en-US.yml
+++ b/i18n/en-US.yml
@@ -265,6 +265,7 @@ components:
   DefaultItinerary:
     clickDetails: Click to view details
     multiModeSummary: "{accessMode} + {transitMode}"
+    nonTransit: Without Transit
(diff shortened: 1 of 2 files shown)


AdrianaCeric (approved):
I like this, good fix!

Thinking if "Single-mode" would be a better way to describe "Without Transit", but I'm fine with either

miles-grant-ibigroup:
Am open to other ideas, but I'm trying to avoid saying "mode". it's an OTP term that most people don't use!

amy-corson-ibigroup (commented):
Great fix! I'll be honest  "without transit" feels a little clunky? I still like "Other" but also maybe "Non-Transit" or "Other Options"? Maybe other folks have ideas?

miles-grant-ibigroup:
I agree! I originally had non-transit but didn't like it because the negation was confusing... I still think no header at all and making it aria-label only is the way to go, but I think we decided we couldn't do that for some reason

amy-corson-ibigroup:
Can't remember where I stood on that decision before so maybe I'm contradicting myself but I think you're right that no label would be fine? Aria-label for screen readers, but for visual users the different grouping of the smaller buttons seems like enough.

miles-grant-ibigroup:
Is transit ever actually inferior? Driving will always be more expensive and if a user can't afford a transit fare they probably won't be using a transit trip planner. Either way I like "alternative options" as a hidden label. Will implement that!

PR #842, opened 2023-03-31
fix(route-viewer): support fuzzy matching for non-numeric routes

miles-grant-ibigroup:
**Description:**

This PR adjusts the route search filtering to support fuzzy searching when the route doesn't have numbers. The library is the same one that's already included with otp-ui, so no bundle size increase.

**PR Checklist:**


lib/util/state.js
```diff
@@ -585,6 +586,24 @@ const getAgencyIdsForFilter = (name, transitOperators) => {
     .map((x) => x.agencyId)
 }
 
+/**
+ * Determine if a string matches a query. Uses direct matching if the string is only numbers,
+ * uses fuzzy matching otherwise.
+ */
+const isFilterMatch = (string, query) => {
+  const normalizedQuery = query.toLowerCase()
+  const normalizedString = string.toLowerCase()
```
amy-corson-ibigroup, line 595: I'm wondering could we also filter dashes and spaces out of the search query + results? Just looking at the "1 line" search I know this PR inspired, and the results don't seem quite as intuitive as we'd hope. Removing spaces and punctuation could also allow for upping the similarity ratio from .4 to something a little higher? 
image

amy-corson-ibigroup: Also so glad we talked about the correlation between hovering and these filters today, I thought I was going insane while reviewing this.

amy-corson-ibigroup (approved):
Great add! Looks solid to me, just a quick question 👍

lib/util/state.js
amy-corson-ibigroup: Sure sure, to be clear I think this works great as-is! But right now for this search, every result has a match of `.75`. If you remove spaces and dashes, `1-Line` has a match of `1`, but `T-Line` would still be `.75`

amy-corson-ibigroup: Although I guess then you'd have to sort the results based on this percentage

miles-grant-ibigroup: Yeah this issue is part of the reason why I was against doing fuzzy search way back when this was originally built. I think a text similar algorithm that could be tweaked for each set of routes might get the results that we're looking for, but yet another instance-specific method like that doesn't make me happy. I think our solution needs to be general

lib/util/state.js
```diff
@@ -603,14 +622,9 @@ export const getFilteredRoutes = createSelector(
         (!filter.mode || filter.mode === route.mode) &&
         // If user search is active, filter by either the long or short name
```
binh-dam-ibigroup, line 623: Update this comment to include any route description.

binh-dam-ibigroup (approved):
Cool stuff. Please add some examples in JSDoc.

lib/util/state.js
```diff
@@ -603,14 +622,9 @@ export const getFilteredRoutes = createSelector(
...
         (!filter.search ||
-          (route.longName &&
-            route.longName
-              .toLowerCase()
-              .includes(filter.search.toLowerCase())) ||
-          (route.shortName &&
-            route.shortName
-              .toLowerCase()
-              .includes(filter.search.toLowerCase())))
+          (route.longName && isFilterMatch(route.longName, filter.search)) ||
```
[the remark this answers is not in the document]
miles-grant-ibigroup: It shouldn't! OR is commutative is it not

PR #844, opened 2023-03-31
Add transitions between route and pattern viewer

amy-corson-ibigroup:
Adds subtle animated transitions from route to pattern viewer including content fade in/out and background color transition.
Creates a util to check for prefers-reduced-motion browsers.

**PR Checklist:**

### DIFF
--- a/lib/components/viewers/RouteRow.js
+++ b/lib/components/viewers/RouteRow.js
@@ -184,16 +184,11 @@ export class RouteRow extends PureComponent {
     }
   }
 
-  _patternButtonClicked = () => {
-    const { route, setViewedRoute } = this.props
+  render() {
+    const { intl, isActive, onPatternButtonClick, operator, route } = this.props
+    const { ModeIcon, RouteRenderer } = this.context
     const { id, patterns } = route
     const firstPattern = route && patterns && Object.values(patterns)[0]?.id
-    setViewedRoute({ patternId: firstPattern, routeId: id })
-  }
-
-  render() {
-    const { intl, isActive, operator, route } = this.props
-    const { ModeIcon, RouteRenderer } = this.context
 
     const patternViewerButtonText = intl.formatMessage({
       description: 'identifies the purpose of the pattern viewer button',
@@ -242,7 +237,7 @@ export class RouteRow extends PureComponent {
         <PatternButton
           aria-label={patternViewerButtonText}
           display={isActive}
-          onClick={this._patternButtonClicked}
+          onClick={() => onPatternButtonClick(firstPattern, id)}
           // Cannot keyboard navigate to the button unless it is visible
           tabIndex={isActive ? 0 : -1}
           title={patternViewerButtonText}
--- a/lib/components/viewers/route-viewer.js
+++ b/lib/components/viewers/route-viewer.js
@@ -26,10 +26,12 @@ import {
   getSortedFilteredRoutes
 } from '../../util/state'
 import { getFormattedMode } from '../../util/i18n'
+import { prefersReducedMotion } from '../../util/a11y'
 import { StyledIconWrapper } from '../util/styledIcon'
 import PageTitle from '../util/page-title'
 
 import { RouteName, RouteRow } from './RouteRow'
+import { StyledRouteViewer } from './styled'
 import RouteDetails from './route-details'
 
 class RouteViewer extends Component {
@@ -56,6 +58,7 @@ class RouteViewer extends Component {
   }
 
   state = {
+    fade: false,
     /** Used to track if all routes have been rendered */
     initialRender: true
   }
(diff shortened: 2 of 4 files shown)


lib/components/viewers/RouteRow.js
```diff
@@ -262,7 +257,7 @@ export class RouteRow extends PureComponent {
           aria-label={patternViewerButtonText}
           display={isActive}
           id={`open-route-button-${route.shortName || route.longName}`}
-          onClick={this._patternButtonClicked}
+          onClick={() => onPatternButtonClick(firstPattern, id)}
```
binh-dam-ibigroup, line 260: Consider making the event handler a class method to avoid this function to be recreated on each render.

lib/components/viewers/styled.js
```diff
@@ -1,18 +1,58 @@
...
+100 { opacity: 100%; }
+`
+
+export const StyledRouteViewerContainer = styled.div`
+  animation: ${(props) => props.patternView && backgroundColorTransition} 150ms
+      linear,
+    ${(props) => (props.fade ? fade : 'none')} 100ms reverse linear forwards;
+  background-color: ${(props) =>
+    props.routeColor ? props.routeColor : '#fff'};
+  color: ${(props) => (props.textColor ? props.textColor : 'inherit')};
```
binh-dam-ibigroup, line 31: This and the other props can be written as:
```suggestion
  color: ${(props) => (props.textColor || 'inherit')};
```

lib/components/viewers/styled.js
```diff
@@ -1,18 +1,58 @@
...
+    100 { opacity: 100%; }
+`
+const fadeDelay = keyframes`
+0% { opacity: 0%; }
+50% { opacity: 0% }
+100 { opacity: 100%; }
+`
+
+export const StyledRouteViewerContainer = styled.div`
+  animation: ${(props) => props.patternView && backgroundColorTransition} 150ms
```
binh-dam-ibigroup, line 26: This sounds like it should be a `transition` attribute on `background-color`, the original state being the white of the original render and the ending state being the background color of the route. Also, `transition` includes a delay parameter that can replace the delay emulation you added in the JS code.

lib/components/viewers/styled.js
```diff
@@ -1,18 +1,58 @@
...
+  display: flex;
+  flex-direction: column;
+  flex-flow: column;
+  height: 100%;
+
+  & > * {
+    animation: ${(props) => props.patternView && fadeDelay} 300ms linear
+      forwards;
+
+    @media (prefers-reduced-motion) {
```
binh-dam-ibigroup, line 42: Clever consideration.

lib/components/viewers/styled.js
```diff
@@ -1,18 +1,58 @@
...
+  animation: ${(props) => props.patternView && backgroundColorTransition} 150ms
+      linear,
+    ${(props) => (props.fade ? fade : 'none')} 100ms reverse linear forwards;
+  background-color: ${(props) =>
+    props.routeColor ? props.routeColor : '#fff'};
+  color: ${(props) => (props.textColor ? props.textColor : 'inherit')};
+  fill: ${(props) => (props.textColor ? props.textColor : 'inherit')};
+  display: flex;
+  flex-direction: column;
+  flex-flow: column;
```
binh-dam-ibigroup, line 35: What is the use of these attributes? (Block elements naturally flow vertically.)

binh-dam-ibigroup (commented):
Cool effect. Could you try using the `transition` attribute (or add a comment why that doesn't work)?

lib/components/viewers/styled.js
amy-corson-ibigroup: Good catch thank you! I originally copied them from the `.route-viewer` CSS but then decided just to add the `className` and forgot to remove.

lib/components/viewers/styled.js
amy-corson-ibigroup: Good call! Not sure why I jumped straight to `keyframes`, `transitions` work just as well. Thanks!

lib/components/viewers/route-viewer.js
```diff
@@ -176,20 +191,21 @@ class RouteViewer extends Component {
       const textColor = getMostReadableTextColor(routeColor, route?.textColor)
       const fill = vehicleIconHighlight === false ? null : textColor
 
+      setTimeout(() => this.setState({ fade: false }), 15)
```
binh-dam-ibigroup, line 194: Move this action out of the render method. (You don't want to allocate resources during `render`, and who knows how many times `render` gets called.)

binh-dam-ibigroup (changes requested):
Move timeout creation out of the `render` method. Also nice-to-have: can we get rid of all the timeouts and `fade`?

lib/components/viewers/route-viewer.js
```diff
@@ -147,6 +150,18 @@ class RouteViewer extends Component {
     this.props.setRouteViewerFilter({ search: value })
   }
 
+  onPatternButtonClick = (firstPattern, routeId) => {
+    const { setViewedRoute } = this.props
+    this.setState({ fade: true })
+    prefersReducedMotion
+      ? setViewedRoute({ patternId: firstPattern, routeId })
+      : setTimeout(
```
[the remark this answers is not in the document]
amy-corson-ibigroup: Without them, the component unmounts before the opacity transition occurs. Is there a way to delay without the `setTimeout`?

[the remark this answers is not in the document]
binh-dam-ibigroup: Ah good point. Then keep it.

binh-dam-ibigroup (approved):
LGTM, thanks for the tweaks!

amy-corson-ibigroup:
Closing this  in favor of a fresh start but will reference it moving forward for route to pattern viewer transitions

PR #845, opened 2023-04-04
fix(route-viewer): always fetch route type

miles-grant-ibigroup:
**Description:**

This PR fixes the bug where routes would jump around on hover. The issue is the routing sort included `type` which wasn't initially fetched. As it's very inexpensive to fetch this extra field, this is done in the initial route call, which both improves the sort and ensures that there is no jumping around on hover

**PR Checklist:**

### DIFF
--- a/lib/actions/apiV2.js
+++ b/lib/actions/apiV2.js
@@ -641,6 +641,7 @@ export function findRoutes() {
             longName
             shortName
             mode
+            type
             color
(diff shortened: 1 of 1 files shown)


binh-dam-ibigroup (approved):
LGTM!

AdrianaCeric (approved):
Great fix!

PR #849, opened 2023-04-06
refactor: upgrade trip-details, use its translation strings

miles-grant-ibigroup:
**Description:**

Makes use of https://github.com/opentripplanner/otp-ui/pull/569. Uses translation strings from otp ui instead of duplicating them.

**PR Checklist:**


lib/components/narrative/default/default-itinerary.js
```diff
@@ -123,9 +123,7 @@ const ITINERARY_ATTRIBUTES = [
       const fareCurrency = itinerary.fare?.fare?.regular?.currency?.currencyCode
       const fare = fareInCents === null ? null : fareInCents / 100
```
binh-dam-ibigroup, line 124: I suppose this logic could be reused in the trip-details component to handle cases where cents show as `-1`. (If that's the case, trip details and the itinerary summary render different things).

binh-dam-ibigroup (approved):
One small change.

lib/components/narrative/default/default-itinerary.js
binh-dam-ibigroup: Noted in https://github.com/opentripplanner/otp-ui/issues/576.

miles-grant-ibigroup: Good call thanks. Yes I think this should be in core-utils eventually

AdrianaCeric (approved):
This is a great change, good catch

PR #850, opened 2023-04-06
Avoid Route Viewer Crash on Refresh

miles-grant-ibigroup:
**Description:**

Currently clicking on a route viewer route and reloading the page causes a crash. This PR fixes this.

**PR Checklist:**


miles-grant-ibigroup:
That looks correct to me. I'm only noticing it on some routes where the agency id contains non-numbers

binh-dam-ibigroup (approved):
Good fix, I was able to reproduce using a call-taker config.

PR #859, opened 2023-04-13
Clarify Links That Open in New Window

amy-corson-ibigroup:
**Description:**
Creates both the `NewWindowIconA11y` and `LinkOpensNewWindow`: 
-  `NewWindowIconA11y` uses the ExternalLink icon from styled-icons and passes it a title and alt text for screen readers.
- `LinkOpensNewWindow` takes `NewWindowIconA11y` and wraps it in an anchor tag with the `target="_blank"` attribute.

Applies the two above elements to external links across the site as needed. 

**PR Checklist:**

| Before | After |
|--------|-------|
| image | image|

### DIFF
--- a/i18n/en-US.yml
+++ b/i18n/en-US.yml
@@ -154,6 +154,7 @@ common:
     tripDurationFormat: >-
       {hours, plural, =0 {} other {# hr }}{minutes} min { seconds, plural, =0 {}
       other {# sec}}
+  linkOpensNewWindow: "(Opens new window)"
 components:
   A11yPrefs:
     accessibilityRoutingByDefault: Prefer accessible trips by default
--- a/lib/components/admin/field-trip-details.js
+++ b/lib/components/admin/field-trip-details.js
@@ -28,6 +28,7 @@ import {
   TICKET_TYPES
 } from '../../util/call-taker'
 import { IconWithText, StyledIconWrapper } from '../util/styledIcon'
+import { NewWindowIconA11y } from '../util/externalLink'
 
 import {
   Bold,
@@ -42,7 +43,6 @@ import {
   Text,
   Val
 } from './styled'
-import { NewWindowIconA11y } from '../util/externalLink'
 import DraggableWindow from './draggable-window'
 import EditableSection from './editable-section'
 import FieldTripNotes from './field-trip-notes'
@@ -99,7 +99,7 @@ class FieldTripDetails extends Component {
     const { request, sessionId } = this.props
     const cancelled = request.status === 'cancelled'
     const printFieldTripLink = `/#/printFieldTrip/?requestId=${request.id}&sessionId=${sessionId}`
-    const newWindowIconWithStyles = (
+    const StyledNewWindowIcon = () => (
       <NewWindowIconA11y size={10} style={{ margin: '-3px 0 0 4px' }} />
     )
     return (
@@ -116,15 +116,15 @@ class FieldTripDetails extends Component {
             target="_blank"
           >
             <IconWithText Icon={CommentDots}>Feedback link</IconWithText>
-            {newWindowIconWithStyles}
+            {StyledNewWindowIcon}
           </MenuItem>
           <MenuItem href={this._getRequestLink('receipt')} target="_blank">
             <IconWithText Icon={FileAlt}>Receipt link</IconWithText>
-            {newWindowIconWithStyles}
+            {StyledNewWindowIcon}
           </MenuItem>
           <MenuItem href={printFieldTripLink} target="_blank">
             <IconWithText Icon={Print}>Printable trip plan</IconWithText>
-            {newWindowIconWithStyles}
+            {StyledNewWindowIcon}
           </MenuItem>
         </DropdownButton>
         <Button
--- a/lib/components/admin/field-trip-details.js
+++ b/lib/components/admin/field-trip-details.js
@@ -99,7 +99,7 @@ class FieldTripDetails extends Component {
     const { request, sessionId } = this.props
     const cancelled = request.status === 'cancelled'
     const printFieldTripLink = `/#/printFieldTrip/?requestId=${request.id}&sessionId=${sessionId}`
-    const newWindowIconStyles = (
+    const newWindowIconWithStyles = (
       <NewWindowIconA11y size={10} style={{ margin: '-3px 0 0 4px' }} />
     )
     return (
@@ -116,15 +116,15 @@ class FieldTripDetails extends Component {
             target="_blank"
           >
             <IconWithText Icon={CommentDots}>Feedback link</IconWithText>
-            {newWindowIconStyles}
+            {newWindowIconWithStyles}
           </MenuItem>
           <MenuItem href={this._getRequestLink('receipt')} target="_blank">
             <IconWithText Icon={FileAlt}>Receipt link</IconWithText>
-            {newWindowIconStyles}
+            {newWindowIconWithStyles}
           </MenuItem>
           <MenuItem href={printFieldTripLink} target="_blank">
             <IconWithText Icon={Print}>Printable trip plan</IconWithText>
-            {newWindowIconStyles}
+            {newWindowIconWithStyles}
           </MenuItem>
         </DropdownButton>
         <Button
--- a/lib/components/admin/field-trip-details.js
+++ b/lib/components/admin/field-trip-details.js
@@ -116,15 +116,15 @@ class FieldTripDetails extends Component {
             target="_blank"
           >
             <IconWithText Icon={CommentDots}>Feedback link</IconWithText>
-            {StyledNewWindowIcon}
+            <StyledNewWindowIcon />
           </MenuItem>
           <MenuItem href={this._getRequestLink('receipt')} target="_blank">
             <IconWithText Icon={FileAlt}>Receipt link</IconWithText>
-            {StyledNewWindowIcon}
+            <StyledNewWindowIcon />
           </MenuItem>
           <MenuItem href={printFieldTripLink} target="_blank">
             <IconWithText Icon={Print}>Printable trip plan</IconWithText>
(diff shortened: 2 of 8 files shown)


lib/components/viewers/route-details.js
```diff
@@ -145,20 +144,15 @@ class RouteDetails extends Component {
               </>
             )}
             {url && (
-              <a
-                href={url}
-                rel="noreferrer"
+              <LinkOpensNewWindow
+                contents={
+                  <FormattedMessage id="components.RouteDetails.moreDetails" />
```
AdrianaCeric, line 149: (From WAVE a11y tool) "More Details" can be ambiguous as per [WCAG](https://www.w3.org/TR/WCAG21/#link-purpose-in-context)
image

Maybe we can change it to something like "View Agency Website"?

AdrianaCeric (approved):
Overall awesome!!!

lib/components/admin/field-trip-list.js
```diff
@@ -185,8 +186,12 @@ class FieldTripList extends Component {
               type="date"
               value={date}
             />
-            <Button bsSize="xsmall" href={this._getReportUrl()} target="_blank">
-              View report
+            <Button bsSize="xsmall">
+              <LinkOpensNewWindow
+                contents="View report"
```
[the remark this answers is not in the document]
amy-corson-ibigroup: Good question, I'm not sure if there's been any discussion of i18ning Field Trip, but either way I think that would call for another PR! Most of the strings in this file are english by default.

i18n/en-US.yml
```diff
@@ -154,6 +154,7 @@ common:
     tripDurationFormat: >-
       {hours, plural, =0 {} other {# hr }}{minutes} min { seconds, plural, =0 {}
       other {# sec}}
+  linkOpensNewWindow: "(Opens new window)"
```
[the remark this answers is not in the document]
amy-corson-ibigroup: I thought maybe it would be nice for both sighted and AT users to have this broken up a bit with punctuation! Otherwise you get lines like "go to agency site opens new window" which feels a bit rough.

lib/components/util/externalLink.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+export const LinkOpensNewWindow = ({
+  contents,
+  gap = 6,
+  inline,
+  size = 14,
+  style,
+  url
+}: LinkProps): JSX.Element => {
+  console.log(size)
```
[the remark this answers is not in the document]
amy-corson-ibigroup: AH

lib/components/user/terms-of-use-pane.js
```diff
@@ -1,44 +1,50 @@
...
+import {
+  TERMS_OF_SERVICE_PATH,
+  TERMS_OF_STORAGE_PATH
+} from '../../util/constants'
 
 /**
  * User terms of use pane.
  */
+/* eslint-disable react/prop-types */
+/* TODO: Prop Types */
```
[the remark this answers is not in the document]
amy-corson-ibigroup: So the reason for the `eslint-disable` and the TODO is because the CI tests were failing! You can see them on [6267208](https://github.com/opentripplanner/otp-react-redux/pull/859/commits/6267208024300eb070dbcf60ed88a1a99e3bed7c). I wanted to impact this file as little as possible outside of the external links, but I suppose if we're already in there and we can change the types, maybe we should? Not sure what the best practice is so I'll defer to 2nd reviewer!!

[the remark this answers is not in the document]
miles-grant-ibigroup: Yes we generally typescript non typescript files whenever we make major changes to them

lib/components/admin/field-trip-list.js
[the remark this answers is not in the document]
miles-grant-ibigroup: Field trip does not need to be internationalized

lib/components/util/externalLink.tsx
```diff
@@ -0,0 +1,63 @@
...
+  return (
+    <a
+      href={url}
+      rel="noreferrer"
+      style={{
+        alignItems: 'center',
+        display: `${isInline ? 'inline-flex' : 'flex'}`,
+        flexDirection: 'row',
+        gap: `${gap}px`,
+        textDecoration: 'underline',
```
miles-grant-ibigroup, line 53: Is this intentional? It's always underlined and there's no hover effect. This is really confusing and doesn't behave the way I'd expect a link to behave. Is there any way to make it only underline on hover?

lib/components/util/externalLink.tsx
```diff
@@ -0,0 +1,63 @@
...
+}: LinkProps): JSX.Element => {
+  return (
+    <a
+      href={url}
+      rel="noreferrer"
+      style={{
+        alignItems: 'center',
+        display: `${isInline ? 'inline-flex' : 'flex'}`,
+        flexDirection: 'row',
+        gap: `${gap}px`,
```
miles-grant-ibigroup, line 52:
```suggestion
        gap,
```
This should work!

lib/components/util/externalLink.tsx
```diff
@@ -0,0 +1,63 @@
...
+        textDecoration: 'underline',
+        whiteSpace: 'nowrap',
+        ...style
+      }}
+      target="_blank"
+    >
+      {contents}
+      <NewWindowIconA11y size={size} />
```
miles-grant-ibigroup, line 60: We generally put the icon before links. Why is it flipped here?

miles-grant-ibigroup (changes requested):
Some initial comments for now -- I agree "more details" is a bit of a problem in the route details. Could we change it to "route details"?

lib/components/util/externalLink.tsx
```diff
@@ -0,0 +1,63 @@
+import { ExternalLinkAlt } from '@styled-icons/fa-solid'
+import { useIntl } from 'react-intl'
+import React, { HTMLAttributes } from 'react'
+
+interface LinkProps extends HTMLAttributes<HTMLElement> {
+  contents: JSX.Element | string
+  gap?: number
+  isInline?: boolean
```
miles-grant-ibigroup, line 8: This should be changed to `inline`

miles-grant-ibigroup (changes requested):
One more nitpick

lib/components/viewers/route-details.js
amy-corson-ibigroup: @miles-grant-ibigroup @AdrianaCeric spoke to the external a11y auditors today and they seemed to think this was alright for sighted users, since it is the only link of its kind on this page, but that we may want to include an invisible span with additional details for AT. 

I think I'd personally prefer keeping "more details" with the external site icon & invisible span as opposed to changing the link text, but I wanted to get both of your thoughts!

lib/components/util/externalLink.tsx
amy-corson-ibigroup: I think it reads a little easier to see the title of the link before the indication that it'll open in a new window!

lib/components/admin/field-trip-details.js
```diff
@@ -99,7 +99,7 @@ class FieldTripDetails extends Component {
     const { request, sessionId } = this.props
     const cancelled = request.status === 'cancelled'
     const printFieldTripLink = `/#/printFieldTrip/?requestId=${request.id}&sessionId=${sessionId}`
-    const newWindowIconWithStyles = (
+    const StyledNewWindowIcon = () => (
       <NewWindowIconA11y size={10} style={{ margin: '-3px 0 0 4px' }} />
     )
```
miles-grant-ibigroup: Wait this still isn't a component! I'm actually confused how this works at all -- the definition looks ok but the new component should be used like `<StyledNewWindowIcon />`

lib/components/user/new-account-wizard.js
```diff
@@ -6,7 +6,7 @@ import AccountSetupFinishPane from './account-setup-finish-pane'
 import FavoritePlaceList from './places/favorite-place-list'
 import NotificationPrefsPane from './notification-prefs-pane'
 import SequentialPaneDisplay from './sequential-pane-display'
-import TermsOfUsePane from './terms-of-use-pane'
+import TermsOfUsePane from './terms-of-use-pane.tsx'
```
miles-grant-ibigroup, line 9: I don't think the tsx is needed!

lib/components/user/terms-of-use-pane.tsx
```diff
@@ -77,10 +84,10 @@ const TermsOfUsePane = ({
           <FormattedMessage
             id="components.TermsOfUsePane.termsOfStorageStatement"
             values={{
-              termsOfStorageLink: (contents) => (
+              termsOfStorageLink: (contents: any) => (
```
miles-grant-ibigroup, line 87: I believe `contents` is `JSX.Element`

miles-grant-ibigroup (changes requested):
Looking better! Just a bit more cleanup

lib/components/viewers/route-details.js
AdrianaCeric: I think `InvisibleA11yLabel` could work here too! I'm not too biased towards either

miles-grant-ibigroup (approved):
Great change! Just one last nitpick and we're ready to merge

lib/components/user/terms-of-use-pane.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { LinkOpensNewWindow } from '../util/externalLink'
+import {
+  TERMS_OF_SERVICE_PATH,
+  TERMS_OF_STORAGE_PATH
+} from '../../util/constants'
+
+/**
+ * User terms of use pane.
+ */
+/* TODO: Prop Types */
```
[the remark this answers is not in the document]
amy-corson-ibigroup: good catch thank you!

PR #860, opened 2023-04-14
Add search route input label

amy-corson-ibigroup:
**Description:**

Route search input was previously un-labled. This adds an aria-label for ATs.

**PR Checklist:**


AdrianaCeric (approved):
Good catch!

binh-dam-ibigroup (approved):

PR #861, opened 2023-04-14
Fix Location Field Invalid State

binh-dam-ibigroup:
## Description

This PR initializes location field invalid states to false, and lets the location fields validate after clicking Plan Trip or after modifying fields.

**PR Checklist:**


lib/components/form/connect-location-field.js
```diff
@@ -61,7 +62,11 @@ export default function connectLocationField(
     const activeSearch = getActiveSearch(state)
     const query = activeSearch ? activeSearch.query : currentQuery
 
-    const isValid = hasValidLocation(query, ownProps.locationType)
+    const isValid =
+      !ownProps.selfValidate ||
+      (includeLocation
+        ? hasValidLocation(query, ownProps.locationType)
+        : !isBlank(ownProps.location?.name))
```
amy-corson-ibigroup, line 69: What about if `isRequired` is false? Does that impact the field validation?

amy-corson-ibigroup (approved):
Looks good to me! Just a couple questions:

lib/components/mobile/batch-search-screen.tsx
```diff
@@ -39,19 +46,23 @@ class BatchSearchScreen extends Component<Props> {
         <main tabIndex={-1}>
           <div className="batch-search-settings mobile-padding">
             <LocationField
+              isRequired
```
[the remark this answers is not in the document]
binh-dam-ibigroup: I am also unsure about this part of the mobile workflow. Technically, on this screen, there is no form (and no form submit button) involved, so my thought was `isRequired` does not apply. But when you pick a location (or go back), then `is Required` is true on both location field components.

lib/components/app/batch-routing-panel.tsx
```diff
@@ -12,14 +11,30 @@ import SwitchButton from '../form/switch-button'
...
+}
+
 /**
  * Main panel for the batch/trip comparison form.
  */
-class BatchRoutingPanel extends Component {
-  handleSubmit = (e) => e.preventDefault()
+class BatchRoutingPanel extends Component<Props> {
+  state = {
+    planTripClicked: false
```
miles-grant-ibigroup, line 26: Do we need to create a new state for this? Is it not possible to determine this info from the redux store? Also, does this handle cases where a trip is loaded in via url parameters?

miles-grant-ibigroup (changes requested):
Good changes, I just have a quick question about how we're storing the "trip planned" state

lib/components/app/batch-routing-panel.tsx
binh-dam-ibigroup: My understanding is that the user must have clicked the Plan Trip button first before the invalid states are activated, regardless of the values in the fields. It didn't make sense to me at this time to add a global redux state for that. I don't see a way to track that the Plan Trip button has been clicked from the existing redux state.

If a URL is entered in the browser with one of the locations missing, the initial `aria-invalid` will still be `false` for both locations, until the user clicks Plan Trip.

miles-grant-ibigroup: That doesn't sound right to me but I trust you.

miles-grant-ibigroup (approved):
Working well thanks for the cleanup!

PR #865, opened 2023-04-19
Adjust font size for longer route short names

amy-corson-ibigroup:
Adds a check for long words inside `routeShortName`s, and adjust the font size accordingly, to avoid clipping.

**PR Checklist:**

| Before | After |
|--------|-------|
|image|image|


AdrianaCeric (approved):
Def think this is a great change, especially for words with the hyphens

binh-dam-ibigroup (approved):

PR #868, opened 2023-04-26
Navbar Agency Logo 2.0

AdrianaCeric:
**Description:**
Navbar agency logo can now be configured to either redirect to a specified link or refresh the page!

**Testing:**
Add The following to a configuration to test for refresh:
```
navBar:
  logoClickAction: start-over
```
And the following to redirect to new website
```
navBar:
  logoClickAction: https://google.com
```

**PR Checklist:**

### DIFF
--- a/i18n/en-US.yml
+++ b/i18n/en-US.yml
@@ -209,6 +209,7 @@ components:
         other {# spaces available}
       }
   AppMenu:
+    agencyLogoUrl: Home page of {agencyName} (External site)
     callHistory: Call History
     closeMenu: Close Menu
     fieldTrip: Field Trip
--- a/i18n/en-US.yml
+++ b/i18n/en-US.yml
@@ -209,8 +209,9 @@ components:
         other {# spaces available}
       }
   AppMenu:
-    agencyLogoReset: Reset trip search and return to "Plan Trip" page
-    agencyLogoResetWarning: This will reset your trip search.
+    agencyLogoReset: '{agencyName}: Reset trip search and return to "Plan Trip" page'
+    agencyLogoResetWarning: This action will reset your trip search.
+    agencyLogoRedirectWarning: This action will redirect you to the {agencyName} website.
     agencyLogoUrl: Home page of {agencyName} (External site)
     callHistory: Call History
     closeMenu: Close Menu
--- a/i18n/en-US.yml
+++ b/i18n/en-US.yml
@@ -209,6 +209,8 @@ components:
         other {# spaces available}
       }
   AppMenu:
+    agencyLogoReset: Reset trip search and return to "Plan Trip" page
+    agencyLogoResetWarning: This will reset your trip search.
     agencyLogoUrl: Home page of {agencyName} (External site)
     callHistory: Call History
     closeMenu: Close Menu
--- a/lib/components/app/desktop-nav.tsx
+++ b/lib/components/app/desktop-nav.tsx
@@ -73,14 +73,30 @@ const DesktopNav = ({
   } = otpConfig
   const showLogin = Boolean(getAuth0Config(persistence))
   const intl = useIntl()
+  const isExternalLink = logoClickAction?.startsWith('http')
+  const willStartOver = logoClickAction === 'start-over'
 
-  const handleStartOver = () => {
+  const handleClick = () => {
     if (
+      willStartOver &&
       window.confirm(
-        intl.formatMessage({ id: 'components.AppMenu.agencyLogoResetWarning' })
+        intl.formatMessage(
+          { id: 'components.AppMenu.agencyLogoResetWarning' },
+          { agencyName: operatorName || title }
+        )
       )
     ) {
       window.location.href = startOver(reactRouterConfig?.basename)
+    } else if (
+      isExternalLink &&
+      window.confirm(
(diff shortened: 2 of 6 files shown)


AdrianaCeric:
Apparently I don't know how to spell the word `parameter`

amy-corson-ibigroup (changes requested):
This looks really smooth! Thank you so much for adding. Just a few thoughts.

I'm not sure about allowing the logo to link to an external site. The logo directing to home is [standard practice](https://www.nngroup.com/articles/homepage-links/) across almost all webpages. Even if the trip planner is embedded within an agency site, I think this would be unexpected behavior. Do you have a use case in mind?

lib/components/app/desktop-nav.tsx
```diff
@@ -55,9 +95,24 @@ const DesktopNav = ({ otpConfig, popupTarget, setPopupContent }: Props) => {
               className={branding && `with-icon icon-${branding}`}
               style={{ marginLeft: 50 }}
             >
-              {/* A title is always rendered (e.g.for screen readers)
-                  but is visually-hidden if a branding icon is used. */}
               <div className="navbar-title">{title}</div>
+              {logoClickAction === 'start-over' && (
+                <TransparentButton bsStyle="link" onClick={handleStartOver}>
+                  <InvisibleA11yLabel>
+                    <FormattedMessage id="components.BatchSearchScreen.header" />
```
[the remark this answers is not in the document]
AdrianaCeric: Maybe "Reset and return to home page"?

lib/components/app/desktop-nav.tsx
```diff
@@ -55,9 +96,28 @@ const DesktopNav = ({ otpConfig, popupTarget, setPopupContent }: Props) => {
...
+                  <InvisibleA11yLabel>
+                    <FormattedMessage id="components.AppMenu.agencyLogoReset" />
+                  </InvisibleA11yLabel>
+                </TransparentButton>
+              )}
+              {logoClickAction?.startsWith('http') && (
+                <TransparentButton
+                  bsStyle="link"
+                  href={logoClickAction}
+                  title="components.AppMenu.agencyLogoUrl"
```
amy-corson-ibigroup, line 111: Missing an `intl.formatMessage` here!

lib/components/app/desktop-nav.tsx
```diff
@@ -55,9 +96,28 @@ const DesktopNav = ({ otpConfig, popupTarget, setPopupContent }: Props) => {
...
+              {logoClickAction === 'start-over' && (
+                <TransparentButton bsStyle="link" onClick={handleStartOver}>
+                  <InvisibleA11yLabel>
+                    <FormattedMessage id="components.AppMenu.agencyLogoReset" />
+                  </InvisibleA11yLabel>
+                </TransparentButton>
+              )}
+              {logoClickAction?.startsWith('http') && (
+                <TransparentButton
+                  bsStyle="link"
```
amy-corson-ibigroup, line 109: Just realizing that these buttons technically work as links, which means for a11y purposes the `TransparentButton` component needs to be changed to an anchor tag (this would be ideal!), or these two buttons need to have a `role="link"`!

i18n/en-US.yml
```diff
@@ -209,6 +209,9 @@ components:
         other {# spaces available}
       }
   AppMenu:
+    agencyLogoReset: Reset trip search and return to "Plan Trip" page
```
amy-corson-ibigroup, line 212: I do still think there needs to be a `{site title}` here, so that AT users know they're on the site logo and not just a reset trip button!

lib/components/app/desktop-nav.tsx
```diff
@@ -55,9 +95,24 @@ const DesktopNav = ({ otpConfig, popupTarget, setPopupContent }: Props) => {
...
+                    <FormattedMessage id="components.BatchSearchScreen.header" />
+                  </InvisibleA11yLabel>
+                </TransparentButton>
+              )}
+              {logoClickAction?.startsWith('http') && (
+                <TransparentButton bsStyle="link" href={logoClickAction}>
+                  <InvisibleA11yLabel>
+                    <FormattedMessage
+                      id="components.AppMenu.agencyLogoUrl"
+                      values={{ agencyName: operatorName || title }}
```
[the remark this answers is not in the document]
amy-corson-ibigroup: Title attribute is good, but perhaps we could also add a `window.confirm` here letting users know they're about to be directed to an external site? We could repurpose the `handleStartOver` function above to achieve both types of link. I'm comfortable checking with the second reviewer on this one too, if it feels like overkill!

amy-corson-ibigroup (approved):
I'll approve pending a few additional comments! Looks really good, although I am still wondering whether linking to an external site is more trouble than it's worth!

i18n/en-US.yml
AdrianaCeric: I'm not sure how to pass a value to an `intl.formatMessage`

AdrianaCeric:
I feel like I have bloated this, but I think that was worth adding the warning before redirect. It complained a bunch when I changed the `TransparentButton` to an anchor tag, so I stuck with adding the role="link". Thanks for the good suggestions

PR #869, opened 2023-04-26
Add Column Context to stop schedule for AT

amy-corson-ibigroup:
Adds `scope="col"` to column headers in stop viewer so that they are navigable by screen readers and other ATs 

**PR Checklist:**

### DIFF
--- a/lib/components/viewers/stop-schedule-table.tsx
+++ b/lib/components/viewers/stop-schedule-table.tsx
@@ -112,17 +112,17 @@ class StopScheduleTable extends Component<{
         <thead>
           <tr>
             {showBlockIds && (
-              <th>
+              <th scope="col">
                 <FormattedMessage id="components.StopScheduleTable.block" />
               </th>
             )}
-            <th>
+            <th scope="col">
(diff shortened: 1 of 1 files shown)


binh-dam-ibigroup (approved):
LGTM

AdrianaCeric (approved):
This is a great change and good catch. Would it be more efficient to do something like:
```
table th {
  scope: col;
}
```
in the relevant css file?

binh-dam-ibigroup:
> This is a great change and good catch. Would it be more efficient to do something like:
> 
> ```
> table th {
>   scope: col;
> }
> ```
> 
> in the relevant css file?

@AdrianaCeric This is not a CSS attribute.

AdrianaCeric:
> > This is a great change and good catch. Would it be more efficient to do something like:
> > ```
> > table th {
> >   scope: col;
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > in the relevant css file?
> 
> @AdrianaCeric This is not a CSS attribute.

Wait I think you're right on that one haha 👍 really looked like it was

PR #876, opened 2023-04-27
correct `FIELD_TRIP_DATE_TIME_FORMAT`

miles-grant-ibigroup:
**Description:**

This bug has been in field trip since it was first added to OTP-RR. The date format on server requests was incorrect, causing trips to be stored incorrectly, which in turn caused all sorts of display issues. I've done lots of testing with the new format and it seems to fix most issues the field trip was having

**PR Checklist:**


amy-corson-ibigroup (approved):

binh-dam-ibigroup (approved):
Oh this is one of the old date formats from momentjs we didn't port.

PR #893, opened 2023-05-09
Support Clickable Header Logo

miles-grant-ibigroup:
**Description:**
This PR adds a new configuration option `brandingClickable` that when set to `true` will make the header icon clickable. When clicking it, otp-rr will reset to the trip plan page and clear out any itineraries shown.

Closes #868

**PR Checklist:**


miles-grant-ibigroup:
I tried implementing this but almost think it makes more sense to not? I like that there's some distinction between the start over button and the logo click

lib/components/app/desktop-nav.tsx
```diff
@@ -46,9 +46,29 @@ const DesktopNav = ({
...
-  const { branding, persistence, title = DEFAULT_APP_TITLE } = otpConfig
+  const {
+    branding,
+    persistence,
+    title = DEFAULT_APP_TITLE,
+    brandClickable
+  } = otpConfig
   const showLogin = Boolean(getAuth0Config(persistence))
 
+  const BrandingElement = brandClickable ? 'a' : 'div'
```
binh-dam-ibigroup, line 57: Maybe using a `<span>` instead of `<div>` would allow to move all the CSS below to `commonStyles`.

lib/components/app/desktop-nav.tsx
```diff
@@ -46,9 +46,29 @@ const DesktopNav = ({
...
+
+  const commonStyles = { marginLeft: 50 }
+  const brandingProps = brandClickable
+    ? {
+        href: '/#/',
```
binh-dam-ibigroup, line 62: There is an existing issue with the router or the code that causes some parts of the UI to not pick up all the changes, so this will not be the same as "Start Over". For instance, after you plan an itinerary and click the logo, most of the map will clear, however the locations in the location fields and location markers will remain.

To trigger a page reload, the URL should not include the hash. Reuse [this method to compute the restart URL](https://github.com/opentripplanner/otp-react-redux/blob/dev/lib/components/util/start-over.ts) that is used by "Start Over" button in itinerary and app menu. It also accommodates clients that have the UI in a subpath of their site.

binh-dam-ibigroup (changes requested):
Reuse the start over URL. The CSS could probably be simplified too.

lib/components/app/desktop-nav.tsx
miles-grant-ibigroup: I used `div` because it was what the element was previously using. `span` doesn't seem to affect anything, but I'm still a bit nervous about the change and I don't see which styles can be retired

binh-dam-ibigroup (approved):
Explanations were good enough, although I suspect this is an area of code that will need to be revisited.

AdrianaCeric:
That's a lot cleaner 😭 ❤️

PR #897, opened 2023-05-12
Replace trip sort select with accessible dropdown

amy-corson-ibigroup:
**Description:**

Select to sort trip results updates page without warning (3.2.2 On Input). This PR repurposes the locale selector dropdown to provide a better approach to sorting in terms of accessibility.

**PR Checklist:**

| Before | After |
|--------|-------|
|image|image| -->

### DIFF
--- a/lib/components/app/locale-selector.tsx
+++ b/lib/components/app/locale-selector.tsx
@@ -4,9 +4,9 @@ import { useIntl } from 'react-intl'
 import React from 'react'
 
 import * as uiActions from '../../actions/ui'
+import { Dropdown } from '../util/dropdown'
 import { getLanguageOptions } from '../../util/i18n'
 import { UnstyledButton } from '../util/unstyled-button'
-import Dropdown from '../util/dropdown'
 
 interface LocaleSelectorProps {
   // Typescript TODO languageOptions based on configLanguage type.
@@ -25,7 +25,6 @@ const LocaleSelector = (props: LocaleSelectorProps): JSX.Element | null => {
       id="locale-selector"
       label={intl.formatMessage({ id: 'components.SubNav.selectALanguage' })}
       listLabel={intl.formatMessage({ id: 'components.SubNav.languages' })}
-      locale
       name={
         <span
           style={{
@@ -35,6 +34,7 @@ const LocaleSelector = (props: LocaleSelectorProps): JSX.Element | null => {
           <GlobeAmericas height="18px" />
         </span>
       }
+      nav
       style={{ display: 'block ruby' }}
       // TODO: How to make this work without block ruby?
     >
--- a/lib/components/app/locale-selector.tsx
+++ b/lib/components/app/locale-selector.tsx
@@ -25,6 +25,7 @@ const LocaleSelector = (props: LocaleSelectorProps): JSX.Element | null => {
       id="locale-selector"
       label={intl.formatMessage({ id: 'components.SubNav.selectALanguage' })}
       listLabel={intl.formatMessage({ id: 'components.SubNav.languages' })}
+      locale
       name={
         <span
           style={{
--- a/lib/components/app/locale-selector.tsx
+++ b/lib/components/app/locale-selector.tsx
@@ -21,35 +21,35 @@ const LocaleSelector = (props: LocaleSelectorProps): JSX.Element | null => {
 
   // Only render if two or more languages are configured.
   return languageOptions ? (
-    <Dropdown
-      id="locale-selector"
-      label={intl.formatMessage({ id: 'components.SubNav.selectALanguage' })}
-      listLabel={intl.formatMessage({ id: 'components.SubNav.languages' })}
-      name={
-        <span
-          style={{
-            color: 'rgba(255, 255, 255, 0.85)'
-          }}
-        >
-          <GlobeAmericas height="18px" />
-        </span>
-      }
-      nav
-      style={{ display: 'block ruby' }}
-      // TODO: How to make this work without block ruby?
-    >
-      {Object.keys(languageOptions).map((locale: string) => (
-        <li key={locale} lang={locale} role="none">
-          <UnstyledButton
-            aria-selected={locale === currentLocale || undefined}
-            onClick={() => setLocale(locale)}
-            role="option"
+    <li>
+      <Dropdown
+        id="locale-selector"
+        label={intl.formatMessage({ id: 'components.SubNav.selectALanguage' })}
+        listLabel={intl.formatMessage({ id: 'components.SubNav.languages' })}
+        name={
+          <span
+            style={{
+              color: 'rgba(255, 255, 255, 0.85)'
+            }}
           >
-            {languageOptions[locale].name}
-          </UnstyledButton>
-        </li>
-      ))}
-    </Dropdown>
+            <GlobeAmericas height="18px" />
+          </span>
+        }
+        style={{ display: 'block ruby' }}
+      >
+        {Object.keys(languageOptions).map((locale: string) => (
+          <li key={locale} lang={locale} role="none">
+            <UnstyledButton
+              aria-selected={locale === currentLocale || undefined}
+              onClick={() => setLocale(locale)}
+              role="option"
+            >
+              {languageOptions[locale].name}
+            </UnstyledButton>
+          </li>
+        ))}
+      </Dropdown>
+    </li>
   ) : null
 }
 
--- a/lib/components/narrative/narrative-itineraries-header.tsx
+++ b/lib/components/narrative/narrative-itineraries-header.tsx
@@ -5,7 +5,7 @@ import { FormattedMessage, useIntl } from 'react-intl'
 import { Itinerary } from '@opentripplanner/types'
 import { SortAmountDown } from '@styled-icons/fa-solid/SortAmountDown'
 import { SortAmountUp } from '@styled-icons/fa-solid/SortAmountUp'
-import React from 'react'
+import React, { useCallback } from 'react'
 import styled from 'styled-components'
 
 import { IconWithText, StyledIconWrapper } from '../util/styledIcon'
@@ -89,7 +89,15 @@ export default function NarrativeItinerariesHeader({
     ? searching
     : intl.formatList([itinerariesFound, numIssues], { type: 'conjunction' })
 
-  const sortText = sortOptions(intl).find((x) => x.value === sort.type)?.text
+  const sortOptionsArr = sortOptions(intl)
+  const sortText = sortOptionsArr.find((x) => x.value === sort.type)?.text
+
+  const handleSortClick = useCallback(
+    (value) => {
+      onSortChange(value)
+    },
+    [onSortChange]
+  )
 
   return (
     <div
@@ -200,17 +208,12 @@ export default function NarrativeItinerariesHeader({
               name={sortText}
               title={sortResultsLabel}
             >
-              {sortOptions(intl).map((x) => {
+              {sortOptionsArr.map((x) => {
                 return (
-                  <li
-                    key={x.value}
-                    style={{
-                      fontWeight: sortText === x.text ? 'bold' : undefined
-                    }}
-                  >
+                  <li className="sort-option" key={x.value}>
                     <UnstyledButton
                       aria-selected={sortText === x.text || undefined}
-                      onClick={() => onSortChange(x.value)}
+                      onClick={() => handleSortClick(x.value)}
                       role="option"
                     >
                       {x.text}
--- a/lib/components/narrative/narrative-itineraries-header.tsx
+++ b/lib/components/narrative/narrative-itineraries-header.tsx
@@ -8,6 +8,9 @@ import React from 'react'
 import styled from 'styled-components'
 
 import { IconWithText, StyledIconWrapper } from '../util/styledIcon'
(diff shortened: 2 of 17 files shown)


binh-dam-ibigroup (changes requested):
I think it is heading in the right direction. I would create s separate styled component for the locale selector (so that you don't have to pass that extra `locale` prop. I also indicated some i18n and a11y tweaks too.

lib/components/narrative/narrative-itineraries-header.tsx
```diff
@@ -172,48 +176,25 @@ export default function NarrativeItinerariesHeader({
...
-                })}
-              </option>
-            </select>
+              {sortOptions.map((x) => {
+                return (
+                  <li key={x.value}>
+                    <UnstyledButton onClick={() => onSortChange(x.value)}>
+                      <FormattedMessage id={x.locale} />
+                    </UnstyledButton>
+                  </li>
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Are you suggessting that whether a page is accessible is a config parameter?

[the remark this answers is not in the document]
miles-grant-ibigroup: I'm also not entirely sure why the locale dropdown work is in this PR? Maybe it should be split out

lib/components/util/dropdown.tsx
```diff
@@ -12,23 +12,28 @@ interface Props extends HTMLAttributes<HTMLElement> {
...
   listLabel?: string
-  name: JSX.Element
+  locale?: boolean
+  name: JSX.Element | string
   pullRight?: boolean
 }
 
 // TODO: make this a button once bootstrap is removed
-const DropdownButton = styled.a`
+const DropdownButton = styled.a<{ locale?: boolean }>`
```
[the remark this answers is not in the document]
amy-corson-ibigroup: Good idea! Would you want to keep that inside `dropdown.tsx`?

lib/components/narrative/narrative-itineraries-header.tsx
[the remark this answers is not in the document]
amy-corson-ibigroup: This shouldn't be any different than the current existing behavior! It should just be taking what was a `select` and making it a dropdown menu. Can you elaborate one what you want to make configurable?

lib/components/util/dropdown.tsx
```diff
@@ -180,10 +172,9 @@ const Dropdown = ({
         aria-expanded={open}
         aria-haspopup="listbox"
         aria-label={label}
-        as={!locale ? 'button' : ''}
+        as={nav ? '' : 'button'}
```
binh-dam-ibigroup, line 175: These should always be buttons. Indeed, right now the navbar dropdown (language and account) buttons render into `<a ... role="button">`. See if you can get rid of the `nav` prop by writing instead `<Dropdown as="li"` (or `"span"`)?

lib/components/util/sortOptions.ts
```diff
@@ -1,26 +1,44 @@
...
-  {
-    locale: 'components.NarrativeItinerariesHeader.selectWalkTime',
-    value: 'WALKTIME'
-  },
-  {
-    locale: 'components.NarrativeItinerariesHeader.selectCost',
-    value: 'COST'
+import { useIntl } from 'react-intl'
+
+export const sortOptions = (): {
```
binh-dam-ibigroup, line 3: Better to pass `intl` to `sortOptions` so you can write attributes as below and get rid of `OptionText`:
```
text: intl.formatMessage('components.NarrativeItinerariesHeader...')
```

binh-dam-ibigroup (changes requested):
Please see the tweaks mentioned, I think that'll make things much cleaner.

lib/components/util/dropdown.tsx
```diff
@@ -192,8 +195,25 @@ const Dropdown = ({
...
 
-export default Dropdown
+export const SortResultsDropdown = styled(Dropdown)`
+    position: relative;
+
+    ${DropdownButton} {
+      border-radius: 5px;
+      padding: 3px 7px;
+      min-width: 130px;
+      text-align: right;
```
binh-dam-ibigroup, line 209: Why `text-align: right`? With the min width, this leaves unexplained padding to the left.

lib/components/util/dropdown.tsx
```diff
@@ -192,8 +195,25 @@ const Dropdown = ({
...
 }
 
-export default Dropdown
+export const SortResultsDropdown = styled(Dropdown)`
+    position: relative;
+
+    ${DropdownButton} {
+      border-radius: 5px;
+      padding: 3px 7px;
+      min-width: 130px;
```
binh-dam-ibigroup, line 208: What is the intended effect for min-width?

amy-corson-ibigroup: To keep the button size consistent instead of resizing according to contents. Although I do agree for options such as "cost" there's a lot of padding. I can remove both the `min-width` and the `text-align`!

lib/components/util/dropdown.tsx
amy-corson-ibigroup: Good point about the button! I have removed the above line in [8f77197](https://github.com/opentripplanner/otp-react-redux/pull/897/commits/8f77197c5c9c1808fa359627deb383facf91a36c) and refactored the CSS to match. 

I changed the outer DropdownContainer to a traditional html element because the `SortResultsDropdown` [needs to pass a `className` to a DOM element](https://styled-components.com/docs/basics#styling-any-component) in order to work. Even if it was changed back to a styled component, passing `as` to `SortResultsDropdown` unfortunately doesn't work. 

I can't think of a way to conditionally render that container as either a `span` or `li` other than using props, but please let me know if I'm missing something!

lib/components/narrative/narrative-itineraries-header.tsx
```diff
@@ -94,10 +89,7 @@ export default function NarrativeItinerariesHeader({
     ? searching
     : intl.formatList([itinerariesFound, numIssues], { type: 'conjunction' })
 
-  const onClickSort = (x: { text: string; value: string }) => {
-    onSortChange(x.value)
-    setSortSelectedText(x.text)
-  }
+  const sortText = sortOptions(intl).find((x) => x.value === sort.type)?.text
```
binh-dam-ibigroup, line 92: Extract and reuse a `const` for `sortOptions(intl)`.

lib/components/narrative/narrative-itineraries-header.tsx
```diff
@@ -205,21 +197,20 @@ export default function NarrativeItinerariesHeader({
...
-                      fontWeight:
-                        sortSelectedText === x.text ? 'bold' : undefined
+                      fontWeight: sortText === x.text ? 'bold' : undefined
                     }}
                   >
                     <UnstyledButton
-                      aria-selected={sortSelectedText === x.text || undefined}
-                      onClick={() => onClickSort(x)}
+                      aria-selected={sortText === x.text || undefined}
+                      onClick={() => onSortChange(x.value)}
```
binh-dam-ibigroup, line 213: Consider `useCallback` for this.

lib/components/narrative/narrative-itineraries-header.tsx
```diff
@@ -205,21 +197,20 @@ export default function NarrativeItinerariesHeader({
...
             >
-              {sortOptions().map((x) => {
+              {sortOptions(intl).map((x) => {
                 return (
                   <li
                     key={x.value}
                     style={{
-                      fontWeight:
-                        sortSelectedText === x.text ? 'bold' : undefined
+                      fontWeight: sortText === x.text ? 'bold' : undefined
```
binh-dam-ibigroup, line 208: Move this to CSS (use `button[aria-selected="true"]`) as with the locale selector.

binh-dam-ibigroup (approved):
Chef's kiss. Much cleaner!

lib/components/app/locale-selector.tsx
```diff
@@ -34,6 +34,7 @@ const LocaleSelector = (props: LocaleSelectorProps): JSX.Element | null => {
           <GlobeAmericas height="18px" />
         </span>
       }
+      nav
```
binh-dam-ibigroup, line 37: Do we still need this?

lib/components/util/dropdown.tsx
amy-corson-ibigroup: That's it!!! This works great, thank you so much!

lib/components/user/nav-login-button.tsx
```diff
@@ -55,50 +55,52 @@ const NavLoginButton = ({
...
+          name={
+            <span>
+              <Avatar
+                alt={displayedName}
+                src={profile.picture}
+                title={`${displayedName}\n(${profile.email})`}
+              />
+            </span>
+          }
+          nav
```
binh-dam-ibigroup, line 71: Just remove the unused `nav ` prop.

miles-grant-ibigroup (changes requested):
Very clever re-use of components and great cleanup!

I have a few cosmetic notes before I dig into the code -- I'm sad to see the locale selector dropdown has lost its background on click. I would be sad to lose it but if it's not possible to keep then I think it's worth it based on how good the new dropdown looks!

The second thing is that because the dropdown is now non-native the leave feedback button that's sometimes to the left of it looks very out of place. Do you think it's worth creating a matching button class?

lib/components/narrative/narrative-itineraries-header.tsx
```diff
@@ -172,48 +202,26 @@ export default function NarrativeItinerariesHeader({
...
-                })}
-              </option>
-              <option value="COST">
-                {intl.formatMessage({
-                  id: 'components.NarrativeItinerariesHeader.selectCost'
-                })}
-              </option>
-            </select>
+              {sortOptionsArr.map((x) => {
+                return (
```
miles-grant-ibigroup, line 212: Nitpick: this can probably be inlined

lib/components/util/dropdown.tsx
```diff
@@ -157,11 +155,12 @@ const Dropdown = ({
   )
 
   return (
-    <DropdownContainer
+    <span
+      className={className}
       id={`${id}-wrapper`}
       onKeyDown={_handleKeyDown}
       ref={containerRef}
-      style={{ float: pullRight ? 'right' : 'left' }}
```
miles-grant-ibigroup, line 164: I think removing this is causing some issues on mobile

Screenshot 2023-05-19 at 9 45 10 AM

lib/components/app/locale-selector.tsx
```diff
@@ -21,34 +21,35 @@ const LocaleSelector = (props: LocaleSelectorProps): JSX.Element | null => {
...
-      style={{ display: 'block ruby' }}
-      // TODO: How to make this work without block ruby?
-    >
-      {Object.keys(languageOptions).map((locale: string) => (
-        <li key={locale} lang={locale} role="none">
-          <UnstyledButton
-            aria-selected={locale === currentLocale || undefined}
-            onClick={() => setLocale(locale)}
-            role="option"
+    <li>
```
miles-grant-ibigroup, line 24: I think this might be the reason I can no longer open the dropdown using keyboard navigation

miles-grant-ibigroup (commented):
A few more bugs to iron out

lib/components/narrative/narrative-itineraries-header.tsx
amy-corson-ibigroup: [Conflicting reviews on this one](https://github.com/opentripplanner/otp-react-redux/pull/897#discussion_r1198265068)

miles-grant-ibigroup: Go with Binh's! That's clean

miles-grant-ibigroup: I just meant removing the `return`!

lib/components/app/locale-selector.tsx
amy-corson-ibigroup: Good catch! Fixed this with a `preventDefault` inside `dropdown`!

amy-corson-ibigroup:
Background on the locale button has been added again! 

Regarding the feedback button, would it be okay to follow up on this in a separate PR?

miles-grant-ibigroup (changes requested):
I'm not sure if this was introduced by this PR but still seeing some problems using arrow keys to navigate the menu. Agreed we can deal with the rough edges in a future PR. Should be ready to go once merge conflicts are sorted

amy-corson-ibigroup:
Fixed merge conflicts! I am not seeing any issues with the keyboard navigation on the dropdown, can you let me know if you still are?

miles-grant-ibigroup (changes requested):
Keyboard navigation isn't working for the leave feedback button or the profile log in button! The color of the login button also looks off on mobile: 
Screenshot 2023-05-26 at 7 20 57 PM

miles-grant-ibigroup (changes requested):
Code is looking so much cleaner this is fantastic thank you so much!

Only 1 last case that's having problems and that's a browser window that's really slim, but not a mobile browser: 
Screenshot 2023-06-01 at 8 33 35 AM

miles-grant-ibigroup (approved):
Great work! Just 2 tiny nitpicks

amy-corson-ibigroup:
Thank you both for all the help with this one!

PR #898, opened 2023-05-12
Add Gitter link to readme

daniel-heppner-ibigroup:


miles-grant-ibigroup (changes requested):
Looks great! Would it be possible to add a paragraph further down in the readme describing what this is and why people should join in more detail?

PR #903, opened 2023-05-18
Enhance Login and Feedback Navbar Buttons

binh-dam-ibigroup:
## Description

This PR sets the visual content of the "Give Feedback" (or other configured popup) and login buttons to be icons, so that text in some languages doesn't bleed into other content. A popup is also added to these buttons and the language selector. The login button English text is coordinated with the text from the Auth0 login screen.

## PR Checklist
- [x] Are all languages supported (Internationalization/Localization)? => improves i18n by preventing text to bleed into other content.

| Before | After |
|--------|-------|
| Before image | After image |

### DIFF
--- a/lib/components/app/desktop-nav.tsx
+++ b/lib/components/app/desktop-nav.tsx
@@ -7,9 +7,10 @@ import styled from 'styled-components'
 import * as uiActions from '../../actions/ui'
 import { accountLinks, getAuth0Config } from '../../util/auth'
 import { DEFAULT_APP_TITLE } from '../../util/constants'
+import InvisibleA11yLabel from '../util/invisible-a11y-label'
 import NavLoginButtonAuth0 from '../user/nav-login-button-auth0'
 
-import AppMenu from './app-menu'
+import AppMenu, { Icon } from './app-menu'
 import LocaleSelector from './locale-selector'
 import ViewSwitcher from './view-switcher'
 
@@ -19,6 +20,13 @@ const NavItemOnLargeScreens = styled(NavItem)`
     display: none !important;
   }
 `
+
+const StyledNav = styled(Nav)`
+  & > li svg {
+    height: 18px;
+  }
+`
+
 // Typscript TODO: otpConfig type
 export type Props = {
   locale: string
@@ -91,12 +99,15 @@ const DesktopNav = ({
 
           <ViewSwitcher sticky />
 
-          <Nav pullRight>
+          <StyledNav pullRight>
             {popupTarget && (
               <NavItemOnLargeScreens
                 onClick={() => setPopupContent(popupTarget)}
               >
-                <FormattedMessage id={`config.popups.${popupTarget}`} />
+                <Icon iconType={popupTarget} />
+                <InvisibleA11yLabel>
+                  <FormattedMessage id={`config.popups.${popupTarget}`} />
+                </InvisibleA11yLabel>
               </NavItemOnLargeScreens>
             )}
             <LocaleSelector />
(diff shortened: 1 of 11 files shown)


lib/components/app/desktop-nav.tsx
```diff
@@ -91,12 +105,14 @@ const DesktopNav = ({
...
 
-          <Nav pullRight>
+          <StyledNav pullRight>
             {popupTarget && (
               <NavItemOnLargeScreens
                 onClick={() => setPopupContent(popupTarget)}
+                title={popupButtonText}
               >
-                <FormattedMessage id={`config.popups.${popupTarget}`} />
+                <Icon iconType={popupTarget} />
```
amy-corson-ibigroup, line 114: You have the chat bubble icon in the screenshot and I think I prefer it to this popout icon. We use this popout icon in other places to indicate that something will open in a new window/tab which feels a little inconsistent with the survey iframe behavior.

amy-corson-ibigroup (changes requested):
Really great improvement and everything seems to be working well! Just a couple of styling adjustments.

lib/components/app/desktop-nav.tsx
binh-dam-ibigroup: I am not sure to understand... It's the same (configured) icon that is also used at the top of itinerary results, at the bottom of an expanded itinerary, and in the app menu. Maybe the a11y text needs to say "Opens in a popup" or something?

amy-corson-ibigroup: I don't think we need additional text! Using this icon for the feedback button in the app menu makes sense, because in mobile view this will open a brand new window in the browser. But in desktop view, it only opens a modal. And since generally when we use this icon (https://github.com/opentripplanner/otp-react-redux/pull/859) it means it's opening a new window, it seems more consistent to have a different icon for the desktop nav.

I suppose I don't mind keeping this icon for both, I can defer to the second reviewer! But I do prefer the chat bubble, if that's an option.

binh-dam-ibigroup: Per our call earlier, I will add additional a11y text to warn of a tab/window opening on mobile. We will keep using any configured icon if any.

lib/components/app/desktop-nav.tsx
```diff
@@ -19,6 +20,13 @@ const NavItemOnLargeScreens = styled(NavItem)`
     display: none !important;
   }
 `
+
+const StyledNav = styled(Nav)`
+  & > li svg {
+    height: 18px;
+  }
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Right now, the login button is flush with the right edge of the viewport, and the app menu button appears to be flush with the left edge of the viewport. Canceling the -15px margin would create an imbalance IMO (and also reduce the available space for the title bar text on mobile view).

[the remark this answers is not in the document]
binh-dam-ibigroup: See screenshot with the -15px margin canceled:

image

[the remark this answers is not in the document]
amy-corson-ibigroup: That's fair! I'm fine to keep the margin as is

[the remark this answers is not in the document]
binh-dam-ibigroup: For both the button width and the right margin, I was able to find a good balance with https://github.com/opentripplanner/otp-react-redux/pull/903/commits/7f761676bf80e394bf0a5cfa84e55d08bdb2a071.

lib/components/app/desktop-nav.tsx
binh-dam-ibigroup: I added the a11y text in https://github.com/opentripplanner/otp-react-redux/pull/903/commits/63ba3ffd4a11cee222f58914c03cabf8a1bc1bd9 and subsequent commits.

amy-corson-ibigroup (approved):
Looks good! I like the layout! Thanks

PR #907, opened 2023-05-26
Render selected active itinerary when pasting URL

binh-dam-ibigroup:
## Description

With this PR, the particular itinerary (`ui_activeItinerary`) being viewed when copying the current URL will also be shown when that URL is pasted in a new browser tab or when refreshing the page. This does not account for side effects from the OTP routing engine processing real-time data.

## PR Checklist


amy-corson-ibigroup (approved):
Tested and seems to work well! Thanks so much

PR #916, opened 2023-06-05
Check existence of the correct itinerary

binh-dam-ibigroup:
## Description

This PR makes the Save Trip screen send the correct itinerary to OTP-middleware for checking.

When requests are pending and itineraries are trickling in, the UI grabs the first one available as the itinerary to check. It might be a walk-only trip, which is incorrect. The PR makes the UI wait until all itineraries are fetched before displaying the desired one and setting that as the itinerary to check. This improvement is also applied to the narrative itineraries container so avoid superfluous renderings when the full itinerary is shown.

As a result, there should be ** *fewer* ** cases where the itinerary check results in no days available, compared to now.
Note that this PR does not address the active itinerary index in grouped itineraries or any issues with the middleware checking itineraries that could cause the remaining cases with no days available.

## PR Checklist

### DIFF
--- a/lib/components/narrative/narrative-itineraries.js
+++ b/lib/components/narrative/narrative-itineraries.js
@@ -435,13 +435,13 @@ class NarrativeItineraries extends Component {
                     </S.ModeResultContainer>
                   )
                 })
-              ) : !pending ? (
+              ) : (
                 <ListContainer>
                   {itinerariesWithCo2.map((itin) =>
                     this._renderItineraryRow(itin)
                   )}
                 </ListContainer>
-              ) : null}
+              )}
               {this._renderLoadingDivs()}
               {groupItineraries &&
                 !itineraryIsExpanded &&
--- a/lib/components/narrative/narrative-itineraries.js
+++ b/lib/components/narrative/narrative-itineraries.js
@@ -435,13 +435,13 @@ class NarrativeItineraries extends Component {
                     </S.ModeResultContainer>
                   )
                 })
-              ) : (
+              ) : !pending ? (
                 <ListContainer>
                   {itinerariesWithCo2.map((itin) =>
                     this._renderItineraryRow(itin)
                   )}
                 </ListContainer>
-              )}
+              ) : null}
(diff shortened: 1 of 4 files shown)


lib/components/narrative/narrative-itineraries.js
```diff
@@ -435,13 +435,13 @@ class NarrativeItineraries extends Component {
                     </S.ModeResultContainer>
                   )
                 })
-              ) : (
+              ) : !pending ? (
```
miles-grant-ibigroup, line 438: I think this is causing results not to "trickle" in. On slow searches I think this is a bit annoying? Or am I just seeing things

lib/components/user/monitored-trip/saved-trip-screen.js
```diff
@@ -244,6 +244,7 @@ class SavedTripScreen extends Component {
 const mapStateToProps = (state, ownProps) => {
   const activeSearch = getActiveSearch(state)
   const activeItinerary = activeSearch && activeSearch.activeItinerary
+  const pending = activeSearch ? Boolean(activeSearch.pending) : false
```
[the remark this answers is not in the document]
binh-dam-ibigroup: I copied this line that is used elsewhere, although I don't disagree.

lib/components/narrative/narrative-itineraries.js
binh-dam-ibigroup: Oh I see. When viewing one itinerary, however, not having that will flash multiple itineraries until `pending` is `false`, and even show a different itinerary than the one originally being saved.

binh-dam-ibigroup: Reverting that change for this PR (https://github.com/opentripplanner/otp-react-redux/pull/916/commits/ac129ae71f6e174ff0a865cf682c7bc30714a26f).

miles-grant-ibigroup (approved):
Looking good thanks for these changes!

amy-corson-ibigroup (approved):
Makes sense! Thank you

PR #917, opened 2023-06-07
Support External Help Link

amy-corson-ibigroup:
**Description:**
If there is an external "help" link supplied in the config, the "help" option inside the account dropdown will direct users to the external web address. An icon will also be added indicating the link is an external link.

image

**PR Checklist:**

### DIFF
--- a/lib/components/mobile/navigation-bar.js
+++ b/lib/components/mobile/navigation-bar.js
@@ -113,7 +113,7 @@ const mapStateToProps = (state) => {
   return {
     auth0Config: getAuth0Config(state.otp.config.persistence),
     configLanguages: state.otp.config.language,
-    extraMenuItems: state.otp.config.extraMenuItems,
+    extraMenuItems: state.otp?.config?.extraMenuItems,
     locale: state.otp.ui.locale
   }
 }
--- a/lib/components/mobile/navigation-bar.js
+++ b/lib/components/mobile/navigation-bar.js
@@ -24,6 +24,7 @@ class MobileNavigationBar extends Component {
   static propTypes = {
     auth0Config: PropTypes.object,
     backScreen: PropTypes.number,
+    config: PropTypes.object,
     configLanguages: PropTypes.object,
     defaultMobileTitle: PropTypes.string,
     headerAction: PropTypes.element,
@@ -45,6 +46,7 @@ class MobileNavigationBar extends Component {
   render() {
     const {
       auth0Config,
+      config,
       configLanguages,
       defaultMobileTitle,
       headerAction,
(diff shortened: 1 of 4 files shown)


binh-dam-ibigroup (approved):
LGTM

lib/util/auth.js
```diff
@@ -5,19 +5,24 @@ import { isOtpMiddleware } from './user'
...
-  },
-  {
-    messageId: 'help',
-    // Add a target attribute if you need the link to open in a new window, etc.
-    // (supports the same values as <a target=... >).
-    // target: '_blank',
-    url: '/help'
-  }
-]
+export const accountLinks = (config) => {
```
miles-grant-ibigroup, line 8: nitpick: can we refactor this to take in `extraMenuItems` instead of the entire config?

lib/util/auth.js
```diff
@@ -5,19 +5,24 @@ import { isOtpMiddleware } from './user'
...
+  const helpLink =
+    config?.extraMenuItems.find((link) => link.id === 'help') || undefined
+  const helpLinkUrl = helpLink ? helpLink.href : '/help'
```
miles-grant-ibigroup, line 11: Can we inline this?

lib/util/auth.js
```diff
@@ -5,19 +5,24 @@ import { isOtpMiddleware } from './user'
...
+  return [
+    {
+      messageId: 'myAccount',
+      url: ACCOUNT_PATH
+    },
+    {
+      messageId: 'help',
+      // Add a target attribute if you need the link to open in a new window, etc.
+      // (supports the same values as <a target=... >).
+      // target: '_blank',
```
miles-grant-ibigroup, line 21: Is this supposed to be uncommented?

lib/components/mobile/navigation-bar.js
```diff
@@ -110,6 +112,7 @@ class MobileNavigationBar extends Component {
 const mapStateToProps = (state) => {
   return {
     auth0Config: getAuth0Config(state.otp.config.persistence),
+    config: state.otp.config,
```
miles-grant-ibigroup, line 115: Once accountLinks is extracted this can be refactored to only extract the part of the config we need

miles-grant-ibigroup (changes requested):
Works! Just a few nitpicks

lib/util/auth.js
amy-corson-ibigroup: This exists in dev, but I think it's just a demonstration? Seems like weird behavior for this to open a new tab by default

miles-grant-ibigroup: Good call. We can leave it commented out for now!

PR #921, opened 2023-06-08
fix(form styles): Use better style import.

### LINKED ISSUE #919: Date/old settings picker styles broken
#911 broke the CSS for the date time selector and the trip settings. The commit before #911 was merged does not show the defect. Note that the broken date time selection styling also appears in #700 at time of writing.

image

binh-dam-ibigroup:
Fix #919 for the new mode selector branch.

Feel free to merge after you agree with the proposed changes.


miles-grant-ibigroup (approved):
Much cleaner thanks

daniel-heppner-ibigroup (approved):
Thank you for the fix!

PR #922, opened 2023-06-08
Fix itinerary existence

binh-dam-ibigroup:
## Description

This PR is against the mode selector branch to take advantage of improved graphql querying.

This PR makes the Save Trip screen send the correct itinerary to OTP-middleware for checking.

When requests are pending and itineraries are trickling in, the UI grabs the first one available as the itinerary to check. It might be a walk-only trip, which is incorrect. The PR makes the UI wait until all itineraries are fetched before displaying the desired one and setting that as the itinerary to check. This improvement is also applied to the narrative itineraries container so avoid superfluous renderings when the full itinerary is shown.

As a result, there should be ** fewer ** cases where the itinerary check results in no days available, compared to now.
Note that this PR does not address any issues with the middleware checking itineraries that could cause the remaining cases with no days available.

**PR Checklist:**

### DIFF
--- a/i18n/en-US.yml
+++ b/i18n/en-US.yml
@@ -510,6 +510,8 @@ components:
     pause: Pause
     resume: Resume
   SavedTripScreen:
+    itineraryLoaded: Itinerary loaded
+    itineraryLoading: Loading itinerary
     tooManyTrips: >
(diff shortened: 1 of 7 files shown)


binh-dam-ibigroup:
Closing in favor of #1178.

PR #923, opened 2023-06-08
Support Custom Stop Color

amy-corson-ibigroup:
**Description:**

Takes custom stop color from config and passes to OTP-UI map overlay

**PR Checklist:**

### DIFF
--- a/lib/components/map/default-map.tsx
+++ b/lib/components/map/default-map.tsx
@@ -254,7 +254,7 @@ class DefaultMap extends Component {
     } = this.props
     const { getCustomMapOverlays, getTransitiveRouteLabel, ModeIcon } =
       this.context
-    const { baseLayers, maxZoom, overlays } = mapConfig || {}
+    const { baseLayers, mapColors, maxZoom, overlays } = mapConfig || {}
     const { lat, lon, zoom } = this.state
     const vectorTilesEndpoint = makeApiUrl(config, 'vectorTiles', {})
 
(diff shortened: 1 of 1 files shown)


lib/components/map/default-map.tsx
```diff
@@ -334,7 +336,9 @@ class DefaultMap extends Component {
               case 'park-and-ride':
                 return <ParkAndRideOverlay {...namedLayerProps} />
               case 'stops':
-                return <StopsOverlay {...namedLayerProps} />
+                return (
+                  <StopsOverlay {...namedLayerProps} stopsColor={stopsColor} />
```
miles-grant-ibigroup, line 340: I don't think you need to explicitly pass in `stopsColor`! `namedLayerProps` already comes from the config, so you can just define it there. This PR is not needed

miles-grant-ibigroup (changes requested):
I think you may not need this PR!

lib/components/map/default-map.tsx
amy-corson-ibigroup: Good catch, although it's not passing through the whole `mapConfig` so I think we will need a pr! I can alternatively add the entire mapColors section of the config to the `namedLayerProps` if you think that would be better

miles-grant-ibigroup: You just have to define `stopsColor` as a property of the map layer rather than as a global

amy-corson-ibigroup: Gotcha, no you're right! Thanks, I'll close this and adjust accordingly!

amy-corson-ibigroup:
Whoops! Closing as it turns out this PR is unnecessary!

PR #925, opened 2023-06-09
(Mailables): Support Simultaneous Small and Large Print

miles-grant-ibigroup:
**Description:**

The mailables large print feature currently disables the small print on the resulting pdf. This PR corrects this, allowing for simultaneous large and small prints. Unfortunately, there is no way to change the count for the second size. Not sure how best to solve this.

**PR Checklist:**
- [no] Does the code follow accessibility standards (WCAG 2.1 AA Compliant)?
- [no] Are all languages supported (Internationalization/Localization)?
- [no] Are appropriate Typescript types implemented?

### DIFF
--- a/lib/util/mailables.js
+++ b/lib/util/mailables.js
@@ -113,13 +113,22 @@ async function writePDF(formData, imageData, otpConfig) {
 
   const tableData = {
     headers: ['Item', 'Quantity'],
-    rows: mailables.map((mailable) => {
-      let { largeFormat, largePrint, name, quantity } = mailable
-      if (largePrint && largeFormat) {
-        name += ' (LARGE PRINT)'
-      }
-      return [name, quantity]
-    })
+    rows: mailables
+      .map((mailable) => {
+        let { largeFormat, largePrint, name, quantity, smallFormat } = mailable
+        const rows = []
+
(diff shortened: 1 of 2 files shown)


philip-cline (approved):
lgtm

lib/util/mailables.js
```diff
@@ -113,13 +113,22 @@ async function writePDF(formData, imageData, otpConfig) {
...
+          rows.push([name, quantity])
+        }
+
+        if (largePrint && largeFormat) {
+          name += ' (LARGE PRINT)'
+          rows.push([name, quantity])
+        }
+        return rows
+      })
+      .flat()
```
amy-corson-ibigroup, line 131: Nitpicky, but could just change the above `map` to `flatMap`

amy-corson-ibigroup (commented):
Could you get rid of the checkboxes and just have two inputs? One for small format one for large? 
Having trouble understanding why you'd want to select both checkboxes without being able to control the number for each, but it it makes sense in context I trust you

lib/util/mailables.js
miles-grant-ibigroup: This is much cleaner thanks!

PR #928, opened 2023-06-19
Route Row Background Color Indicates Active

amy-corson-ibigroup:
**Description:**

Route row now has a background color to indicate to users it is an active row. 

**PR Checklist:**

| Before | After |
|--------|-------|
|image|image|

### DIFF
--- a/lib/components/viewers/RouteRow.js
+++ b/lib/components/viewers/RouteRow.js
@@ -34,7 +34,8 @@ export const RouteRowButton = styled(Button)`
   width: 100%;
 
   &:before {
-    background-color: white;
+    background-color: ${(props) =>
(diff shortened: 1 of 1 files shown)


miles-grant-ibigroup (approved):
Could maybe rename `patternActive` to just `active`

daniel-heppner-ibigroup (approved):
Agree with Miles on just called it `active`, but at the end of the day it doesn't really matter.

daniel-heppner-ibigroup:
Also, I feel like we should be thinking about how to not have all these random colors hard coded all over the place.

amy-corson-ibigroup:
Agree with the naming feedback but `active` is unfortunately an existing button prop, so renaming it this error: 
image

PR #932, opened 2023-06-28
Restore walk icon to short trips

amy-corson-ibigroup:
**Description:**

Fixes a small issue that was keeping the walk icon from showing up on shorter trips in itinerary results

**PR Checklist:**

| Before | After |
|--------|-------|
|image|image|


lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -309,7 +305,6 @@ class MetroItinerary extends NarrativeItinerary {
     )
 
     const firstTransitStop = getFirstTransitLegStop(itinerary)
-    const routeLegs = itinerary.legs.filter(removeInsignifigantWalkLegs)
```
miles-grant-ibigroup, line 312: I'm nervous about removing this for all route blocks -- this is used to remove a walking leg icon from a say 3 minute walking, 20 minute transit, 2 minute walking itinerary

miles-grant-ibigroup (changes requested):
Spotted one potential regression! Could be fixed by probably making the insignificant walk leg removal dependent on the number of legs in the itinerary

lib/components/narrative/metro/metro-itinerary.tsx
amy-corson-ibigroup: This was the only place in the file it was used, which is why I took it out, but I think `removeInsignifigantWalkLegs` is used other places for what you're talking about!

miles-grant-ibigroup: Yes unfortunately this change affects Walk + Transit itineraries as well. 3 minute walk legs are being shown when they shouldn't be

miles-grant-ibigroup (approved):
Looks like this may have been a refactor gone wrong. Let's merge this and fix any fallout later

lib/components/narrative/metro/metro-itinerary.tsx
daniel-heppner-ibigroup: Can we do this instead, to address Miles' comment (which I agree with)
```
const routeLegs = itinerary.legs.length > 1 ? itinerary.legs.filter(removeInsignifigantWalkLegs) : itinerary.legs
```

daniel-heppner-ibigroup (changes requested):

lib/components/narrative/metro/metro-itinerary.tsx
amy-corson-ibigroup: Just want to clarify that the routeLegs in this file does not affect transit itineraries - Miles & I discussed this offline and I think that's making this thread confusing (for me at least!!) so apologies for bad transparency!

PR #933, opened 2023-06-28
Call Taker Time Picker Support External Updates

miles-grant-ibigroup:
**Description:**

There were some regressions that got introduced here. This PR resolves these with a second state for the text input, separate from the actual time input.

**PR Checklist:**

### DIFF
--- a/lib/components/form/call-taker/date-time-picker.tsx
+++ b/lib/components/form/call-taker/date-time-picker.tsx
@@ -5,7 +5,7 @@ import { IntlShape, useIntl } from 'react-intl'
 import { isMatch, parse } from 'date-fns'
 import { OverlayTrigger, Tooltip } from 'react-bootstrap'
 import coreUtils from '@opentripplanner/core-utils'
-import React, { useEffect, useState } from 'react'
+import React, { useEffect, useRef, useState } from 'react'
 
 const { getCurrentDate, OTP_API_DATE_FORMAT, OTP_API_TIME_FORMAT } =
   coreUtils.time
@@ -104,6 +104,8 @@ const DateTimeOptions = ({
   const [time, setTime] = useState<string | undefined>(initialTime)
   const [typedTime, setTypedTime] = useState<string | undefined>(initialTime)
 
+  const timeRef = useRef(null)
+
   const intl = useIntl()
 
   /**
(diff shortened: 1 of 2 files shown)


binh-dam-ibigroup (changes requested):
Please elaborate on what problem was supposed to be fixed. Also, this PR introduces a regression that prevents entering shorthand time notation such as "9p" (for 9 pm) or "1145" (military format), that regression should be addressed.

miles-grant-ibigroup:
Good catch thanks! Let me know how f28dc50e4fd1c6121512d0732dfd225347f5e10a works for you

lib/components/form/call-taker/date-time-picker.tsx
```diff
@@ -134,7 +137,20 @@ const DateTimeOptions = ({
   // Update state when external state is updated
   useEffect(() => {
     if (initialDate !== date) setDate(initialDate)
-    if (initialTime !== time) setTime(initialTime)
+    if (initialTime !== time) {
+      setTime(initialTime)
+
+      // Don't update if still typing
+      if (timeRef.current !== document.activeElement) {
```
binh-dam-ibigroup, line 144: Seems to work now, but would it be simpler to handle `onBlur` instead?

lib/components/form/call-taker/date-time-picker.tsx
```diff
@@ -134,7 +137,20 @@ const DateTimeOptions = ({
...
+        setTypedTime(
+          safeFormat(dateTime, timeFormat, {
```
binh-dam-ibigroup, line 146: This needs to happen on initial render of the component (for situations such as pasting the URL into the browser), otherwise the "typed" time can be different from the status time as shown in the screenshot below.

image

binh-dam-ibigroup (commented):
The fix seems to work, however there remains one glitch when refreshing the page/pasting URL in the browser.

lib/components/form/call-taker/date-time-picker.tsx
miles-grant-ibigroup: I tried this, but some edge cases are missed when doing this

lib/components/form/call-taker/date-time-picker.tsx
miles-grant-ibigroup: Good catch thanks! Let me add an initial load update

miles-grant-ibigroup: Try 3ccd8d23

lib/components/form/call-taker/date-time-picker.tsx
```diff
@@ -201,24 +223,22 @@ const DateTimeOptions = ({
           className="datetime-slim"
           // TODO: Why does this no longer work when set as `value`? Is this a
```
binh-dam-ibigroup, line 224: Is this comment still relevant?

binh-dam-ibigroup (approved):
Works much better, thanks for the changes!

lib/components/form/call-taker/date-time-picker.tsx
miles-grant-ibigroup: Good catch thanks!

PR #935, opened 2023-06-29
fix(apiv2): adds 'headsign' to pattern request

daniel-heppner-ibigroup:
**Description:**
This PR adds one line to the query used in the route/pattern viewer so that the pattern headsigns are correctly displayed.

**PR Checklist:**

### DIFF
--- a/lib/actions/apiV2.js
+++ b/lib/actions/apiV2.js
@@ -609,6 +609,7 @@ export const findRoute = (params) =>
       
           patterns {
             id
+            headsign
             name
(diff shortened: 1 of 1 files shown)


daniel-heppner-ibigroup:
I don't think the Percy tests should be failing to even run from a change like this. I'll need to look into that.

binh-dam-ibigroup (approved):

miles-grant-ibigroup (approved):
Oh my gosh I can't believe this was the issue!

PR #936, opened 2023-06-30
a11y button fixes

amy-corson-ibigroup:
**Description:**
- Adds the role "image" to mode SVGs, to ensure the aria-labels aren't overlooked by AT
- Refactors "Copy Link" button `aria-live` to be more consistent across AT

**PR Checklist:**

### DIFF
--- a/lib/components/narrative/trip-tools.js
+++ b/lib/components/narrative/trip-tools.js
@@ -17,6 +17,7 @@ import React, { Component, useContext, useMemo } from 'react'
 import * as uiActions from '../../actions/ui'
 import { ComponentContext } from '../../util/contexts'
 import { IconWithText } from '../util/styledIcon'
+import InvisibleA11yLabel from '../util/invisible-a11y-label'
 import PopupTriggerText from '../app/popup-trigger-text'
 import startOver from '../util/start-over'
 
@@ -55,6 +56,12 @@ class CopyUrlButton extends Component {
   render() {
     return (
       <div>
+        {/* Announces copy button status to AT */}
+        <InvisibleA11yLabel aria-live="assertive">
+          {this.state.showCopied && (
+            <FormattedMessage id="components.TripTools.linkCopied" />
+          )}
+        </InvisibleA11yLabel>
         <Button
           aria-live={this.state.showCopied ? 'assertive' : undefined}
           className="tool-button"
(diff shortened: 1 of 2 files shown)


lib/components/narrative/trip-tools.js
```diff
@@ -55,6 +56,12 @@ class CopyUrlButton extends Component {
   render() {
     return (
       <div>
+        {/* Announces copy button status to AT */}
+        <InvisibleA11yLabel aria-live="assertive">
+          {this.state.showCopied && (
+            <FormattedMessage id="components.TripTools.linkCopied" />
```
miles-grant-ibigroup, line 62: The conditional should probably include the label too right

lib/components/narrative/trip-tools.js
```diff
@@ -55,11 +56,13 @@ class CopyUrlButton extends Component {
...
+          )}
+        </InvisibleA11yLabel>
+        <Button className="tool-button" onClick={this._onClick}>
           {this.state.showCopied ? (
             <span>
               <IconWithText Icon={Check}>
```
miles-grant-ibigroup, line 68: isn't this the same button again?

miles-grant-ibigroup (changes requested):
Looks good, some nitpicks

lib/components/narrative/trip-tools.js
amy-corson-ibigroup: Actually, conditionally rendering the `aria-live` causes inconsistent announcements on AT. The `aria-live` `div` needs to be present even when it's empty, its like a tiny status region

lib/components/narrative/trip-tools.js
amy-corson-ibigroup: Yes the button itself hasn't changed just removing the `aria-live` element from the button itself and adding to the `div`!

lib/components/narrative/trip-tools.js
miles-grant-ibigroup: Yikes that is UGLY! The a11y spec really has a ways to go...

lib/components/narrative/trip-tools.js
miles-grant-ibigroup: Yes but if you look at the file in the source code, it's there twice in a row!

miles-grant-ibigroup: Never mind that's just us handling state poorly

lib/components/narrative/trip-tools.js
amy-corson-ibigroup: Long term we may want to look at creating a global status region that handles little announcements like this as well as the results from the geocoder, trip results, etc!!

daniel-heppner-ibigroup (approved):

PR #939, opened 2023-07-10
Return focus to feedback button on modal close

amy-corson-ibigroup:
**Description:**

On modal close, focus resets to the button that the user pressed to open the modal, instead of resetting to the top of the page. 

Example: user presses the "Leave Feedback" button inside the sliding pane. With this PR, the sliding pane stays open on modal open, and when the user closes the modal, the focus is reset back to the "Leave feedback" button in the sliding pane.

This approach feels like a hack, so I'm very open to scrapping this in favor of another approach!

**PR Checklist:**

### DIFF
--- a/lib/components/app/popup.tsx
+++ b/lib/components/app/popup.tsx
@@ -62,6 +62,21 @@ const PopupWrapper = ({ content, hideModal }: Props): JSX.Element | null => {
 
   const title = intl.formatMessage({ id: `config.popups.${id}` })
 
+  /* HACK: Since Bootstrap 3.x does not support adding id or name to navItem, 
+  we have to grab a list of all navItems by className and find the correct button.
+  Since the sliding pane "Leave Feedback" button will always be in the DOM after
+  the navbar button, reverse + find should always find the correct button to return to. */
+
+  const navItemList = Array.from(
+    document.getElementsByClassName('navItem')
+  ).reverse() as HTMLElement[]
+  const focusElement = navItemList.find((el) => el.innerText === title)
+
+  const closeModal = () => {
+    hideModal()
+    focusElement?.focus()
+  }
+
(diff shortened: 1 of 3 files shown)


lib/components/app/popup.tsx
```diff
@@ -62,19 +62,34 @@ const PopupWrapper = ({ content, hideModal }: Props): JSX.Element | null => {
 
   const title = intl.formatMessage({ id: `config.popups.${id}` })
 
+  /* HACK: Since Bootstrap 3.x does not support adding id or name to navItem, 
+  we have to grab a list of all navItems by className and find the correct button.
+  Since the sliding pane "Leave Feedback" button will always be in the DOM after
+  the navbar button, reverse + find should always find the correct button to return to. */
+
+  const navItemList = Array.from(
+    document.getElementsByClassName('navItem')
```
binh-dam-ibigroup, line 71: We've done something like this for handling the keyboard up and down arrows in the menu dropdowns, so this should be fine and it seems to work.

lib/components/app/popup.tsx
```diff
@@ -62,19 +62,34 @@ const PopupWrapper = ({ content, hideModal }: Props): JSX.Element | null => {
...
+  ).reverse() as HTMLElement[]
+  const focusElement = navItemList.find((el) => el.innerText === title)
+
+  const closeModal = () => {
```
binh-dam-ibigroup, line 75: Use `useCallback` with the correct dependencies.

binh-dam-ibigroup (approved):
It seems to be working! I just have one minor change.

miles-grant-ibigroup (approved):
This should probably use refs but it's working for me so I'd say if we're feeling ambitious we leave a comment and resolve this another time!

PR #945, opened 2023-07-17
New fare product impl

daniel-heppner-ibigroup:
**Description:**

**PR Checklist:**


binh-dam-ibigroup:
@daniel-heppner-ibigroup or @miles-grant-ibigroup, in this PR, could you add the strings for "No fare info" that were in OTP-UI, and now removed from OTP-UI, into OTP-RR?

daniel-heppner-ibigroup:
I believe the percy tests are failing because the test server is not updated. @miles-grant-ibigroup

miles-grant-ibigroup:
Tests are still failing even with an updated server... Let me look into it

miles-grant-ibigroup (approved):
I made some of these changes, so we should probably get a third reviewer. But looks good and is working! Yay

binh-dam-ibigroup (changes requested):
Code is working, however there is some more cleanup needed (renaming `defaultFareKey`, removing the old default currency setting.

example-config.yml
```diff
@@ -320,7 +320,7 @@ itinerary:
   #  electronicRegular: "SmartCard Fare"
   #  student: "Student Fare"
   # One fare will always be shown by default
-  #defaultFareKey: electronicRegular
+  defaultFareType: { mediumId: null, riderCategoryId: null }
```
miles-grant-ibigroup, line 323: This is an awesome default, but we need it to also apply if the key is not set!

miles-grant-ibigroup (changes requested):
Not seeing fares by default -- problematic

example-config.yml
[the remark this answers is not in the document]
daniel-heppner-ibigroup: Yes!

lib/util/state.js
```diff
@@ -78,10 +84,13 @@ const DEFAULT_COSTS = {
 export function getTotalFare(
   itinerary,
   configCosts = {},
-  defaultFareKey = 'regular'
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: Wow, good catch

lib/util/state.js
```diff
@@ -45,16 +45,22 @@ function getDriveTime(itinerary) {
...
     coreUtils.itinerary.calculateTncFares(itinerary)
 
-  const transitFares = itinerary.fare?.fare
-  const transitFare =
-    (transitFares?.[defaultFareKey] || transitFares?.regular)?.cents || null
-  const fareCurrency = transitFare?.currency?.symbol || currency
+  const itineraryCost = coreUtils.itinerary.getItineraryCost(
+    itinerary?.legs,
+    defaultFareType?.mediumId || null,
+    defaultFareType?.riderCategoryId || null
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: Good catch. They are needed, since it's different from undefined. The fix in `default-itinerary` is in the `mapStateToProps` function.

example-config.yml
```diff
@@ -320,7 +320,7 @@ itinerary:
   #  electronicRegular: "SmartCard Fare"
   #  student: "Student Fare"
   # One fare will always be shown by default
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: Yeah, I think so. Functionality should be the same to the user

daniel-heppner-ibigroup:
Actually this raises an interesting question about when to display "no fare info" message. I think I will add this to the tech call.

example-config.yml
miles-grant-ibigroup: I'm seeing the fare in the metro itinerary, but not in the batch ui/calltaker ui or the trip details!

binh-dam-ibigroup: Fares were removed from trip-details as of version 5.0, right? Btw @daniel-heppner-ibigroup you could use the non-alpha trip-details 5.0.0.

miles-grant-ibigroup: Oh I wasn't aware...

miles-grant-ibigroup: Oh yes I know but I'm still not seeing "new-school" fares in the trip-details when I am seeing them in the metro itinerary. I think something is wrong with the data being passed around

daniel-heppner-ibigroup: Fixed the bug with a new alpha version of trip-details. 🙃

binh-dam-ibigroup:
@miles-grant-ibigroup @daniel-heppner-ibigroup Are we interpreting the fares correctly? (see screenshot with incorrect $10 total fares)

image

yarn.lock
```diff
@@ -2703,9 +2721,9 @@
     "@opentripplanner/core-utils" "^8.2.1"
 
 "@opentripplanner/types@^6.0.0-alpha.8":
-  version "6.0.0-alpha.8"
-  resolved "https://registry.yarnpkg.com/@opentripplanner/types/-/types-6.0.0-alpha.8.tgz#92a13c09b8be55ec83eba1d405bac5a791bba9d2"
-  integrity sha512-38YX9HJiYg/FonohTPrPdBoHV3Wff8GOgz/QgWMU7iiKdgDZvneYU3jJosuoAL3k035DCbA0NL4ngR7VyTbJmA==
+  version "6.0.0"
```
binh-dam-ibigroup, line 2724: Can we just use the latest types package version too?

binh-dam-ibigroup (approved):
Looks good now, thanks for the changes.

miles-grant-ibigroup (approved):
Thanks for these changes! Much cleaner. Glad this is finally getting merged!

PR #946, opened 2023-07-18
Support current location in PlaceEditor

binh-dam-ibigroup:
## Description

This PR fixes the `PlaceEditor` component so that when Current Location is selected from the location field,
the current position gets fetched, and the address of that location is populated in the component.

## PR Checklist


lib/components/user/places/place-editor.tsx
```diff
@@ -67,18 +84,49 @@ function makeLocationFieldLocation(favoriteLocation: UserSavedLocation) {
...
+    const { category, lat, lon, name } = location
     setValues({
       ...values,
-      address: name,
+      address:
+        // If the raw current location is passed without a name attribute (i.e. the address),
+        // set the "address" as the formatted coordinates of the current location at that time.
+        category === 'CURRENT_LOCATION'
+          ? intl.formatMessage({ id: 'common.coordinates' }, { lat, lon })
+          : name,
```
amy-corson-ibigroup, line 97: To confirm grabbing the current location worked, it could be good to append the current location string so the users sees `41.9746601, -87.6636681 (current location)`. Alternatively you may be able to remove the `map.currentLocation` i18n string since I don't think it gets used anywhere else.

binh-dam-ibigroup: I don't think I want to show the text "current location" because the favorite place will not necessarily be the current location when the user travels. Thoughts?

amy-corson-ibigroup: > 

I'd agree with this, I'd say the only time I'd show the text "current location" is when the user clicks this button in the location field. 
image

But I don't mind disregarding this idea, I think having the coordinates populate and the map marker is probably enough confirmation that the location was fetched.

lib/actions/location.tsx
daniel-heppner-ibigroup, line 1: yay! Thanks for adding TS!

lib/components/user/places/place-editor.tsx
```diff
@@ -67,18 +83,51 @@ function makeLocationFieldLocation(favoriteLocation: UserSavedLocation) {
...
+        this._setLocation({
+          category: 'CURRENT_LOCATION',
+          lat,
+          lon
+        })
+        getGeocoder(geocoderConfig)
+          .reverse({ point: coords })
+          .then(this._setLocation)
+          .catch((err: Error) => {
+            console.warn(err)
```
daniel-heppner-ibigroup, line 125: Will this prevent bugsnag from detecting errors resulting from failed geocoding requests? Do we care?

daniel-heppner-ibigroup (approved):
This is nice code! Good fix

lib/components/user/places/place-editor.tsx
binh-dam-ibigroup: Do you have an example in mind you can point me to?

binh-dam-ibigroup: I think I copied some other geocoder code that does the same thing.

binh-dam-ibigroup:
@amy-corson-ibigroup and @daniel-heppner-ibigroup, https://github.com/opentripplanner/otp-react-redux/pull/946/commits/6e5bd2131165ede99562236df2da537baceb7522 now addresses the "invalid address" error after picking a recently searched location in the PlaceEditor, let me know if that works.

PR #952, opened 2023-07-24
Release: July 24, 2023

amy-corson-ibigroup:
Includes the following PRs:


miles-grant-ibigroup (changes requested):
We should wait until the Percy tests are working on dev before merging. As for generating the list of PRs, in the past we would copy the list of commits into sublime text and run a bunch of macros. We could probably replicate that with a quick python script?

miles-grant-ibigroup:
Percy is passing again. I will mark as approved then this can be merged

miles-grant-ibigroup (approved):
Can confirm this commit works

PR #953, opened 2023-07-25
Fix: Create mode strings consistently across i18n

amy-corson-ibigroup:
**Description:**

Sorting itineraries by filtering based on a space inadvertently sorts itineraries differently in different languages. Example: 

| English | French |
|--------|-------|
|image|image| 

This PR instead sorts based on the presence of a plus sign, which should be a little more consistent although perhaps we may want to sort regardless of the mode string naming convention, in case that changes!

**PR Checklist:**

### DIFF
--- a/lib/components/narrative/narrative-itineraries.js
+++ b/lib/components/narrative/narrative-itineraries.js
@@ -427,12 +427,12 @@ class NarrativeItineraries extends Component {
         })
 
         // We determine which "bucket" to put this itinerary into based on what
-        // is currently a hack hack hack: looking at the presence of a space
+        // is currently a hack hack hack: looking at the presence of a "+"
         // in the mode string.
         //
         // Another approach may be cleaner, but this includes the least overhead
         const modeContainer =
-          modeString.indexOf(' ') > -1 ? modeItinMap.multi : modeItinMap.single
+          modeString.indexOf('+') > -1 ? modeItinMap.multi : modeItinMap.single
 
         // Now that we know the mode container to place our itinerary in, we do so
         if (!modeContainer[modeString]) modeContainer[modeString] = []
--- a/lib/components/narrative/narrative-itineraries.js
+++ b/lib/components/narrative/narrative-itineraries.js
@@ -2,7 +2,7 @@
 import { connect } from 'react-redux'
 import { differenceInDays } from 'date-fns'
 import { FormattedMessage, injectIntl } from 'react-intl'
-import { isFlex } from '@opentripplanner/core-utils/lib/itinerary'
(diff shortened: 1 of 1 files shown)


daniel-heppner-ibigroup (approved):
Oh my gosh, I knew this was a bad hack when I first saw it but didn't even consider i18n as a way to break it.

miles-grant-ibigroup (changes requested):
I don't think we can use this as a fix. Should we maybe go for an explicit list of modes that we check against?

lib/components/narrative/narrative-itineraries.js
```diff
@@ -427,12 +427,12 @@ class NarrativeItineraries extends Component {
...
 
         // We determine which "bucket" to put this itinerary into based on what
-        // is currently a hack hack hack: looking at the presence of a space
+        // is currently a hack hack hack: looking at the presence of a "+"
         // in the mode string.
         //
         // Another approach may be cleaner, but this includes the least overhead
         const modeContainer =
-          modeString.indexOf(' ') > -1 ? modeItinMap.multi : modeItinMap.single
+          modeString.indexOf('+') > -1 ? modeItinMap.multi : modeItinMap.single
```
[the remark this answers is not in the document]
amy-corson-ibigroup: Yeah I think you're right, see [130b5d7](https://github.com/opentripplanner/otp-react-redux/pull/953/commits/130b5d71eb269a499271316b9b3d842f9d46fd6e) for a better solution!

lib/components/narrative/narrative-itineraries.js
```diff
@@ -426,13 +426,18 @@ class NarrativeItineraries extends Component {
           itinerary: cur
         })
 
+        const containsTransit = cur.legs
+          .map((leg) => isTransit(leg.mode))
+          .includes(true)
```
miles-grant-ibigroup, line 432:
```javascript
const containsTransit = cur.legs
           .filter((leg) => isTransit(leg.mode))
``` 

You can probably clean this up even more!

miles-grant-ibigroup (changes requested):
Feel a bit embarrassed I didn't come up with this when first built this. Great fix! Just one nitpick

lib/components/narrative/narrative-itineraries.js
amy-corson-ibigroup: Thanks!! Added in [f1f73c6](https://github.com/opentripplanner/otp-react-redux/pull/953/commits/f1f73c66000a8f59142cf8190739e0e2898bab30)

miles-grant-ibigroup (approved):
Love it!

PR #954, opened 2023-07-25
Fix: stop viewer crash when no company.modes

amy-corson-ibigroup:
**Description:**
Fixes a bug where companies without modes crash the stop viewer.

To test: open any stop through the route viewer


miles-grant-ibigroup (approved):
Good catch thanks!

daniel-heppner-ibigroup (approved):
we love the ?

PR #955, opened 2023-07-25
Date Time Popup

binh-dam-ibigroup:
## Description

This PR changes the date/time popup to use the same popups as the mode selector.
Improvements are as follows:
- Hover effect - with a mouse, no need to click to open the date/time settings
- a11y: Date/time settings use a (non-modal) dialog with full keyboard suport, that can be dismissed with ESC or moving the mouse away.

## PR Checklist

| Before | After |
|--------|-------|
|  before screenshot      |    after screenshot   |


lib/components/form/batch-settings.tsx
```diff
@@ -186,63 +182,65 @@ function BatchSettings({
...
 
   return (
-    <>
-      <MainSettingsRow>
-        <StyledDateTimePreviewContainer
-          aria-controls="date-time-modal"
-          aria-expanded={dateTimeExpanded}
-          aria-label={intl.formatMessage({
-            id: 'components.BatchSettings.dateTimeSettingsLabel'
+    <MainSettingsRow onMouseEnter={checkModeSelectorPopup}>
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Looks like `mouseenter` wasn't enough, so I changed to `mousemove` in https://github.com/opentripplanner/otp-react-redux/pull/955/commits/25c6f549c757106f0b136a090a2f7ba0c5c403f2.

lib/components/form/batch-settings.tsx
```diff
@@ -186,63 +182,65 @@ function BatchSettings({
...
+      <DateTimeButton
+        open={dateTimeOpen}
+        setOpen={setDateTimeOpen}
+        // Prevent, in many cases (not all), the hover on date/time selector when mode selector has a popup open.
```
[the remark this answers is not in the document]
binh-dam-ibigroup: `mouseenter` was a little greedy. It looks like using `mousemove` in https://github.com/opentripplanner/otp-react-redux/pull/955/commits/25c6f549c757106f0b136a090a2f7ba0c5c403f2 takes care of everything, I think?

amy-corson-ibigroup (approved):
Looking great! Thanks for the changes!

lib/components/form/batch-styled.ts
```diff
@@ -1,35 +1,34 @@
...
-} from '@opentripplanner/trip-form'
 import styled, { css } from 'styled-components'
 
-import { commonInputCss, modeButtonButtonCss } from './styled'
-import DateTimePreview from './date-time-preview'
+export const buttonPixels = 51
 
-const activeCss = css`
+export const activeCss = css`
   background: #d5d5d5;
```
miles-grant-ibigroup, line 6: The mode selector gets darker on hover, while this gets lighter. Can we have this go darker instead?

lib/components/form/date-time-button.tsx
```diff
@@ -0,0 +1,190 @@
...
+      // Enable hover only if no popup has been triggered via keyboard.
+      // (This keeps the keyboard-triggered popups open while moving mouse around.)
+      enabled: !openWithKeyboard,
+      handleClose: safePolygon({
+        blockPointerEvents: false,
+        buffer: 0,
+        restMs: 500
+      })
+    }),
+    useClick(context),
```
miles-grant-ibigroup, line 123: I'm not sure I agree with a click opening the popup, as this doesn't match the mode selector behavior

miles-grant-ibigroup (changes requested):
Really great change! Shows how a small adjustment can make a huge different. Just some small UX nitpicks, but otherwise this is looking great! And working fantastic across all device types too

binh-dam-ibigroup:
> [...] Shows how a small adjustment can make a huge different.

Lol this "small" adjustment touched 15 files!

lib/components/form/date-time-button.tsx
binh-dam-ibigroup: > I'm not sure I agree with a click opening the popup, as this doesn't match the mode selector behavior

What do you mean by that? Hover is already enabled. Click is still needed for keyboard access.

lib/components/form/batch-styled.ts
binh-dam-ibigroup: Oh I see, it is a Safari thing! Addressed in https://github.com/opentripplanner/otp-react-redux/pull/955/commits/cfd7ee8eb08fc8b774a680fd636b9a23b9d2547d.

lib/components/form/date-time-button.tsx
miles-grant-ibigroup: The hover is good, but having the popup stick on click is behavior I think we should disable

binh-dam-ibigroup: Thanks for clarifying, I figured out that fix in https://github.com/opentripplanner/otp-react-redux/pull/955/commits/97a0eb06172395b7cfb7113a081703c32dc4d380.

miles-grant-ibigroup (approved):
New changes look great thanks! Only issue is now the Percy tests are failing because they can't click the button anymore, but this should be repairable.

PR #957, opened 2023-07-27
Fix endpoint save links

binh-dam-ibigroup:
## Description

This PR brings a fix so that when you click "Save as/Forget home/work" from an endpoint on the map, the resulting saved locations are correctly shown in the main pane under "My saved places" and in the account page under "Favorite Locations".

## PR Checklist

### DIFF
--- a/lib/components/form/user-settings.js
+++ b/lib/components/form/user-settings.js
@@ -211,7 +211,6 @@ class UserSettings extends Component {
   }
 
   render() {
-    console.log('user settings render')
     const {
       className = '',
       forgetSearch,
--- a/lib/components/form/user-settings.js
+++ b/lib/components/form/user-settings.js
@@ -184,7 +184,11 @@ class UserSettings extends Component {
 
   _getLocations = (user) => {
     const locations = [...user.savedLocations]
-    if (!locations.find(isWork)) {
+
+    const workLocation = locations.find(isWork)
+    if (workLocation) {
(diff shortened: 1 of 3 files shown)


lib/components/form/user-settings.js
```diff
@@ -65,6 +63,24 @@ function formatElapsedTime(timestamp, intl) {
...
 }
 
+/**
+ * Helper to ensure the home and work locations are correctly populated,
+ * or a placeholder is created.
+ */
+function addOrEditPlaceIfNeeded(locations, type, icon) {
+  const editedLocation = locations.find((loc) => loc.type === type)
+  if (editedLocation) {
+    if (!editedLocation.address) editedLocation.address = editedLocation.name
```
daniel-heppner-ibigroup, line 73: What's the purpose of setting the address to equal the name? Can you add a comment explaining why we do this?

daniel-heppner-ibigroup (approved):
Looks good, although I don't totally get how this fixes the bug. Can you explain in a comment?

lib/components/form/user-settings.js
binh-dam-ibigroup: Good point, I added the comment in https://github.com/opentripplanner/otp-react-redux/pull/957/commits/457cf905218b8195b3a0fcaf9dcae5299342b468.

amy-corson-ibigroup (approved):
Looks great!

PR #960, opened 2023-08-01
fix(default-itin): fix crash when currency is undefined

daniel-heppner-ibigroup:
**Description:**
Default itinerary had a bug that caused a crash when the currency was undefined. This is now handled correctly.

**PR Checklist:**


miles-grant-ibigroup (approved):
Thanks so much for figuring this out... What a terrible error

binh-dam-ibigroup (approved):

PR #961, opened 2023-08-02
fix(RoutePreviewOverlay): Recenter map when overlay is shown.

binh-dam-ibigroup:
## Description

This PR makes the map center on the endpoints when the `RoutePreviewOverlay` is shown, but is that too much?

## PR Checklist


miles-grant-ibigroup (approved):
I think this is a nice touch! And well implemented

amy-corson-ibigroup (approved):
Very nice change!!

PR #963, opened 2023-08-04
CallTaker: support persisting call taker mode to URL

miles-grant-ibigroup:
**Description:**

Calltaker mode selector wasn't persisting its state to the URL. This PR resolves this. It doesn't use the main URL parameter handler. Instead, it uses the `userQueryParam`

**PR Checklist:**

### DIFF
--- a/lib/components/form/call-taker/mode-dropdown.tsx
+++ b/lib/components/form/call-taker/mode-dropdown.tsx
@@ -52,7 +52,7 @@ function ModeDropdown({ modeDropdownOptions, onChangeModes }: Props) {
       onBlur={onChange}
       onChange={onChange}
       value={
-        typeof selectedMode === 'string' && selectedMode !== undefined
+        typeof selectedMode === 'string'
           ? selectedMode
           : modeDropdownOptions?.[0].label
       }
--- a/lib/components/form/call-taker/mode-dropdown.tsx
+++ b/lib/components/form/call-taker/mode-dropdown.tsx
@@ -1,6 +1,7 @@
 import { connect } from 'react-redux'
 import { TransportMode } from '@opentripplanner/types'
-import React, { useState } from 'react'
+import { useQueryParam } from 'use-query-params'
+import React, { useEffect } from 'react'
 
 import { defaultDropdownConfig } from '../../../util/call-taker'
 import { getModuleConfig, Modules } from '../../../util/config'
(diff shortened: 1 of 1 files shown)


lib/components/form/call-taker/mode-dropdown.tsx
```diff
@@ -23,21 +24,39 @@ type Props = {
...
-      const newModes = modeDropdownOptions.find(
-        (mdo) => mdo.label === e.target.value
-      )?.combination
-      onChangeModes(newModes || [])
     },
     []
   )
 
+  useEffect(() => {
+    if (selectedMode === undefined) {
```
amy-corson-ibigroup, line 39: Could you do `!selectedMode` here? Or would that be different?

lib/components/form/call-taker/mode-dropdown.tsx
```diff
@@ -23,21 +24,39 @@ type Props = {
...
+    onChangeModes(newModes || [])
+  }, [selectedMode, modeDropdownOptions, onChangeModes])
+
   return (
-    <select onBlur={onChange} onChange={onChange} value={selectedMode}>
+    <select
+      onBlur={onChange}
+      onChange={onChange}
+      value={
+        typeof selectedMode === 'string' && selectedMode !== undefined
```
amy-corson-ibigroup, line 55: same question here

amy-corson-ibigroup (approved):
Looks great works well! Quick question but I think it's more for my sake than yours.

lib/components/form/call-taker/mode-dropdown.tsx
miles-grant-ibigroup: `!selectedMode` is also true when `selectedMode` is `null`, `0`, and a few other javascript weirdnesses

lib/components/form/call-taker/mode-dropdown.tsx
miles-grant-ibigroup: if selectedMode is null or 0 we don't want this behavior!

binh-dam-ibigroup: Is the `string` type check enough here?

binh-dam-ibigroup (approved):
LGTM

lib/components/form/call-taker/mode-dropdown.tsx
miles-grant-ibigroup: Good catch thanks!

PR #966, opened 2023-08-07
refactor: improve OTP error messages

miles-grant-ibigroup:
**Description:**

Some of the existing messages are not accurate in all cases. This PR improves the language 

**PR Checklist:**

### DIFF
--- a/i18n/en-US.yml
+++ b/i18n/en-US.yml
@@ -375,12 +375,12 @@ components:
       body: >-
         No transit connection was found between your origin and destination on
         the selected day of service, using the vehicle types you selected.
-      header: No transit connections for chosen vehicle types
+      header: No transit connections
     NO_TRANSIT_CONNECTION_IN_SEARCH_WINDOW:
       body: >-
         A transit connection was found, but it was outside the search window,
         try adjusting your search window, using the vehicle types you selected.
-      header: No transit connection for chosen vehicle types in search window
+      header: No transit connection in search window
     OUTSIDE_BOUNDS:
       body: >-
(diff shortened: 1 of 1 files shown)


i18n/en-US.yml
```diff
@@ -374,13 +374,13 @@ components:
     NO_TRANSIT_CONNECTION:
       body: >-
         No transit connection was found between your origin and destination on
-        the selected day of service.
-      header: No transit connections
+        the selected day of service, using the vehicle types you selected.
+      header: No transit connections for chosen vehicle types
```
amy-corson-ibigroup, line 378: Could we keep just "No Transit Connections" for the header? The result here is looking a little wordy & I think the description does a good job of clarifying.
image

i18n/en-US.yml
```diff
@@ -374,13 +374,13 @@ components:
...
     NO_TRANSIT_CONNECTION_IN_SEARCH_WINDOW:
       body: >-
         A transit connection was found, but it was outside the search window,
-        try adjusting your search window.
-      header: No transit connection in search window
+        try adjusting your search window, using the vehicle types you selected.
+      header: No transit connection for chosen vehicle types in search window
```
amy-corson-ibigroup, line 383: Same thing here I think!

amy-corson-ibigroup (changes requested):

i18n/en-US.yml
miles-grant-ibigroup: Sure!

daniel-heppner-ibigroup (approved):

PR #969, opened 2023-08-09
Include Route Details in Route Button

amy-corson-ibigroup:
**Description:**
- Includes route details in buttons for AT so buttons aren't just read to users as "toggle on map"
- Adds invisible indication to AT users that a route is already selected
- Corrects `role="img"` tag in svg

**PR Checklist:**


lib/components/viewers/RouteRow.js
```diff
@@ -69,11 +60,11 @@ export const ModeIconElement = styled.span`
...
 const RouteDetailsContainer = styled.div`
   align-items: center;
-  display: grid;
-  grid-template-columns: repeat(4, auto);
+  display: flex;
   gap: 12px;
-  justify-items: center;
   margin-left: 5px;
+  // Make sure route details always leaves enough room for pattern button
+  max-width: 87%;
```
binh-dam-ibigroup, line 67: How about, instead, setting a `padding-right` on `RouteRowButton` with value the `PatternButton` width plus the margins around it?

lib/components/viewers/RouteRow.js
```diff
@@ -83,15 +74,16 @@ const PatternButton = styled.button`
...
   height: 40px;
   margin-right: 8px;
-  min-width: 40px;
-  opacity: ${(props) => (props.display ? '80%' : 0)};
-  position: relative;
+  position: absolute;
+  right: 5px;
   transition: all ease-out 0.1s;
   z-index: ${(props) => (props.display ? 2 : -2)};
+  width: 40px;
```
binh-dam-ibigroup, line 81: Sort CSS

lib/components/viewers/RouteRow.js
```diff
@@ -172,42 +164,51 @@ export class RouteRow extends PureComponent {
...
-          title={routeMapToggleText}
-        />
+        >
+          <InvisibleA11yLabel>
+            {isActive &&
+              intl.formatMessage({
+                description: 'Identifies that route is active on map',
+                id: 'components.RouteViewer.selected'
+              })}
+          </InvisibleA11yLabel>
```
binh-dam-ibigroup, line 181: Should this be an `aria-selected` attribute on the button?

binh-dam-ibigroup (approved):
LGTM as is with a few suggestions. If you decide to keep the "Route selected" text, let me know for adding the French text.

lib/components/viewers/RouteRow.js
amy-corson-ibigroup: Replaced in [865becf](https://github.com/opentripplanner/otp-react-redux/pull/969/commits/865becfc193681477669da87330c8446c3c747f3)! The `aria-pressed` doesn't seem to have a consistent impact on VoiceOver and I'm not sure why, but I think you're right it's more appropriate.

lib/components/viewers/RouteRow.js
amy-corson-ibigroup: Great idea, thank you! Changed in [6384960](https://github.com/opentripplanner/otp-react-redux/pull/969/commits/638496002b716f8d199ae115bd56a122f175a677)

daniel-heppner-ibigroup (approved):
Thanks, LGTM!

PR #972, opened 2023-08-10
Fix percy tests

binh-dam-ibigroup:
## Description

This PR changes Percy tests to use the mock OTP server with a curated `mock.har` file checked into the repo,
avoiding test failures resulting from external factors such as schedule changes.
The tests cover OTP-RR desktop and mobile UI.

A few things to note:
- The HAR file is no longer downloaded. Went from ~50MB down to ~2MB.
- Calltaker is left out for now, we can put them back once we have a handle of how these tests perform.
- To save on allowance, I have removed some screenshots we used to do, e.g. those with a blank UI.
- This process unveiled a number of undesired, extra web queries made by the UI and other oddities. They are noted with FIXME in the Percy test script. I have not attempted to fix them in this PR.

## PR Checklist


.github/workflows/percy.yml
```diff
@@ -25,36 +25,12 @@ jobs:
...
-        env:
-          PERCY_OTP2_CONFIG_URL: ${{ secrets.PERCY_OTP2_CONFIG_URL_METRO_MODE_SELECTOR }}
-      - name: Take Percy Snapshots
-        # Move everything from latest commit back to staged
-        run: npx percy exec -- npx jest percy/percy.test.js --force-exit
-        env:
-          PERCY_TOKEN: ${{ secrets.PERCY_TOKEN_CALL_TAKER_OTP2 }}
-          PERCY_OTP_CONFIG_OVERRIDE: /tmp/otp2config.yml
-          OTP_RR_PERCY_CALL_TAKER: true
+  # TODO: Reinstate calltaker UI tests.
```
miles-grant-ibigroup, line 36: I'm ok with re-adding the call taker tests separately, but ti seems like the tests are now running in mobile mode only (at least according to the Percy results)

percy/README.md
```diff
@@ -0,0 +1,45 @@
+# Percy Tests
+
+Percy tests are end-to-end tests of the OTP-RR desktop, mobile, and call-taker UIs
+with screenshots taken at key steps of the user interaction.
+
+To run locally, uncomment the line in `percy.test.js` that says `headless: false`,
+then enter (Linux syntax):
+```sh
+PERCY_OTP_CONFIG_OVERRIDE=<path-to-yml> npx percy exec -- npx jest percy/percy.test.js --force-exit
```
miles-grant-ibigroup, line 9: There is a default Percy yml that we might want to update instead? That way we won't need to supply an override

percy/percy.test.js
```diff
@@ -6,7 +6,7 @@ import puppeteer from 'puppeteer'
 const percySnapshot = require('@percy/puppeteer')
 
 const OTP_RR_TEST_CONFIG_PATH = '../percy/har-mock-config.yml'
-const { OTP_RR_PERCY_CALL_TAKER, OTP_RR_PERCY_MOBILE } = process.env
```
miles-grant-ibigroup, line 9: The mobile env variable was used to enter mobile and leave mode -- it's more than changing the screen size, the user agent needs to be faked

percy/percy.test.js
```diff
@@ -440,18 +366,70 @@ test('OTP-RR', async () => {
...
-  )
+test('OTP-RR Mobile', async () => {
+  const page = await loadPath('/')
+  await page.setUserAgent('android')
+  await page.setViewport({
+    height: 1134,
+    width: 750
+  })
+  // Need to reload to load mobile view properly
+  await page.reload()
```
miles-grant-ibigroup, line 390: This trick isn't working for me. I think the mobile tests need to be executed separately for them to register correctly in percy

miles-grant-ibigroup (changes requested):
Tests are running, and code looks good, but some small overall issues. 

Annoyed at the call taker tests disappearing, as that's where we often introduce regressions but I see how it makes things complicated with the request-based graphql queries

percy/README.md
binh-dam-ibigroup: Do you mean a YML that uses all localhosts? The geocoder would have to be mocked eventually too.

percy/percy.test.js
binh-dam-ibigroup: Not sure how Percy is reporting snapshots, but I have re-separated the tests with [c4fd913](https://github.com/opentripplanner/otp-react-redux/pull/972/commits/c4fd913c4c53bc752ec01444ed449d20bd5f580d).

percy/percy.test.js
binh-dam-ibigroup: That was always the case.

.github/workflows/percy.yml
binh-dam-ibigroup: Call-taker is back with https://github.com/opentripplanner/otp-react-redux/pull/972/commits/8b388e3b41bc342c9cf1774c91b44ca1324ddfbb (although only one screenshot is taken of the main panel, the other steps are the same as desktop so they are skipped).

percy/README.md
binh-dam-ibigroup: I made the configs explicit now in `percy.yml` for the regular vs. calltaker builds.

percy/percy.test.js
binh-dam-ibigroup: So far, I think there 18 snapshots taken total for each commit, but it is not obvious to list them all in Percy.

.github/workflows/percy.yml
miles-grant-ibigroup: I'm not sure if we can just skip them -- we need the screenshots to make sure that there are no regressions in the trip planning form or the batch itinerary

binh-dam-ibigroup: The first Calltaker screenshot occurs after a trip query is made (but it is the same query result as the regular config). 

How about letting the Calltaker test run without taking screenshots? 
The other steps after that are identical to the ones for regular desktop and would result in duplicate screenshots.

miles-grant-ibigroup (approved):
Locally this is looking good, in Percy I'm still not quite seeing what I was expecting but we can configure that after this is merged

daniel-heppner-ibigroup (approved):
Looks fine, thanks for doing the unfun work of Percy fixes. Hopefully it works well!

PR #973, opened 2023-08-15
Reset focus to top on each account step

amy-corson-ibigroup:
**Description:**
In account creation, each account setup step will set the focus to the `h1` element at the top of the step and announce it to AT with an `aria-live="assertive"`

**PR Checklist:**


daniel-heppner-ibigroup (approved):
Wonderful work. you are becoming an expert on this stuff!

lib/components/user/sequential-pane-display.tsx
```diff
@@ -50,6 +50,12 @@ class SequentialPaneDisplay<T> extends Component<Props<T>> {
     routeTo(`${parentPath}/${nextId}`)
   }
 
+  h1Ref = React.createRef<HTMLHeadingElement>()
+
+  _handleFocus = () => {
```
miles-grant-ibigroup, line 55: Can we rename this to be `focusHeader`, since we're not handling a focus event!

miles-grant-ibigroup (changes requested):
One nitpick but otherwise really cool!

PR #974, opened 2023-08-17
OTP2 field trip support (Trip Hashes)

miles-grant-ibigroup:
**Description:**

Trip hashes were previously generated with OTP1 syntax. This doesn't work with OTP2, as the trip id location changes in the graphQL api. This PR resolves this by using the new syntax where applicable. TODO: needs a lot more testing.

There is also a null error repair included that seems to be specific to trimet and field trip.

**PR Checklist:**


daniel-heppner-ibigroup (approved):
Thanks for the fix!

philip-cline (approved):
LGTM

PR #975, opened 2023-08-17
Fix(StopViewer): avoid repeating Related Stops

amy-corson-ibigroup:
**Description:**

An issue where `nearbyStops` can contain duplicates is leading to stops getting rendered 10+ times in the Related Stops panel. This PR seeks to filter duplicates out of that `nearbyStops` array but it may be worth investigating why we're seeing duplicates in the first place.

To recreate bug:
- Use ST config
- Go to the Route Viewer.
- Choose Route 545.
- Select the 5th and Pine stop - Stop ID 760.
- Show Extra stops. Scroll down the list.
You can see stop ID 700 gets returned like 10 times.

**PR Checklist:**


miles-grant-ibigroup (approved):
This PR changes behavior to send an empty array instead of undefined which happens to actually fix an unrelated bug where the related stops panel shows up when there's no related stops. well done! Snapshots need to be updated though (`yarn unit -u`)

lib/components/viewers/stop-viewer.js
```diff
@@ -481,7 +481,9 @@ const mapStateToProps = (state) => {
   const { autoRefreshStopTimes = true, favoriteStops } = state.user.localUser
   const stopLookup = transitIndex.stops
   const stopData = stopLookup[ui.viewedStop.stopId]
-  const nearbyStops = stopData?.nearbyStops?.map((stopId) => stopLookup[stopId])
+  const nearbyStops = Array.from(new Set(stopData?.nearbyStops))?.map(
```
philip-cline, line 484: Maybe it could be a bit cleaner if we broke the set onto a different line? 
like: `uniqueNearbyStops = Array.from(new Set(stopData?.nearbyStops))` 
but I'm not too sure, so up to you

philip-cline (approved):
LGTM, one tentative suggestion to make things a bit more readable but I'm not picky

PR #977, opened 2023-08-18
Fix View Switcher

binh-dam-ibigroup:
## Description

This PR fixes issues around the "Plan Trip" view switcher button.
With this PR, the following actions are now possible:
- From "Account" > "Trips" > "Edit a trip", "Plan Trip" is no longer highlighted, and clicking "Plan Trip" returns to the plan view, showing any previously active itineraries.
- After returning to a plan view from the trips (or other account page), clicking the browser's "Back" button once returns to that account page.

These actions should still be possible:
- With an active itinerary shown, after planning an other itinerary (e.g. by changing the start/end point), clicking the browser's "Back" button once still brings back the previously shown itinerary.
- Navigating to and from the Route Viewer is still possible, with UI response by clicking the browser's "Back" button once.

## PR Checklist


amy-corson-ibigroup (approved):
This looks good! Having trouble testing some of the potential actions in the description due to what looks like an unrelated bug with saving trips, but everything around view switcher highlight on account page and back button navigation LGTM!

daniel-heppner-ibigroup (approved):
Thanks for the improvement, some cleanup in here that I recently was dealing with.

PR #978, opened 2023-08-22
Support Configuring Batch UI Sort

miles-grant-ibigroup:
**Description:**

This PR adds 2 new configuration options around the batch UI sorting mechanism, allowing for full control over what is displayed and in what order.

Default behavior is unchanged

## Seeking help with:
- Typescript types, they are a bit funky at the moment
- Avoiding repeating code, especially around the typescript

**PR Checklist:**


miles-grant-ibigroup:
@daniel-heppner-ibigroup I could use your help with these TS errors

lib/components/narrative/narrative-itineraries.js
```diff
@@ -577,7 +580,8 @@ const mapStateToProps = (state) => {
     groupByMode: groupItineraries,
     groupTransitModes,
     mergeItineraries,
-    showHeaderText
+    showHeaderText,
+    sortModes
```
binh-dam-ibigroup, line 584: Would that be better if `NarrativeItinerariesHeader` became a connected component and derive the props it needs from redux, removing those from `NarrativeItineraries`?

lib/components/util/sortOptions.ts
```diff
@@ -45,5 +61,8 @@ export const sortOptions = (
     }
   ]
 
-  return sortOptionsArray
+  return sortOptionsArray.filter((sortOption) =>
+    // @ts-expect-error TODO: why is this failing
```
binh-dam-ibigroup, line 65: Create `type SortOptionEntry = { text: string, value: SortOptionsType }`, and declare `sortOptionsArray` to be of type `SortOptionEntry[]`.

lib/reducers/create-otp-reducer.js
```diff
@@ -162,7 +162,7 @@ export function getInitialState(userDefinedConfig) {
       sort: {
         direction: 'ASC',
         // Only apply custom sort if batch routing is enabled.
-        type: 'BEST'
+        type: config?.itinerary?.defaultSort || 'BEST'
```
binh-dam-ibigroup, line 165: `config` is defined at line 90 so you don't need to write `config?`.

binh-dam-ibigroup (approved):
Works well! See the few suggested tweaks.

lib/components/narrative/narrative-itineraries.js
miles-grant-ibigroup: Yes, but I don't think that's possible since both of those components get subclassed

lib/reducers/create-otp-reducer.js
miles-grant-ibigroup: Good catch thanks!

lib/components/util/sortOptions.ts
miles-grant-ibigroup: Great fix thanks!

lib/components/narrative/narrative-itineraries.js
miles-grant-ibigroup: I think it's a good architecture!

daniel-heppner-ibigroup (approved):
Perfect! thank you for adding example-config updates too

PR #980, opened 2023-08-22
Update dependencies

amy-corson-ibigroup:
**Description:**
- Updates OTP-UI dependencies
- Updates `create-OTP-reducer` to avoid crash

**PR Checklist:**


daniel-heppner-ibigroup (approved):
LGTM, thanks for the updates!

lib/components/narrative/metro/default-route-renderer.tsx
```diff
@@ -34,7 +34,7 @@ const DefaultRouteRenderer = ({
   leg,
   style
 }: RouteRendererProps): JSX.Element => {
-  const routeTitle = leg.routeShortName || leg.route || leg.routeLongName
+  const routeTitle = leg.routeShortName || leg.routeLongName
```
miles-grant-ibigroup, line 37: I think `leg.route` is sometimes a value. Can we do `leg?.route` instead?

miles-grant-ibigroup: and while we're at it `leg?.routeShortName` and `leg?.routeShortName`

package.json
```diff
@@ -39,7 +39,7 @@
     "@bugsnag/plugin-react": "^7.17.0",
     "@floating-ui/react": "^0.19.2",
     "@opentripplanner/base-map": "^3.0.13",
-    "@opentripplanner/core-utils": "^9.0.3",
+    "@opentripplanner/core-utils": "^11.0.0",
```
miles-grant-ibigroup, line 42: Once we're upgrading can we please also update `itinerary-body`? `yarn upgrade -S @opentripplanner --latest` should give us all we need

miles-grant-ibigroup (changes requested):
Just a few comments!

miles-grant-ibigroup:
Also snapshots need to be updated!

PR #981, opened 2023-08-22
Fix itinerary order

binh-dam-ibigroup:
## Description

Responses for the different mode combinations requested to OTP were stored in the order they are returned.
The order of the responses could change when performing the same search again.

This PR ensures that the responses are stored in the same order as the mode combinations sent to OTP,
and also preserves (and fixes some) itinerary searches functionality with the Field Trip module.

## PR Checklist

### DIFF
--- a/lib/actions/apiV2.js
+++ b/lib/actions/apiV2.js
@@ -919,7 +919,7 @@ export function routingQuery(searchId = null, updateSearchInReducer) {
 
     dispatch(setItineraryView(ItineraryView.LIST))
 
-    combinations.forEach((combo) => {
+    combinations.forEach((combo, index) => {
       const query = generateOtp2Query(combo)
       dispatch(
         createGraphQLQueryAction(
@@ -982,6 +982,7 @@ export function routingQuery(searchId = null, updateSearchInReducer) {
               }
 
               return {
+                index,
                 response: {
                   plan: {
                     ...response.data?.plan,
--- a/lib/actions/field-trip.js
+++ b/lib/actions/field-trip.js
@@ -687,7 +687,7 @@ function checkValidityAndCapacity(state, request) {
 
     // iterate through itineraries to check validity and assign field trip
     // groups
-    response.plan.itineraries.forEach((itinerary) => {
+    response.plan.itineraries.forEach((itinerary, itinIdx) => {
       let itineraryCapacity = Number.POSITIVE_INFINITY
 
       // check each individual trip to see if there aren't any trips in this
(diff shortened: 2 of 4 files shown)


binh-dam-ibigroup:
Taking this back to draft and assigning to me, this PR broke the itinerary sorting!

miles-grant-ibigroup (changes requested):
This works well for the metro UI, but breaks the field trip module

binh-dam-ibigroup:
> This works well for the metro UI, but breaks the field trip module

@miles-grant-ibigroup Which issue are you seeing that doesn't occur on the dev branch? If it is "Unhandled Rejection (TypeError): Cannot read properties of undefined (reading 'pending')" while planning itineraries for a field trip, this error occurs on the dev branch too.

miles-grant-ibigroup (approved):
We will fix the field trip separately

miles-grant-ibigroup (changes requested):
Actually the bug is dependent on this PR

binh-dam-ibigroup:
> breaks the field trip module

How would changes from https://github.com/opentripplanner/otp-react-redux/pull/981/commits/31c000b6b2ff458f9985ef882f41520e45752c99 work?

miles-grant-ibigroup:
No unfortunately field trip is still broken. I think the culprit is using `$set` instead of `$push` which means that the old itinerary results are overwritten

binh-dam-ibigroup:
> No unfortunately field trip is still broken. I think the culprit is using `$set` instead of `$push` which means that the old itinerary results are overwritten

@miles-grant-ibigroup Could you provide steps to reproduce, please?

miles-grant-ibigroup:
If you launch a field trip configuration with an OTP2 server, go to the new tab of the field trip requests window, select a trip with a date int he future and with a class size of over 40, and then click plan, you'll see that without this PR many results appear in the results pane. With this PR, only one result appears (and the number of passengers per vehicle is also hidden)

binh-dam-ibigroup:
> If you launch a field trip configuration with an OTP2 server, go to the new tab of the field trip requests window, select a trip with a date int he future and with a class size of over 40, and then click plan, you'll see that without this PR many results appear in the results pane. With this PR, only one result appears (and the number of passengers per vehicle is also hidden)

@miles-grant-ibigroup Ok the correct redux update process was tricky to find, but I think this one is the good one.

daniel-heppner-ibigroup (approved):
Add this to the list of insane code written to make field trip continue working. Thank you for figuring this out, Binh.

PR #982, opened 2023-08-23
refactor: avoid un-needed time chop-off

miles-grant-ibigroup:
**Description:**

Itinerary body was sometimes breaking time text when it shouldn't be. This PR resolves this. Will leave PR open for now to collect other styling regressions.

**PR Checklist:**


binh-dam-ibigroup:
Just came across this one. Could setting the time-column component to have `min-width: 60px` and remove `width: 60px` do the trick regardless of the width of the containers?

miles-grant-ibigroup:
Possibly! I know there are a lot of nesting styles going on here and then quite a bit of a11y changes happened as well so I'll have to spend some proper time with it to double check that doesn't introduce any regressions

AdrianaCeric:
I found that removing the width also did the trick. I think this is a lazy hack, but I've yet to find something else. Will look into it more, but let me know if you have any comments/suggestions!

binh-dam-ibigroup (approved):

miles-grant-ibigroup:
Removing styles is always a good thing!

PR #985, opened 2023-08-28
Fix pattern viewer

binh-dam-ibigroup:
## Description

This PR fixes
- the pattern dropdown in the Patter Viewer, so that the correct name for the viewed pattern is always shown,
- the stops list, so it can handle looped patterns.

**PR Checklist:**


miles-grant-ibigroup (approved):
A lot of good fixes here, thanks! Working great no regressions

daniel-heppner-ibigroup (approved):
Works well, thanks for the fixes. I hope we can generate some types from the schema soon to eliminate all these types scattered around.

PR #989, opened 2023-08-29
Refine desktop location fetch

binh-dam-ibigroup:
## Description

Undoes #984 and instead, displays an alert on desktop when attempting to use the current location for the first time after loading the page, and location access is blocked.

i18n messages are provided when the user denied location access.

## PR Checklist

### DIFF
--- a/lib/actions/location.tsx
+++ b/lib/actions/location.tsx
@@ -57,9 +57,19 @@ export function getCurrentPosition(
               })
             )
           }
-          // FIXME, analyze error code to produce better error message.
-          // See https://developer.mozilla.org/en-US/docs/Web/API/GeolocationPositionError
-          dispatch(receivedPositionError({ error }))
+          const newError = { ...error }
+          if (error.code === 1) {
+            // i18n for user-denied location message (error.code = 1 on secure origins).
+            if (
+              window.location.protocol === 'https:' ||
+              window.location.host.startsWith('localhost:')
+            ) {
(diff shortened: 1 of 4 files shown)


lib/actions/location.tsx
```diff
@@ -47,9 +48,28 @@ export function getCurrentPosition(
         // On error
         (error) => {
           console.log('error getting current position', error)
-          // FIXME, analyze error code to produce better error message.
-          // See https://developer.mozilla.org/en-US/docs/Web/API/GeolocationPositionError
-          dispatch(receivedPositionError({ error }))
+          // On desktop, after user clicks "Use location" from the location fields,
+          // show an alert and explain if location is blocked.
+          if (!isMobile() && error.code === 1) {
+            window.alert(
```
miles-grant-ibigroup, line 54: A bit clunky, but I think good enough for now! Could we possibly add a TODO here to move this code into `location-field` and handle this more elgantly

miles-grant-ibigroup (approved):
Much better thank you!

lib/actions/location.tsx
binh-dam-ibigroup: I added the TODO in https://github.com/opentripplanner/otp-react-redux/pull/989/commits/933d34a0c8fc1838653a27433651b5e21b92c218.

philip-cline (approved):
LGTM

PR #990, opened 2023-08-29
Alert if user clicks Plan Trip with invalid mode selection

binh-dam-ibigroup:
## Description

This PR adds an alert shown when user clicks "Plan Trip" with invalid modes selected (e.g. "car" alone).
No new alerts are shown when planning is triggered outside of clicking the "Plan Trip" button.

## PR Checklist


lib/actions/api-constants.ts
```diff
@@ -0,0 +1,5 @@
+export const RoutingQueryResult = {
```
amy-corson-ibigroup, line 1: Would it be appropriate to rename to `ROUTING_QUERY_RESULT`? Even just to make it a bit more distinct from `routingQueryResult` in `batch-settings.tsx`

i18n/en-US.yml
```diff
@@ -227,6 +227,7 @@ components:
     modeSelectorLabel: Select a transit mode
   BatchSettings:
     destination: destination
+    invalidModeSelection: Cannot plan a trip using the selected modes.
```
amy-corson-ibigroup, line 230: Could we add another sentence like "Try including transit in your mode selection" just to point people in the right direction?

amy-corson-ibigroup (approved):
Very helpful! Thanks Binh! A couple suggestions but nothing blocking

lib/actions/api-constants.ts
binh-dam-ibigroup: Instead of changing the case, I picked a (better) name in https://github.com/opentripplanner/otp-react-redux/pull/990/commits/8a0e6c906d37de4c0a8359da6fde1f7d77544b44.

i18n/en-US.yml
binh-dam-ibigroup: Good idea! Added in https://github.com/opentripplanner/otp-react-redux/pull/990/commits/c72cbc70eeb78434188d524794b78dd25c406370.

AdrianaCeric:
What is the expected flow before producing this error? When planning a trip using car-only on the mode selector, I do not receive an alert
image

AdrianaCeric (approved):
This works great! I'm wondering if this alert could also be shown when the plan trip button isn't selected. For instance, when planning trips I almost never click plan trip. I mostly select the start and end positions on the map

PR #992, opened 2023-09-07
Improve feedback when saving locations, add toast notifications

binh-dam-ibigroup:
## Description

This PR improves the user feedback from setting the home or work locations, from the map and the places editor, notably with toast notifications or alerting logged out users they need to log in. Some other confirmation alert messages are also converted into toast notifications.

Under the hood, the added `react-hot-toast` package improves the "politeness" of the confirmation messages shown.

## PR Checklist


lib/util/user.js
```diff
@@ -138,7 +139,7 @@ export function convertToLegacyLocation(place) {
     lon,
     // HACK: If a place name and address are provided, put the address in parentheses
     // to mimic the existing LocationField behavior for "work" and "home".
-    // TODO: use addInParentheses from location-field (requires passing an intl contexty).
+    // TODO: use addInParentheses from location-field (requires passing an intl context).
```
amy-corson-ibigroup, line 142: 👍

amy-corson-ibigroup (approved):
Looks good! Tested on mobile and AT. 

It would be great to move all `window.alerts` over to Toast, as this does make the user alerts a little inconsistent, but I think this is a great place to start!!

i18n/en-US.yml
```diff
@@ -52,6 +52,8 @@ actions:
     emailVerificationResent: The email verification message has been resent.
     genericError: "An error was encountered: {err}"
     itineraryExistenceCheckFailed: Error checking whether your selected trip is possible.
+    mustBeLoggedInToSavePlace: Please log in to save locations.
```
AdrianaCeric, line 55: [Testing] How is it possible to save locations without logging in?

i18n/en-US.yml
```diff
@@ -52,6 +52,8 @@ actions:
...
+    placeRemembered: "Your {placeType} has been set to: {address}."
```
AdrianaCeric, line 56: `placeType` is either "custom" or "dining" which results in the alert becoming either:
- "Your custom has been set to: Sesame Street."
- "Your dining has been set to: Sesame Street."

Adding something like "place type" would help make this alert more clear
```suggestion
    placeRemembered: "Your {placeType} place type has been set to: {address}."
```

AdrianaCeric (changes requested):
I think Toast is a cool package, good work! Just one comment -- also not sure how to test the login problem

i18n/en-US.yml
binh-dam-ibigroup: Good point! I changed the confirmation popup in https://github.com/opentripplanner/otp-react-redux/pull/992/commits/e1c64536b1f836a7597349d1e54636fe62043861 so it shows the correct place name, and did away with showing the address as there can be other settings for places later.

AdrianaCeric (approved):
Looks good with the change!

PR #993, opened 2023-09-07
Percy: mobile snapshots, separate key for calltaker, mock geocoder

binh-dam-ibigroup:
## Description

This PR changes the percy test so it runs and keeps the mobile snapshots, and uses a separate percy key for calltaker mode and runs calltaker snapshots in a separate job to avoid race conditions. Geocoder is mocked too now. Feedback welcome on how to reduce run time or reuse the checkout steps.

**PR Checklist:**


miles-grant-ibigroup (approved):
Works great and snapshots look fantastic. I think this is as clean as something like this can get well done!

daniel-heppner-ibigroup (approved):
Very good assortment of screenshots, thank you! This should make the percy tests more useful.

PR #994, opened 2023-09-08
Remove OTP1 Stop Time Filtering

miles-grant-ibigroup:
**Description:**

When we switched to the new stop viewer, we forgot to remove a filter that was only relevant for the old stop viewer and OTP1. The result was no stop times appearing at the second to last stop! This PR removes this filter and fixes things.

**PR Checklist:**

### DIFF
--- a/lib/components/viewers/live-stop-times.tsx
+++ b/lib/components/viewers/live-stop-times.tsx
@@ -219,6 +219,7 @@ class LiveStopTimes extends Component<Props, State> {
                   this.renderDay(homeTimezone, time.day, now)}
                 <PatternRow
                   homeTimezone={homeTimezone}
+                  key={pattern.id}
                   pattern={pattern}
                   route={{
                     ...route,
--- a/lib/util/viewer.js
+++ b/lib/util/viewer.js
@@ -136,11 +136,8 @@ export function getStopTimesByPattern(stopData) {
           times: []
         }
       }
-      // Exclude the last stop, as the stop viewer doesn't show arrival times to a terminus stop.
-      const filteredTimes = times.filter(excludeLastStop)
 
-      stopTimesByPattern[id].times =
-        stopTimesByPattern[id].times.concat(filteredTimes)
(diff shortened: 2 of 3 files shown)


lib/components/viewers/live-stop-times.tsx
```diff
@@ -219,6 +219,7 @@ class LiveStopTimes extends Component<Props, State> {
                   this.renderDay(homeTimezone, time.day, now)}
                 <PatternRow
                   homeTimezone={homeTimezone}
+                  key={pattern.id}
```
binh-dam-ibigroup, line 222: Should the key be applied to the surrounding fragment instead?

lib/util/viewer.js
```diff
@@ -136,11 +136,8 @@ export function getStopTimesByPattern(stopData) {
           times: []
         }
       }
-      // Exclude the last stop, as the stop viewer doesn't show arrival times to a terminus stop.
-      const filteredTimes = times.filter(excludeLastStop)
```
binh-dam-ibigroup, line 140: This change results in the display of arrival times (in addition to departures) at terminal stops. Per our discussion,. this is not desired.

binh-dam-ibigroup (changes requested):
Not sure why displaying arrivals in addition to departures at terminal stops has been enabled.

lib/components/viewers/live-stop-times.tsx
miles-grant-ibigroup: I thought so too but this is what stopped the console warnings

lib/util/viewer.js
miles-grant-ibigroup: I can't replicate this. Do you have any examples?

miles-grant-ibigroup: Ok never mind I can see it on some bus routes where the last stop of one pattern is not identical to the first stop of another. Looking at the behavior I am not against that

miles-grant-ibigroup:
This will be addressed by the nearby view

PR #997, opened 2023-09-13
Un-de-duplicate headsigns in pattern viewer

miles-grant-ibigroup:
**Description:**

Headsign de-duplication previously removed headsigns from the pattern viewer dropdown that were duplicated. However, this caused deviations or short runs to sometimes not be selectable. The long term solution should obviously be to show all patterns, grouped by direction. For now, we append the last stop of the pattern so that they are differentiable. 

**PR Checklist:**


lib/components/viewers/styled.ts
```diff
@@ -52,7 +52,6 @@ export const PatternContainer = styled.div`
       align-items: center;
       display: flex;
       justify-content: space-between;
-      max-width: 300px;
```
amy-corson-ibigroup, line 55: Good catch this was making things weird!

lib/components/viewers/route-details.tsx
```diff
@@ -85,27 +86,29 @@ class RouteDetails extends Component<Props> {
...
           geometryLength: pat.patternGeometry?.length || 0,
           headsign: extractHeadsignFromPattern(pat, shortName),
-          id
+          id,
+          lastStop: pat.stops?.[pat.stops?.length - 1]?.name
         })
       )
-      // Remove duplicate headsigns. Using a reducer means that the first pattern
-      // with a specific headsign is the accepted one. TODO: is this good behavior?
+      // Remove duplicate headsigns. Replaces duplicate headsigns with the last stop name
```
amy-corson-ibigroup, line 93: Nitpicky but we're not technically removing anything anymore right, just appending the last stop

amy-corson-ibigroup (approved):
Looks nice!

AdrianaCeric (changes requested):
Ok so I am sort of understanding what's going on here, but how does appending the last stop help in the case that two patterns start at different stops and end at the same stop (e.g. South Lake Union Streetcar)?
image
image

miles-grant-ibigroup:
Yes @AdrianaCeric it doesn't cover this use case but it does cover other cases

AdrianaCeric (approved):
👍

PR #1005, opened 2023-09-20
Fix stop viewer next days departures

binh-dam-ibigroup:
## Description

This PR fixes the departure days in the Stop Viewer for trips that only have departures in the next days and not the current day. Specifically, trip departures are grouped by departure day (or service day) based on the data returned by OTP instead of an arbitrary rule.

## PR Checklist


amy-corson-ibigroup (approved):
Much cleaner and great fixes! Thank you!

lib/util/arrays.ts
```diff
@@ -0,0 +1,12 @@
+/** Retrieves a key from a dictionary or adds it with the desired value if absent. */
+export function getOrPutEntry<K extends string | number | symbol, V>(
+  entries: Record<K, V>,
+  key: K,
+  newValue: (newKey: K) => V
+): V {
+  let entry = entries[key]
+  if (!entry) {
+    entries[key] = entry = newValue(key)
```
daniel-heppner-ibigroup, line 9: I can't say I've seen this done before in Javascript. Had to look run an experiment to see what happens here. I wonder if it would be more clear if it were on two lines?

lib/components/util/types.ts
```diff
@@ -66,6 +65,17 @@ export interface Time {
   tripId: string
 }
 
+export interface PatternStopTimes {
```
daniel-heppner-ibigroup, line 68: This type is different from the one returned in the GraphQL API, which is an issue for the Nearby View. Let's get this merged in and then we can figure out what to do in the Nearby View branch before merging it.

lib/util/stop-times.ts
```diff
@@ -0,0 +1,82 @@
...
+}
+
+/** Filter any times according to time range set in config. */
+function isWithinDaysAhead(time: Time, now: Date, daysAhead: number) {
+  const departureTime = time.serviceDay + time.realtimeDeparture
+  return isBefore(departureTime, addDays(now, daysAhead))
+}
+
+/** Helper to sort and group stop times by pattern by service day */
+export function groupAndSortStopTimesByPatternByDay(
```
daniel-heppner-ibigroup, line 38: This logic is super clean, but since the shape of the graphql stoptimes response is different it might have to be adapted in the future. I wonder if we shouldn't build it for the new API and then transform the old API into the new shape for now?

daniel-heppner-ibigroup: This can come later

daniel-heppner-ibigroup (approved):
Clean code, nice work here

PR #1011, opened 2023-09-22
Remove calltaker index imports

binh-dam-ibigroup:
**Description:**

This PR unbundles calltaker from non-calltaker instances.

For calltaker configs, import calltaker components as follows (as needed):
```js

import CallTakerControls from '../lib/components/admin/call-taker-controls'
import CallHistoryWindow from '../lib/components/admin/call-history-window'
import CallTakerPanel from '../lib/components/app/call-taker-panel'
import FieldTripWindows from '../lib/components/admin/field-trip-windows'
import MailablesWindow from '../lib/components/admin/mailables-window'
```

**PR Checklist:**


miles-grant-ibigroup (changes requested):
I'm seeing an error when trying to run a call-taker config. Is there something I have to change in the config? 

Screenshot 2023-09-26 at 2 51 37 PM

binh-dam-ibigroup:
> I'm seeing an error when trying to run a call-taker config. Is there something I have to change in the config?
> 
Yes, instead of importing from `../lib`, import the component from the module that defines it.

miles-grant-ibigroup (approved):
Working great, thanks so much for the changes

daniel-heppner-ibigroup (approved):
Nice change! Do you think we could add some documentation to explain how to enable the call taker module? Perhaps we need some better docs on config.js in general

PR #1017, opened 2023-09-28
Add configuration types.

binh-dam-ibigroup:
## Description

This PR introduces types used in the app configuration (config.yml), and a partial redux state type is added.
The new types are not necessarily exhaustive and may not include obsolete config properties still in example-config.yml.
A few component files have been changed just to give an idea of the direction taken.

 Note that calltaker config types are still to be implemented.

## PR Checklist


lib/components/map/enhanced-stop-marker.tsx
```diff
@@ -166,12 +163,11 @@ class EnhancedStopMarker extends Component<Props> {
...
 }
 
-const mapStateToProps = (state: any, ownProps: OwnProps) => {
+const mapStateToProps = (state: AppReduxState, ownProps: OwnProps) => {
   const { highlightedStop } = state.otp.ui
 
-  const transitModes = state.otp.config.modes.transitModes
   const modeColors: ModeColors = {}
-  transitModes.forEach((mode: ConfiguredTransitMode) => {
+  state.otp.config.modes.transitModes.forEach((mode) => {
```
AdrianaCeric, line 170: Do you want to also add the
`const { config: otpConfig } = state.otp` here?

lib/components/util/types.ts
```diff
@@ -95,10 +95,3 @@ export type SetViewedRouteHandler = (route?: ViewedRouteState) => void
 export type SetViewedStopHandler = (payload: { stopId: string } | null) => void
 
 export type SetLocationHandler = (payload: MapLocationActionArg) => void
-
-export interface ConfiguredTransitMode {
```
AdrianaCeric, line 99: Why is this interface being removed?

AdrianaCeric (approved):
I think this looks good. Grouping the types in a consistent way and in one file is a good idea. I just had a few questions:

lib/components/app/app-menu.tsx
```diff
@@ -89,7 +90,7 @@ class AppMenu extends Component<
 
   _triggerPopup = () => {
     const { popupTarget, setPopupContent } = this.props
-    setPopupContent(popupTarget)
+    if (popupTarget) setPopupContent(popupTarget)
```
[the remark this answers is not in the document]
binh-dam-ibigroup: `popupTarget` is optional, so TypeScript forces this truthy check before triggering the popup.

lib/components/map/enhanced-stop-marker.tsx
binh-dam-ibigroup: Not needed, in this file, it is used only on this line.

lib/components/util/types.ts
binh-dam-ibigroup: I moved it to `config-types.ts`, and it is now called `TransitModeConfig`.

example-config.yml
```diff
@@ -349,8 +349,6 @@ itinerary:
   disableMetroSeperatorDot: false
   # Shows the duration of a leg below the leg in the metro itinerary summary
   showLegDurations: false
-  # Allows calorie counts to be hidden
-  displayCalories: true
```
miles-grant-ibigroup, line 353: Was this behavior removed?

lib/components/narrative/connected-trip-details.js
```diff
@@ -20,10 +20,6 @@ const mapStateToProps = (state) => {
       mediumId: null,
       riderCategoryId: null
     },
-    displayCalories:
-      typeof itinerary?.displayCalories === 'boolean'
```
miles-grant-ibigroup, line 24: Why was this removed?

miles-grant-ibigroup (approved):
This is some great cleanup, thanks for all your work!

example-config.yml
binh-dam-ibigroup: Displaying calories was completely removed from OTP-UI, so this doesn't apply anymore.

lib/components/narrative/connected-trip-details.js
binh-dam-ibigroup: Displaying calories was completely removed from OTP-UI, so this doesn't apply anymore.

PR #1018, opened 2023-09-28
Nearby view

daniel-heppner-ibigroup:
**Description:**
This PR adds a new Nearby View component which replaces the stop viewer. 

**PR Checklist:**

 Before | After |
    |--------|-------|
    | image |   image|


miles-grant-ibigroup (changes requested):
A few initial observations:
- view routes crashes everything (`lat` fails)
- nearby view fails un-gracefully (no error message, just small blue)
- there should be a view switcher option for nearby view that launches it at the current center of the map (I think you can pull this from `useMap`, let me know if you'd like to pair program this)
- existing icon functions should be used where possible (again let me know if you'd like to pair program this)
- existing stop route sorting/day de-duplication needs to be imported (I'd be less happy to pair program this but would also be available)
- shadow on hover needs to be more prominent. can we have the current one be the default and then make it darker and more spread on hover?

That's all for now, I think the biggest push is getting it working at all. There's also some merge conflicts here :(

miles-grant-ibigroup (changes requested):
future PR: more map support

miles-grant-ibigroup (changes requested):
I'm seeing some cases where the old stop viewer shows a stop id but the new one does not. Any idea what's going on there?

miles-grant-ibigroup (changes requested):
Still seeing stop id issue, still seeing some stops not loading (although this is a core-utils version issue I think, please update package.json and yarn lock/generate new alpha versions)

overall this is definitely getting there, I think we're almost ready for an initial merge!

daniel-heppner-ibigroup:
Please use https://github.com/opentripplanner/otp-ui/pull/655 in your testing for this. The alpha package versions are broken and I do not know why.

daniel-heppner-ibigroup:
I accidentally merged this GraphQL Mocking PR into this one somehow. Don't know how that happened, I have no memory of doing it! Maybe I accidentally dragged it over in Git Tower or something. Anyway, take a look here to see the GraphQL Mocking changes. https://github.com/opentripplanner/otp-react-redux/pull/1063

lib/components/map/default-map.tsx
```diff
@@ -308,7 +308,6 @@ class DefaultMap extends Component {
           <EndpointsOverlay />
           <RouteViewerOverlay />
           <TransitVehicleOverlay ModeIcon={ModeIcon} />
-          <StopViewerOverlay />
```
miles-grant-ibigroup, line 311: Can we delete it?

lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -0,0 +1,132 @@
...
+    case 'RentalVehicle':
+      return <Vehicle key={place.id} vehicle={place} />
+    case 'Stop':
+      return <Stop showOperatorLogo stopData={place} />
+    case 'VehicleParking':
+      return <VehicleParking place={place} />
+    case 'BikeRentalStation':
+      return <RentalStation place={place} />
+    default:
+      return `${place.__typename}you are from the future and have a cool new version of OTP2 let me know how it is mlsgrnt@icloud.com`
```
miles-grant-ibigroup, line 45: We should probably modify this

lib/components/viewers/nearby/rental-station.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const { bikesAvailable, spacesAvailable } = place
+  return (
+    <Card onMouseEnter={() => zoomToPlace(map, place)}>
+      <CardHeader>
+        {place.name}
+        <CardSubheader>
+          {company || (
+            <FormattedMessage id="components.NearbyView.bikeRentalStation" />
+          )}
+        </CardSubheader>
```
miles-grant-ibigroup, line 34: shouldn't the company come first?

lib/components/viewers/pattern-row.tsx
```diff
@@ -65,18 +68,19 @@ const PatternRow = ({
                 'gtfs'
               )}
               leg={generateFakeLegForRouteRenderer(route, true)}
-              style={{ fontSize: routeNameFontSize(routeName) }}
+              style={{ fontSize: routeNameFontSize(routeName), maxHeight: 90 }}
```
miles-grant-ibigroup, line 71: This will cause a lot of problems. What is this for?

lib/components/viewers/nearby/nearby-view.tsx
daniel-heppner-ibigroup: evidence of your code laundering

lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -0,0 +1,132 @@
...
+  const firstItemRef = useRef<HTMLDivElement>(null)
+  useEffect(() => {
+    firstItemRef.current?.scrollIntoView({ behavior: 'smooth' })
+    if (nearbyViewCoords) {
+      fetchNearby(nearbyViewCoords, map)
+      setLoading(true)
+      const interval = setInterval(() => {
+        fetchNearby(nearbyViewCoords, map)
+        setLoading(true)
+      }, 15000)
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: I can add it as a const for now but I don't really want to add a config for it, I don't see anyone caring enough to want to change the value

lib/components/viewers/nearby/rental-station.tsx
daniel-heppner-ibigroup: I don't know what you mean? I think this is right

lib/components/viewers/pattern-row.tsx
daniel-heppner-ibigroup: image

daniel-heppner-ibigroup: looks like it wasn't doing anything? I removed it

daniel-heppner-ibigroup:
I wasn't able to replicate the bugs you mentioned in your response, except for the bug with clicking on view stop from an itinerary. This was because the fields were missing from the stop type in the response. I added them to the GraphQL query but now we have another alpha version. :(

daniel-heppner-ibigroup:
What is going on with Typescript? This error doesn't make sense to me and it isn't happening locally

miles-grant-ibigroup:
> I wasn't able to replicate the bugs you mentioned in your response, except for the bug with clicking on view stop from an itinerary. This was because the fields were missing from the stop type in the response. I added them to the GraphQL query but now we have another alpha version. :(

I've DMed you instructions. these are show-stopping bugs.

lib/actions/apiV2.js
```diff
@@ -446,6 +448,140 @@ const findNearbyAmenities = ({ lat, lon, radius = 300, stopId }) => {
   )
 }
 
+const mergeSameStops = (nearbyResponse) => {
+  console.log(nearbyResponse)
```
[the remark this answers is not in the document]
miles-grant-ibigroup: It's still here...

miles-grant-ibigroup (changes requested):
A few final issues for us to work on:

- Schedule view doesn’t pull entire schedule (we’ve had this before)
- Highlight view doesn’t show scooters or bikes or bike stations
- A lot of duplicated code in the various cards (can be tabled for another time)
 - Layers are wrong. We should show the bikes and scooters on load
 - stop viewer tests need to be modified

miles-grant-ibigroup (approved):
Alright I think we're about ready to go. There are still a few rough edges

- scooter/bikeshare view on the map
- various cleanup
- a11y
- stop viewer tests
- french

but I think we are good enough to go for now

daniel-heppner-ibigroup:
I'd love some help on accessibility improvements and French translations.

I'll add scooter/bikeshare layer auto-enable to the follow up ticket

lib/components/viewers/nearby/vehicle-rent.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const StationIcon = () => {
+    const CompanyIcon = getCompanyIcon(vehicle.network)
+    return CompanyIcon ? (
+      <Suspense fallback={<span>{company}</span>}>
+        <CompanyIcon height={22} style={{ marginRight: '5px' }} width={22} />
+      </Suspense>
+    ) : (
+      <span>{getVehicleIcon(formFactor)}</span>
+    )
+  }
```
binh-dam-ibigroup, line 110: Surround with `useMemo`.

lib/components/viewers/nearby/vehicle-rent.tsx
```diff
@@ -0,0 +1,146 @@
...
+  return (
+    <Card onMouseEnter={() => zoomToPlace(map, vehicle)}>
```
binh-dam-ibigroup, line 112: Maybe surround the 3 event handlers with `useCallback`

lib/components/viewers/nearby/vehicle-rent.tsx
```diff
@@ -0,0 +1,146 @@
...
+    </Card>
+  )
+}
+
+const mapDispatchToProps = {
+  setLocation: mapActions.setLocation,
+  zoomToPlace: mapActions.zoomToPlace
+}
+
+const mapStateToProps = (state: any) => {
```
binh-dam-ibigroup, line 139: Use type `AppReduxState`.

lib/components/viewers/route-details.tsx
```diff
@@ -66,12 +66,14 @@ class RouteDetails extends Component<Props> {
     setViewedRoute({ patternId: id, routeId: route.id })
   }
 
+  // FIXME: this should be the Stop type but
+  // we can't do that because Stop is a duplicate identifier (imported from styled)
```
binh-dam-ibigroup, line 70: Can you rename the imported type instead?

lib/components/viewers/stop-viewer.js
```diff
@@ -368,7 +367,17 @@ class StopViewer extends Component {
...
+            className="link-button pull-right"
+            onClick={this._viewNearby}
+            style={{ fontSize: 'small' }}
+          >
+            {/* FIXME: What icon should we use? */}
+            <IconWithText Icon={Calendar}>
+              <FormattedMessage id="components.StopViewer.viewNearby" />
+            </IconWithText>
+          </button>
+          {/* {!isFlex && (
```
binh-dam-ibigroup, line 380: Do we still need the block below?

lib/components/viewers/stop-viewer.js
```diff
@@ -368,7 +367,17 @@ class StopViewer extends Component {
           >
             <Icon Icon={Search} style={{ marginLeft: '0.2em' }} />
           </button>
-          {!isFlex && (
+          <button
+            className="link-button pull-right"
+            onClick={this._viewNearby}
+            style={{ fontSize: 'small' }}
+          >
+            {/* FIXME: What icon should we use? */}
```
binh-dam-ibigroup, line 375: Any thoughts for https://fontawesome.com/icons/magnifying-glass-location?f=classic&s=solid?

lib/components/viewers/view-stop-button.tsx
```diff
@@ -45,7 +45,7 @@ class ViewStopButton extends Component {
   }
 }
 
-const mapStateToProps = (state, ownProps) => {
+const mapStateToProps = () => {
```
binh-dam-ibigroup, line 48: Delete variable and pass `null` as first arg to the `connect` function.

lib/components/viewers/viewers.css
```diff
@@ -323,6 +323,13 @@
   color: #000080;
 }
 /* end of child-stop-details */
+.stop-view .route-name {
+  align-items: baseline;
+  display: grid;
+  align-items: center;
```
binh-dam-ibigroup, line 329: Sort CSS and remove duplicate attribute.

lib/components/viewers/viewers.css
```diff
@@ -562,4 +569,4 @@
   align-items: center;
   justify-content: center;
   padding-top: 10px;
-}
\ No newline at end of file
+}
```
binh-dam-ibigroup, line 572: Revert this change.

package.json
```diff
@@ -39,26 +40,26 @@
...
     "@opentripplanner/base-map": "^3.0.15",
-    "@opentripplanner/core-utils": "^11.1.3",
+    "@opentripplanner/core-utils": "^12.0.0-alpha.6",
     "@opentripplanner/endpoints-overlay": "^2.0.11",
     "@opentripplanner/from-to-location-picker": "^2.1.11",
     "@opentripplanner/geocoder": "^1.4.2",
     "@opentripplanner/humanize-distance": "^1.2.0",
     "@opentripplanner/icons": "^2.0.6",
-    "@opentripplanner/itinerary-body": "^5.1.3",
+    "@opentripplanner/itinerary-body": "^5.2.0",
```
binh-dam-ibigroup, line 49: Update itinerary-body to 5.2.1.

package.json
```diff
@@ -39,26 +40,26 @@
     "@bugsnag/plugin-react": "^7.17.0",
     "@floating-ui/react": "^0.19.2",
     "@opentripplanner/base-map": "^3.0.15",
-    "@opentripplanner/core-utils": "^11.1.3",
+    "@opentripplanner/core-utils": "^12.0.0-alpha.6",
```
binh-dam-ibigroup, line 43: Is there a non-alpha version for this?

package.json
```diff
@@ -243,4 +245,4 @@
       "percy-css": ".percy-hide { opacity: 0!important; } "
     }
   }
-}
+}
```
binh-dam-ibigroup, line 248: Revert this change.

lib/components/map/enhanced-stop-marker.tsx
```diff
@@ -140,7 +141,8 @@ class EnhancedStopMarker extends Component<Props> {
           activeStopId !== stop.id && (
             <BaseMapStyled.MapOverlayPopup id={activeContentId}>
               <StopPopup
-                entity={stop}
+                // FIXME: We need to fix the stop type
+                entity={{ ...stop, lat: stop.lat || 0, lon: stop.lon || 0 }}
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: I added that to make Typescript happy, you're right that it isn't necessary. But I just removed it and Typescript seems okay, so maybe whatever was causing that was fixed elsewhere.

lib/components/map/nearby-view-dot-overlay.tsx
```diff
@@ -0,0 +1,43 @@
+import { connect } from 'react-redux'
+import { Location } from '@opentripplanner/types'
+import { Marker } from 'react-map-gl'
+import React from 'react'
+import styled from 'styled-components'
+
+import { AppReduxState } from '../../util/state-types'
+
+type Props = {
+  location: Location | null
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: Null is okay, since the prop should always be defined, either as null or some value. It's initialized to null in the reducer.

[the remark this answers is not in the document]
daniel-heppner-ibigroup: Null is okay, since the prop should always be defined, either as null or some value. It's initialized to null in the reducer.

lib/components/util/types.ts
```diff
@@ -41,6 +48,15 @@ export interface Pattern {
     length: number
     points: string
   }
+  route: {
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: hopefully!

lib/components/viewers/nearby/stop.tsx
```diff
@@ -0,0 +1,237 @@
...
+          {
+            operatorName: operator.name
+          }
+        )}
+        operator={operator}
+      />
+    ) : (
+      <MapPin />
+    )
+  }
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: Cleaned this up by adding a new prop to the styled icon components. Take a look at that file too.

lib/components/viewers/stop-viewer.js
daniel-heppner-ibigroup: I feel weird removing it without removing all the code that made the stop viewer a stop viewer, although we are now just using this for schedule viewing until we can make a new stop schedule viewer. Should I rip out all the code for the old stop viewer now that there's no way to access it?

daniel-heppner-ibigroup: It will be deleted at some point when this file gets deleted entirely.

lib/components/viewers/nearby/vehicle-rent.tsx
daniel-heppner-ibigroup: Moved to its own component in the file

lib/components/viewers/nearby/vehicle-rent.tsx
```diff
@@ -0,0 +1,146 @@
...
+}): JSX.Element => {
+  const map = useMap().default
+  const intl = useIntl()
+  const company = companies.find((c) => c.id === vehicle.network)?.label ?? ''
+  const { formFactor } = vehicle.vehicleType
+  const name =
+    vehicle.name === 'Default vehicle type'
+      ? getVehicleText(formFactor, company, intl)
+      : vehicle.name
+  const setLocationFromPlace = (locationType: 'from' | 'to') => {
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: This function will be cached where it is used in the callbacks, so it doesn't need to be cached here as well.

lib/components/map/enhanced-stop-marker.tsx
```diff
@@ -140,7 +141,8 @@ class EnhancedStopMarker extends Component<Props> {
           activeStopId !== stop.id && (
             <BaseMapStyled.MapOverlayPopup id={activeContentId}>
               <StopPopup
-                entity={stop}
+                // FIXME: We need to fix the stop type
+                entity={{ ...stop, lat: stop.lat, lon: stop.lon }}
                 setLocation={setLocation}
                 setViewedStop={setViewedStop}
```
binh-dam-ibigroup, line 147: By triggering the nearby view from the map or an itinerary body, the nearby view should either hide the other stuff and offer a button to view the rest if the user desires, or visually distinguish the first card so it draws more attention than the other ones with respect to the selected stop. Alternately, we can consider renaming the "Stop Viewer" prompts to something else such as "Explore here".

lib/components/viewers/nearby/stop.tsx
```diff
@@ -0,0 +1,240 @@
...
+    zoomToStop()
+    setHoveredStop(stopData.gtfsId)
+  }
+
+  const onMouseLeave = () => {
+    setHoveredStop(undefined)
+  }
+
+  const operator = agencies.size
+    ? transitOperators?.find((o) => o.agencyId === Array.from(agencies)[0])
```
binh-dam-ibigroup, line 179: In the case of shared stops, does it make any sense to show the first operator found in the list? An alternative is to show all operator logos, but maybe do it after the text about the stop ID.

daniel-heppner-ibigroup:
Looks like the Percy tests broke again...

lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -0,0 +1,183 @@
...
+  const onMouseLeave = useCallback(() => {
+    setHighlightedLocation(null)
+  }, [setHighlightedLocation])
+
+  const nearbyItemList = nearby?.map((n: any) => (
+    <div
+      key={n.place.id}
+      onMouseEnter={() => onMouseEnter(n.place)}
+      onMouseLeave={onMouseLeave}
+      role="menuitem"
```
binh-dam-ibigroup, line 114: These are not menu items (i.e. they are not menu items that come from an OS menu bar). It is better to make each entry an` <li>` within an ordered list `<ol>` (given that the view shows closes stops first).

lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -0,0 +1,183 @@
...
+      tabIndex={0}
+    >
+      {getNearbyItem(n.place)}
+    </div>
```
binh-dam-ibigroup, line 118: How about rendering the common footer with the from/to picker here?

lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -0,0 +1,183 @@
...
+            id: 'components.NearbyView.header'
+          })}
+          onBackClicked={goBack}
+        />
+      )}
+      <NearbySidebarContainer
+        className="base-color-bg"
+        style={{ marginTop: mobile ? '50px' : 0 }}
+      >
+        <div ref={firstItemRef} />
```
binh-dam-ibigroup, line 145: We need to introduce the list for screen reader users.

lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -0,0 +1,183 @@
...
+    [setHighlightedLocation]
+  )
+  const onMouseLeave = useCallback(() => {
+    setHighlightedLocation(null)
+  }, [setHighlightedLocation])
+
+  const nearbyItemList = nearby?.map((n: any) => (
+    <div
+      key={n.place.id}
+      onMouseEnter={() => onMouseEnter(n.place)}
```
binh-dam-ibigroup, line 112: Add a similar behavior for `onFocus` and `onBlur` (otherwise, it is impossible to highlight the stop using the keyboard alone).

binh-dam-ibigroup (changes requested):
Second round of changes for now. See the a11y issues and links on mobile view that don't work, among other things.

lib/components/viewers/nearby/styled.tsx
```diff
@@ -0,0 +1,77 @@
...
+  margin: 5px;
+  padding: 3px;
+  position: fixed;
+`
+
+export const NearbySidebarContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  gap: 3rem;
+  overflow-y: scroll;
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: Ah ha! That's where that's from

lib/components/viewers/nearby/stop.tsx
```diff
@@ -0,0 +1,240 @@
...
+      </IconWithText>
+    </StyledAlert>
+  )
+}
+
+type Props = {
+  homeTimezone: string
+  setHoveredStop: (stopId: string | undefined) => void
+  setLocation: (args: any) => void
+  setViewedStop: (stop: any, nearby: string) => void
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: It's actually not... `StopEventHandler` is missing the optional string as a second argument, which is only used here.

lib/components/map/enhanced-stop-marker.tsx
daniel-heppner-ibigroup: I'm interested in the visual distinction approach. It probably wouldn't be too hard to add an entity ID to the URL query which we can use to highlight something.

lib/components/viewers/nearby/stop.tsx
daniel-heppner-ibigroup: I think the intent was to check the size equals 1 rather than is greater than zero. That way for shared stops it'll display the generic stop icon.

lib/components/viewers/nearby/nearby-view.tsx
daniel-heppner-ibigroup: What does this mean?

binh-dam-ibigroup: It means adding a visually-hidden ~paragraph~ heading that says something like: "10 stops found near xxxxx" where xxxxx is the feature or other center of the nearby query.

lib/components/viewers/nearby/stop.tsx
binh-dam-ibigroup: As written, this (or something else?) is showing the logo of the first operator for "shared" stops, i.e. stops with the same `stop_code`. The first operator changes depending on the order of stops from the nearby query, so the result is not consistent visually. Try for instance around Peachtree+10th in Atlanta.

lib/components/viewers/nearby/nearby-view.tsx
daniel-heppner-ibigroup: Does 7b89edf work or does it need to be outside the ol?

lib/components/viewers/nearby/stop.tsx
daniel-heppner-ibigroup: Oh yes, I was saying that the behavior you are seeing was a bug. The condition here was supposed to be `agencies.size === 1` but it wasn't. I changed it to that so that the behavior is as intended. Let me know if that works. I think this is fine vs listing out all the agency's icons, but let me know!

lib/components/viewers/nearby/styled.tsx
```diff
@@ -0,0 +1,77 @@
...
+  left: 0px;
+  margin: 5px;
+  padding: 3px;
+  position: fixed;
+`
+
+export const NearbySidebarContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  gap: 3rem;
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: hm... looks proportionate to me. Are you wanting less padding on the x direction?

lib/components/viewers/nearby/nearby-view.tsx
binh-dam-ibigroup: Outside please.

lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -0,0 +1,177 @@
...
+      const interval = setInterval(() => {
+        fetchNearby(nearbyViewCoords, map)
+        setLoading(true)
+      }, AUTO_REFRESH_INTERVAL)
+      return function cleanup() {
+        clearInterval(interval)
+      }
+    } else {
+      if (getPosition && viewNearby) {
+        getPosition(intl, null, (pos) => {
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: I made some improvements here but I think it's not perfect yet. The location now falls back to the map center, but I think it needs a few more things to be good:
- Should show a marker at the current position. What symbol should we use?
- When you move the map, it should offer to update the location. 
- When you leave Nearby and come back to it, it currently shows the stale data since it's saved in Redux. The redux data should be cleared or otherwise hidden when you reopen the page.

lib/components/map/default-map.tsx
```diff
@@ -304,12 +304,12 @@ class DefaultMap extends Component {
           zoom={zoom}
         >
           <PointPopup />
+          <NearbyViewDotOverlay />
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Maplibre has this feature to show the user's current position that could be useful: https://maplibre.org/maplibre-gl-js/docs/examples/locate-user/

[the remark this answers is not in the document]
daniel-heppner-ibigroup: So yes, I was looking at this, but you can't manually trigger the blue dot. You have to add the control the map, then when the user clicks on it it will show the blue dot. 

Maybe we can do this:
- If a user opens Nearby without coordinates (eg, by clicking on the menu bar icon), we can use the map center as the location.
- While nearby view is open, we show the geolocate control on the map. Then `onGeolocate` we navigate to nearby view with the coordinates specified. 

The downsides:  
- it would be not be very obvious that clicking on the map's geolocate control will change the nearby view to center on that location. 
- when opening the nearby view without coordinates it isn't always clear where the map center is. I guess we could add a map layer that reads the current nearby view coordinates and puts a dot there? How will people know the dot corresponds with nearby view centerpoint?

lib/components/map/enhanced-stop-marker.tsx
daniel-heppner-ibigroup: I added a ring highlighted the current entity ID.

lib/components/map/default-map.tsx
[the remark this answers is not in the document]
binh-dam-ibigroup: >     * If a user opens Nearby without coordinates (eg, by clicking on the menu bar icon), we can use the map center as the location.
 I think this is necessary.

>     * While nearby view is open, we show the geolocate control on the map. Then `onGeolocate` we navigate to nearby view with the coordinates specified.

Probably for another PR. I think the geolocate control should be shown all the time so that users can find themselves regardless of view.

> The downsides:
> 
>     * it would be not be very obvious that clicking on the map's geolocate control will change the nearby view to center on that location.

Showing the geolocate control all the time would address this downside imo.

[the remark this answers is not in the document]
daniel-heppner-ibigroup: Then I think we're good here. Part one is done, Nearby View uses the current map coords when it doesn't have anything else to use.

package.json
```diff
@@ -39,26 +40,26 @@
     "@bugsnag/plugin-react": "^7.17.0",
     "@floating-ui/react": "^0.19.2",
     "@opentripplanner/base-map": "^3.0.15",
-    "@opentripplanner/core-utils": "^11.1.3",
+    "@opentripplanner/core-utils": "^12.0.0-alpha.6",
     "@opentripplanner/endpoints-overlay": "^2.0.11",
     "@opentripplanner/from-to-location-picker": "^2.1.11",
     "@opentripplanner/geocoder": "^1.4.2",
     "@opentripplanner/humanize-distance": "^1.2.0",
     "@opentripplanner/icons": "^2.0.6",
```
binh-dam-ibigroup, line 48: Update the icons package to 2.0.8 (2.0.7 at least for company icon matching)

lib/components/map/enhanced-stop-marker.tsx
binh-dam-ibigroup: Oh I see, that is for when clicking "Stop Viewer" from a stop popup on the map. Can you do the same when clicking the "Stop Viewer" from an itinerary body?

i18n/tr.yml
binh-dam-ibigroup, line 1: The "delete" entry involved in a merge conflict has been removed from `dev`, so go ahead and remove it from your version.

lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -0,0 +1,266 @@
...
+  const location = {
+    lat: stopData.lat ?? 0,
+    lon: stopData.lon ?? 0,
+    name: stopData.name
+  }
+  return (
+    <span role="group">
+      <FromToLocationPicker
+        label
+        onFromClick={() => {
```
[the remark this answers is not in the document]
binh-dam-ibigroup: In mobile view, clicking one of these has no visual effect. The "Plan Trip" view should be triggered when clicking on these, as it does on desktop view.

binh-dam-ibigroup (changes requested):
Code-wise it looks much better. There are some quirks to resolve around clicking the from/to on mobile view, the use of `role="button"` for each card, and asking for the user's geolocation when refreshing from path `/#/nearby` on desktop view.

lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -0,0 +1,266 @@
...
+        (n.place.gtfsId ?? n.place.id) === entityId ? 'highlighted' : ''
+      }
+      key={n.place.id}
+    >
+      <div
+        onBlur={onMouseLeave}
+        onFocus={() => onMouseEnter(n.place)}
+        onMouseEnter={() => onMouseEnter(n.place)}
+        onMouseLeave={onMouseLeave}
+        role="button"
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: ESLint wanted me to add something to it since it has interaction handlers. I looked through the ARIA roles but wasn't sure what to give it

lib/components/viewers/nearby/stop.tsx
```diff
@@ -0,0 +1,212 @@
...
+          >
+            <IconWithText Icon={Calendar}>
+              <FormattedMessage id="components.StopViewer.viewSchedule" />
+            </IconWithText>
+          </button>
+        </div>
+        {fromToSlot}
+      </CardBody>
+      <div>
+        <div>{timezoneWarning}</div>
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: I'd like to show the card still. I could see myself being confused if I wanted to check to see if I missed the last bus and the card just wasn't there. No departures is good

lib/components/viewers/nearby/nearby-view.tsx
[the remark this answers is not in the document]
binh-dam-ibigroup: Let's use an eslint exception this time.

i18n/tr.yml
daniel-heppner-ibigroup: Should I just check out the `dev` version of this file into this branch?

lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -0,0 +1,266 @@
...
+        tabIndex={0}
+      >
+        {getNearbyItem(n.place, setLocation)}
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: Maybe it should turn on the map layer when you hover over it?

i18n/tr.yml
binh-dam-ibigroup: Sure

lib/components/viewers/nearby/nearby-view.tsx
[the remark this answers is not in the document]
daniel-heppner-ibigroup: I think I fixed this. There's some really weird behavior with routing on mobile, but I added a call to `setMainPanelContent` to 0 to force it back onto the trip planner page.

lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -0,0 +1,266 @@
...
+        setLoading(true)
+        const interval = setInterval(() => {
+          fetchNearby(mapCoords)
+          setLoading(true)
+        }, AUTO_REFRESH_INTERVAL)
+        return function cleanup() {
+          clearInterval(interval)
+        }
+      }
+      getPosition(intl, null, (pos) => {
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: OK, I removed this code and instead added an event handler that uses the map's geolocate button (which I also added). When this button is clicked, we load that location on nearby view. I struggled to find good UX here, so let me know what you think.

lib/components/map/enhanced-stop-marker.tsx
daniel-heppner-ibigroup: Got it fixed. The ID being passed was wrong.

i18n/tr.yml
binh-dam-ibigroup: I reviewed @AdrianaCeric's PR #1097 that moves the delete button for monitored trips.

lib/actions/ui.js
```diff
@@ -168,7 +168,14 @@ export function setViewedNearbyCoords(payload) {
       'nearby',
       [payload.lat, payload.lon].join(',')
     )
-    dispatch(routeTo(path, payload.id ? `?entityId=${payload.id}` : undefined))
+    dispatch(
+      routeTo(
+        path,
+        payload.gtfsId || payload.id
```
binh-dam-ibigroup, line 174: Not holding this against this PR, but the search is getting cleared as a result of passing this second param. Although using the browser back navigation command does restore the search params, the UI is not picking those params. I'll file this as a separate issue.

lib/components/map/default-map.tsx
```diff
@@ -310,6 +310,7 @@ class DefaultMap extends Component {
           <EndpointsOverlay />
           <RouteViewerOverlay />
           <TransitVehicleOverlay ModeIcon={ModeIcon} />
+          <GeolocateControl position="bottom-left" />
```
binh-dam-ibigroup, line 313: How about top-left? Currently, in the desktop view, the geolocate button position conflicts with the attribution (the attribution text appears on top, so the geolocate button cannot be activated through the mouse. In the mobile view, once the results are shown, the geolocate button is hidden by the Expand/Collapse Map button.

binh-dam-ibigroup: I think the geolocate button is very useful.

lib/components/viewers/nearby/styled.tsx
```diff
@@ -0,0 +1,85 @@
...
+  margin: 5px;
+  padding: 3px;
+  position: fixed;
+`
+
+export const NearbySidebarContainer = styled.ol`
+  display: flex;
+  flex-direction: column;
+  gap: 3rem;
+  padding: 0 3rem;
```
binh-dam-ibigroup, line 18: Use 1em for horizontal padding to be consistent with the search results.

binh-dam-ibigroup (changes requested):
We are getting very close! I'd move the geolocate button to top-left for now to avoid positioning conflicts, and use 1em horizontal padding for the nearby view cards.

binh-dam-ibigroup:
Also @daniel-heppner-ibigroup please resolve conflicts.

lib/components/viewers/nearby/styled.tsx
daniel-heppner-ibigroup: Oops, I did this but forgot to push it.

binh-dam-ibigroup (approved):
Nice work with all of these changes!

miles-grant-ibigroup:
why do some of the stops have a big yellow border?

binh-dam-ibigroup:
@daniel-heppner-ibigroup I updated the translations where i could, except for 2 strings in Tagalog that were not provided.

daniel-heppner-ibigroup:
@miles-grant-ibigroup I think you broke the tests

miles-grant-ibigroup:
Snapshots are updated. @daniel-heppner-ibigroup Not sure why the a11y tests are failing. It works fine locally. Should we just disable the /nearby route for the a11y tests?

PR #1021, opened 2023-10-03
Fix schedule view input

binh-dam-ibigroup:
## Description

This PR ensures that web requests to get stop times are made only when the user enters a valid date
within the service range available in OTP. Stop times requests when the date is invalid or out of range are no longer made.
Bonus UI: the browser will show the valid dates and sometimes prevent users from entering a date outside the range.

**PR Checklist:**


miles-grant-ibigroup (approved):
Not sure how crazy I am about using the invisible component instead of baking this into a relevant redux action/graphql query, but it works well and code is clean! Thanks for this fix

lib/components/viewers/stop-viewer.js
```diff
@@ -292,11 +306,21 @@ class StopViewer extends Component {
...
+      // In schedule view, the time zone code should be that of the entered date,
+      // or the current day in the live view.
+      // This is to account for daylight time changes, especially when the live and
+      // schedule views are in different daylight saving periods.
+      const timezoneCode = formatTz(
+        isShowingSchedule && date
+          ? parse(date, inputDateFormat, new Date())
+          : new Date(), // TODO: mock for percy tests,
+        'z',
+        {
```
amy-corson-ibigroup, line 318: This is causing the timezone code to change & update in the alert if the year starts with 0, which is a little jarring as you're typing.
image

amy-corson-ibigroup (changes requested):
I really like this change. I think it's a little confusing from a user perspective. See comment about the timezone alert bug. 

If a user enters an invalid date, they should be made aware the request was invalid and the schedule did not update. Or the schedule should disappear, so users do not mistakenly assume the schedule represents the invalid date. 
image

lib/components/viewers/stop-viewer.js
binh-dam-ibigroup: Good catch! Fixed in https://github.com/opentripplanner/otp-react-redux/pull/1021/commits/4d1e6153d292e5f4458d7e873458fe3afd3442f2.

binh-dam-ibigroup:
> If a user enters an invalid date, they should be made aware the request was invalid and the schedule did not update. Or the schedule should disappear, so users do not mistakenly assume...

Good catch, fixed in https://github.com/opentripplanner/otp-react-redux/pull/1021/commits/cc7e5038fc25fb883d8d4a657e695e3bf2bd653e.

amy-corson-ibigroup (approved):
Working really well! Thank Binh, this looks great!

PR #1022, opened 2023-10-04
fix(AppMenuItem): Make app menu external links open in new browser tab.

binh-dam-ibigroup:
## Description

This PR makes external links in the app menu open in a new browser tab.
(Relative links would still open in the same window.)

**PR Checklist:**

### DIFF
--- a/lib/components/app/app-menu-item.tsx
+++ b/lib/components/app/app-menu-item.tsx
@@ -68,7 +68,9 @@ export default class AppMenuItem extends Component<Props, State> {
   render(): JSX.Element {
     const { icon, id, onClick, subItems, text, ...otherProps } = this.props
     const { isExpanded } = this.state
-    const Element = otherProps.href ? 'a' : 'button'
+    const hasHref = !!otherProps.href
+    const isAbsolute = otherProps.href?.startsWith('http')
+    const Element = hasHref ? 'a' : 'button'
     const containerId = `${id}-container`
     return (
       <>
@@ -80,6 +82,7 @@ export default class AppMenuItem extends Component<Props, State> {
           onClick={subItems ? this._toggleSubmenu : onClick}
           onKeyDown={this._handleKeyDown}
           {...otherProps}
+          target={hasHref && isAbsolute ? '_blank' : undefined}
         >
           <span aria-hidden>{icon}</span>
           <span>{text}</span>


lib/components/app/app-menu-item.tsx
```diff
@@ -68,7 +68,9 @@ export default class AppMenuItem extends Component<Props, State> {
   render(): JSX.Element {
     const { icon, id, onClick, subItems, text, ...otherProps } = this.props
     const { isExpanded } = this.state
-    const Element = otherProps.href ? 'a' : 'button'
+    const hasHref = !!otherProps.href
+    const isAbsolute = otherProps.href?.startsWith('http')
```
amy-corson-ibigroup, line 72: If a url gets passed without an http should it still open a new browser tab?

amy-corson-ibigroup (approved):
Thanks!

lib/components/app/app-menu-item.tsx
binh-dam-ibigroup: Then it would be a relative URL (like the route viewer or another resource), but that's a good question.

binh-dam-ibigroup: Right now, relative URLs open in the same window per PR description.

AdrianaCeric, line 72: Looks good! One suggestion, but not needed:

```suggestion
    const isAbsolute = hasHref && otherProps.href?.startsWith('http')
```

lib/components/app/app-menu-item.tsx
```diff
@@ -80,6 +82,7 @@ export default class AppMenuItem extends Component<Props, State> {
           onClick={subItems ? this._toggleSubmenu : onClick}
           onKeyDown={this._handleKeyDown}
           {...otherProps}
+          target={hasHref && isAbsolute ? '_blank' : undefined}
```
AdrianaCeric, line 85:
```suggestion
          target={isAbsolute ? '_blank' : undefined}
```

AdrianaCeric (approved):

lib/components/app/app-menu-item.tsx
binh-dam-ibigroup: It is not needed. That condition would be redundant, as `otherProps.href?.startsWith(...)` is null if `otherProps.href` is null, and can't be non-null unless `otherProps.href`is non null. Same for your other suggestion.

PR #1024, opened 2023-10-06
Add configurable strict mode to itinerary filter

amy-corson-ibigroup:
**Description:**

Adds support for a config item that will filter itineraries based on exactly what modes the user has selected (eliminates walk itineraries being shown by default if user did not specifically select a walk mode)

**PR Checklist:**

@daniel-heppner-ibigroup also wrote this and I forgot to put him in the commit!!!

### DIFF
--- a/lib/actions/apiV2.js
+++ b/lib/actions/apiV2.js
@@ -45,7 +45,8 @@ import { RoutingQueryCallResult } from './api-constants'
 import { setItineraryView } from './ui'
 import { zoomToPlace } from './map'
 
-const { generateCombinations, generateOtp2Query } = coreUtils.queryGen
+const { generateCombinations, generateOtp2Query, SIMPLIFICATIONS } =
+  coreUtils.queryGen
 const { getTripOptionsFromQuery, getUrlParams } = coreUtils.query
 const { convertGraphQLResponseToLegacy } = coreUtils.itinerary
 const { randId } = coreUtils.storage
@@ -824,6 +825,8 @@ export function routingQuery(searchId = null, updateSearchInReducer) {
       config?.modes?.initialState?.enabledModeButtons ||
       {}
 
+    const strictModes = config?.itinerary?.strictItineraryFiltering
+
     // Filter mode definitions based on active mode keys
     const activeModeButtons = config.modes?.modeButtons.filter((mb) =>
       activeModeKeys.includes(mb.key)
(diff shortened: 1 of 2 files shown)


lib/actions/apiV2.js
```diff
@@ -920,11 +923,28 @@ export function routingQuery(searchId = null, updateSearchInReducer) {
...
+
+              if (strictModes) {
+                filteredItineraries = itineraries.filter((itin) =>
+                  itin.legs.some((leg) =>
+                    activeModeStrings.includes(SIMPLIFICATIONS[leg.mode])
+                  )
+                )
+              } else {
+                filteredItineraries = itineraries
+              }
```
miles-grant-ibigroup, line 940: Wouldn't it be easier to just always run the filter and return true if `strictModes`?

lib/actions/apiV2.js
```diff
@@ -920,11 +923,28 @@ export function routingQuery(searchId = null, updateSearchInReducer) {
...
+
+              const withCollapsedShortNames = filteredItineraries.map(
+                (itin) => ({
```
miles-grant-ibigroup, line 943: What are we doing with this new variable?

miles-grant-ibigroup (changes requested):
I think it could be helpful to have a larger discussion about what exactly we're doing here... 

But in the meantime I have a few quick cleanup suggestions. Also, the new config item needs to go in the example config!

lib/actions/apiV2.js
daniel-heppner-ibigroup: withCollapsedShortNames isn't new

lib/actions/apiV2.js
daniel-heppner-ibigroup: easier sure, but less efficient.

lib/actions/apiV2.js
miles-grant-ibigroup: Ah might be a weird diff then. Either way please comment!

amy-corson-ibigroup: This and example config updated in [15b71fc](https://github.com/opentripplanner/otp-react-redux/pull/1024/commits/15b71fc61007ac7ecc59175566d53e96f758d3cb)!

lib/actions/apiV2.js
```diff
@@ -920,11 +923,33 @@ export function routingQuery(searchId = null, updateSearchInReducer) {
...
-              const withCollapsedShortNames =
-                response.data?.plan?.itineraries?.map((itin) => ({
+              const itineraries = response.data?.plan?.itineraries
+
+              // Convert user-selected transit modes from mode selector into modes recognized by OTP.
+              const activeModeStrings = activeModes.map(
+                (am) => SIMPLIFICATIONS[am.mode]
+              )
+
+              let filteredItineraries
```
miles-grant-ibigroup, line 933: This can be a const, then you don't need an else block
```suggestion
              const filteredItineraries = clone(response.data?.plan?.itineraries)
```

miles-grant-ibigroup (changes requested):
One last cleanup!

miles-grant-ibigroup (changes requested):
We're almost ready to go! We just need to get rid of that else block

miles-grant-ibigroup (approved):
We got there! Thanks for all your work on this

AdrianaCeric (approved):
Works great, nice job!

PR #1028, opened 2023-10-11
Optional Lateness Display

miles-grant-ibigroup:
**Description:**
Hides the lateness below the striked-through time display


example-config.yml
```diff
@@ -349,6 +349,8 @@ itinerary:
   # This flag disables the dot between legs in the metro itnerary summary, replacing it with
   # a gray background
   disableMetroSeperatorDot: false
+  # Show the number of minutes of delay or early beneath a time in the itinerary body
+  showLateness: true
```
amy-corson-ibigroup, line 353: nitpicky but i think this should be something like `showScheduleDeviation` (Not married to this) since setting this to false would remove late or early.

lib/components/viewers/realtime-status-label.tsx
```diff
@@ -117,45 +119,48 @@ const RealtimeStatusLabel = ({
...
-              values={{
-                originalTime: (
-                  <FormattedTime timeStyle="short" value={originalTime} />
-                )
-              }}
-            />
-          </InvisibleAdditionalDetails>
-        )}
-      </MainContent>
+      {showLateness && (
```
amy-corson-ibigroup, line 122: This is going to hide it by default on every config!

amy-corson-ibigroup (changes requested):
I think you'll either need to set this new variable to `true` on default or switch the config to `hideLateness`!

miles-grant-ibigroup:
Going to add a discussion item about making this change default. I think it should be.

miles-grant-ibigroup:
Moving forward with off by default as agreed

lib/components/viewers/realtime-status-label.tsx
```diff
@@ -117,45 +119,50 @@ const RealtimeStatusLabel = ({
...
+      )}
     </Container>
   )
 }
 
 const mapStateToProps = (state: {
   // Typescript TODO: type state
-  otp: { config: { onTimeThresholdSeconds: any } }
+  otp: {
+    config: { onTimeThresholdSeconds: any; showScheduleDeviation: boolean }
```
binh-dam-ibigroup, line 161: Add the new field to the config type instead.

binh-dam-ibigroup (changes requested):
Add the new config parameter to `lib/util/config-types.ts`, and use that type in `RealtimeStatusLabel`.

lib/components/viewers/realtime-status-label.tsx
```diff
@@ -117,45 +120,49 @@ const RealtimeStatusLabel = ({
...
+              />
+            </InvisibleAdditionalDetails>
+          )}
+        </MainContent>
+      )}
     </Container>
   )
 }
 
 const mapStateToProps = (state: {
```
binh-dam-ibigroup, line 159: Write `(state: AppState)` (import from state-types.js), then you don't need to declare all the details.

lib/components/viewers/realtime-status-label.tsx
```diff
@@ -117,45 +120,49 @@ const RealtimeStatusLabel = ({
...
+              <FormattedMessage
+                id="components.MetroUI.originallyScheduledTime"
+                values={{
+                  originalTime: (
+                    <FormattedTime timeStyle="short" value={originalTime} />
+                  )
+                }}
+              />
+            </InvisibleAdditionalDetails>
+          )}
```
binh-dam-ibigroup, line 152: This block still needs to be rendered in all cases. (It makes voice over say what the original time was, after enunciating the updated time first.)

binh-dam-ibigroup (changes requested):
The voice over tag for the original time needs to be present.

binh-dam-ibigroup (approved):
Nice, thanks for the changes

amy-corson-ibigroup (approved):
Looking good!

PR #1042, opened 2023-10-19
repair `getTransitiveRouteLabel` behavior

miles-grant-ibigroup:
**Description:**

Currently `getTransitiveRouteLabel` does nothing. This PR repairs this and makes it useful again

**PR Checklist:**
- [ ] Does the code follow accessibility standards (WCAG 2.1 AA Compliant)?
- [ ] Are all languages supported (Internationalization/Localization)?
- [ ] Are appropriate Typescript types implemented?

PR #1046, opened 2023-10-24
Add new  `onlyShowCountdownForRealtime` Option

miles-grant-ibigroup:
**Description:**

Blocked by #1042 

Supports better countdown clock behavior where it only shows up if there's realtime data! Way better.

**PR Checklist:**


AdrianaCeric (approved):
Great change!

lib/components/viewers/stop-time-cell.tsx
```diff
@@ -109,7 +114,10 @@ const StopTimeCell = ({
           status: getTripStatus(realtime, stopTime.departureDelay, 30) as status
         })}
       >
-        {showCountdown ? (
+        {/* Not the cleanest boolean, but makes it very clear what we're doing
+        in all cases. */}
+        {(onlyShowCountdownForRealtime === true && realtime) ||
+        (onlyShowCountdownForRealtime === false && showCountdown) ? (
```
amy-corson-ibigroup, line 120: Can this not be 
```suggestion
        {(onlyShowCountdownForRealtime && realtime) ||
        (!onlyShowCountdownForRealtime && showCountdown) ? (
```

amy-corson-ibigroup (approved):
I like this change! Just a question about the boolean but otherwise looks good.

lib/components/viewers/stop-time-cell.tsx
miles-grant-ibigroup: The explicit `true` helps us avoid false/null/undefined confusion

PR #1050, opened 2023-10-25
Prevent phone defaulting to verified when new account is created

amy-corson-ibigroup:
**Description:**

Fixes a small bug where, in account set up wizard, the notifications editor defaulted to the "verified/change number" state before the user had a chance to input their number.

**PR Checklist:**


miles-grant-ibigroup (approved):
Great fix thanks!

PR #1052, opened 2023-10-26
fix: Filter transit itineraries by including routeModeOverrides

amy-corson-ibigroup:
**Description:**

Includes checking `routeModeOverrides` when sorting transit itineraries. There may be a better/already existing way to do this in which case let me know!

Prevents a live bug where transit trips with override modes are being returned as walk trips.

image

**PR Checklist:**


lib/components/narrative/narrative-itineraries.js
```diff
@@ -444,8 +445,12 @@ class NarrativeItineraries extends Component {
           itinerary: cur
         })
 
+        const configModeOverrides = Object.values(routeModeOverrides)
+
         // Identify whether an itinerary uses transit & sort into appropriate bucket
-        const transitLegs = cur.legs.filter((leg) => isTransit(leg.mode))
+        const transitLegs = cur.legs.filter(
+          (leg) => isTransit(leg.mode) | configModeOverrides.includes(leg.mode)
```
binh-dam-ibigroup, line 452: Just curious if there are other places we need to make this change.

lib/components/narrative/narrative-itineraries.js
```diff
@@ -444,8 +445,12 @@ class NarrativeItineraries extends Component {
           itinerary: cur
         })
 
+        const configModeOverrides = Object.values(routeModeOverrides)
```
binh-dam-ibigroup, line 448: This line is making Percy tests fail. At least pass `{}` if `routeModeOverrides` is not set.

binh-dam-ibigroup (changes requested):

amy-corson-ibigroup:
Closed in favor of #1055

PR #1053, opened 2023-10-27
fix: Avoid reseting search when autoPlan is false

amy-corson-ibigroup:
**Description:**

Avoids reseting the search results in cases where `autoPlan` is false and `activeSearchId` might be null.

Also fixes a bug on mobile view where adjusting the search would revert to results without running the new search

To repro bug
- Use config with `autoPlan` set to false
- Plan trip
- Move location on map
- Go into mode selector and deselect a transit submode
- You should see trip search reverts back to trip results from step 2.

**PR Checklist:**


lib/actions/ui.js
```diff
@@ -279,7 +279,7 @@ export function handleBackButtonPress(e) {
     if (previousSearch) {
       // If back button pressed and active search has changed, set search to
       // previous search ID.
-      if (activeSearchId !== previousSearchId) {
+      if (activeSearchId && activeSearchId !== previousSearchId) {
```
binh-dam-ibigroup, line 282: How does this have to do with `autoplan = false`?

amy-corson-ibigroup: I think if the search changes, the activeSearchId changes, but if autoPlan is false, the activeSearchId still changes but is set to null, which is triggering this check.

percy/percy.test.js
```diff
@@ -121,17 +121,9 @@ async function executeTest(page, isMobile, isCallTaker) {
     await page.click('label[title="Transit"]')
     await page.waitForTimeout(200)
 
-    // FIXME: Must click Edit again [mobile-specific]
```
binh-dam-ibigroup, line 124: Nice fix!

binh-dam-ibigroup (approved):
Nice job fixing the percy tests!

PR #1057, opened 2023-11-01
Stricter itinerary order

binh-dam-ibigroup:
## Description

This PR makes the active itinerary index (`ui_activeItinerary` and related variables) refer to the same itinerary, regardless of sort order selected from the UI.

## PR Checklist


miles-grant-ibigroup (changes requested):
Immediate reactions before reviewing the code: "Best option" sorting doesn't seem to take late-arriving itineraries into account, those always get sent to the bottom. This doesn't seem to be an issue with other sort modes.

The activeItinerary ids also seem dependent on which order the requests return in, although I don't blame this PR for that issue.

Otherwise things look excellent, thanks for these fixes!

binh-dam-ibigroup:
> The activeItinerary ids also seem dependent on which order the requests return in, although I don't blame this PR for that issue.

I have that as a discussion topic for next week that could improve loading speed.

binh-dam-ibigroup:
> "Best option" sorting doesn't seem to take late-arriving itineraries into account, those always get sent to the bottom. 

@miles-grant-ibigroup Do you have an example of that please?

binh-dam-ibigroup:
> "Best option" sorting doesn't seem to take late-arriving itineraries into account, those always get sent to the bottom. This doesn't seem to be an issue with other sort modes.

@miles-grant-ibigroup As discussed, we have some missing itinerary attributes. For now, https://github.com/opentripplanner/otp-react-redux/pull/1057/commits/08642add89b7f4ab1ba15ef1cce2f80d752a7e9c defaults those to zero, so the result is not NaN, which previously messed up the sorting order for "Best Option".

miles-grant-ibigroup (approved):
I've discovered some more unrelated bugs while reviewing this, but code is much cleaner and intended functionality works. Thanks for the changes!

PR #1061, opened 2023-11-07
add `mutedErrors` itinerary config feature

miles-grant-ibigroup:
**Description:**

Adds a new config feature that allows the OTP2 error renderer to be disabled for errors a config does not want to display.

Co-authored by @AdrianaCeric 

**PR Checklist:**


amy-corson-ibigroup (approved):
Works great! Thank you!

lib/components/narrative/metro/metro-error-renderer.tsx
```diff
@@ -57,6 +65,9 @@ const ErrorRenderer = ({ errors }: { errors: Error }): JSX.Element => {
           return null
         }
 
+        // Don't show errors that have been muted in the config
```
binh-dam-ibigroup, line 68: Maybe move the two `return null` to the top of the mapping function.

binh-dam-ibigroup (approved):
Should be good to go after resolving conflicts.

lib/components/narrative/metro/metro-error-renderer.tsx
miles-grant-ibigroup: Good catch thanks!

PR #1063, opened 2023-11-08
Percy: GraphQL Mock Server

daniel-heppner-ibigroup:
**Description:**
This PR adds a new express.js server that handles GraphQL requests separately from REST requests. REST is handled by the HAR library that we already had, while GraphQL requests are handled by a GraphQL handler so that we can mock individual types on the GraphQL query and make decisions based on the arguments to it.

| Before | After |
    |--------|-------|
    | image | image  |


miles-grant-ibigroup:
Assigning back to you until the fixes we did together are in

miles-grant-ibigroup:
What are you stuck on right now again?

PR #1067, opened 2023-11-16
fallback for when pattern filtering filters all patterns

miles-grant-ibigroup:
**Description:**

The filtering was sometimes filtering too much. This PR makes sure that if there was too much filtering, we don't filter at all :)

**PR Checklist:**

### DIFF
--- a/lib/actions/apiV2.js
+++ b/lib/actions/apiV2.js
@@ -13,8 +13,8 @@ import { generateModeSettingValues } from '../util/api'
 import {
   getActiveItineraries,
   getActiveItinerary,
-  isValidSubsequence,
   getRouteOperator,
+  isValidSubsequence,
   queryIsValid
 } from '../util/state'
 import { getRouteColorBasedOnSettings } from '../util/viewer'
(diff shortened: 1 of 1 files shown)


lib/actions/apiV2.js
```diff
@@ -682,7 +682,12 @@ export const findRoute = (params) =>
                 })
               })
 
-            filteredPatterns.forEach((pattern) => {
+            // Fallback for if the filtering removes all patterns
+            // If this happens, it is not possible to know which pattern to keep
+            ;(filteredPatterns.length > 0
```
philip-cline, line 687: Is the semi colon intended?

philip-cline (approved):
Small formatting question, but otherwise looks like a good fix

lib/actions/apiV2.js
miles-grant-ibigroup: Unfortunately yes this is the one time the looney goose C non compliant semicolon free style we love these days shows its issues

philip-cline: Did you mean *Loosey Goose* ?

philip-cline: Could the semi colon go on the line previous?

miles-grant-ibigroup: Prettier put it here

amy-corson-ibigroup (commented):
Wait wouldn't it be more efficient to fix `filteredPatterns` so it never returns an empty array? For example, it shouldn't ever filter out the longest pattern, right?

miles-grant-ibigroup:
> Wait wouldn't it be more efficient to fix `filteredPatterns` so it never returns an empty array? For example, it shouldn't ever filter out the longest pattern, right?

You're right, but the issues we're seeing are only showing up on malformed GTFS, and this feels like a better approach to dealing with that than to work with it

amy-corson-ibigroup (approved):

PR #1072, opened 2023-11-17
Restore auto-highlight feature

miles-grant-ibigroup:
**Description:**
Some previous work broke the auto highlight behavior. We can fix this by moving the auto-highlight responsibility to React from Redux, which allows us to do so even when the array order doesn't match the display order.

**PR Checklist:**


binh-dam-ibigroup (approved):
Seems to work!

PR #1077, opened 2023-11-21
CallTaker: Remove Bike Tolerance Slider, add Bike Tolerance Dropdown

miles-grant-ibigroup:
**Description:**

Creates a more unified merged interface. 

**PR Checklist:**

image

### DIFF
--- a/lib/components/form/call-taker/advanced-options.js
+++ b/lib/components/form/call-taker/advanced-options.js
@@ -167,9 +167,9 @@ class AdvancedOptions extends Component {
     })
   }
 
-  _setBikeTolerance = ({ bikeTolerance }) => {
+  _setBikeReluctance = ({ bikeReluctance }) => {
     this.props.setUrlSearch({
-      bikeTolerance
+      bikeReluctance
     })
   }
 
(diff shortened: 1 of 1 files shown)


lib/components/form/call-taker/advanced-options.js
```diff
@@ -231,34 +235,32 @@ class AdvancedOptions extends Component {
...
             style={{ display: 'block', width: '33ch' }}
-            value={this.props.modeSettingValues.walkReluctance}
+            value={
+              bikeEnabled
+                ? this.props.modeSettingValues.bikeReluctance
+                : this.props.modeSettingValues.walkReluctance
+            }
           />
 
-          {hasBike(currentModes?.map((m) => m.mode).join(',') || '') ? (
```
daniel-heppner-ibigroup, line 248: Are you able to remove the hasBike function now?

daniel-heppner-ibigroup (approved):
Looks good, but I think we should make the values more configurable. This change also changes the available values for bike reluctance as a side effect.

amy-corson-ibigroup (approved):
I kinda think it would be less confusing to just have two `DropdownSelector`s instead of so many ternaries, but could just be personal preference! This works!

PR #1078, opened 2023-11-21
Use existing trip existence

binh-dam-ibigroup:
## Description

This PR makes the saved trip editor reuse the `itineraryExistence` object stored in the corresponding trip, instead of starting a new check that may not complete. (When saving new trips, a new itinerary existence check will continue to be initiated.) (That's the first commit in this PR.)

The PR also introduce `ItineraryExistence` and `MonitoredTrip` types and applies them to select files.

## PR Checklist


miles-grant-ibigroup (approved):
Not sure if this PR is supposed to fix this bug but I'm still getting the bug where you save a trip and then it tells you the trip doesn't exist 

Screenshot 2023-11-28 at 10 42 45 AM

Code looks clean!

binh-dam-ibigroup:
> Not sure if this PR is supposed to fix this bug but I'm still getting the bug where you save a trip and then it tells you the trip doesn't exist

This PR does not address this bug, it is a backend issue.

daniel-heppner-ibigroup (approved):
Nice improvement! Thanks

PR #1080, opened 2023-11-27
Add TOS Link from Config

amy-corson-ibigroup:
**Description:**

Allows a URL override for the Terms of Service that will direct to an external link 

**PR Checklist:**

### DIFF
--- a/lib/components/user/terms-of-use-pane.tsx
+++ b/lib/components/user/terms-of-use-pane.tsx
@@ -16,11 +16,13 @@ const TermsOfUsePane = ({
   disableCheckTerms,
   handleBlur,
   handleChange,
+  termsOfServiceLink,
   values: userData
 }: {
   disableCheckTerms: boolean
   handleBlur: () => void
   handleChange: FormEventHandler<Checkbox>
+  termsOfServiceLink: string
   values: {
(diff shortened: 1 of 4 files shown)


lib/components/user/terms-of-use-pane.tsx
```diff
@@ -16,11 +16,15 @@ const TermsOfUsePane = ({
   disableCheckTerms,
   handleBlur,
   handleChange,
+  locale,
+  termsOfServiceLink,
```
binh-dam-ibigroup, line 20: Please add `termsOfServiceLink` to the config types. I guess `termsOfStorageSet` needs to be added too.

lib/components/user/terms-of-use-pane.tsx
```diff
@@ -99,6 +103,8 @@ const TermsOfUsePane = ({
 }
 const mapStateToProps = (state: any) => {
```
binh-dam-ibigroup, line 104: Use type `AppReduxState`.

binh-dam-ibigroup (changes requested):
Could you please update the config types? Looks good otherwise.

AdrianaCeric (approved):
Great work!

PR #1081, opened 2023-11-27
Add alert tag on saved trips

AdrianaCeric:
**Description:**
It's not intuitive that alerts are shown in the edit button on saved trips. An alert notif/tag is shown as a seperate component. This is the first iteration before we change more styling

image

**PR Checklist:**

Q: Should I be using `LinkWithQuery` rather than button?
Which red colours are appropriate for this?

### DIFF
--- a/lib/components/user/monitored-trip/saved-trip-list.tsx
+++ b/lib/components/user/monitored-trip/saved-trip-list.tsx
@@ -99,7 +99,7 @@ class TripListItem extends Component<ItemProps, ItemState> {
     const { renderData, trip } = this.props
     const { itinerary } = trip
     const { legs } = itinerary
-    const alerts = renderData.alerts
+    const { alerts } = renderData
     // Assuming the monitored itinerary has at least one leg:
     // - get the 'from' location of the first leg,
     // - get the 'to' location of the last leg.
--- a/lib/components/user/monitored-trip/saved-trip-list.tsx
+++ b/lib/components/user/monitored-trip/saved-trip-list.tsx
@@ -36,6 +36,7 @@ import TripSummaryPane from './trip-summary-pane'
 interface ItemProps {
   confirmAndDeleteUserMonitoredTrip: (id: string, intl: IntlShape) => void
   intl: IntlShape
+  renderData: any
   routeTo: (url: any) => void
   togglePauseTrip: (trip: MonitoredTrip, intl: IntlShape) => void
   trip: MonitoredTrip
@@ -179,7 +180,7 @@ class TripListItem extends Component<ItemProps, ItemState> {
 }
 
 // connect to the redux store
-const itemMapStateToProps = (state, ownProps) => {
+const itemMapStateToProps = (ownProps: ItemProps) => {
   const { trip } = ownProps
   const renderData = getRenderData({
     monitoredTrip: trip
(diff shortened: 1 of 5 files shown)


lib/components/user/styled.ts
```diff
@@ -48,6 +48,22 @@ export const TripPanelHeading = styled(Panel.Heading)`
   background-color: white !important;
 `
 
+export const TripPanelAlert = styled(Button)`
+  border: none;
+  color: red;
```
amy-corson-ibigroup, line 53: We have a color variable in OTP-RR you can import here called RED_ON_WHITE!

lib/components/user/styled.ts
```diff
@@ -48,6 +48,22 @@ export const TripPanelHeading = styled(Panel.Heading)`
   background-color: white !important;
 `
 
+export const TripPanelAlert = styled(Button)`
```
amy-corson-ibigroup, line 51: For accessibility purposes this shouldn't be a button, [it should be a link! ](https://www.digitala11y.com/links-vs-buttons-a-perennial-problem/#:~:text=Use%20links%20when%20the%20user%2Daction%20doesn%E2%80%99t%20affect%20the%20website%20at%20all.%20In%20this%2C%20the%20users%20are%20just%20readers%20or%20spectators%20of%20the%20site.%20For%20example%2C%20to%20navigate%20to%20the%20next%20page%20or%20an%20external%20source%20after%20viewing%20the%20content%20of%20the%20page.)

lib/components/user/styled.ts
```diff
@@ -48,6 +48,22 @@ export const TripPanelHeading = styled(Panel.Heading)`
...
+  color: red;
+  cursor: pointer;
+  float: right;
+  text-decoration: underline;
+
+  :hover,
+  :focus,
+  :active {
+    background-color: transparent;
+    color: #d9534f;
```
amy-corson-ibigroup, line 62: Once we switch over to using an `a` (and therefore don't need to override bootstrap's style sheets), I'd recommend making this an `opacity: 90%` on hover instead of a different color!

amy-corson-ibigroup (changes requested):
Looks really good and works well! The only thing is instead of using a `button` for this, we should be using a link because clicking it takes the users to a different part of the site.

lib/components/user/styled.ts
AdrianaCeric: good catch!

lib/components/user/monitored-trip/saved-trip-list.tsx
```diff
@@ -92,9 +96,10 @@ class TripListItem extends Component<ItemProps, ItemState> {
   }
 
   render() {
-    const { trip } = this.props
+    const { renderData, trip } = this.props
     const { itinerary } = trip
     const { legs } = itinerary
+    const alerts = renderData.alerts
```
miles-grant-ibigroup, line 102:
```suggestion
    const { alerts } = renderData
```
Quick cleanup

miles-grant-ibigroup (approved):
Hard to test. I'm not crazy about the styling but since this whole thing is going to be rebuilt anyway, it's good enough for now!

AdrianaCeric:
(Re: alerts not showing up) There's probably an issue with monitored trip data here @binh-dam-ibigroup?

PR #1082, opened 2023-11-30
add GeoJSONOverlay

miles-grant-ibigroup:
**Description:**

Adds a new overlay that renders any geojson on the map!

to configure:
```yaml
 - type: geojson
      name: Custom Layer
      visible: true
      url: <geojson url here>
```

**PR Checklist:**

### DIFF
--- a/lib/components/map/connected-geojson-layer.js
+++ b/lib/components/map/connected-geojson-layer.js
@@ -1,0 +1,79 @@
+import {
+  Styled as BaseMapStyled,
+  MarkerWithPopup
+} from '@opentripplanner/base-map'
+import { connect } from 'react-redux'
+import FromToLocationPicker from '@opentripplanner/from-to-location-picker'
+import React, { useEffect, useState } from 'react'
+
+import { setLocation } from '../../actions/map'
+
+const GeoJSONOverlay = (props) => {
+  const { setLocation, url } = props
+
+  const [locations, setLocations] = useState([])
+  useEffect(() => {
+    async function downloadLocations() {
+      const json = await (await fetch(url)).json()
+      setLocations(json)
+    }
+    if (url) downloadLocations()
+  }, [])
+
+  return (
+    <>
+      {locations?.features?.map((feature, k) => {
+        const { geometry, properties } = feature
+        if (!geometry || !geometry.coordinates) return null
+        return (
+          <MarkerWithPopup
+            key={k}
+            popupContents={
+              <BaseMapStyled.MapOverlayPopup>
+                {properties.Name && (
+                  <BaseMapStyled.PopupTitle>
+                    {properties.Name}
+                  </BaseMapStyled.PopupTitle>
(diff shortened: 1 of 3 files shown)


lib/components/map/connected-geojson-layer.js
```diff
@@ -0,0 +1,79 @@
...
+                </BaseMapStyled.PopupRow>
+              </BaseMapStyled.MapOverlayPopup>
+            }
+            popupProps={{ offset: 10 }}
+            position={[geometry.coordinates[1], geometry.coordinates[0]]}
+          >
+            <img
+              alt={properties.Name}
+              src={properties.icon}
+              style={{ height: 40 }}
```
amy-corson-ibigroup, line 67: This is coming across as very tall!!

amy-corson-ibigroup (approved):
Seems to work well! I'm wondering though, aren't all these map layers geoJSON? Would it make sense to call this like "custom map layer"? Your call!

lib/components/map/connected-geojson-layer.tsx
```diff
@@ -0,0 +1,82 @@
...
+    async function downloadLocations() {
+      const json = await (await fetch(url)).json()
+      setLocations(json)
+    }
+    if (url) downloadLocations()
+  }, [url])
+
+  return (
+    <>
+      {/* @ts-expect-error TODO: geojson types */}
```
binh-dam-ibigroup, line 30: Try setting the type of `locations` to `GeoJSON.FeatureCollection` (do it where you call `useState`).

lib/components/map/connected-geojson-layer.js
binh-dam-ibigroup: I think a `className` should still be set on the `` tag so that the configs can style the images if needed.

binh-dam-ibigroup (changes requested):
Could you clean up the effect, the reference to map actions, and the types?

lib/components/map/connected-geojson-layer.tsx
```diff
@@ -0,0 +1,82 @@
...
+const GeoJSONOverlay = (props: Props) => {
+  const { setLocation, url } = props
+
+  const [locations, setLocations] = useState([])
+  useEffect(() => {
+    async function downloadLocations() {
+      const json = await (await fetch(url)).json()
+      setLocations(json)
+    }
+    if (url) downloadLocations()
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Unfortunately main useEffect functions must be sync

lib/components/map/connected-geojson-layer.tsx
```diff
@@ -0,0 +1,82 @@
...
+  }, [url])
+
+  return (
+    <>
```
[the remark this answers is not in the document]
miles-grant-ibigroup: It's not! The comment counts as an element

PR #1083, opened 2023-11-30
Import building blocks + Start Overriding Colors

amy-corson-ibigroup:
**Description:**

Import color variables from building-blocks and begin to add colors to route viewer, account settings, and itinerary.

**PR Checklist:**


miles-grant-ibigroup (changes requested):
Screenshot 2023-12-21 at 8 59 12 AM
Seeing some bizarre CSS issues when switching to this branch.

Also, I think I would prefer all the colors to be exported as `colors` and then `colors.blue` etc, but that one is on me for not clocking that during the OTP UI PR.

Otherwise the code looks good!

amy-corson-ibigroup:
I'm not seeing the weird CSS, could you let me know if you still are!!

amy-corson-ibigroup:
There are other style libraries that use this same structure! For example, building-blocks color exports were modeled after [material ui](https://github.com/mui/material-ui/tree/3322cbbd479881304e93b23708dd6c199eaf1464/packages/mui-material/src/colors)

lib/components/viewers/realtime-status-label.tsx
```diff
@@ -27,7 +28,7 @@ const Container = styled.div<{ withBackground?: boolean }>`
 
 const TimeStruck = styled.div`
   text-decoration: line-through;
-  opacity: 0.5;
+  opacity: 0.5'
```
miles-grant-ibigroup, line 31: 😓

miles-grant-ibigroup (approved):
We did it! This is lovely

lib/components/user/notification-prefs-pane.tsx
```diff
@@ -40,7 +42,7 @@ const NotificationOption = styled(ListGroupItem)`
     text-transform: uppercase;
   }
   label + span {
-    color: ${GRAY_ON_WHITE};
+    color: ${grey[700]};
```
binh-dam-ibigroup, line 45: Is there any indication in the code, other than `700`, to tell that this particular shade is meant to be used for, in this instance, gray-on-white text, as opposed to, say `600` and the possibility that a11y tests could fail? I think it would make sense to define aliases for some of the most used indexes/colors, so that we use them consistently long-term.

lib/components/viewers/realtime-status-label.tsx
```diff
@@ -37,11 +40,11 @@ const TimeBlock = styled.div`
 
 const STATUS = {
   EARLY: {
-    color: '#337ab7',
+    color: blue[700],
     label: 'early'
   },
   LATE: {
-    color: '#D92923',
+    color: red[700],
```
binh-dam-ibigroup, line 47: Why 700 over 800 that is used in the forms?

lib/components/viewers/route-details.tsx
```diff
@@ -240,7 +243,7 @@ class RouteDetails extends Component<Props> {
                   onClick={() => this._stopLinkClicked(stop)}
                   onMouseOver={() => setHoveredStop(stop.id)}
                   routeColor={
-                    routeColor.includes('ffffff') ? '#333' : routeColor
+                    routeColor.includes('ffffff') ? grey[900] : routeColor
```
binh-dam-ibigroup, line 246: Why 900 over the 800 used as the default route color in `connected-transit-vehicles`?

binh-dam-ibigroup (changes requested):
It is difficult to know which color indexes should be used in which context. To help with that, I would reinstate `/lib/util/colors.ts` and import the building blocks colors there, and define/add in that file aliases for the colors we use the most and how/where they are used.

lib/components/user/notification-prefs-pane.tsx
amy-corson-ibigroup: Really good point, reinstated the color variables in #1207!

lib/components/viewers/realtime-status-label.tsx
amy-corson-ibigroup: In general I was trying to match the original colors to the new variables as closely as possible, but I agree this is a good opportunity to consolidate some of these

amy-corson-ibigroup:
Closing in favor of #1207 because I agree with @binh-dam-ibigroup that these should be imported through the color.tsx file and wanted to avoid removing each instance of destructuring `colors` in this PR 😅

PR #1085, opened 2023-11-30
Prevent mobile view change on location clear

amy-corson-ibigroup:
**Description:**
Currently when you clear a location field in mobile view, it resets the view back to the form. This PR changes that so now using the clear location button just clears the location.

This PR also updates the mobile bar back button to be web accessible, and turns it on for the location field. 

**PR Checklist:**


daniel-heppner-ibigroup (changes requested):
I think we do need a close or back button here. The browser back button doesn't work to go back to the trip form. :( Sorry

daniel-heppner-ibigroup (approved):
Ooo yes nicely done! Thanks!

binh-dam-ibigroup:
@amy-corson-ibigroup Do we need to fix some of the Percy tests?

amy-corson-ibigroup:
@binh-dam-ibigroup Fixed!

amy-corson-ibigroup:
As @binh-dam-ibigroup pointed out, the mobile nav bar already has a back arrow, so I've used that instead of creating my own! I've also updated the back arrow to be web accessible but changing it from a `div` to a `button` and adding an `aria-label`

lib/components/mobile/mobile.css
```diff
@@ -52,9 +52,10 @@
 }
 
 .otp .navbar .mobile-back {
+  background: transparent;
+  border: none;
   position: fixed;
-  top: 12px;
-  left: 15px;
+  left: 7px;
   color: white;
```
binh-dam-ibigroup, line 59: Sort CSS

lib/components/mobile/navigation-bar.js
```diff
@@ -51,21 +53,28 @@ class MobileNavigationBar extends Component {
...
+    const backButtonText = intl.formatMessage({ id: 'common.forms.back' })
+
     return (
       <header>
         <Navbar className="mobile-navbar-container" fixedTop fluid>
           <Navbar.Header>
             <Navbar.Brand>
               {showBackButton ? (
-                <div className="mobile-back">
+                <button
```
binh-dam-ibigroup, line 69: Nice

binh-dam-ibigroup (approved):
> As @binh-dam-ibigroup pointed out, the mobile nav bar already has a back arrow, so I've used that instead of creating my own! I've also updated the back arrow to be web accessible but changing it from a `div` to a `button` and adding an `aria-label`

Nice fix!

PR #1087, opened 2023-12-04
Stricter cases to enable pattern filtering

miles-grant-ibigroup:
**Description:**

If only one pattern is returned, it's possible the pattern filtering is filtering too many patterns!

**PR Checklist:**


amy-corson-ibigroup (approved):
Can we update the comment since this no longer just applies to if the filter removes all patterns?

AdrianaCeric (approved):
👍

PR #1088, opened 2023-12-05
Add Itinerary flag to enable error on accessible trips

miles-grant-ibigroup:
**Description:**

Accessible trips can be funky, and this PR allows us to create a new error whenever we get accessible results. This can be used to warn users of poor accessible routing quality.

**PR Checklist:**


example-config.yml
```diff
@@ -362,6 +362,8 @@ itinerary:
   showScheduleDeviation: true
   # Shows the duration of a leg below the leg in the metro itinerary summary
   showLegDurations: false
+  # Whether to add a OTP_RR_A11Y_ROUTING_ENABLED error to all itineraries with accessibility scores
```
amy-corson-ibigroup, line 365: Can we add a note here that `OTP_RR_A11Y_ROUTING_ENABLED` will also need to be added in config?

amy-corson-ibigroup (approved):
Looks good!

example-config.yml
miles-grant-ibigroup: I think it's clear if it's missing (big error in the html()

lib/components/narrative/metro/metro-error-renderer.tsx
amy-corson-ibigroup, line 1: Nitpicky but is it possible to bold this link? it's a little difficult at first glance to tell what's going on in the error

miles-grant-ibigroup: It's possible, but this is a generic feature so we can't bold things in every case

amy-corson-ibigroup: Sure, but probably in every case if you're adding a link you're going to want that link to stand out from the rest of the error

lib/actions/apiV2.js
```diff
@@ -1011,6 +1011,24 @@ export function routingQuery(searchId = null, updateSearchInReducer) {
                   response.data?.plan?.routingErrors.filter(
                     (re) => re?.code !== 'NO_TRANSIT_CONNECTION'
                   )
+
+                // Add accessibility error, if it is turned on
+                const state = getState()
+                const { displayA11yError } = state.otp.config?.itinerary
+                if (displayA11yError) {
+                  if (
```
AdrianaCeric, line 1019: this is a lot of if statements maybe use `&&`?

AdrianaCeric (approved):
👍

lib/actions/apiV2.js
miles-grant-ibigroup: I think it's cleaner like this!

PR #1089, opened 2023-12-05
Update itinerary body packages

amy-corson-ibigroup:
**Description:**
Update itinerary body to include fixes for non-collapsible alerts & accessibility labels.

Are there other packages that should be included while we're in here?

**PR Checklist:**

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "@opentripplanner/geocoder": "^1.4.2",
     "@opentripplanner/humanize-distance": "^1.2.0",
     "@opentripplanner/icons": "^2.0.6",
-    "@opentripplanner/itinerary-body": "^5.1.3",
+    "@opentripplanner/itinerary-body": "^5.1.5",
     "@opentripplanner/location-field": "^2.0.13",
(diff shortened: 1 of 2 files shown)


miles-grant-ibigroup (changes requested):
I think this has to be blocked by the nearby view, which is causing the errors we're seeing in ci

package.json
```diff
@@ -45,7 +45,7 @@
     "@opentripplanner/geocoder": "^1.4.2",
     "@opentripplanner/humanize-distance": "^1.2.0",
     "@opentripplanner/icons": "^2.0.6",
-    "@opentripplanner/itinerary-body": "^5.1.3",
+    "@opentripplanner/itinerary-body": "^5.1.5",
```
binh-dam-ibigroup, line 48: Could you use the latest itinerary body package? As of writing, it is now 5.2.1 and includes a Tagalog language fix.

amy-corson-ibigroup:
Closed in favor of https://github.com/opentripplanner/otp-react-redux/pull/1102

PR #1092, opened 2023-12-14
Utilize OTP A11y score

amy-corson-ibigroup:
**Description:**
Replaces the `getAccessibilityScoreForItinerary` function with the `accessibilityScore` that is returned in the OTP itinerary.

**PR Checklist:**


miles-grant-ibigroup (approved):
So much better!

lib/util/accessibility-routing.js
```diff
@@ -2,17 +2,5 @@
  * Determine if an itinerary has accessibility scores
  */
 export const itineraryHasAccessibilityScores = (itinerary) => {
```
binh-dam-ibigroup, line 4: Because we are looking for an accessiblity score for all legs, we should rename this function to `itineraryHasAccessibilityScoresOnAllLegs` or `hasAccessibilityScoresOnAllLegs`.

lib/util/accessibility-routing.js
```diff
@@ -2,17 +2,5 @@
...
- * Calculates the total itinerary score based on leg score by weighting
- * each leg equally
- */
-export const getAccessibilityScoreForItinerary = (itinerary) => {
-  const scores = itinerary.legs
-    .map((leg) => leg.accessibilityScore || null)
-    .filter((score) => score !== null)
-
-  return scores.reduce((prev, cur) => prev + (cur * (1 / scores.length)), 0)
+  return !!itinerary.legs.find((leg) => !!leg.accessibilityScore)
```
binh-dam-ibigroup, line 5: Can the accessibility score of a leg be zero? if so, the whole itinerary score will NOT be displayed, and in such cases it should be displayed.

binh-dam-ibigroup (commented):
See my two comments at first glance.

binh-dam-ibigroup (changes requested):
Please advise on cases where a leg accessibility score is zero.

amy-corson-ibigroup:
I think the accessibility score would only be zero if one of the legs of the trip is absolutely confirmed to be inaccessible to wheelchair users. Wouldn't that be a trip we want to exclude?

binh-dam-ibigroup:
> I think the accessibility score would only be zero if one of the legs of the trip is absolutely confirmed to be inaccessible to wheelchair users. Wouldn't that be a trip we want to exclude?

Do you mean OTP would not return such an itinerary?

lib/util/accessibility-routing.js
amy-corson-ibigroup: Since the previous version of this function was just checking for itinerary scores on any of the legs, I repurposed it to check for the itinerary score itself! Which should make the name of this function still make sense but let me know if you disagree.

lib/util/accessibility-routing.js
amy-corson-ibigroup: Yes, understand what you mean now! Thank you, fixed in [92694f2](https://github.com/opentripplanner/otp-react-redux/pull/1092/commits/92694f2f351b69949b3ed95a9f9607c432f3fff5)

lib/util/accessibility-routing.js
binh-dam-ibigroup: Sounds good. But it should be singular (`itineraryHasAccessibilityScore`) since you are only checking one score!

lib/util/accessibility-routing.js
```diff
@@ -2,5 +2,7 @@
  * Determine if an itinerary has accessibility scores
  */
 export const itineraryHasAccessibilityScores = (itinerary) => {
-  return !!itinerary.legs.find((leg) => !!leg.accessibilityScore)
+  return (
+    itinerary.accessibilityScore !== null && itinerary.accessibilityScore > -1
```
binh-dam-ibigroup, line 6: Please add a comment regarding the -1 value.

binh-dam-ibigroup (approved):
Approving after internal discussion, and subject to the tweaks indicated. Sorry for holding up this work.

lib/util/accessibility-routing.js
amy-corson-ibigroup: Good point, changed in [ab6574e](https://github.com/opentripplanner/otp-react-redux/pull/1092/commits/ab6574ee9fe335fddcb9706508ebf9933ebc74a2)!

lib/util/accessibility-routing.js
amy-corson-ibigroup: Added in [ab6574e](https://github.com/opentripplanner/otp-react-redux/pull/1092/commits/ab6574ee9fe335fddcb9706508ebf9933ebc74a2)!

amy-corson-ibigroup:
Not at all! These were helpful changes!

lib/util/accessibility-routing.js
binh-dam-ibigroup: The comment talks about zero which is good, how about the -1?

amy-corson-ibigroup: I missed this comment before I merged, I'm sorry! I will sneak an updated comment into another otp-rr pr

PR #1094, opened 2023-12-15
add support for custom settings in calltaker

daniel-heppner-ibigroup:
**Description:**
This PR adds support for custom mode settings in the calltaker advanced options panel. 

**PR Checklist:**

image


daniel-heppner-ibigroup:
Currently relies on an alpha build

daniel-heppner-ibigroup:
Actually, I'm going to have @miles-grant-ibigroup review this first because he did some work on these files recently, and I removed some code he wrote.

miles-grant-ibigroup (changes requested):
Code looks great, nice and clean! Only issue is I don't see how to hide walk reluctance on bike+transit or bikesahre+transit. what do I put in the config? Is this possible?

Also is it possible to go back to the old layout? There's almost no room for results now on high zoom levels/low screen resolutions

daniel-heppner-ibigroup:
Hm, the layout thing is a problem. I'm not sure how the old layout would work given that there's the ability to have multiple settings. Isn't it fine if the results aren't visible while advanced options are open? 

Otherwise, the reason that walk reluctance is visible is because it's associated with the "transit" mode. We need to add "walk" to the transit only option, then assign walk reluctance mode setting to "walk" instead of "transit". This is the same as the regular mode selector.

daniel-heppner-ibigroup:
OK, so I made it show the first one in the old position and any extras are displayed below.

miles-grant-ibigroup (approved):
Great stuff!

daniel-heppner-ibigroup:
I updated the version of trip-details in this PR to get the new fare changes.

lib/components/form/call-taker/advanced-options.js
```diff
@@ -234,33 +223,14 @@ class AdvancedOptions extends Component {
...
-              bikeEnabled
-                ? this.props.modeSettingValues.bikeReluctance
-                : this.props.modeSettingValues.walkReluctance
-            }
-          />
-
+          {/* Show the first mode setting */}
+          {applicableModeSettings.length >= 1 && (
+            <ModeSettingRenderer
+              key={applicableModeSettings[0].key}
```
binh-dam-ibigroup, line 229: What is the need for the `key` prop here?

binh-dam-ibigroup (approved):
LGTM

PR #1097, opened 2023-12-19
Move delete saved trips functionality within edit functionality

AdrianaCeric:
**Description:**
Previously, the delete button on a saved trips card was seperate from the edit button, but it makes more sense if it were part of the edit page

**PR Checklist:**

image


lib/components/user/form-navigation-buttons.tsx
```diff
@@ -15,19 +15,24 @@ const RightButton = styled(Button)`
...
+export interface ButtonType {
+  /** The component to use for the button. */
+  ButtonComponent?: ReactNode
+  /** Whether the button is disabled. */
   disabled?: boolean
   /** Triggered when the button is clicked. */
   onClick?: MouseEventHandler<Button>
   /** The text to display on the button (JSX elements accepted). */
-  text: ReactElement
+  text?: ReactElement
```
miles-grant-ibigroup, line 26: Why is this optional?

lib/components/user/form-navigation-buttons.tsx
```diff
@@ -15,19 +15,24 @@ const RightButton = styled(Button)`
   float: right;
 `
 
-interface ButtonType {
+export interface ButtonType {
+  /** The component to use for the button. */
+  ButtonComponent?: ReactNode
```
miles-grant-ibigroup, line 20: Can we use `children` instead?

miles-grant-ibigroup (changes requested):
Looks good! Just a few nitpicks

lib/components/user/form-navigation-buttons.tsx
AdrianaCeric: this button text was already defined in our separate `DeleteForm` component with i18n, so we didn't need it for this button and then we decided to make it optional

lib/components/user/form-navigation-buttons.tsx
binh-dam-ibigroup: Or `content`?

AdrianaCeric: Good suggestions I think either make sense. I'll change to `content`

miles-grant-ibigroup: No I meant using `children` so you don't have to pass a prop at all and can instead just use the child of the component!

lib/components/user/form-navigation-buttons.tsx
miles-grant-ibigroup: Hm I'm not sure I love that... Can we name it `textOverride` or something?

AdrianaCeric:
`sequential-pane-display` is being removed??

lib/components/user/form-navigation-buttons.tsx
```diff
@@ -23,7 +23,7 @@ export interface ButtonType {
   /** Triggered when the button is clicked. */
   onClick?: MouseEventHandler<Button>
   /** The text to display on the button (JSX elements accepted). */
-  text?: ReactElement
+  textOverride?: ReactElement
```
binh-dam-ibigroup, line 26: Not sure why you'd want to rename this field, as there is no default text that is set if this attribute is not defined.

miles-grant-ibigroup: I thought @AdrianaCeric said that there was a default?

AdrianaCeric: I think naming can go either way. what matters is functionality. We can rename and refactor some things in FormNavigationButtons but for now its circling around the point of this PR

lib/components/user/form-navigation-buttons.tsx
AdrianaCeric: I understood how to implement this only halfway, but having trouble subsequently rendering children of extraButton in SavedTripEditor

miles-grant-ibigroup (approved):
With all our conversation noted here, I think this is good enough for now!

binh-dam-ibigroup (approved):

PR #1100, opened 2023-12-28
Make Auth0 cacheLocation Local Storage

amy-corson-ibigroup:
**Description:**
Sets the auth0Provider cache location to localstorage. This prevents users from being signed out on page refresh.

**PR Checklist:**

### DIFF
--- a/example-config.yml
+++ b/example-config.yml
@@ -68,6 +68,7 @@ persistence:
 #  enabled: true
 #  strategy: otp_middleware
 #  auth0:
+#    cacheLocation: localstorage (optional - default is "memory")
 #    domain: your-auth0-domain
 #    clientId: your-auth0-client-id
 #  otp_middleware:
--- a/example-config.yml
+++ b/example-config.yml
@@ -68,7 +68,6 @@ persistence:
 #  enabled: true
 #  strategy: otp_middleware
 #  auth0:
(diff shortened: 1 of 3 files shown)


example-config.yml
```diff
@@ -68,6 +68,7 @@ persistence:
 #  enabled: true
 #  strategy: otp_middleware
 #  auth0:
+#    cacheLocation: localstorage (optional - default is "memory")
```
binh-dam-ibigroup, line 71: Please add this new attribute to `Auth0Config` in config-types.ts.

binh-dam-ibigroup (approved):
It is working as intended on Chrome at least. We'll have to figure out something for Firefox and Safari. Please add the new config property to config-types before proceeding.

example-config.yml
binh-dam-ibigroup: Make the values a type that accepts two strings, i.e. `cacheLocation: 'memory' | 'localstorage'` or the equivalent type from the Auth0 library.

binh-dam-ibigroup:
> We'll have to figure out something for Firefox and Safari.

Correction: That settings works (I testing using the wrong config file...).

miles-grant-ibigroup (approved):
What could go wrong!

PR #1101, opened 2023-12-29
add `origColor` route color when route color is replaced

miles-grant-ibigroup:
**Description:**
We do a cool thing where we allow overriding the route color. This is useful, but there are cases when we might want the original one anyway, especially within a custom route renderer. This PR allows for this by appending an `origColor` field when the color is overwritten.

**PR Checklist:**
- [X] Does the code follow accessibility standards (WCAG 2.1 AA Compliant)?
- [X] Are appropriate Typescript types implemented?

### DIFF
--- a/lib/actions/apiV2.js
+++ b/lib/actions/apiV2.js
@@ -702,6 +702,18 @@ export const findRoute = (params) =>
                 stops: patternStops
               }
             })
+            newRoute.origColor = newRoute.color
+            newRoute.color = getRouteColorBasedOnSettings(
+              getRouteOperator(
+                {
+                  agencyId: newRoute?.agency?.id,
+                  id: newRoute?.route?.id
+                },
+                getState().otp.config.transitOperators
+              ),
+              { color: newRoute?.route?.color, mode: newRoute.mode }
+            ).split('#')?.[1]
+
             newRoute.patterns = routePatterns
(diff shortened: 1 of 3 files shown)


daniel-heppner-ibigroup (approved):
Looks good to me, this is a nice place to implement this.

lib/actions/apiV2.js
```diff
@@ -753,8 +765,30 @@ export function findRoutes() {
...
+                    agencyId: route?.agency?.id,
+                    id: route?.route?.id
+                  },
+                  config.transitOperators
+                ),
+                {
+                  color: route?.route?.color,
+                  mode: route.mode
+                }
+              ).split('#')?.[1]
```
amy-corson-ibigroup, line 790: This method looks almost the same as the one above (and I think there's a third one in this file as well) is there a way to pull this out so you don't have to write it out twice?

amy-corson-ibigroup (approved):
Looks good! Just had one question

lib/actions/apiV2.js
miles-grant-ibigroup: Yeah I thought about this but the problem is this is adapted from a core-utils method that's already slightly too wrong to work here so I was thinking you know let's just let this sit for a while, iron out all the changes and make sure that no tiny differences are needed and then we can clean this up in a future refactor!

PR #1102, opened 2023-12-29
Third pattern filtering patch

miles-grant-ibigroup:
**Description:**
This should hopefully be the last patch to the pattern filtering. This is what I get for not writing unit tests for this... This PR resolves the edge case where we have 2 patterns with the same length that are different. Currently, they're both thrown out (both subsets of each other) but this is clearly wrong. We correct this by adjusting one condition!

Unit tests failing until nearby view is merged

**PR Checklist:**


amy-corson-ibigroup (approved):
Why stop at a third, keep them coming! 🤠

daniel-heppner-ibigroup (approved):
Looks fine to me, do you want to merge it into nearby view first?

miles-grant-ibigroup:
Yes please @daniel-heppner-ibigroup. I will change target branch

PR #1103, opened 2023-12-30
some updates to example-config.yml

Pi-Cla:
**Description:**
Fix issue with ModeSelector for bicycle not having the label be recognized due to configuration using key `bike` instead of `bicycle`. See here: https://github.com/opentripplanner/otp-ui/blob/423502b5a281befbf0a0c0d84c4ba0aba6636b18/packages/trip-form/src/MetroModeSelector/i18n.ts#L29

Stamen Toner Lite is now being hosted on [Stadia](https://docs.stadiamaps.com/guides/migrating-from-stamen-map-tiles/) so update url to point to it now

**PR Checklist:**

Before
before
After
after


daniel-heppner-ibigroup (approved):
Good catch on this, thank you for the contribution!

miles-grant-ibigroup (approved):
This looks great thanks so much for your changes!

PR #1105, opened 2024-01-02
Convert monitored trip alert timestamps to seconds

binh-dam-ibigroup:
## Description

This PR adds a timestamp conversion to seconds for monitored trip alerts so that their effective date
displays with the correct year in the monitored trip editor.

For the first PR of the year, some snapshots needed to be updated.

**PR Checklist:**


miles-grant-ibigroup (approved):
How was this not previously caught ???

binh-dam-ibigroup:
> How was this not previously caught ???

See how much time we have been working on alerts lately!

lib/components/user/monitored-trip/trip-status-rendering-strategies/base-renderer.js
```diff
@@ -50,8 +62,11 @@ export default function baseRenderer(monitoredTrip) {
     )
   }
 
-  // set some alert data if the matching itinerary exists
-  data.alerts = data.matchingItinerary && data.matchingItinerary.alerts
+  // Set some alert data if the matching itinerary exists.
+  // (Convert alert timestamps in milliseconds to seconds for display using alerts body.)
+  data.alerts =
+    data.matchingItinerary &&
+    data.matchingItinerary.alerts.map(withAlertTimestampsInSeconds)
```
daniel-heppner-ibigroup, line 69: could you write this using null coalescing operator?

daniel-heppner-ibigroup (approved):
Wow, good catch!

lib/components/user/monitored-trip/trip-status-rendering-strategies/base-renderer.js
binh-dam-ibigroup: Did you mean optional chaining? As written, `data.alerts` can be `null` and the existing code supports that.

PR #1106, opened 2024-01-04
fix(stop-viewer): Fix camel case on realTime

amy-corson-ibigroup:
**Description:**

An issue with camelcase was accidentally filtering out ALL times in the stop viewer, leaving the departures pane blank! 

Is there an easier way to change the `stopTimes` variable names when we query so we don't have to change all of these fields?

**PR Checklist:**


binh-dam-ibigroup (approved):
Good catch on these OTP breaking changes!

miles-grant-ibigroup (changes requested):
I would request this be closed. I'll defer to others, but I see realtime as a single word.

miles-grant-ibigroup (approved):
Wait I'm understanding what's going on now. This is resolved by the nearby view anyway, but if we'd like a patch in time then this doesn't hurt

PR #1108, opened 2024-01-10
feat(routingQuery): Add banned routes to routing query

amy-corson-ibigroup:
**Description:**

In cases where we're using custom route mode overrides, we use the `banned` parameter in the routing query to make sure we are only returning mode results in line with the custom modes.

Uses https://github.com/opentripplanner/otp-ui/pull/699

**PR Checklist:**


daniel-heppner-ibigroup (approved):
I've made some changes to work with the OTP UI changes. We need to update version numbers though

lib/actions/apiV2.js
```diff
@@ -909,17 +910,23 @@ export function routingQuery(searchId = null, updateSearchInReducer) {
...
     const queryParamData = {
       modeButtons: rawModeButtonQP,
       ...modeSettingValues
     }
 
+    const excludedRoutes = getBannedRoutesFromSubmodes(
+      modeSettings,
+      config.routeModeOverrides
+    ).join(',')
+    console.log(excludedRoutes)
```
miles-grant-ibigroup, line 924: Whoops!

lib/actions/apiV2.js
```diff
@@ -909,17 +910,23 @@ export function routingQuery(searchId = null, updateSearchInReducer) {
...
+
     const baseQuery = {
       arriveBy,
       banned: {
+        routes: excludedRoutes || undefined,
```
miles-grant-ibigroup, line 929: Are we sure this is the right place to put this? Don't we have an instrument inside of the mode selector or something to do this at a more contextually relevant place?

miles-grant-ibigroup (changes requested):
A little worried about putting this inside the apiv2 file

lib/actions/apiV2.js
amy-corson-ibigroup: Do you know specifically what it could be? I'm looking in OTP-UI and the only reference to `banned` where we're passing this to the `PlanQuery`!

miles-grant-ibigroup: calltaker uses the `setQueryParam` method. Could we use that here?

miles-grant-ibigroup: Willing to let a third party adjudicate. @binh-dam-ibigroup

amy-corson-ibigroup: To summarize, apiv2 may not be ideal place to put it, but the alternative would be to add this to `batchSettings` which would require a pretty massive overhaul, or send more props to the mode selector, which feels excessive for what feels like an OTP-RR feature

binh-dam-ibigroup: This PR introduces a bug that crashes the UI upon making a trip search, so I am not able to see the intended effects of the proposed change.

amy-corson-ibigroup: @binh-dam-ibigroup Should be fixed! Sorry about that

binh-dam-ibigroup: @miles-grant-ibigroup The difference with calltaker is that in calltaker, users set each banned route individually, and there may be few. Here, we have to set many based on a UI setting, but it looks like the URL param for the selected submode from the mode selector already does that, so @amy-corson-ibigroup's approach is fine and it doesn't add extra stuff in the URL.

binh-dam-ibigroup (approved):

lib/actions/apiV2.js
daniel-heppner-ibigroup: I strongly support putting this code alongside where we generate the rest of the query. As I see it, the mode override functionality should be something that's built into OTP-RR at a low level (right before we send the query) so that any component that is using mode overrides doesn't need to worry about working around OTP's API limitations.

PR #1109, opened 2024-01-10
Fix colors in route viewer

amy-corson-ibigroup:
**Description:**

Fixing regression that is causing all route colors in route viewer to be set to #333333

Also took off a couple of pixels that were causing a weird crop on some routes with long names.

**PR Checklist:**

| Before | After |
|--------|-------|
|image|image|

### DIFF
--- a/lib/components/narrative/metro/default-route-renderer.tsx
+++ b/lib/components/narrative/metro/default-route-renderer.tsx
@@ -36,7 +36,7 @@ const DefaultRouteRenderer = ({
   const routeTitle = leg.routeShortName || leg.routeLongName
   return (
     <Block
-      color={leg.origColor || '333333'}
+      color={leg.origColor || leg.routeColor || '333333'}
       isOnColoredBackground={leg.onColoredBackground}
       style={style}
       title={routeTitle}
--- a/lib/components/narrative/metro/default-route-renderer.tsx
+++ b/lib/components/narrative/metro/default-route-renderer.tsx
@@ -25,7 +25,7 @@ const Block = styled.span<{ color: string; isOnColoredBackground?: boolean }>`
 `
 
 export type RouteRendererProps = {
-  leg: Leg & { onColoredBackground?: boolean; origColor?: string }
+  leg: Leg & { onColoredBackground?: boolean }
   style?: CSSProperties
 }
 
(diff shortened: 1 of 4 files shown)


lib/components/narrative/metro/default-route-renderer.tsx
```diff
@@ -37,7 +36,7 @@ const DefaultRouteRenderer = ({
   const routeTitle = leg.routeShortName || leg.routeLongName
   return (
     <Block
-      color={leg.routeColor || '333333'}
+      color={leg.origColor || '333333'}
```
miles-grant-ibigroup, line 39: Woah there! We need to include `routeColor` as a fallback!

lib/components/narrative/metro/default-route-renderer.tsx
```diff
@@ -26,7 +25,7 @@ const Block = styled.span<{ color: string; isOnColoredBackground?: boolean }>`
 `
 
 export type RouteRendererProps = {
-  leg: Leg & { onColoredBackground?: boolean }
+  leg: Leg & { onColoredBackground?: boolean; origColor: string }
```
miles-grant-ibigroup, line 28:
```suggestion
  leg: Leg & { onColoredBackground?: boolean; origColor?: string }
```

`origColor` is optional!

miles-grant-ibigroup (changes requested):
I have a few questions about this PR. Doesn't this defeat the point of route color overrides entirely? Since we're always using `origColor`

amy-corson-ibigroup:
> I have a few questions about this PR. Doesn't this defeat the point of route color overrides entirely? Since we're always using `origColor`

Definitely open to other suggestions! Using the color variables that are available on `leg` this seemed like the easiest way to make sure the appropriate colors were showing up on each route.

miles-grant-ibigroup:
> > I have a few questions about this PR. Doesn't this defeat the point of route color overrides entirely? Since we're always using `origColor`
> 
> Definitely open to other suggestions! Using the color variables that are available on `leg` this seemed like the easiest way to make sure the appropriate colors were showing up on each route.

Yes but doesn't this PR disable the route color override functionality? Since we're always using the `origColor`

amy-corson-ibigroup:
I see what you mean, I think I fixed it!

miles-grant-ibigroup (changes requested):
This still breaks the functionality! If we set a route color in the config, then don't we want it applied in the route viewer?

amy-corson-ibigroup:
Okay this should fix the route viewer and allow color overrides by mode

miles-grant-ibigroup (approved):
Works perfectly in every situation I could throw at it. Well done! Thanks for dealing with all this back and forth. Should be good to go once the snapshots are updated!

binh-dam-ibigroup (approved):
LGTM

PR #1112, opened 2024-01-17
Add support for mode icon overrides by route

amy-corson-ibigroup:
**Description:**
Makes mode icon overrides configurable! Changes mode icon displayed based on config without changing anything else.

**PR Checklist:**


miles-grant-ibigroup (changes requested):
I'm not a huge fan of all this new code... Could we instead just change each `<ModeIcon ...` to include a new routeId prop? And then inside the modeIcon component we can do all the logic we need. That way we don't have to write all this new code and don't have to add any new features to config.yaml?

amy-corson-ibigroup:
So would you add all the key value pairs to the custom mode icon renderer in this config.js file? That seems like it would get messy? 

I thought of adding a prop to the `ModeIcon` but the problem I ran into is the `LegIconWithAlly` component we pass to OTP-UI. I guess we could add the prop in OTP-UI too but keeping it in otp-rr seemed a little easier?

miles-grant-ibigroup:
> So would you add all the key value pairs to the custom mode icon renderer in this config.js file? That seems like it would get messy?
> 
It's messier than yours that's true but I think I prefer a cleaner otp-rr to a cleaner config. Happy to punt this to the tech call though if you think we should get everyone's feedback

> I thought of adding a prop to the `ModeIcon` but the problem I ran into is the `LegIconWithAlly` component we pass to OTP-UI. I guess we could add the prop in OTP-UI too but keeping it in otp-rr seemed a little easier?

Oh grr... I still think it's cleaner than this though! An entire new set of methods for a feature like this is giving me the ick

lib/components/narrative/metro/route-block.tsx
```diff
@@ -142,13 +149,15 @@ const RouteBlock = ({
...
+  console.log(leg)
+
   return (
     <>
       {showDivider && previousLegMode && <Divider>•</Divider>}
       <Wrapper className="route-block-wrapper">
         {leg.mode !== previousLegMode && (
           <LegIconWrapper>
-            <LegIcon height={28} leg={leg} />
+            <LegIcon height={28} leg={leg} routeId={leg.routeId || null} />
```
miles-grant-ibigroup, line 160: Why are we explicitly adding `leg.routeId` when we're already sending the entire `leg`?

lib/components/narrative/metro/route-block.tsx
```diff
@@ -142,13 +149,15 @@ const RouteBlock = ({
     }
   }
 
+  console.log(leg)
```
miles-grant-ibigroup, line 152: oops!

lib/components/viewers/RouteRow.js
```diff
@@ -192,6 +192,7 @@ export class RouteRow extends PureComponent {
                   }}
                   mode={getModeFromRoute(route)}
                   role="img"
+                  routeId={route.id}
```
miles-grant-ibigroup, line 195: Could we just add this to the `leg` like we do with the long and short names?

miles-grant-ibigroup (changes requested):
Some architecture questions

lib/components/narrative/metro/route-block.tsx
amy-corson-ibigroup: you dont want to keep ?

lib/components/narrative/metro/route-block.tsx
amy-corson-ibigroup: Yes good point!!

amy-corson-ibigroup:
The leg prop was not making it through `ClassicLegIcon`, so this fix now also relies on https://github.com/opentripplanner/otp-ui/pull/707

miles-grant-ibigroup (changes requested):
A few more cleanup possibilities!

lib/components/narrative/line-itin/LegIconWithA11y.tsx
```diff
@@ -19,7 +19,7 @@ const LegIconWithA11y = (props: any) => {
   const ariaLabel = isTransit(mode) ? getFormattedMode(mode, intl) : null
   return (
     <>
-      <LegIcon {...props} />
+      <LegIcon {...props} leg={leg} />
```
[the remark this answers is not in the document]
amy-corson-ibigroup: Good catch thank you!

lib/components/narrative/line-itin/LegIconWithA11y.tsx
```diff
@@ -17,6 +17,7 @@ const LegIconWithA11y = (props: any) => {
   const { leg } = props
   const { mode } = leg
   const ariaLabel = isTransit(mode) ? getFormattedMode(mode, intl) : null
+  console.log(props)
```
miles-grant-ibigroup, line 20: Whoops!

miles-grant-ibigroup (approved):
One last clean up and then I think we're finally good to go! What a tiny PR this wound up being haha

lib/components/narrative/line-itin/LegIconWithA11y.tsx
amy-corson-ibigroup: what is going on with me!!!

binh-dam-ibigroup (approved):

PR #1113, opened 2024-01-17
Fetch Auth0 Token on reload

amy-corson-ibigroup:
**Description:**
Fixes a little bug where if you are logged in, refresh the page, and then go to account page, you'd hit an endless loading spinner. 

**PR Checklist:**

### DIFF
--- a/lib/components/user/with-logged-in-user-support.js
+++ b/lib/components/user/with-logged-in-user-support.js
@@ -78,10 +78,19 @@ class UserLoaderScreen extends Component {
   }
 
   render() {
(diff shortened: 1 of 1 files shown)


binh-dam-ibigroup (approved):
Oh yes, great, thank you!

miles-grant-ibigroup (changes requested):
Woaaah... I'm a little nervous about putting this in the render method. It looks like we have this same code block inside `componentDidUpdate` already. If that's not working should we maybe consider moving it to `componentDidMount`? Or somehow make sure `componentDidUpdate` fires?

amy-corson-ibigroup:
Works even better in `componentDidMount`!

PR #1116, opened 2024-01-23
fix: don't use the old otp-ui query param framework

daniel-heppner-ibigroup:
**Description:**
There was a really subtle bug where in instances where the query didn't succeed, any query parameters that come from otp-ui's list of query parameters would overwrite the parameters set in the UI in the url search string. 

This PR removes that functionality entirely, since I don't think it's needed anymore with the new mode selector. 

There *could* be some edge case or config I can't checked that breaks with this, so it should require a lot of testing.

### DIFF
--- a/lib/actions/api.js
+++ b/lib/actions/api.js
@@ -17,7 +17,7 @@ import v2Actions from './apiV2'
 
 if (typeof fetch === 'undefined') require('isomorphic-fetch')
 
-const { getRoutingParams, getUrlParams } = coreUtils.query
+const { getUrlParams } = coreUtils.query
 
 // Generic API actions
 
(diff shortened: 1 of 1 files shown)


miles-grant-ibigroup (approved):
Got this working across all instances!

binh-dam-ibigroup (changes requested):
This PR would make trip searches un-bookmarkable. We and clients and users use trip bookmarks extensively, so this would be a major feature loss. I recommend closing this PR, and instead work on removing obsolete params from the search URL.

daniel-heppner-ibigroup:
@binh-dam-ibigroup You're right. I hadn't considered that. The mode selector state is stored in the URL directly and is not affected by this, but indeed the from/to fields are and some other params.

PR #1117, opened 2024-01-23
Update GraphQL Endpoint

miles-grant-ibigroup:
**Description:**
The GraphQL endpoint changed (see https://github.com/opentripplanner/OpenTripPlanner/pull/5581), so we need to also change.

**PR Checklist:**


daniel-heppner-ibigroup (approved):
LGTM except for all the old graphql queries in the HAR. I can remove them if you want

percy/mock.har
```diff
@@ -10,7 +10,7 @@
       {
```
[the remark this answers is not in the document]
miles-grant-ibigroup: Feel free

lib/actions/api.js
```diff
@@ -263,7 +263,11 @@ export function createQueryAction(
       }
     }
 
-    const url = makeApiUrl(config, endpoint, options)
+    const url = options?.url
+      ? `${config?.api?.host}${config?.api?.port ? ':' + config.api.port : ''}${
```
binh-dam-ibigroup, line 267: The host/port of the URL is repeated code with `makeApiUrl` and should be extracted, and why are we introducing a new `options.url` that duplicates the `endpoint` parameter?

lib/actions/apiV2.js
```diff
@@ -94,16 +94,16 @@ function createGraphQLQueryAction(
...
     headers: { 'Content-Type': 'application/json' },
     method: 'POST'
   }
-  return createQueryAction(endpoint, responseAction, errorAction, {
+  return createQueryAction(null, responseAction, errorAction, {
     ...options,
     fetchOptions,
-    noThrottle: true
+    noThrottle: true,
+    url: '/otp/gtfs/v1'
```
binh-dam-ibigroup, line 106: Can `url` be transfered to the `endpoint` argument?

percy/mock.har
```diff
@@ -10,7 +10,7 @@
...
         "request": {
           "method": "POST",
-          "url": "http://localhost:9999/otp2/routers/default/index/graphql",
+          "url": "http://localhost:9999//otp/gtfs/v1",
```
binh-dam-ibigroup, line 13: Remove the double slash before `otp` (multiple instances).

binh-dam-ibigroup (changes requested):
Please cleanup some of the changes.

lib/actions/api.js
miles-grant-ibigroup: The problem is that makeApiUrl appends `router/default` which we can't have here. I thought about adapting it to handle this new use case but once we're fully on graphQL it'll be deleted anyway!

lib/actions/apiV2.js
miles-grant-ibigroup: `endpoint` appends `routers/default`

lib/actions/api.js
binh-dam-ibigroup: I still think a function that deals with this portion: `${config?.api?.host}${config?.api?.port ? ':' + config.api.port : ''}${<arg>}` would be helpful.

binh-dam-ibigroup (changes requested):
There are now 3 identical pieces of code that assemble the OTP base path, they should be extracted.

lib/actions/api.js
miles-grant-ibigroup: I've removed `makeApiUrl`

lib/components/map/default-map.tsx
```diff
@@ -269,7 +269,9 @@ class DefaultMap extends Component {
       this.context
     const { baseLayers, maxZoom, overlays } = mapConfig || {}
     const { lat, lon, zoom } = this.state
-    const vectorTilesEndpoint = makeApiUrl(config, 'vectorTiles', {})
+    const vectorTilesEndpoint = `${assembleBasePath(
+      config
+    )}/otp/routers/default/vectorTiles`
```
binh-dam-ibigroup, line 274: Wow, `/otp/routers/default` should come from `config`.

binh-dam-ibigroup (commented):
@miles-grant-ibigroup Your last commit broke stuff.

lib/components/map/default-map.tsx
binh-dam-ibigroup: It is already the case in most of them.

miles-grant-ibigroup: What do you mean? This route is going to not follow the `/routers/default` pattern very soon. Therefore the `path` config item should be deprecatable soon. We don't allow configuration of the graphql path, so why this one?

binh-dam-ibigroup: But right now it does. It will also be easier to not miss this if referring to `config.api.path` once we switch to the new paths and get rid of `config.api.path`.

miles-grant-ibigroup: What about a global constant that can be overridden in the config?

miles-grant-ibigroup:
Not sure what is going on with the trip view request... Something has gone haywire with the mocks

lib/components/map/default-map.tsx
binh-dam-ibigroup:
```
    const vectorTilesEndpoint = `${assembleBasePath(
      config
    )}${config.api.path}/vectorTiles`
```

miles-grant-ibigroup: We should discuss further. I want to deprecate `path`

yarn.lock
```diff
@@ -2510,6 +2519,24 @@
     lodash.isequal "^4.5.0"
     qs "^6.9.1"
 
+"@opentripplanner/core-utils@^11.2.3":
```
binh-dam-ibigroup, line 2522: Can the dependencies de deduplicated?

lib/components/map/connected-transit-vehicle-overlay.js
```diff
@@ -138,7 +138,8 @@ const mapStateToProps = (state) => {
           ? '#' + route.color
           : route?.color || '#5A5A5A'
       // Try to populate this attribute, which is required for the vehicle popup to appear.
-      vehicle.routeShortName = vehicle.routeShortName || route?.shortName
+      vehicle.routeShortName =
+        vehicle.routeShortName || route?.shortName || true
```
binh-dam-ibigroup, line 142: Why add `true` as opposed to something like "unknown"?

lib/components/viewers/nearby/stop.tsx
```diff
@@ -105,11 +106,12 @@ const Stop = ({
...
   )
 
-  // TODO: We need to bring back the day break-up we had with the old stop viewer
   const patternRows = stopData.stoptimesForPatterns
     ?.reduce<PatternStopTime[]>((acc, cur) => {
       const dupe = acc.findIndex(
-        (p) => p.pattern.headsign === cur.pattern.headsign
+        (p) =>
+          extractHeadsignFromPattern(p.pattern) ===
+          extractHeadsignFromPattern(cur.pattern)
```
binh-dam-ibigroup, line 114: Maybe move `extractHeadsignFromPattern(cur.pattern)` to before line 111.

binh-dam-ibigroup (approved):
Did you accidentally stuff nearby view fixes? Could you also deduplicate the OTP-UI packages.

lib/components/map/connected-transit-vehicle-overlay.js
miles-grant-ibigroup: In case one day the overlay did decide to render the string, I didn't want it to possibly even have an unlocalized string to deal wtih

lib/components/viewers/nearby/stop.tsx
miles-grant-ibigroup: Great clean up thanks!

miles-grant-ibigroup:
Ugh I have no idea how that merge happened... I'm going to make this a piggyback PR

PR #1122, opened 2024-01-30
Support Node 21

miles-grant-ibigroup:
**Description:**

Really the solution is to upgrade web pack. However, this works too! Maybe it's wiser to spend that dev effort on something else otp-rr related... t🤔

**PR Checklist:**

## (Optional) Before and after screenshots for visual changes:
| Before | After |
    |--------|-------|
    |Screenshot 2024-01-30 at 3 23 56 PM|Screenshot 2024-01-30 at 3 23 56 PM|


daniel-heppner-ibigroup (approved):
I don't even think this is that much of a hack, lgtm

binh-dam-ibigroup (approved):
Oohh I have used `NODE_OPTIONS` to trick webpack it was running on Linux while actually being on Android!

PR #1123, opened 2024-01-30
fix sticky highlighted location

daniel-heppner-ibigroup:
**Description:**
Sometimes the highlighted location from nearby view would remain highlighted after leaving nearby view. It happened when clicking on a link in one of the cards, since the mouse-out handler never fired. We clean this up with a useEffect that ensures that the highlighted item is cleared when nearby view is unmounted.

Also fixes a crash when the OTP server is down.


miles-grant-ibigroup (approved):
Beautiful work

binh-dam-ibigroup (approved):

PR #1124, opened 2024-01-31
fix(DropdownOptions): Pass onChange to Field only when it exists

amy-corson-ibigroup:
**Description:**
Fixes a bug in Trip Notifications Settings where an undefined `onChange` was overriding Formik `Field`'s inherent `onChange`, making it so users could not change their trip notification settings.

**PR Checklist:**

### DIFF
--- a/lib/components/user/common/dropdown-options.tsx
+++ b/lib/components/user/common/dropdown-options.tsx
@@ -28,20 +28,16 @@ export const Select = ({
     componentClass: 'select',
     defaultValue,
     id: name,
-    name
+    name,
+    onChange
   }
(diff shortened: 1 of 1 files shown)


binh-dam-ibigroup (approved):
Good finding

miles-grant-ibigroup (changes requested):
This is a lot of new code for something small... Could we not set `fieldProps`, set `onChange` inside `fieldProps`, and then if onChange is null we can just `delete fieldProps.onChange`. That way we don't have to have a ternarhy

miles-grant-ibigroup (approved):
Wonderful!

PR #1132, opened 2024-02-06
Add titles to app menu items

amy-corson-ibigroup:
**Description:**

Add title prop to app menu items. Extra information for users, especially for menu items represented by icons. Also helpful for if a menu item overflow menu container.

**PR Checklist:**


lib/components/app/app-menu.tsx
```diff
@@ -154,6 +154,7 @@ class AppMenu extends Component<
             role={isSelected !== undefined ? 'option' : undefined}
             subItems={this._addExtraMenuItems(children) || undefined}
             text={label}
+            title={typeof label === 'string' ? label : ''}
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Actually, could you add this check inside of `AppMenuItem` and apply it to the `text` prop, so that if `text` is a string and `title` is not specified, then use `text` as the title. This would reduce the duplicate values passed to `AppMenuItem` below.

binh-dam-ibigroup (changes requested):
Can you deduplicate the `text`/`title` props, please?

binh-dam-ibigroup (approved):
Much better!

binh-dam-ibigroup:
@amy-corson-ibigroup Remember to update snapshots.

lib/components/app/app-menu.tsx
```diff
@@ -238,57 +238,71 @@ class AppMenu extends Component<
...
+              })}
             />
             {popupTarget && (
               <AppMenuItem
                 icon={<SvgIcon iconName={popupTarget} />}
                 onClick={this._triggerPopup}
-                text={<PopupTriggerText popupTarget={popupTarget} />}
+                text={intl.formatMessage({
+                  id: `config.popups.${popupTarget}`
+                })}
```
miles-grant-ibigroup, line 278: Why are we not using `PopupTriggerText` anymore?

miles-grant-ibigroup (changes requested):
Is the only reason we're moving from `FormattedMessage` to `intl.formatMessage` so we can do that type check?

lib/components/app/app-menu.tsx
amy-corson-ibigroup: That was a mistake! I will fix

miles-grant-ibigroup:
Ah ok that makes more sense thanks. Why are we setting the title and body the same? That doesn't feel normal or am I missing something

amy-corson-ibigroup:
@miles-grant-ibigroup  It's redundant in most existing cases, but 1) helps contextualize the menu items that use icons and 2) it's a safety net for if there's ever an instance where the text overflows the app menu container

miles-grant-ibigroup (approved):
That sounds good to me!

PR #1136, opened 2024-02-07
suport `/otp` override

miles-grant-ibigroup:
**Description:**

`path` will soon be deprecated once the router api is removed in OTP. However, we can't just hardcode `/otp` either, as some users have otp behind a proxy. This PR allows customizing the base path so that the graphql endpoint and the vector tile server can be hit with a customizable endpoint.

cc @fpurcell 

This adds a new (optional) configuration item `api.basePath`


package.json
```diff
@@ -25,7 +25,7 @@
     "parser": "babel-eslint"
   },
   "engines": {
-    "node": ">=16"
+    "node": ">=17"
```
binh-dam-ibigroup, line 28: Isn't it 21 we require elsewhere in the repo?

example-config.yml
```diff
@@ -1,6 +1,11 @@
 api:
   host: http://localhost
+  # Path will be deprecated once the REST API is deprecated, which will be soon.
+  # If the OTP server is hosted at /otp, then no path needs to be configured
   path: /otp/routers/default
+  # If your OTP server is at a path other than "/otp" (usually due to a proxy)
+  # Then you can set the OPTIONAL property basePath. The default is "/otp"
+  basePath: /otp
```
binh-dam-ibigroup, line 8: Please add `basePath` to the `ApiConfig` type.

binh-dam-ibigroup (changes requested):
LGTM

package.json
miles-grant-ibigroup: Yes but 17 is the minimum

example-config.yml
miles-grant-ibigroup: Good catch thank you!

lib/util/config-types.ts
```diff
@@ -31,7 +31,9 @@ export interface AccessibilityScoreConfig {
 
 /** OTP URL settings */
 export interface ApiConfig {
+  basePath?: string
   host: string
+  // Soon to be deprecated
   path: string
   port: number
   v2?: boolean
```
binh-dam-ibigroup, line 39: Let's add a deprecating comment to v2 too.

binh-dam-ibigroup (approved):

fpurcell (commented):
Hey @miles-grant-ibigroup  

Looking good.  See https://rtp.trimet.org/rtp. I'm seeing the basePath working for the /rtp proxy, fully talking (e.g., graphql and vectorTile urls) to OTP.

BTW, ignore my earlier feedback, ala:
~~A bit of mixed results (but mostly positive, going in the right direction):~~
~~that said, OTP's tilejson.json request for vector tile overlays is seemingly needing~~

The only thing that's a bit whacked right now is OTP's response of the tile URL, which is something I need to fix on my end.  Your fix here is allowing OTP-RR to talk with OTP via the /rtp proxy as it should.  And once I fix a few things on that proxy, the vector layers should show up too.

Ship It!
-Frank

p.s., the latest code (with the node version bump) is building on CentOS-7/Node 17 (won't build on Node 16 .. then again, code from a week ago won't build on 17).  FWIW, I can't install  versions greater than Node 17 on CentOS 7 due to GLIBC dependencies (might be possible, but "painful" according to a co-worker) seemingly hard-wired into the CentOS 7 release, and would have to move to CentOS 9 (which would also be painful).

fpurcell (approved):
lgtm

miles-grant-ibigroup:
We'll be sticking with 17 as a minimum version for a while now. No fear!

PR #1137, opened 2024-02-08
fix: critical bug fix merging routes in redux

daniel-heppner-ibigroup:
**Description:**
There was a bug in our reducer that caused route info to be overridden when the fetch routes response came back before the fetch route, overriding the extra details we ask for in fetch route. This deep merge will ensure that nothing gets overridden. The $merge utility that we used before was not deep.


miles-grant-ibigroup (approved):
Fix is so clean and works so well! Great job

binh-dam-ibigroup (approved):
Good find!

PR #1138, opened 2024-02-13
Small config and README changes for out-of-the-box use

abyrd:
I've recently been discussing OTP frontends with people evaluating components for new, large-scale OTP deployments. I've summarized the various frontend options in a PR to the main OTP docs: https://github.com/opentripplanner/OpenTripPlanner/pull/5679/files

Most people I talked to were under the impression that it would be very difficult to get any "real" (non-debug) UI up and running, and that it might take days of work by an experienced specialist. I took a chance tinkering with OTP-RR and after a few false starts, in the end it basically works out of the box with only a couple of changes to the default example config file.

Some simple ambiguities about indentation levels of settings and the meaning of "api: v2", as well as the internationalization behavior, made this seem much more complex than it actually is.

This PR proposes a few small changes to the config file and README that I believe will make this project immediately usable by most people getting started with OTP. This could really affect the perception and approachability of OTP among people evaluating it.

After realizing how straightforward this project now is to set up, I also think there's potential for OTP-RR to be considered the de facto stock frontend for OTP, and to be referred to as such in the documentation. At the time people began working on it, I recall OTP-RR being referred to as the new default frontend, and if I'm not mistaken there was an official steering committee decision to place these UI projects directly under the OpenTripPlanner organization (where they remain to this day).

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ An example of an OTP-RR application is included in the repository. The example p
 
 To run, first clone the repo and install [yarn](https://yarnpkg.com/) if needed.
 
-Update `example-config.yml` with the needed API keys, and optionally, the OTP endpoint and initial map origin. (The default values are for a test server for Portland, OR.).
+Update `example-config.yml` with the needed API keys, and optionally, the OTP endpoint and initial map origin. (The default values are for a test server for Portland, OR.). See the comments at the head of the config file for further details.
 
 Install the dependencies and start a local instance using the following script:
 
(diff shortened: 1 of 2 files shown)


binh-dam-ibigroup (approved):

miles-grant-ibigroup (approved):
There's a lot of legacy/needlessly complicated stuff in the example config. I dream of some js tool/script/similar that will allow otp-rr to run with just an OTP url. But for now this is a great start

PR #1140, opened 2024-02-14
fix(nearby-view): bug in location highlighting

daniel-heppner-ibigroup:
**Description:**
There was a bug causing the highlighted location to go away when the nearby view refreshed. This was caused by the unmount handler which was clearing the highlighted location when the view refreshed. I solved this by adding the router location as a dependency so it only clears when the location changes.

**PR Checklist:**


miles-grant-ibigroup (approved):
Not bad!

binh-dam-ibigroup (approved):

PR #1141, opened 2024-02-15
Fix Mobility Devices Input

binh-dam-ibigroup:
## Description

This PR fixes the behavior of the checkboxes in the pane for selecting mobility devices:
- "No assistive device" is checked if no other device is selected.
- "No assistive device" gets unchecked if user checks other devices.
- If user checks "No assistive device", other devices that were checked are unchecked.

## PR Checklist

### DIFF
--- a/lib/util/user.js
+++ b/lib/util/user.js
@@ -12,7 +12,7 @@ import { isBlank } from './ui'
 
 const { toSentenceCase } = coreUtils.itinerary
 
-export const noneSingleton = ['none']
+export const NONE_SINGLETON = ['none']
 
 /**
  * Determines whether a loggedInUser is a new user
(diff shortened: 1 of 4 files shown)


amy-corson-ibigroup (approved):
Makes sense and works well! 

The only thing that occurs to me is if mobility devices is set to "none" because the user neglected to choose any options, then I think next time they open the mobility editor, "none" should be checked. But that's a very small note, I'm happy to revisit it down the line.

binh-dam-ibigroup:
> The only thing that occurs to me is if mobility devices is set to "none" because the user neglected to choose any options, then I think next time they open the mobility editor, "none" should be checked.

Oh, it is not?

amy-corson-ibigroup:
Yeah, it looks like if a user goes through the sign up process and doesn't select anything on the mobility devices screen, the box isn't checked when they go into the editor. It does show up on the account page as "Mobility devices: No assistive device"

binh-dam-ibigroup:
> Yeah, it looks like if a user goes through the sign up process and doesn't select anything on the mobility devices screen, the box isn't checked when they go into the editor. It does show up on the account page as "Mobility devices: No assistive device"

Good catch! I made the change in https://github.com/opentripplanner/otp-react-redux/pull/1141/commits/f118e73b8dafb77b71f90c74c4131f2a0aaf7115 so a default is set for all new accounts when `mobilityProfile` is enabled in config.

lib/util/user.js
```diff
@@ -12,6 +12,8 @@ import { isBlank } from './ui'
 
 const { toSentenceCase } = coreUtils.itinerary
 
+export const noneSingleton = ['none']
```
miles-grant-ibigroup, line 15: should this be an all caps const?

miles-grant-ibigroup (approved):
One tiny nitpick. Overall very clean!

PR #1142, opened 2024-02-15
Further tweaks to config and readme

abyrd:
This is a follow-up to #1138 after figuring out a few more details on configuration and deployment.

Although all modes are selected by default, only the modes present in the UI are sent in the request. This does not include subway or ferry in the example configuration. In places where these modes exist, the resulting itineraries will look "broken" (always using buses instead of subways), and it's not immediately apparent that other modes need to be added or how to do so. I added some explanation on this, as well as subway and ferry modes. These can be commented out in the example YAML if it causes any problem, as their use is explained in other new comments.

The comments on i18n in my previous PR were not complete - I eventually realized i18n does not exactly work out of the box and has to be enabled in the config. The comments on the language section in the config begin by saying the section serves to override strings (rather than to enable languages). Further down it does say that the UI element for selecting languages will only appear if languages are defined in this section, but I eventually realized the languages are not even available for activation by navigator.language unless configured here. I have tried to make this clearer in the config comments.

I also added a few words to the instructions for starting a server and performing a production build.

All of my statements here are inferred from experimenting with OTP-RR while deploying it in an evaluation situation, so please correct me if I've reached any incorrect conclusions.
 
Aside: I noticed that the dev branch is not protected. It looks like the current workflow is for dev to always be updated from pull requests, so protection against direct pushes should probably be enabled on dev.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,11 +15,13 @@ To run, first clone the repo and install [yarn](https://yarnpkg.com/) if needed.
 
 Update `example-config.yml` with the needed API keys, and optionally, the OTP endpoint and initial map origin. (The default values are for a test server for Portland, OR.). See the comments at the head of the config file for further details.
 
-Install the dependencies and start a local instance using the following script:
+Install the dependencies and start a local development server using the following script:
 
 ```bash
 yarn start
 ```
+
+The port on which the development server listens can be configured in `.env`.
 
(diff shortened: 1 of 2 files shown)


miles-grant-ibigroup (approved):
These are some great clarifications thanks! Glad the readme and example config is getting some love. A real step in the right direction. As for branch protection I remember we turned it off for some reason but I can't remember why... might be good to look into that thanks

PR #1143, opened 2024-02-16
Handle 2-pattern-direction-with-same-headsign edge case

miles-grant-ibigroup:
**Description:**
There was a strange edge case where there could be 2 patterns with the same head sign. we fixed this by appending the stop info to the second head sign but this looked really weird. this PR fixes this by addressing this one specific edge case

**PR Checklist:**


amy-corson-ibigroup (approved):

lib/components/viewers/route-details.tsx
```diff
@@ -108,9 +108,27 @@ class RouteDetails extends Component<Props> {
...
+            amended[0].headsign = intl.formatMessage(
+              { id: 'components.RouteDetails.headsignTo' },
+              { ...amended[0] }
+            )
+          }
         }
 
-        amended.push(cur)
+        if (alreadyExistingIndex >= 0) {
+          // Only replace if new pattern has greater geometry
```
binh-dam-ibigroup, line 123: I recall that if we fuse patterns, at one point there was an issue navigating from the route viewer to the pattern viewer. Specifically, the pattern viewer would not show a pattern by default. Can you check that the issue is not reintroduced?

daniel-heppner-ibigroup: Are you referring to the bug that I fixed here? https://github.com/opentripplanner/otp-react-redux/pull/1137

miles-grant-ibigroup: Yes I can confirm that this bug isn't happening anymore with this branch

lib/components/viewers/route-details.tsx
```diff
@@ -108,9 +108,27 @@ class RouteDetails extends Component<Props> {
...
+          if (amended.length === 1 && Object.entries(patterns).length === 2) {
+            amended[0].headsign = intl.formatMessage(
+              { id: 'components.RouteDetails.headsignTo' },
+              { ...amended[0] }
+            )
+          }
         }
 
-        amended.push(cur)
+        if (alreadyExistingIndex >= 0) {
```
binh-dam-ibigroup, line 122: Add a comment regarding showing only one pattern if there were two patterns originally and the two patterns have the same text.

binh-dam-ibigroup (approved):
Add the comment as indicated.

PR #1144, opened 2024-02-16
Nearby View: Various small tweaks

miles-grant-ibigroup:
**Description:**
The old sorting didn't work at all. But the new one is smarter and sorts by departure time! Wonderful 

Adds support for multiple operator logos per stop

**PR Checklist:**

### DIFF
--- a/lib/components/util/operator-logo.tsx
+++ b/lib/components/util/operator-logo.tsx
@@ -3,6 +3,7 @@ import React from 'react'
 import styled from 'styled-components'
 
 const OperatorImg = styled.img`
+  margin-right: 0.25ch;
   width: 25px;
 `
 
--- a/lib/components/util/operator-logo.tsx
+++ b/lib/components/util/operator-logo.tsx
@@ -4,7 +4,7 @@ import styled from 'styled-components'
 
 const OperatorImg = styled.img`
   &:not(:last-of-type) {
-    margin-right: 0.25ch;
+    margin-right: 0.5ch;
   }
   width: 25px;
 `
--- a/lib/components/util/operator-logo.tsx
+++ b/lib/components/util/operator-logo.tsx
@@ -3,7 +3,9 @@ import React from 'react'
 import styled from 'styled-components'
 
 const OperatorImg = styled.img`
-  margin-right: 0.25ch;
+  &:not(:last-of-type) {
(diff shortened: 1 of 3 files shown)


lib/components/viewers/nearby/stop.tsx
```diff
@@ -159,18 +159,23 @@ const Stop = ({
...
       <CardHeader>
         <CardTitle>
-          <IconWithText icon={<Operator operator={operator} />}>
+          <IconWithText
+            icon={
+              <>
+                {transitOperators
+                  ?.filter((to) => Array.from(agencies).includes(to.agencyId))
+                  .map((to) => (
+                    <Operator key={to.agencyId} operator={to} />
```
amy-corson-ibigroup, line 174: Can we filter out duplicate operators here? 
image

lib/components/util/operator-logo.tsx
```diff
@@ -3,6 +3,9 @@ import React from 'react'
 import styled from 'styled-components'
 
 const OperatorImg = styled.img`
+  &:not(:last-of-type) {
+    margin-right: 0.25ch;
```
amy-corson-ibigroup, line 7: I think these are a little a little tight together, I wouldn't mind upping this to like .5!

amy-corson-ibigroup (changes requested):
Looks good, just some notes about the multiple operator logos

lib/components/viewers/nearby/stop.tsx
miles-grant-ibigroup: How would you suggest doing that? They're different operators with different IDs, but the same image. We could filter on image url? But isn't that a bit hacky

amy-corson-ibigroup: Hmm in this example, even though they're different feeds, it looks like both the operators have the same `name`?

miles-grant-ibigroup: Good idea! Filtering on name works great

miles-grant-ibigroup:
Great changes thanks!

amy-corson-ibigroup (approved):
Looks good! Thank you!

daniel-heppner-ibigroup (approved):
Great changes! thank you!!

PR #1155, opened 2024-03-01
make /stop route redirect to nearby view after fetcing stop location

daniel-heppner-ibigroup:
**Description:**
This PR moves the /stop route (which now only displays the schedule) to /schedule, and adds a new flow for /stop that first grabs the position of the stop, then redirects to the nearby view at those coordinates and highlights the requested stop.


miles-grant-ibigroup (approved):
It's not perfect and it crashes when there's an `entityId` in the url, but it works well enough

daniel-heppner-ibigroup:
Crash should be fixed before merging... I am wondering how you ended up on a URL with that in the path though

binh-dam-ibigroup (approved):
LGTM

PR #1156, opened 2024-03-04
OTP-RR 4.2.0

miles-grant-ibigroup:
**Description:**

There might be breaking change in here so maybe 5.0.0. Nearby view is stable enough now that I think it should be included in main branch deployments!


binh-dam-ibigroup:
> There might be breaking change in here so maybe 5.0.0.

@miles-grant-ibigroup Yes, the Nearby View removed support for some OTP1 calls.

binh-dam-ibigroup (approved):

PR #1159, opened 2024-03-07
nearby view: react to user location

miles-grant-ibigroup:
**Description:**

The nearby view was ignoring the user location. This is not a good thing! This PR makes use of the user current position, regardless of how we get that position.

**PR Checklist:**


daniel-heppner-ibigroup (approved):
Works much better, thank you! The only other thing that would be a nice to have is zooming the map to the user's location when you navigate to the nearby view without a set location. (as is, the map doesn't move, so the nearby view center could be off screen)

miles-grant-ibigroup:
> Works much better, thank you! The only other thing that would be a nice to have is zooming the map to the user's location when you navigate to the nearby view without a set location. (as is, the map doesn't move, so the nearby view center could be off screen)

I think this behavior is covered by another PR!

amy-corson-ibigroup (approved):
Looks good!

PR #1161, opened 2024-03-07
Shrink Down to one Start Over Method

miles-grant-ibigroup:
**Description:**
Uses the one start over method in all places it can be used.

**PR Checklist:**


binh-dam-ibigroup (changes requested):
Good fix, could you apply to the "Start Over" button under Trip Tools, please?

binh-dam-ibigroup:
> Good fix, could you apply to the "Start Over" button under Trip Tools, please?

@miles-grant-ibigroup As you do that, let's remove the remaining traces of `reactRouterConfig`.

lib/components/narrative/trip-tools.js
```diff
@@ -218,15 +216,11 @@ const TripTools = ({
...
         break
       case 'START_OVER':
-        const startOverUrl = startOver(
-          reactRouterConfig?.basename,
-          location?.search
-        )
         buttonComponents.push(
           <LinkButton
             Icon={Undo}
+            onClick={() => startOverFromInitialUrl()}
```
binh-dam-ibigroup, line 222: Can you just pass the function reference?

binh-dam-ibigroup (approved):
LGTM

lib/components/narrative/trip-tools.js
miles-grant-ibigroup: I tried but it didn't like it for some reason. Something with the import?

PR #1162, opened 2024-03-07
Fix route viewer crash

amy-corson-ibigroup:
**Description:**

Fixes a regression that was causing a crash when trying to use the route viewer! 

**PR Checklist:**

### DIFF
--- a/a11y/a11y.test.js
+++ b/a11y/a11y.test.js
@@ -14,7 +14,7 @@ const OTP_RR_TEST_CONFIG_PATH = '../a11y/test-config.yml'
 
 const MOCK_SERVER_PORT = 9999
 
-const DISABLED_ROUTES = ['/nearby', '/nearby/:latLon']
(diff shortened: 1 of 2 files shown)


miles-grant-ibigroup (approved):
We should probably put a better check here but for now this is ok. @daniel-heppner-ibigroup this is the solution to the crash I was talking about

daniel-heppner-ibigroup (approved):
Looks fine, we use this approach elsewhere in this file so I think it's ok. But why are the a11y tests failing?

daniel-heppner-ibigroup:
I was able to fix the a11y tests. It looks like this change uncovered another fix I should have made in a recent PR. Thanks for that!

daniel-heppner-ibigroup:
I was able to fix the a11y tests. It looks like this change uncovered another fix I should have made in a recent PR. Thanks for that!

PR #1165, opened 2024-03-08
CallTaker: first/last buttons affect sort

miles-grant-ibigroup:
**Description:**
It makes more sense for the first and last buttons to a affect sort so the option you really care about is right up front

**PR Checklist:**


amy-corson-ibigroup (approved):
Makes sense!

binh-dam-ibigroup (approved):

PR #1166, opened 2024-03-12
Add Default Error Message in Cases Where there are No Itins

amy-corson-ibigroup:
**Description:**

Adds a default OTP-style error message encouraging a user to change their search if we are returning a blank itinerary results pane. 

**PR Checklist:**

### DIFF
--- a/i18n/en-US.yml
+++ b/i18n/en-US.yml
@@ -431,6 +431,9 @@ components:
       minutes).
     verified: Verified
     verify: Verify
+    verifySms: >-
+      Please complete the verification process in order to set up SMS
+      notifications.
   Place:
     deleteThisPlace: Delete this place
     enterAlert: >
--- a/i18n/fr.yml
+++ b/i18n/fr.yml
@@ -444,7 +444,7 @@ components:
     requestNewCode: Envoyer un nouveau code
     sendVerificationText: Envoyer le SMS de vérification
     smsConsent: >
-      En donnant votre numéro téléphone, vous acceptez de recevoir des SMS de
+      En donnant votre numéro de téléphone, vous acceptez de recevoir des SMS de
       vérification et de suivi de vos trajets. Frais applicables selon votre
       opérateur téléphonique.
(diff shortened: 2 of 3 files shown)


miles-grant-ibigroup (changes requested):
I don't like the approach of putting this data decision in the presentation part of the error renderer. Could we instead dynamically add the error to the error set the way we do for `OTP_RR_A11Y_ROUTING_ENABLED`?

miles-grant-ibigroup (changes requested):
Also, could we change this from `DEFAULT_ERROR_MESSAGE` to something like `NO_ITINS_RETURNED`?

amy-corson-ibigroup:
@miles-grant-ibigroup We can add this to the error set in cases where the we actually return a response from the server! The issue is when the query is invalid (such as when a user only selects "rental" in the mode selector), we don't actually dispatch the routing request, and so there's no error set to add to.

miles-grant-ibigroup:
> @miles-grant-ibigroup We can add this to the error set in cases where the we actually return a response from the server! The issue is when the query is invalid (such as when a user only selects "rental" in the mode selector), we don't actually dispatch the routing request, and so there's no error set to add to.

That makes sense. But it definitely has to be added somewhere in the routing request process rather than at render time. What we have now is not ok

miles-grant-ibigroup (approved):
Beautiful work!  Excellently architected. This should be in a textbook

i18n/en-US.yml
```diff
@@ -384,7 +384,7 @@ components:
       body: >-
         No transit connection was found between your origin and destination on
         the selected day of service, using the vehicle types you selected.
-      header: No transit connections
+      header: No itineraries found
```
binh-dam-ibigroup, line 387: Do the other languages need to be updated too?

binh-dam-ibigroup (approved):

i18n/en-US.yml
amy-corson-ibigroup: I think so!

PR #1170, opened 2024-03-14
Recenter Nearby View on Map Pan

miles-grant-ibigroup:
**Description:**

Uses the maplibre `moveend` event to load new nearby view data when the map pans. Refactors some things to avoid too many calls

**PR Checklist:**


lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -126,19 +126,36 @@ function NearbyView({
...
-    map?.on('moveend', listener)
+
+    const dragListener = (e: any) => {
+      const coords = {
+        lat: e.viewState.latitude,
+        lon: e.viewState.longitude
+      }
+      setViewedNearbyCoords(coords)
+
+      // Briefly flash the highlight to alert the user that we've moved
```
binh-dam-ibigroup, line 145: Not the biggest fan of that effect, but that can be improved subsequently.

binh-dam-ibigroup (approved):

amy-corson-ibigroup (changes requested):
I think this is much cleaner if you remove the drag listener entirely and put all of the current drag listener logic under the move listener instead 
1. Often when dragging the map, the map physics can overshoot where the drag stopped. So when the map stops moving, the "map center" we just set is actually nowhere near the center. 
2. Using the `moveend` listener instead means the center is set when the map actually stops moving, and other movements like zoom also affect the map center, which I think is more intuitive.

amy-corson-ibigroup (approved):
Okay, I see what you mean!! It'd be nice to revisit this in the future, but for now I agree this is the best approach

PR #1171, opened 2024-03-15
Center map on location selection

miles-grant-ibigroup:
**Description:**

When an origin or destination is set, the map will now center to that point. If both are set, the map will show both. New refactored code makes this happen more efficiently with fewer frame drops.

**PR Checklist:**

### DIFF
--- a/lib/actions/map.js
+++ b/lib/actions/map.js
@@ -17,7 +17,7 @@ export const settingLocation = createAction('SET_LOCATION')
 /**
  * Centers the given map to the coordinates of the specified place.
  */
-export function setMapCenter(map /* MapRef */, location, zoom) {
+export function setMapCenter(map /* MapRef */, location) {
   return function () {
     const { lat, lon } = location
     if (map && !isNaN(lat) && !isNaN(lon)) {
--- a/lib/actions/map.js
+++ b/lib/actions/map.js
@@ -22,9 +22,7 @@ export function setMapCenter(map /* MapRef */, location, zoom) {
     const { lat, lon } = location
     if (map && !isNaN(lat) && !isNaN(lon)) {
       const center = [lon, lat]
-      // Note: passing an undefined zoom to map.jumpTo will trigger a "can't invert matrix" error.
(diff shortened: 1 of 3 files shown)


lib/actions/map.js
```diff
@@ -22,9 +22,7 @@ export function setMapCenter(map /* MapRef */, location, zoom) {
     const { lat, lon } = location
     if (map && !isNaN(lat) && !isNaN(lon)) {
       const center = [lon, lat]
-      // Note: passing an undefined zoom to map.jumpTo will trigger a "can't invert matrix" error.
-      const jumpParams = zoom !== undefined ? { center, zoom } : { center }
-      map.jumpTo(jumpParams)
```
binh-dam-ibigroup, line 27: `zoom` becomes unused and can be removed from the function parameters.

binh-dam-ibigroup (approved):
Nice

lib/components/app/responsive-webapp.js
```diff
@@ -97,9 +96,19 @@ class ResponsiveWebapp extends Component {
...
       if (isMobile() && query.from === null) {
         setLocationToCurrent({ locationType: 'from' }, intl)
-        setMapCenter(map, pt, initZoomOnLocate)
+        setMapCenter(map, pt)
       }
+    } else if (query.from && query.to) {
+      map?.fitBounds([query.from, query.to], {
+        duration: 500,
+        padding: getFitBoundsPadding(map, 0.2)
+      })
```
amy-corson-ibigroup, line 105: I think something we discussed was if both location fields are filled, set the nearby center to the destination location? Setting the nearby center to the center of the trip is what we were looking to change

amy-corson-ibigroup (changes requested):
Nice! Just a question about the behavior when both location fields are set.

lib/components/app/responsive-webapp.js
miles-grant-ibigroup: That happens in a different PR. This PR only handles map panning

PR #1175, opened 2024-03-20
Use query to set focus point

miles-grant-ibigroup:
**Description:**
We use the `from` or `to` query to set the focus point for the other. helps users plan short trips (?)

Screenshot 2024-03-20 at 12 12 13 PM

**PR Checklist:**

### DIFF
--- a/lib/components/form/connect-location-field.js
+++ b/lib/components/form/connect-location-field.js
@@ -80,17 +80,19 @@ export default function connectLocationField(
     }
 
     const geocoderConfig = config.geocoder
-    if (query.to) {
-      const { lat, lon } = query.to
-      geocoderConfig.focusPoint = { lat, lon }
-    }
-    if (query.from) {
-      const { lat, lon } = query.from
-      geocoderConfig.focusPoint = { lat, lon }
-    }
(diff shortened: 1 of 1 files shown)


amy-corson-ibigroup (approved):

lib/components/form/connect-location-field.js
```diff
@@ -80,9 +80,19 @@ export default function connectLocationField(
...
+    if (geocoderConfig) {
+      if (query.to && query.to?.lat && query.to?.lon) {
+        const { lat, lon } = query.to
+        geocoderConfig.focusPoint = { lat, lon }
+      }
+      if (query.from && query.from?.lat && query.from?.lon) {
+        const { lat, lon } = query.from
+        geocoderConfig.focusPoint = { lat, lon }
+      }
+      if (currentPosition?.coords) {
```
binh-dam-ibigroup, line 92: Should these be `else if` blocks? Right now, the order of priority reads like:
1. current position
2. from
3. to

miles-grant-ibigroup: this is the intended priority order

binh-dam-ibigroup (approved):

PR #1178, opened 2024-03-25
Monitored Trips - Middleware GraphQL support

binh-dam-ibigroup:
## Description

Works with https://github.com/ibi-group/otp-middleware/pull/243.

This PR supports OTP-middleware itinerary existence checks using GraphQL. Variables for the corresponding `plan` query are included with the monitored trip data when a trip is saved and the trip existence is checked. A new field `leg.originalMode` is introduced for saving and new/existing trip to revert route mode overrides made by the UI. The PR also updates the UI so that the route numbers are shown properly in the Saved Trips page.

## PR Checklist


lib/actions/user.js
```diff
@@ -169,14 +174,35 @@ function convertRequestToSearch(config) {
   }
 }
 
+// TODO move to a utils file
+function areRequestsSame(qp1, qp2) {
```
amy-corson-ibigroup: I feel like we've talked about this before, but would it be easier to compare these by excluding unwanted values, stringifying the objects, and comparing the strings?

lib/actions/user.js
```diff
@@ -169,14 +174,35 @@ function convertRequestToSearch(config) {
...
+  // Everything except modes.
```
amy-corson-ibigroup: Also, why do we exclude modes from this check?

amy-corson-ibigroup (approved):
Looks good! Thank you for cleaning up some of these `routeModeOverride` functions!

lib/actions/user.js
binh-dam-ibigroup: Better than stringify, I used deep comparison in https://github.com/opentripplanner/otp-react-redux/pull/1178/commits/ed5f0ad46b92ac710b956ccf09ad6b3d9ebf4054 (We already use a library for that!)

lib/actions/user.js
binh-dam-ibigroup: This is for comparing two GraphQL variables from the same trip search, where we send multiple mode combinations. I have clarified the comment in https://github.com/opentripplanner/otp-react-redux/pull/1178/commits/7817c5b9f592a7c4d5d78999d813f28e555d515b.

josh-willis-arcadis (approved):
LGTM!

PR #1179, opened 2024-03-25
fix(route-viewer): Cleanup duplicate headsign filtering

amy-corson-ibigroup:
**Description:**
Fixes bug that was causing one headsign to show up in cases where there are only two headsigns on a pattern and both headsigns have the same name.

Also updates some of the comments for clarity.

**PR Checklist:**


miles-grant-ibigroup (approved):
Working great! Thanks for the fixes

binh-dam-ibigroup (approved):

PR #1182, opened 2024-03-26
feat(connected-itinerary-body): make alert collapse configurable

amy-corson-ibigroup:
**Description:**
Use `alwaysCollapseAlerts` prop in connected itinerary body to configure whether users can collapse alerts.

**PR Checklist:**


example-config.yml
```diff
@@ -347,6 +347,8 @@ itinerary:
   renderRouteNamesInBlocks: true
   # Whether the mode icons should be colored as well
   fillModeIcons: true
+  # Allow user to collapse alerts in itinerary body
+  collapseAlerts: true
```
miles-grant-ibigroup, line 351: This is misleadingly named. Can we name it `allowUserAlertCollapsing`?

miles-grant-ibigroup (changes requested):
One small nitpick!

lib/components/narrative/line-itin/connected-itinerary-body.js
```diff
@@ -116,6 +117,7 @@ class ConnectedItineraryBody extends Component {
       <ItineraryBodyContainer>
         <StyledItineraryBody
           accessibilityScoreGradationMap={accessibilityScoreGradationMap}
+          alwaysCollapseAlerts={allowUserAlertCollapsing}
```
miles-grant-ibigroup, line 120:
```suggestion
          alwaysCollapseAlerts={!allowUserAlertCollapsing}
```
Should this be flipped?

miles-grant-ibigroup (approved):
Looks good! Just one boolean thing to deal with

lib/components/narrative/line-itin/connected-itinerary-body.js
amy-corson-ibigroup: I don't think so, I'm setting `allowUserAlertCollapsing` to `true` in the config!

amy-corson-ibigroup: Unless we want alerts to collapse by default, which might be preferable?

miles-grant-ibigroup: But as it is right now if `allowUserAlertCollapsing` is true, then they won't be allowed to collapse them right? Or am I forgetting how the behavior works

amy-corson-ibigroup: No, `alwaysCollapseAlerts` needs to be true for the alerts to collapse

binh-dam-ibigroup (changes requested):
Please add the new attribute to the corresponding config type.

lib/util/config-types.ts
```diff
@@ -243,6 +243,7 @@ export interface ItineraryCostWeights {
 }
 
 export interface ItineraryConfig {
+  allowUserAlertCollapsing: boolean
```
binh-dam-ibigroup, line 246: Yes, but make it optional:
```suggestion
  allowUserAlertCollapsing?: boolean
```

binh-dam-ibigroup (approved):
Approving subject to https://github.com/opentripplanner/otp-react-redux/pull/1182/files#r1543097570.

PR #1185, opened 2024-04-01
Save trip fixes

amy-corson-ibigroup:
**Description:**
- Fixes an issue where focus styling on checkboxes did not have high enough contrast with monitored day background (both were blue)
- Prevents app from crashing if a user refreshes on the "save new trip" screen. Instead, should show a "trip not found" screen and a link back to the trip planner.

**PR Checklist:**


lib/components/user/monitored-trip/trip-not-found.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+import BackLink from '../back-link'
+import PageTitle from '../../util/page-title'
+import React, { ReactComponentElement } from 'react'
+
+interface Props {
+  intl: IntlShape
+}
+
+const TripNotFound = ({ intl }: Props): JSX.Element => {
```
binh-dam-ibigroup, line 11: Instead of passing `intl` around, call the `useIntl` hook.

lib/components/user/monitored-trip/trip-basics-pane.tsx
```diff
@@ -101,6 +101,19 @@ const AvailableDays = styled(FieldSet)`
...
+    &:focus-visible,
+    &:focus {
+      outline-offset: 0.5px;
+      outline: solid 2px blue;
+      &:checked {
+        outline: solid 2px white;
+      }
+    }
+  }
+`
```
binh-dam-ibigroup, line 116: Could you move these to the `AvailableDays` component? (it is a styled component definition) In doing so you won't need the `CheckboxWrapper`.

binh-dam-ibigroup (changes requested):
See suggestions for cleanup and combining CSS.

lib/components/user/monitored-trip/trip-basics-pane.tsx
amy-corson-ibigroup: Oh perfect! Thank you

binh-dam-ibigroup (approved):
Good quick fix!

lib/components/user/monitored-trip/saved-trip-screen.js
```diff
@@ -163,13 +164,16 @@ class SavedTripScreen extends Component {
...
+    const { isCreating, itinerary, loggedInUser, monitoredTrips, pending } =
+      this.props
     const isAwaiting = !monitoredTrips || (isCreating && pending)
 
     let screenContents
     if (isAwaiting) {
       // Flash an indication while the selected and saved user trips are being loaded.
       screenContents = <AwaitingScreen />
+    } else if (isCreating && !itinerary) {
+      screenContents = <TripNotFound />
```
miles-grant-ibigroup, line 176: Are we sure this doesn't flash while loading?

miles-grant-ibigroup (approved):
Working well, but have one quick thing we should double check. I didn't observe it in testing but better safe than sorry!

PR #1188, opened 2024-04-04
fix(nearby-view): don't flash stale data while loading new response

daniel-heppner-ibigroup:
**Description:**
Nearby view would show stale data from redux when loading a new location. This fixes that by storing the associated coordinates along with the nearby response in Redux, allowing us to hide the results if the coords don't match the URL coords. 

**PR Checklist:**

### DIFF
--- a/lib/components/viewers/nearby/nearby-view.tsx
+++ b/lib/components/viewers/nearby/nearby-view.tsx
@@ -75,7 +75,6 @@ function NearbyView({
   nearbyViewCoords,
   radius,
   setHighlightedLocation,
-  setLocation,
   setMainPanelContent,
   setViewedNearbyCoords,
   zoomToPlace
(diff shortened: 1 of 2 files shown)


miles-grant-ibigroup (approved):
Clean, elegant, and solves all our problems. Thanks!

lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -66,7 +67,26 @@ const getNearbyItem = (place: any) => {
   }
 }
 
+function getNearbyCoords(
```
binh-dam-ibigroup, line 70: Rename to `getNearbyCoordsFromUrlOrMapCenter`

binh-dam-ibigroup (approved):

PR #1191, opened 2024-04-10
Avoid setting ui_itineraryView on desktop

binh-dam-ibigroup:
## Description

This PR removes the use of the `ui_itineraryView` URL parameter on desktop. Its change would previously cause, on desktop view (but not on mobile view), a re-render of the map with a reframing on the entire itinerary.

## PR Checklist


amy-corson-ibigroup (approved):
LGTM!

miles-grant-ibigroup (approved):
Can't believe this wasn't needed! Working great thanks for the cleanup

PR #1197, opened 2024-04-18
Avoid non-itin-view jumping

miles-grant-ibigroup:
**Description:**
We used to jump around a lot and it made everyone nauseous. This PR restricts when we run some of these flyTo events. 

Please review carefully, this may have created some bugs.

### DIFF
--- a/lib/components/app/responsive-webapp.js
+++ b/lib/components/app/responsive-webapp.js
@@ -62,6 +62,7 @@ class ResponsiveWebapp extends Component {
       formChanged,
       intl,
       location,
+      mainPanelContent,
       map,
       matchContentToUrl,
       query,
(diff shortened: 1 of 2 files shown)


daniel-heppner-ibigroup:
I see this has fixed some zooming issues in the route viewer, but I'm still seeing a lot of unnecessary zooming in other parts of the app. I think I'd like to see that fixed in this PR as well.

miles-grant-ibigroup:
> I see this has fixed some zooming issues in the route viewer, but I'm still seeing a lot of unnecessary zooming in other parts of the app. I think I'd like to see that fixed in this PR as well.

I'm unable to replicate. All the movement I see is sensible

daniel-heppner-ibigroup:
How to replicate:
- Plan a trip
- Move the map
- Adjust any of the trip form properties, like selecting a mode setting.

Observe how the map zooms unnecessarily back to the planned trip. Not good ux

miles-grant-ibigroup:
> How to replicate:
> 
> * Plan a trip
> * Move the map
> * Adjust any of the trip form properties, like selecting a mode setting.
> 
> Observe how the map zooms unnecessarily back to the planned trip. Not good ux

This makes sense though doesn't it? If I'm changing the mode, why would I want to be looking at some random part of the map? I want to be looking at the trip I was planning

miles-grant-ibigroup:
Ok we have agreed offline to make the jumping overall configurable.

PR #1202, opened 2024-04-22
fix: mode selector settings sync with url correctly

daniel-heppner-ibigroup:
**Description:**
The root of this bug is that we replace the URL search with the `currentQuery` every time a trip plan is sent off for keys that are defined in `core-utils` `query-params` file. However, in some cases (it seems like when a request isn't successful), the `currentQuery` doesn't get updated with new values from the mode selector. This sync issue means that the old value from `currentQuery` replaces the value in the URL.

I found what I think is the simplest way of fixing this, by ensuring that the `currentQuery` is directly updated at the same time that URL is updated by the mode selector. However, it looks a bit weird because the `setQueryParam` function only updates URL search values stored inside `queryParamData`, while the higher level gets put in currentQuery directly. A bit funky.

**PR Checklist:**


lib/components/form/batch-settings.tsx
```diff
@@ -223,6 +225,10 @@ function BatchSettings({
   )
   const accentColor = tinycolor(baseColor).darken(10)
 
+  const _onSettingsUpdate = (params: any) => {
+    setQueryParam({ queryParamData: params, ...params })
```
binh-dam-ibigroup, line 229: Should `setQueryParam` be used for callback `_toggleModeButton` too?

daniel-heppner-ibigroup: We could use it for that! I don't know of any sync bugs affecting that, I think because those keys are not in conflict with the ones defined in query-params.js in core-utils. But I think it wouldn't hurt to use it and would probably make things more resilient. I'll try it out.

binh-dam-ibigroup (approved):
Nice fix. Keep us posted regarding https://github.com/opentripplanner/otp-react-redux/pull/1202#discussion_r1576806026.

lib/components/form/batch-settings.tsx
```diff
@@ -81,6 +82,7 @@ function BatchSettings({
   modeSettingValues,
   onPlanTripClick,
   routingQuery,
+  setQueryParam,
   setUrlSearch,
```
miles-grant-ibigroup, line 86: Do we still need `setUrlSearch`? Is it used somewhere else? Should we use something else instead?

miles-grant-ibigroup (approved):
Working great!

PR #1210, opened 2024-05-13
Render All Walk Legs in a Trip

josh-willis-arcadis:
**Description:**
Always renders walking icons in the trip options. Pulls a new key from the config file called, `showAllWalkLegs`

**PR Checklist:**


lib/components/narrative/metro/metro-itinerary-routes.tsx
```diff
@@ -131,7 +135,8 @@ const MetroItineraryRoutes = ({
 
 // TODO: state type
 const mapStateToProps = (state: any) => ({
-  enableDot: !state.otp.config.itinerary?.disableMetroSeperatorDot
+  enableDot: !state.otp.config.itinerary?.disableMetroSeperatorDot,
+  showAllWalkLegs: state.otp.config.itinerary?.showAllWalkLegs
```
miles-grant-ibigroup, line 139: Make sure to add this new config option to both the `example-config.yml` and the config typescript type! and while we're at it we should use that new type on line 137

miles-grant-ibigroup (changes requested):
Looks great! Thanks for fixing that typo. Just one last step to get it complete

miles-grant-ibigroup (approved):
Thanks for the changes! And for fixing my typo :)

PR #1213, opened 2024-05-16
fix(NavLoginButton): Make help link valid when running locally

binh-dam-ibigroup:
## Description

This PR fixes the Help link under the Account menu so that it points to the intended URL instead of localhost when running OTP-RR locally.

**PR Checklist:**


amy-corson-ibigroup (approved):
I'm confused about why this works but it definitely does!

binh-dam-ibigroup:
> I'm confused about why this works but it definitely does!

Our own `Link` component is intended for internal links and generates an `href` accordingly pointing to `/#/whatever`.

josh-willis-arcadis (approved):
LGTM!

PR #1216, opened 2024-05-16
Refactor route viewer buttons to links

binh-dam-ibigroup:
## Description

This PR refactors the route viewer buttons to display route shapes and pattern viewer into links.

I have added an underline effect when hovering the cursor over the links.
To simplify the code, "deselecting" a route once selected is removed.

## PR Checklist


daniel-heppner-ibigroup (approved):
Nice one!

binh-dam-ibigroup:
> Code looks good, but behavior I have some issue with. The underline effect implies to the user that they're about to be navigated somewhere and while that may be strictly true it doesn't feel that way. Can we fade the background of the row or something? Maybe we can have the arrow appear? But the underline makes things confusing. Especially now that we can't click to de-select! Nothing happens when I click this link that appears as a link

@miles-grant-ibigroup How about I just remove the underline effect?

miles-grant-ibigroup:
> > Code looks good, but behavior I have some issue with. The underline effect implies to the user that they're about to be navigated somewhere and while that may be strictly true it doesn't feel that way. Can we fade the background of the row or something? Maybe we can have the arrow appear? But the underline makes things confusing. Especially now that we can't click to de-select! Nothing happens when I click this link that appears as a link
> 
> @miles-grant-ibigroup How about I just remove the underline effect?

Could we replace it with some kind of background tint shift? I do like the feedback

binh-dam-ibigroup:
> > > Code looks good, but behavior I have some issue with. The underline effect implies to the user that they're about to be navigated somewhere and while that may be strictly true it doesn't feel that way. Can we fade the background of the row or something? Maybe we can have the arrow appear? But the underline makes things confusing. Especially now that we can't click to de-select! Nothing happens when I click this link that appears as a link
> > 
> > 
> > @miles-grant-ibigroup How about I just remove the underline effect?
> 
> Could we replace it with some kind of background tint shift? I do like the feedback

@miles-grant-ibigroup Sure, I can do a background hover effect.

binh-dam-ibigroup:
> > > Code looks good, but behavior I have some issue with. The underline effect implies to the user that they're about to be navigated somewhere and while that may be strictly true it doesn't feel that way. Can we fade the background of the row or something? Maybe we can have the arrow appear? But the underline makes things confusing. Especially now that we can't click to de-select! Nothing happens when I click this link that appears as a link
> > 
> > 
> > @miles-grant-ibigroup How about I just remove the underline effect?
> 
> Could we replace it with some kind of background tint shift? I do like the feedback

@miles-grant-ibigroup How about this: blue outline on hover, so it plays well with the keyboard focus outline? (see screenshot and https://github.com/opentripplanner/otp-react-redux/pull/1216/commits/a9005e311d2be396d4998529e6ec3b91315c142f). I thought about doing background+outline when active and background on hover in https://github.com/opentripplanner/otp-react-redux/pull/1216/commits/566e514d28d4bb5ed1ccafb4fdc53d62dd88b47e, but that messes up the keyboard focus.

image

daniel-heppner-ibigroup:
Your screenshot looks really good @binh-dam-ibigroup. I didn't have a problem with the underline but this feels better!

lib/components/viewers/route-row.tsx
```diff
@@ -37,6 +37,7 @@ export const StyledRouteRow = styled.li`
 // Route Row Button sits invisible on top of the route name and info.
 export const RouteRowLink = styled(Link)`
   align-items: center;
+  border-radius: 4px;
```
miles-grant-ibigroup, line 40: Any chance we can bump this up to maybe 10? Just to be in line with some of the rest of our UI

miles-grant-ibigroup (approved):
Looking good! One last style nitpick and then I think we're good to go

lib/components/viewers/route-row.tsx
binh-dam-ibigroup: Increassed to 10px in https://github.com/opentripplanner/otp-react-redux/pull/1216/commits/ec9b5527f61e24d4c9a952b5ab9d3754e0582d7d, so it is like the Nearby View.

PR #1218, opened 2024-05-21
fix adding NO_TRANSIT_CONNECTION error

amy-corson-ibigroup:
**Description:**
Fix adding the `NO_TRANSIT_CONNECTION` error. Previously it was returning the error in cases where there were plenty of transit itineraries available to users.

**PR Checklist:**


miles-grant-ibigroup:
Easiest fix might be to replicate what we used to have, where the error instead says "no walk+transit itineraries found"

lib/actions/apiV2.js
```diff
@@ -1105,17 +1105,16 @@ export function routingQuery(searchId = null, updateSearchInReducer) {
...
+                )
               if (
                 getActiveItineraries(getState())?.length > 0 &&
                 response?.data?.plan
               ) {
-                response.data.plan.routingErrors =
-                  response.data?.plan?.routingErrors.filter(
-                    (re) => re?.code !== 'NO_TRANSIT_CONNECTION'
-                  )
-
```
miles-grant-ibigroup, line 1118: I'm a bit confused as to why this was moved. I believe this was inside this block to prevent a crash when there was an empty result from the server?

lib/components/narrative/narrative-itineraries.js
```diff
@@ -604,6 +605,16 @@ const mapStateToProps = (state) => {
   // Sort the merged (displayed) itineraries if needed
   sortItinerariesInPlaceIfNeeded(mergedItineraries, state)
 
+  const otp2Errors = activeSearch?.response?.reduce(
+    reduceErrorsFromResponse,
+    {}
+  )
```
miles-grant-ibigroup, line 611: Inline this

lib/util/state.js
```diff
@@ -118,6 +118,18 @@ export const getActiveItineraries = createSelector(
   }
 )
 
+/**
+ * If there are no itineraries, generate an error
+ * so we're not left with an empty results container.
+ */
+export const createNoItinerariesErr = (state, errors, pending) => {
+  const itineraries = getActiveItineraries(state)
+  if (errors && itineraries.length === 0 && !pending) {
```
miles-grant-ibigroup, line 127: Clever fix with the pending! But isn't pending a number? Should it be `== 0` instead just for clarity? I could be wrong here

miles-grant-ibigroup (changes requested):
Great fix! Just a few nitpicks

lib/actions/apiV2.js
amy-corson-ibigroup: This is causing the `NO_TRANSIT_CONNECTION` error in cases where there aren't walk/transit combinations! I think it's just cleaner to fully avoid having the error unless we know we want to. I did some testing for this PR with trip searches that return 0 itineraries and it seemed to be working okay, but let me know if there's something specific I can test!

lib/util/state.js
amy-corson-ibigroup: By the time it gets here, it's a boolean!

lib/actions/apiV2.js
miles-grant-ibigroup: Great! I can't remember anymore. I played with it a bunch and nothing crashed but maybe we should add a comment saying that this might cause weird behavior?

miles-grant-ibigroup (approved):
Off to the races!

josh-willis-arcadis (approved):
good fix! LGTM

PR #1220, opened 2024-05-23
fix(Link): Keep link active on subpaths of the tracked path

binh-dam-ibigroup:
## Description

This PR fixes the `Link` behavior to remain active if on a subpath of the tracked path. Unit tests added for hardening.

To test:
1. Open the route viewer or the trips list
2. Click a route or edit a trip.
3. Note that previously, the route viewer link or the trip link lose their highlight. This should no longer happen with this PR.

## PR Checklist

### DIFF
--- a/__tests__/components/util/link.ts
+++ b/__tests__/components/util/link.ts
@@ -1,0 +1,18 @@
+import '../../test-utils/mock-window-url'
+
+import { isSubpath } from '../../../lib/components/util/link'
+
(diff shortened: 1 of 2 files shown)


miles-grant-ibigroup (approved):
I can think of a few more unit tests to do but this is simple enough I think we're good!

josh-willis-arcadis (approved):
LGTM

PR #1222, opened 2024-05-24
fix(apiV2): Don't perform query/UI changes on invalid mode selection

binh-dam-ibigroup:
## Description

With this PR, `routingQuery` exits early before performing any OTP calls if the mode selection is invalid.
If the UI was on search, it remains there. If UI had previous search results, those results continue to be shown.

**PR Checklist:**

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -12,6 +12,7 @@ jobs:
       - uses: codespell-project/actions-codespell@master
         with:
           check_filenames: true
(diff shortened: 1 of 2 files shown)


amy-corson-ibigroup (approved):
Such a great fix! It would be nice to eventually reset the Ul back to the initial state, so we don't retain any previous successful search results under an invalid search? But I think that can come later

josh-willis-arcadis (approved):
LGTM

PR #1223, opened 2024-05-29
Make Auto-Jumping Optional, Fix it

miles-grant-ibigroup:
**Description:**
Wow it had so many problems. We don't jump when we don't have to anymore, and we make trip-form jumping optional.

**PR Checklist:**


lib/components/app/responsive-webapp.js
```diff
@@ -98,15 +100,17 @@ class ResponsiveWebapp extends Component {
...
-    } else if (query.from && query.to) {
-      map?.fitBounds([query.from, query.to], {
-        duration: 500,
-        padding: getFitBoundsPadding(map, 0.2)
-      })
-    } else if (query.from && !query.to) {
-      setMapCenter(map, query.from)
-    } else if (query.to && !query.from) {
-      setMapCenter(map, query.to)
+    } else if (mainPanelContent === null && autoFly !== false) {
```
amy-corson-ibigroup, line 103: So this would mean `autoFly` is by default turned on, and you'd have to manually set it to `false` in the config? Just confirming that's what we want. 

I think I'd remove the `!== false` and go with the more conservative map movement by default.

lib/components/map/route-preview-overlay.tsx
```diff
@@ -15,24 +15,31 @@ import {
...
+  geometries,
+  mainPanelContent,
+  to,
+  visible
+}: Props) => {
   // Center the map over the endpoints when this overlay is shown.
   const { current: map } = useMap()
   useEffect(() => {
-    if (visible) {
+    if (visible && mainPanelContent === null) {
```
amy-corson-ibigroup, line 36: What does adding this `mainPanelContent` check do?

example-config.yml
```diff
@@ -115,6 +115,7 @@ persistence:
 map:
   initLat: 45.52
   initLon: -122.682
+  # autoFlyOnTripFormUpdate: false
```
amy-corson-ibigroup, line 118: If we could add a description of the default behavior here too, depending on what we land on

amy-corson-ibigroup (approved):
Looking good!! A few questions but nothing blocking

lib/components/app/responsive-webapp.js
miles-grant-ibigroup: Yes default is on. Too many implementers of otp-rr that do not want to see the behavior changed. Better to avoid a breaking change

lib/components/map/route-preview-overlay.tsx
miles-grant-ibigroup: Allows us to move the map in the route viewer and nearby view!

josh-willis-arcadis (approved):
I think this feels much better! Great job!

PR #1224, opened 2024-05-29
don’t show nearby view shadow dot when disabled

miles-grant-ibigroup:
**Description:**
We have this lovely nearby view shadow dot, but some like it disabled. This allows it to be disabled. Disables it by shutting down the location to give the smallest diff. If you have a better idea, please let me know.

**PR Checklist:**


lib/components/map/nearby-view-dot-overlay.tsx
```diff
@@ -33,8 +33,11 @@ const NearbyViewDotOverlay = ({ location }: Props) => {
 
 const mapStateToProps = (state: AppReduxState) => {
   const { highlightedLocation } = state.otp.ui
+  const { nearbyView } = state.otp.config
+
   return {
-    location: highlightedLocation
+    location:
+      nearbyView?.showShadowDotOnMapDrag === false ? null : highlightedLocation
```
daniel-heppner-ibigroup, line 40: cmon now

`nearbyView?.showShadowDotOnMapDrag ? highlightedLocation : null`

daniel-heppner-ibigroup (approved):
Nice! Thank you!

lib/components/map/nearby-view-dot-overlay.tsx
miles-grant-ibigroup: we need it on my default!

amy-corson-ibigroup: If the shadow dot should be on by default, maybe hideShadowDotOnMapDrag instead?

miles-grant-ibigroup: but then you're dealing with a double negative :(

amy-corson-ibigroup: Maybe, but it would tidy up this line, plus I think as a config item the phrasing of `hideShadowDotOnMapDrag` makes it a little more clear how you're impacting the default configuration.

amy-corson-ibigroup (approved):
config variable name isn't blocking, otherwise this looks good!

lib/components/map/nearby-view-dot-overlay.tsx
miles-grant-ibigroup: We've done this strategy many times before and it keeps things much clearer than getting confused with double negatives

PR #1226, opened 2024-05-31
Remove old URL params [OTP-1233]

binh-dam-ibigroup:
## Description

This PR remove some (not all) params that are no longer used in OTP searches from the URL.

## PR Checklist


amy-corson-ibigroup (approved):
Looks good! Thanks for the changes

lib/util/api.ts
```diff
@@ -57,3 +57,17 @@ export function combineQueryParams(
   }
   return qs.stringify(search, { arrayFormat: 'repeat' })
 }
+
+/**
+ * Drops unused params so they don't show up in URL.
+ * TODO: Remove those params from core-utils.
```
daniel-heppner-ibigroup, line 63: Removing these from core-utils would be a big project, because they're still relied on by some other packages which are used in custom uis by some users. I think we should try to remove the dependency on them from OTP-RR instead.

lib/util/api.ts
```diff
@@ -57,3 +57,17 @@ export function combineQueryParams(
...
+ */
+export function removeUnusedQueryParams(params: any): void {
```
daniel-heppner-ibigroup, line 65: make the type `Record<string, any>`

lib/util/api.ts
binh-dam-ibigroup: You are right. The stuff that comes out of `getDefaultQuery` is not very useful anyway:
```
bikeSpeed: 3.58
date: "2024-06-11"
departArrive: "NOW"
endTime: "09:00"
from: null
intermediatePlaces: Array []
maxBikeTime: 20
maxEScooterDistance: 4828
maxWalkTime: 15
mode: "WALK,TRANSIT"
numItineraries: 3
otp2: true
routingType: "ITINERARY"
startTime: "07:00"
time: "10:36"
to: null
watts: 250
wheelchair: false
```

lib/util/api.ts
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/1226/commits/b3f93e4cd9d3ff48df94812947e2084d112e5f95

PR #1232, opened 2024-06-11
Properly respect `numItineraries` config item

miles-grant-ibigroup:
**Description:**
Previously it would always be ignored as core utils always ships a `3`? Curious what I did here, this doesn't feel entirely right

**PR Checklist:**


daniel-heppner-ibigroup:
What if we just delete that property in Binh's new function? https://github.com/opentripplanner/otp-react-redux/pull/1226

miles-grant-ibigroup:
> What if we just delete that property in Binh's new function? #1226

But it is used

daniel-heppner-ibigroup:
Are you sure? I think it only gets set in the config, never by a user action

miles-grant-ibigroup:
> Are you sure? I think it only gets set in the config, never by a user action

Field trip sets it upon click

daniel-heppner-ibigroup (approved):
Field trip strikes again

lib/actions/apiV2.js
```diff
@@ -995,12 +995,12 @@ export function routingQuery(searchId = null, updateSearchInReducer) {
       from: currentQuery.from,
       modes: modes || activeModes,
       modeSettings,
-      numItineraries: numItineraries || config?.modes?.numItineraries || 7,
       time,
       to: currentQuery.to,
       unpreferred,
       // TODO: Does this break everything?
```
amy-corson-ibigroup, line 1001: Should we delete this one year old TODO

amy-corson-ibigroup (approved):
Can't find any issues!

lib/actions/apiV2.js
miles-grant-ibigroup: I'm still not confident

PR #1235, opened 2024-06-13
Replace core-utils getDefaultQuery

binh-dam-ibigroup:
## Description

This PR is a follow up from https://github.com/opentripplanner/otp-react-redux/pull/1226#pullrequestreview-2108759858 from #1226. I replaced `getDefaultQuery` from core-utils with a simpler version. The code to remove unused params is no longer needed.

## PR Checklist


daniel-heppner-ibigroup (approved):
Yay! Thank you. This is a reasonable solution for now.

lib/util/api.ts
```diff
@@ -71,3 +72,17 @@ export function removeUnusedQueryParams(params: Record<string, any>): void {
...
 }
+
+/** Gets the default URL params when reinitializing search */
+export function getDefaultQuery() {
+  return {
+    date: getCurrentDate(),
+    departArrive: 'NOW',
+    intermediatePlaces: [], // required to avoid crash
+    mode: 'WALK,TRANSIT', // obsolete but required to avoid crash
+    numItineraries: 3,
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Apparently it is, see https://github.com/opentripplanner/otp-react-redux/pull/1232#issuecomment-2165761439

lib/actions/api.js
```diff
@@ -114,7 +114,8 @@ export function updateOtpUrlParams(state, searchId) {
   const { config, currentQuery } = state.otp
   // Get updated OTP params from current query.
   const otpParams = getRoutingParams(config, currentQuery, true)
-  removeUnusedQueryParams(otpParams)
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Yes, thank you, done in https://github.com/opentripplanner/otp-react-redux/pull/1235/commits/3daeb37199f25a362ebdfbb1cdda0838b7a90f20.

lib/util/api.ts
```diff
@@ -71,3 +72,17 @@ export function removeUnusedQueryParams(params: Record<string, any>): void {
...
   delete params.maxBikeDistance
 }
+
+/** Gets the default URL params when reinitializing search */
+export function getDefaultQuery() {
+  return {
+    date: getCurrentDate(),
+    departArrive: 'NOW',
+    intermediatePlaces: [], // required to avoid crash
+    mode: 'WALK,TRANSIT', // obsolete but required to avoid crash
```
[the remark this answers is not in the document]
binh-dam-ibigroup: I added a TODO to cover the extra required params in https://github.com/opentripplanner/otp-react-redux/pull/1235/commits/3daeb37199f25a362ebdfbb1cdda0838b7a90f20

lib/util/api.ts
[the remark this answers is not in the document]
binh-dam-ibigroup: I added a comment for this. Currently, especially for transit, we return 3 itineraries. If `numItineraries` is not set, it would return 7 itineraries.

lib/util/api.ts
```diff
@@ -58,16 +59,16 @@ export function combineQueryParams(
...
-  delete params.maxBikeDistance
+/** Gets the default URL params when reinitializing search */
+export function getDefaultQuery() {
+  return {
+    date: getCurrentDate(),
+    departArrive: 'NOW',
+    // TODO: Rework the crash-related params below so we can remove them.
+    intermediatePlaces: [], // required to avoid crash
+    mode: 'WALK,TRANSIT', // obsolete but required to avoid crash
+    numItineraries: 3, // instead of 7 per apiV2 if no limit defined in config
```
miles-grant-ibigroup, line 70: Won't this override the otp-rr default? Is this a required parameter?

miles-grant-ibigroup (changes requested):
Looks much cleaner overall, just want to make sure we're not doing anything that would break https://github.com/opentripplanner/otp-react-redux/pull/1232

lib/util/api.ts
binh-dam-ibigroup: `numItineraries=3` shows up in the URL bar after a new search. If absent, many more results will be shown, especially on transit results.

daniel-heppner-ibigroup:
Can you update the example config with the latest change?

binh-dam-ibigroup:
> Can you update the example config with the latest change?

I didn't think there were any config changes, any one you have in mind?

lib/util/config-types.ts
```diff
@@ -318,6 +318,7 @@ export interface ModesConfig {
   }
   modeButtons?: ModeButtonDefinition[]
   modeSettingDefinitions?: ModeSetting[]
+  numItineraries?: number
```
daniel-heppner-ibigroup, line 321: Isn't this a config change?

binh-dam-ibigroup: Oh, yes, it is an existing parameter that was undocumented (https://github.com/opentripplanner/otp-react-redux/pull/1235/commits/5adc6a57c6bde9d67d8f022e73bdbba2f5a6494b)

lib/actions/apiV2.js
```diff
@@ -1000,7 +1000,7 @@ export function routingQuery(searchId = null, updateSearchInReducer) {
       unpreferred,
       // TODO: Does this break everything?
       ...currentQuery,
-      numItineraries: config?.modes?.numItineraries || numItineraries || 7
+      numItineraries: numItineraries || config?.modes?.numItineraries || 7
```
binh-dam-ibigroup, line 1003: @miles-grant-ibigroup This is to give preference to the URL parameter if it is specified, instead of the config. Is that a conflict with #1232?

lib/actions/api.js
```diff
@@ -114,7 +114,12 @@ export function updateOtpUrlParams(state, searchId) {
...
+
+  // Remove unneeded URL params.
+  delete otpParams.ignoreRealtimeUpdates
+  if (otpParams.numItineraries === getDefaultNumItineraries(config)) {
+    delete otpParams.numItineraries
```
binh-dam-ibigroup, line 121: @miles-grant-ibigroup  Related to https://github.com/opentripplanner/otp-react-redux/pull/1235/files#r1678246983, with this change, `numItineraries` will be not present in the URL by default, so the config setting should still prevail in most cases.

lib/actions/apiV2.js
binh-dam-ibigroup: Also, should we make the default number of itineraries 3 across the board?

lib/actions/api.js
miles-grant-ibigroup: This should be fine! Thanks for checking

lib/actions/apiV2.js
miles-grant-ibigroup: I've checked and this works fine! I think 7 as a default is fine as well

miles-grant-ibigroup (approved):
Working great! Thanks for the thorough research

lib/actions/apiV2.js
daniel-heppner-ibigroup: I am confused why this can't just use the function you wrote to generate this default? A little worried we're going to look at this in the future and be really confused what is happening with `numItineraries`. Why would we want 7 here but a 3 where the URL parameters are generated?

daniel-heppner-ibigroup (commented):
I'm confused why there are two different default values for numItineraries. Why wouldn't we want the URL param default to match the routing query one?

lib/actions/apiV2.js
binh-dam-ibigroup: @miles-grant-ibigroup Was 7 the default for the old batch UI? In the dev branch, it gets overwritten by the `numItineraries=3` param every time, so we could use 3 here too.

miles-grant-ibigroup: I think 7 is the default for the new batch UI too. 3 is an override for when using combinations

binh-dam-ibigroup: Is there a deployment where 7 itineraries returned is the desired outcome with the default settings?

binh-dam-ibigroup: @daniel-heppner-ibigroup Thoughts?

binh-dam-ibigroup: After discussion with @daniel-heppner-ibigroup and further investigation, it turns out `numItineraries` that is passed in `currentQuery` is initialized from the initial state of `state.otp.currentQuery`, which is set upon creation of the otp (and user) reducer, and updated as needed from the `numItineraries` URL param, if set. In theory, `currentQuery.numItineraries` is always defined. So, the default of 7 is in theory never hit. Just to be sure, I'll set the fallback to be consistent with the other cases by using `getDefaultNumItineraries(config)`.

PR #1237, opened 2024-06-13
nearby view new `alwaysShowLongName` setting

miles-grant-ibigroup:
**Description:**
If this is enabled we will show a route long name always, regardless of if there is a duplicate. Only to be enabled in special use cases.

Screenshot 2024-06-13 at 11 46 18 AM


lib/components/viewers/pattern-row.tsx
```diff
@@ -101,6 +103,11 @@ const PatternRow = ({
             />
           </span>
           <span style={{ wordBreak: 'break-word' }} title={pattern.headsign}>
+            {alwaysShowLongName && (
```
amy-corson-ibigroup, line 106:
```suggestion
            {alwaysShowLongName && pattern.route.longName && (
```

Wouldn't this prevent the threat of a crash??

example-config.yml
```diff
@@ -668,4 +668,7 @@ itinerary:
 ### Setting this to true will use the route viewer sort algorithm on routes
 #   useRouteViewSort: false
 ### Setting this to false will hide the shadowey dot that appears when dragging the map
-#   showShadowDotOnMapDrag: false
\ No newline at end of file
+#   showShadowDotOnMapDrag: false
+### In routes, this will always show the routeLongName before the headsign
+### !!! WARNING: in some feeds, this may show the long name twice. In some feeds, this may crash !!!
```
amy-corson-ibigroup, line 673: 😬😬😬 Adding a config item that could cause a crash feels wrong, especially since `route_long_name` is only conditionally required

amy-corson-ibigroup (changes requested):
Thoughts on adding `pattern.route.longName` to the conditional?

lib/components/viewers/pattern-row.tsx
miles-grant-ibigroup: It would. My thinking was that this would discourage people from using the feature when it didn't make sense, but I think I agree with you. Will leave it to the second reviewer to decide?

example-config.yml
miles-grant-ibigroup: Yeah it's not a feature I'm crazy about, which is why I'm trying to make sure that only people who really know what they're doing use it

amy-corson-ibigroup (approved):
Looking good to me!!

lib/components/viewers/pattern-row.tsx
```diff
@@ -101,6 +103,11 @@ const PatternRow = ({
             />
           </span>
           <span style={{ wordBreak: 'break-word' }} title={pattern.headsign}>
+            {alwaysShowLongName && !!pattern.route.longName && (
```
josh-willis-arcadis, line 106: I like the idea of avoiding the crash and having layout issues. Maybe insert a comment here, warning about the potential layout issues to compliment the comment in the config?

josh-willis-arcadis (approved):
Looks good!

lib/components/viewers/pattern-row.tsx
miles-grant-ibigroup: Good idea

PR #1238, opened 2024-06-14
Saved trip days styles

binh-dam-ibigroup:
## Description

This PR updates the styling of the monitored day checkboxes for the saved trip editor to use rounded disks.

Days for which an itinerary presumably doesn't exist are not denoted with an oblique bar instead of just a different colored background. Let me know what you think!

## PR Checklist

| Before | After |
|--------|-------|
| before screenshot  |   after screenshot  |


binh-dam-ibigroup:
> Is this PR meant as a stop-gap between what we have now and that design overhaul?

It is more like a stop gap to make the visuals consistent with the existing monitored days dots on the trip cards. Did you want to go ahead and update the visuals for both?

lib/components/user/monitored-trip/trip-basics-pane.tsx
```diff
@@ -67,53 +71,59 @@ const AvailableDays = styled(FieldSet)`
...
+    border-radius: 3rem;
     box-sizing: border-box;
-    display: inline-block;
-    height: 3em;
-    max-width: 150px;
-    min-width: 14.28%;
+    display: inline-flex;
+    flex-direction: row-reverse;
+    height: 3rem;
+    min-width: 4.5rem;
```
amy-corson-ibigroup, line 80: Can we adjust the width/padding on these to make them more like pills? They're looking a little squished as is.

lib/components/user/monitored-trip/trip-basics-pane.tsx
```diff
@@ -55,6 +56,9 @@ interface State {
 
 // Styles.
 const AvailableDays = styled(FieldSet)`
+  display: flex;
+  gap: 4px;
+
```
amy-corson-ibigroup, line 61: These labels and inputs aren't appearing centered to me, especially when there's a unavailable day 
image

lib/components/user/monitored-trip/trip-monitored-days.tsx
amy-corson-ibigroup, line 1: Thank you for the cleanup!

amy-corson-ibigroup (approved):
I'm definitely comfortable using this as a stop-gap! Long term I'd like to either use the styles in the design doc, or revisit the design to incorporate more of what we're doing here, but in the meantime I like this a lot and I appreciate the change! The save trip form redesign has taken a bit of a backseat for the time being so I think it makes a lot of sense to make these visuals consistent in the meantime.

lib/components/user/monitored-trip/trip-basics-pane.tsx
binh-dam-ibigroup: How do the changes from https://github.com/opentripplanner/otp-react-redux/pull/1238/commits/8d8715f342c67205bb797ff3263227575670b3a1 look like?

lib/components/user/monitored-trip/trip-basics-pane.tsx
binh-dam-ibigroup: I do see that on Firefox but not Chrome... It is actually the text that is slightly offset vertically, the icon and the checkbox alignment are fine.

binh-dam-ibigroup: You see a similar offset with the regular input controls too:
image

lib/components/user/monitored-trip/trip-basics-pane.tsx
amy-corson-ibigroup: Looks good thank you! I think the margin could be adjusted to something like `0 7px 0 2px` to look a little more even?

binh-dam-ibigroup: Adjusted in https://github.com/opentripplanner/otp-react-redux/pull/1238/commits/7395d62977c3509feb45137950acbb25fe5c1cdd

josh-willis-arcadis (approved):
Looks good!

PR #1231, opened 2024-06-10
Fix overlapping patterns

binh-dam-ibigroup:
## Description

This PR fixes pattern extraction logic where partially overlapping patterns were missed.

## PR Checklist

- [ ] Does the code follow accessibility standards (WCAG 2.1 AA Compliant)?
- [ ] Are all languages supported (Internationalization/Localization)?
- [x] Are appropriate Typescript types implemented?

PR #1241, opened 2024-06-24
refactor: remove util/dropdown. import new building-block dropdown

josh-willis-arcadis:
**Description:**
Remove util/dropdown and use the new building-blocks component.
blocked by: https://github.com/opentripplanner/otp-ui/pull/753

**PR Checklist:**


lib/components/narrative/narrative-itineraries-header.tsx
```diff
@@ -206,7 +206,7 @@ export default function NarrativeItinerariesHeader({
                 )}
               </StyledIconWrapper>
             </button>
-            <SortResultsDropdown
+            <Dropdown
               id="sort-results"
               label={sortResultsLabel}
               name={sortText}
```
amy-corson-ibigroup, line 212: Pretty sure there's no `name` prop on the new `Dropdown` so this is causing an error!

lib/components/user/nav-login-button.tsx
```diff
@@ -64,6 +64,7 @@ const NavLoginButton = ({
     return (
       <li>
         <Dropdown
+          className="navBarItem"
           id="user-selector"
           label={intl.formatMessage({ id: 'components.SubNav.userMenu' })}
           name={
```
amy-corson-ibigroup, line 70: Same here, invalid prop with "name"!

lib/components/app/locale-selector.tsx
```diff
@@ -1,10 +1,10 @@
 import { connect } from 'react-redux'
+import { Dropdown } from '@opentripplanner/building-blocks/lib/dropdown'
```
amy-corson-ibigroup, line 2: this should be just `@opentripplanner/building-blocks`! Same in other files

amy-corson-ibigroup (changes requested):
1) When I update the packages, I'm not able to get the app running. I think it's the import path for the dropdown.
2) Double check some of the props, they're not the same in the new version as the old! Props like "name" and "pullRight" aren't supported anymore
3) I've added the commit updating the building blocks package!

lib/components/viewers/route-details.tsx
amy-corson-ibigroup, line 1: Hmm seems like there was a bit of a weird merge and we're getting a lot of pattern deduplication logic I believe would revert some of #1231?

lib/components/user/nav-login-button.tsx
```diff
@@ -75,7 +76,6 @@ const NavLoginButton = ({
               />
             </span>
           }
-          pullRight
         >
           <li className="header">{displayedName}</li>
```
amy-corson-ibigroup, line 80: When you're adding the `<ul>` to this `Dropdown,`  I would make this one a more traditional header (not an `li`) and insert the `<ul>` underneath! This technically is not a part of the list, it's labeling the list.

lib/components/user/nav-login-button.css
```diff
@@ -25,6 +25,11 @@
   background-color: #080808;
 }
 
+li.header {
```
amy-corson-ibigroup, line 28: Flagging that if we make this into an actual header, this css will have to change

amy-corson-ibigroup (changes requested):
Because the `Dropdown` is no longer inherently a `ul`, we will need to add a `ul` in the children of some of these components! Right now, there are some orphan`li`s.

Also, looks like there may have been a bad merge conflict that resulted in some pattern deduplication logic getting re-added.

lib/components/narrative/narrative-itineraries-header.tsx
```diff
@@ -206,10 +206,10 @@ export default function NarrativeItinerariesHeader({
...
             </button>
-            <SortResultsDropdown
+            <Dropdown
               id="sort-results"
               label={sortResultsLabel}
-              name={sortText}
+              text={sortText}
               title={sortResultsLabel}
             >
               {sortOptionsArr.map((sortOption) => (
```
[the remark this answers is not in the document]
josh-willis-arcadis: The ui component will be rendered as a `<ul>` if the component has `<li>` children, so these will not be orphaned.

lib/components/viewers/route-details.tsx
josh-willis-arcadis: I think I've got the changes added back. I def messed up that file when trying to resolve conflicts.

lib/components/user/nav-login-button.tsx
```diff
@@ -75,51 +76,51 @@ const NavLoginButton = ({
...
             </span>
           }
-          pullRight
         >
-          <li className="header">{displayedName}</li>
-
-          {links &&
-            links.map((link, i) => {
-              if (link.url.startsWith('http')) {
+          <h4>{displayedName}</h4>
```
amy-corson-ibigroup, line 80: I know this was in response to my comment, but maybe we should make this a `<p>` instead? Since this `h4` is in the nav bar, it's showing up semantically before the `<h1>` which is going to be confusing for screenreaders. I think a `p` should be fine in this context!

lib/components/viewers/route-details.tsx
```diff
@@ -149,12 +149,12 @@ class RouteDetails extends Component<Props> {
             <HeadsignSelectLabel htmlFor="headsign-selector-label">
               <FormattedMessage id="components.RouteDetails.stopsTo" />
             </HeadsignSelectLabel>
-            <SortResultsDropdown
+            <Dropdown
```
amy-corson-ibigroup, line 152: This is inheriting the text color of the page, which is white! 
image

lib/components/user/nav-login-button.tsx
```diff
@@ -75,51 +76,51 @@ const NavLoginButton = ({
...
-            <UnstyledButton onClick={onSignOutClick}>
-              <FormattedMessage id="components.NavLoginButton.signOut" />
-            </UnstyledButton>
-          </li>
+            <li>
+              <UnstyledButton onClick={onSignOutClick}>
+                <FormattedMessage id="components.NavLoginButton.signOut" />
+              </UnstyledButton>
+            </li>
+          </ul>{' '}
```
amy-corson-ibigroup, line 123: I think we can remove this `{' '}`

lib/components/narrative/narrative-itineraries-header.tsx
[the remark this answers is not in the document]
amy-corson-ibigroup: Just saw this in the otp-ui branch - VERY cool!!

amy-corson-ibigroup (changes requested):
This is looking REALLY nice! Just a few small cleanup items and I think it should be good to go!

lib/components/narrative/narrative-itineraries-header.tsx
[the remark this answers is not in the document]
daniel-heppner-ibigroup: 100x developer

daniel-heppner-ibigroup:
Uh oh, this breaks the percy tests because it changed one of the buttons that it tries to click on. Let me know if you need help fixing this! @josh-willis-arcadis

lib/components/viewers/route-details.tsx
josh-willis-arcadis: Good catch!

josh-willis-arcadis:
@daniel-heppner-ibigroup looks like percy tests are failing because the submode selector has been moved. I think we should remove the tests for now, and then add more for the advanced panel.

PR #1244, opened 2024-07-02
Add Plan Trip Button to Advanced Mode Settings Page

amy-corson-ibigroup:
**Description:**
- Adds 'back' and "plan trip" buttons to the advanced trip form
- Moves around some of the planQuery logic to allow it to be accessed by the advanced trip form
- Adds a toast notification when the advanced settings close, indicating to the user that the settings have been saved

image

**PR Checklist:**

### DIFF
--- a/lib/components/form/advanced-settings-panel.tsx
+++ b/lib/components/form/advanced-settings-panel.tsx
@@ -82,7 +82,7 @@ const PlanTripButton = styled.button`
   gap: 0.5em;
   background-color: var(--main-base-color, ${blue[900]});
   border: 0;
-  width: 45%;
+  width: 47%;
   height: 51px;
   color: white;
   font-weight: 700;
--- a/lib/components/form/advanced-settings-panel.tsx
+++ b/lib/components/form/advanced-settings-panel.tsx
@@ -24,9 +24,6 @@ import { AppReduxState } from '../../util/state-types'
 import { blue, getBaseColor } from '../util/colors'
 import { ComponentContext } from '../../util/contexts'
 import { generateModeSettingValues } from '../../util/api'
-
-import { setModeButtonEnabled } from './batch-settings'
-import { styledCheckboxCss } from './styled'
 import PageTitle from '../util/page-title'
 
 import {
@@ -38,6 +35,8 @@ import {
   populateSettingWithIcon,
   setModeButton
 } from './util'
+import { setModeButtonEnabled } from './batch-settings'
+import { styledCheckboxCss } from './styled'
 
 const PanelOverlay = styled.div`
   height: 100%;
--- a/lib/components/form/advanced-settings-panel.tsx
+++ b/lib/components/form/advanced-settings-panel.tsx
@@ -7,21 +7,22 @@ import { Close } from '@styled-icons/fa-solid'
 import { connect } from 'react-redux'
 import { decodeQueryParams, DelimitedArrayParam } from 'serialize-query-params'
 import { FormattedMessage, useIntl } from 'react-intl'
-import { Search } from '@styled-icons/fa-solid/Search'
-
-import { generateModeSettingValues } from '../../util/api'
 import {
   ModeButtonDefinition,
   ModeSetting,
   ModeSettingValues
 } from '@opentripplanner/types'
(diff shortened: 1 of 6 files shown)


lib/components/form/advanced-settings-panel.tsx
```diff
@@ -28,20 +20,30 @@ import PageTitle from '../util/page-title'
...
 } from './util'
+import { AppReduxState } from '../../util/state-types'
 import { setModeButtonEnabled } from './batch-settings'
 
+import { blue, getBaseColor } from '../util/colors'
+import { ComponentContext } from '../../util/contexts'
+
 import styled from 'styled-components'
 
+const baseColor = getBaseColor()
```
josh-willis-arcadis, line 43: not working for me. potential timing issues with when the function is called and css is loaded?

lib/components/form/util.tsx
```diff
@@ -69,3 +72,45 @@ export const setModeButton =
       encodeQueryParams(modesQueryParamConfig, { modeButtons: newButtons })
     )
   }
+
+export const alertUserTripPlan = (
+  intl,
+  currentQuery,
+  onPlanTripClick,
+  routingQuery
```
josh-willis-arcadis, line 80: getting eslint warnings and errors for no return type on function and implicit 'any' types.

lib/components/mobile/batch-search-screen.tsx
```diff
@@ -21,6 +21,8 @@ import SwitchButton from '../form/switch-button'
 import MobileContainer from './container'
 import MobileNavigationBar from './navigation-bar'
 
+import toast from 'react-hot-toast'
```
josh-willis-arcadis, line 24: sort imports

josh-willis-arcadis (changes requested):
Looking good. Not sure what is going on with getBaseColor yet.

daniel-heppner-ibigroup (changes requested):
Buttons look good, functionality good, just improve the types please!

daniel-heppner-ibigroup (approved):
I fixed both my and Josh's comments, @josh-willis-arcadis please look closely at my comments too to make sure they're actually fixed, since I was the one who fixed them I'd like to make sure they get a second look.

lib/components/form/advanced-settings-panel.tsx
daniel-heppner-ibigroup: You were right, the CSS variable isn't loaded yet when this got called. I moved it around and used the vanilla CSS variable in the styled components, no need to take a CSS variable, make it a JS variable, just to put it back in the CSS!

lib/components/form/advanced-settings-panel.tsx
```diff
@@ -79,26 +73,57 @@ const HeaderContainer = styled.div`
...
-const accentColor = baseColor || blue[900]
+
+const PlanTripButton = styled.button`
+  align-items: center;
+  display: flex;
+  justify-content: center;
+  gap: 0.5em;
+  background-color: var(--main-base-color, ${blue[900]});
+  border: 0;
+  width: 45%;
```
josh-willis-arcadis, line 84: super nit picky but from widths 992px - 1058px the trip text gets pushed to a new line. change the width to like 47% to fit the text on a single line.

josh-willis-arcadis (approved):
The changes look great! One thing I noticed is that if you uncheck every mode and then click "Plan Trip," you get an error message, followed by the "Preferences Saved" toast, and the advanced panel closes. If there is an error, I believe we shouldn't close the panel or display the toast.

lib/components/form/advanced-settings-panel.tsx
amy-corson-ibigroup: Really clean, thank you!!

amy-corson-ibigroup:
@josh-willis-arcadis  really good point about the toast. I went ahead and adjusted when the toast displays - now it will only happen when the user closes the panel and returns to the home page, and not when they click "plan trip"

amy-corson-ibigroup:
@josh-willis-arcadis @daniel-heppner-ibigroup thank you both so much for the help on this one!

PR #1248, opened 2024-07-11
Add Date Time selector and remove subsettings in mode selector 

amy-corson-ibigroup:
**Description:**
- Adds the `DateTimeModal` to the advanced trip panel and wraps it in styling
- Removes the logic that generates mode subsettings in the mode selector

**PR Checklist:**

### DIFF
--- a/lib/components/form/advanced-settings-panel.tsx
+++ b/lib/components/form/advanced-settings-panel.tsx
@@ -38,6 +38,7 @@ import {
 } from './util'
 import { setModeButtonEnabled } from './batch-settings'
 import { styledCheckboxCss } from './styled'
+import DateTimeModal from './date-time-modal'
 
 const PanelOverlay = styled.div`
   height: 100%;
@@ -68,7 +69,6 @@ const HeaderContainer = styled.div`
   align-items: center;
   display: flex;
   justify-content: space-between;
-  margin-bottom: 2em;
 `
 
 const Subheader = styled.h2`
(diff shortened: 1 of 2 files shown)


lib/components/form/advanced-settings-panel.tsx
```diff
@@ -99,6 +99,24 @@ const ButtonContainer = styled.div`
   margin-top: 2em;
 `
 
+const DtSelectorContainer = styled.div`
+  padding: 1em;
```
josh-willis-arcadis, line 103: I think we should add some left right padding in here somewhere for the buttons. The text is squished at widths 992-1150ish px
image

josh-willis-arcadis (changes requested):
The CSS needs a little cleanup with the buttons and then I think it will be good to go!

lib/components/form/advanced-settings-panel.tsx
amy-corson-ibigroup: Good call! I removed the padding on these actually since the text is centered, now when the buttons resize it should be way more graceful. Fixed in [d4215c5](https://github.com/opentripplanner/otp-react-redux/pull/1248/commits/d4215c58eff30662b2fd78761ec1b3f69bb221fa)

josh-willis-arcadis (approved):
Perfect! Thank you!

daniel-heppner-ibigroup (approved):
Nice, looks good to me

PR #1249, opened 2024-07-12
Advanced Trip Form

amy-corson-ibigroup:
**Description:**

Moves mode subsettings into "Advanced" trip panel. Requires https://github.com/opentripplanner/otp-ui/pull/749

**PR Checklist:**


lib/components/form/advanced-settings-panel.tsx
```diff
@@ -0,0 +1,288 @@
...
+  const { ModeIcon } = useContext(ComponentContext)
+
+  const planTrip = () => {
+    alertUserTripPlan(intl, currentQuery, onPlanTripClick, routingQuery)
+    closeAdvancedSettings()
+  }
+
+  const closeAndAnnounce = () => {
+    closeAdvancedSettings()
+    toast.success(intl.formatMessage({ id: 'actions.user.preferencesSaved' }))
```
miles-grant-ibigroup, line 168: The toast is very attention-grabbing and could be avoided by clearer design

lib/components/form/advanced-settings-panel.tsx
```diff
@@ -0,0 +1,288 @@
...
+        fillModeIcons
+        label="test"
+        modeButtons={processedModeButtons}
+        onSettingsUpdate={onSettingsUpdate(setQueryParam)}
+        onToggleModeButton={setModeButton(
+          enabledModeButtons,
+          onSettingsUpdate(setQueryParam)
+        )}
+      />
+      <ButtonContainer>
```
miles-grant-ibigroup, line 242: Two buttons at the bottom is confusing. Should be one save button for clarity

lib/components/form/advanced-settings-panel.tsx
```diff
@@ -0,0 +1,288 @@
...
+      setModeButtonEnabled(enabledModeButtons)
+    )
+  )
+
+  return (
+    <PanelOverlay className="advanced-settings" ref={innerRef}>
+      <HeaderContainer>
+        <PageTitle title={headerText} />
+        <h1 className="header-text">{headerText}</h1>
+        <CloseButton
```
miles-grant-ibigroup, line 207: There is currently no way to cancel my changes. This button could be repurposed as a "cancel" button. This then also removes the need for a toast. Cancel/Save makes it clear to the user what is going on.

lib/components/form/advanced-settings-panel.tsx
```diff
@@ -0,0 +1,288 @@
...
+          aria-label={closeButtonText}
+          onClick={closeAndAnnounce}
+          title={closeButtonText}
+        >
+          <Close size={22} />
+        </CloseButton>
+      </HeaderContainer>
+      <DtSelectorContainer>
+        <DateTimeModal />
```
miles-grant-ibigroup, line 216: Is it possible to remove the blue background? Very confusing for the date time selector to be separated from other content by background color when all other seconds are demarcated by header text

lib/components/form/styled.ts
```diff
@@ -230,3 +231,125 @@ export const StyledLocationField = styled(LocationField)`
...
+    ${wipeInMixin}
+  }
+
+  .${mainPanelClassName}-exit-active {
+    ${wipeOutMixin(-wipeOffset)}
+    ${transitionMixin}
+  }
+`
+
+const toggleTransition = css`
```
miles-grant-ibigroup, line 290: I am failing to see the utility in these toggles instead of a standard checkbox

lib/components/form/styled.ts
```diff
@@ -230,3 +231,125 @@ export const StyledLocationField = styled(LocationField)`
...
+export const mainPanelClassName = 'main-panel'
+export const transitionDuration = prefersReducedMotion ? 0 : 175
+
+const wipeOffset = 7
+
+const transitionMixin = css`
+  transition: all ${transitionDuration}ms ease-in;
+`
+
+const wipeOutMixin = (offset: number) => css`
```
miles-grant-ibigroup, line 245: Instead of this confusing wipe transition that provides no visual hierarchy, should we consider having the advanced settings panel appear next to the existing panel? Then a slide animation would make more sense, and there would be a clear visual hierarchy

miles-grant-ibigroup (changes requested):
Haven't reviewed code yet, but have some suggestions about behavior

lib/components/form/advanced-settings-panel.tsx
amy-corson-ibigroup: I'm open to alternatives! There needs to be some sort of user feedback that the changes have been applied when you return to the home page, and this toast is in line with how we show that feedback in other places across the app.

lib/components/form/advanced-settings-panel.tsx
amy-corson-ibigroup: Why is it confusing if both of the buttons communicate what they do? It seems like from this page there are two options on how to proceed (returning to trip form, or pulling the trigger on plan trip) and having them both easily available makes that clear

lib/components/form/advanced-settings-panel.tsx
amy-corson-ibigroup: 1. Even if we added a "save" button, there would still need to be some sort of user feedback to indicate that the save was successful. 2. Adding a "cancel" button would require creating an interim state that we can then discard on cancel and I don't think it's necessary? Every setting here is easily reversible and if feels like less user friction for the save to be implicit.

lib/components/form/advanced-settings-panel.tsx
amy-corson-ibigroup: Just to clarify, the header text in most deployments will not be visible to sighted users. The header text you're seeing is a custom configuration. It will be hidden by default. 

The blue I'm happy to nix if that's what folks want, although I think it looks nice and makes the DT modal feel like more of its own tool rather than three different buttons? I added it because the upcoming design utilizes the blue background and this felt like a way to bridge the gap. Curious what other folks think.

lib/components/form/styled.ts
amy-corson-ibigroup: Can you provide an example of what you mean here? I'm having trouble visualizing

lib/components/form/advanced-settings-panel.tsx
miles-grant-ibigroup: The animation can be used as a form of feedback. See https://bootcamp.uxdesign.cc/toast-or-no-toast-tips-on-providing-feedback-in-product-design-210d24a7e547 or https://www.nngroup.com/articles/ten-usability-heuristics/ #1, #8 and #4. There are few products that have a toast confirmation when saving a simple preferences pane

lib/components/form/advanced-settings-panel.tsx
miles-grant-ibigroup: It's a second location for the plan trip button. That's confusing and lacks internal consistency https://www.nngroup.com/articles/ten-usability-heuristics/ #4. Moving the pane next to the other pane should solve this problem even better.

lib/components/form/advanced-settings-panel.tsx
miles-grant-ibigroup: The pane disappearing when you click save is a much cleaner way to provide that feedback. I agree a cancel button is work, but if we want to have a secondary pane like this then I think it's important. See https://www.nngroup.com/articles/ten-usability-heuristics/ #3, #5, #9

lib/components/form/advanced-settings-panel.tsx
miles-grant-ibigroup: Internal consistency https://www.nngroup.com/articles/ten-usability-heuristics/ #8, #6, #4

lib/components/form/styled.ts
miles-grant-ibigroup: Screenshot 2024-07-19 at 12 31 58 PM
I think this change alone would fix most of the usability problems here!

miles-grant-ibigroup:
I want to make clear that all the changes I'm talking about don't have to be the default! But I'd like to see it be made configurable

lib/components/mobile/batch-search-screen.tsx
```diff
@@ -14,6 +23,20 @@ import MobileNavigationBar from './navigation-bar'
 
 const { SET_FROM_LOCATION, SET_TO_LOCATION } = MobileScreens
 
+const MobileSearchSettings = styled.div<{ advancedPanelOpen: boolean }>`
+  background: white;
```
miles-grant-ibigroup, line 27: Screenshot 2024-08-13 at 11 27 45 AM
Seeing a weird gray strip when setting the sidebar background color to a color other than white. My guess is this line is the culprit

lib/components/app/batch-routing-panel.tsx
```diff
@@ -33,6 +61,14 @@ class BatchRoutingPanel extends Component<Props> {
     this.setState({ planTripClicked: true })
   }
 
+  handleOpenAdvanceSettings = () => {
+    this.setState({ showAdvancedModeSettings: true })
+  }
+
+  handleCloseAdvanceSettings = () => {
+    this.setState({ showAdvancedModeSettings: false })
+  }
```
miles-grant-ibigroup, line 70: nitpick: refactor this to be a single method that takes true/false as an argument?

lib/components/app/batch-routing-panel.tsx
```diff
@@ -53,55 +91,113 @@ class BatchRoutingPanel extends Component<Props> {
...
-        >
-          <span className="batch-routing-panel-location-fields">
-            <LocationField
-              inputPlaceholder={intl.formatMessage(
-                { id: 'common.searchForms.enterStartLocation' },
-                { mapAction }
+        <TransitionStyles>
+          {!this.state.showAdvancedModeSettings && (
+            <InvisibleA11yLabel>
+              <h1>
```
miles-grant-ibigroup, line 97: Why is this `h1` if it's invisible?

lib/components/form/styled.ts
```diff
@@ -230,3 +231,78 @@ export const StyledLocationField = styled(LocationField)`
     }
   }
 `
+
+export const advancedPanelClassName = 'advanced-panel'
+export const mainPanelClassName = 'main-panel'
+export const transitionDuration = prefersReducedMotion ? 0 : 175
+export const transitionDelay = prefersReducedMotion ? 0 : 300
```
miles-grant-ibigroup, line 238: Can we tighten this up? 150ms-200ms total animation time feels snappier to me. The delay I'm guessing is for the save button? That feels perfect

lib/components/form/styled.ts
```diff
@@ -230,3 +231,78 @@ export const StyledLocationField = styled(LocationField)`
...
+
+const wipeOffset = 7
+
+const transitionMixin = css`
+  transition: all ${transitionDuration}ms ease-in;
```
miles-grant-ibigroup, line 243: Thoughts on a curve like `cubic-bezier(0.95, 0.21, 0.76, 1.25)`?

miles-grant-ibigroup (changes requested):
I know we've had a lot of back and forth on this but I think the final result is quite good. Definitely more work to do in the future but this is a solid foundation. Just a few small final checks and then I think this is entirely ready to go!

lib/components/mobile/batch-search-screen.tsx
amy-corson-ibigroup: Do you know specifically what panel this is? I've tried resetting the background on a few different sidebar layers and im not seeing this!

lib/components/app/batch-routing-panel.tsx
amy-corson-ibigroup: Good catch, I've gone one step further and eliminated the methods entirely

lib/components/app/batch-routing-panel.tsx
amy-corson-ibigroup: It's for assistive technology navigation! We do the same thing on the home page, there's a hidden "plan trip" `h1`

lib/components/form/styled.ts
amy-corson-ibigroup: The transition animation is only `175`, which is right in that window! And yes, the 300 is for the delay

lib/components/form/styled.ts
amy-corson-ibigroup: Oooh this might need some tweaking, it's shuddering a little at the end of the curve.

lib/components/form/advanced-settings-panel.tsx
```diff
@@ -0,0 +1,274 @@
...
+      addModeButtonIcon(ModeIcon),
+      addSettingsToButton(processedModeSettings),
+      setModeButtonEnabled(enabledModeButtons)
+    )
+  )
+
+  return (
+    <PanelOverlay className="advanced-settings" ref={innerRef}>
+      <HeaderContainer>
+        <PageTitle title={headerText} />
```
miles-grant-ibigroup, line 182: This doesn't seem to be working. It flashes advanced preferences and then goes back to the default. We can probably remove this feature?

lib/components/form/styled.ts
miles-grant-ibigroup: Are you ok with `100`? I've tweaked it and it feels much snappier

lib/components/form/styled.ts
miles-grant-ibigroup: With a higher speed I'm not seeing the shuddering. I'd be ok with `ease-in-out` as well but I feel the overshoot adds a nice human element

amy-corson-ibigroup: Changed it to `ease-in-out` in [8496d37](https://github.com/opentripplanner/otp-react-redux/pull/1249/commits/8496d37e5bc10ed911ba5e7576d91139ef2a9aeb)

lib/components/form/styled.ts
amy-corson-ibigroup: `100` for a full panel moving in and out is really fast. [`175` is already aggressive](https://www.nngroup.com/articles/animation-duration/#:~:text=When%20animation%20involves%20substantial%20screen%20changes%2C%20such%20as%20when%20a%20modal%20window%20moves%20into%20view%2C%20a%20duration%20of%20200%E2%80%93300%20ms%20can%20be%20appropriate.) for how much content is transitioning.

lib/components/form/advanced-settings-panel.tsx
miles-grant-ibigroup: Was seeing this on Firefox but if it's working for you let's hope it's a fluke!

miles-grant-ibigroup (approved):
We're good to go!

lib/components/form/advanced-settings-panel.tsx
amy-corson-ibigroup: @binh-dam-ibigroup Let me know if you notice this!

lib/components/app/batch-routing-panel.tsx
```diff
@@ -53,55 +83,120 @@ class BatchRoutingPanel extends Component<Props> {
...
+                >
+                  <AdvancedSettingsPanel
+                    closeAdvancedSettings={
+                      () => this.setState({ showAdvancedModeSettings: false })
+                      // eslint-disable-next-line react/jsx-curly-newline
+                    }
+                    innerRef={this._advancedSettingRef}
+                    onPlanTripClick={this.handlePlanTripClick}
+                  />
+                </CSSTransition>
```
binh-dam-ibigroup, line 117: Can this block be reused with its analog in batch-search-screen?

binh-dam-ibigroup (changes requested):
Lots of things to like, and some cleanup needed as well as indicated.

lib/components/form/advanced-settings-panel.tsx
```diff
@@ -0,0 +1,274 @@
...
+          </GlobalSettingsContainer>
+        </>
+      )}
+      <Subheader>
+        <FormattedMessage id="components.BatchSearchScreen.modeOptions" />
+      </Subheader>
+      <AdvancedModeSubsettingsContainer
+        accentColor={accentColor}
+        fillModeIcons
+        label="test"
```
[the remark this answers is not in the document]
amy-corson-ibigroup: Wow, incredible catch! But yes, the `AdvancedModeSubsettingsContainer` works exactly how you described, the label prop functions as a `legend` within a `fieldset`.

lib/components/form/styled.ts
```diff
@@ -230,3 +231,78 @@ export const StyledLocationField = styled(LocationField)`
...
+  transform: translateX(${offset}px);
+  opacity: 0;
+`
+const wipeInMixin = css`
+  opacity: 1;
+`
+
+export const TransitionStyles = styled.div`
+  display: contents;
+  .${advancedPanelClassName}-enter {
```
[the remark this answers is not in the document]
amy-corson-ibigroup: Nope, this library just applies class names as the state changes! http://reactcommunity.org/react-transition-group/css-transition

lib/components/mobile/batch-search-screen.tsx
```diff
@@ -44,32 +71,77 @@ class BatchSearchScreen extends Component<Props> {
...
-          </div>
+          <MobileSearchSettings
+            advancedPanelOpen={showAdvancedModeSettings}
+            className={`batch-search-settings mobile-padding ${
+              showAdvancedModeSettings && 'advanced-mode-open'
+            }`}
+          >
+            <TransitionStyles>
+              <TransitionGroup style={{ display: 'content' }}>
+                {!showAdvancedModeSettings && (
```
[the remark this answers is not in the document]
amy-corson-ibigroup: Using ternaries with these `TransitionGroup` blocks is cancelling out the animation, unforunately

lib/components/form/util.tsx
```diff
@@ -0,0 +1,116 @@
+import { DelimitedArrayParam, encodeQueryParams } from 'serialize-query-params'
```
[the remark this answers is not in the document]
daniel-heppner-ibigroup: We're using this library to ensure consistent encoding and decoding of the URL parameters. Maybe we don't need it as much as we used to though, so we could consider writing our own helpers for URL state handling. But let's handle that in a future PR.

lib/components/form/advanced-settings-panel.tsx
amy-corson-ibigroup: I've removed the `PageTitle` component from Advanced Settings but I can look into this further if we feel like it's important to update the website title when this panel is open.

lib/components/form/util.tsx
```diff
@@ -0,0 +1,116 @@
...
+    icon: <ModeIcon mode={msd.iconName} width={16} />
+  })
+
+export const addModeButtonIcon =
+  (ModeIcon: React.ComponentType<{ mode?: string; width?: number }>) =>
+  (def: ModeButtonDefinition): ModeButtonDefinition => ({
+    ...def,
+    Icon: function ModeButtonIcon() {
+      return <ModeIcon mode={def.iconName} />
+    }
```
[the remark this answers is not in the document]
amy-corson-ibigroup: Can you elaborate on what this should look like? This function does return a different structure than `populateSettingWithIcon` so I'm not immediately seeing what you mean

lib/components/form/util.tsx
```diff
@@ -0,0 +1,116 @@
...
+      }
+    }
+    return msd
+  }
+
+/**
+ * Stores parameters in both the Redux `currentQuery` and URL
+ * @param params Params to store
+ */
+export const onSettingsUpdate =
```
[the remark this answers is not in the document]
amy-corson-ibigroup: Daniel updated the types and cleaned this up a little in https://github.com/opentripplanner/otp-react-redux/pull/1249/commits/ddda7fbd52294f4f1cba50ed0d0cdebebd92ae38

lib/components/form/advanced-settings-panel.tsx
binh-dam-ibigroup: A back button is in place now, so this is resolved.

lib/components/form/advanced-settings-panel.tsx
```diff
@@ -0,0 +1,282 @@
...
+        )}
+      />
+      {saveAndReturnButton && (
+        <ReturnToTripPlanButton
+          className="save-settings-button"
+          onClick={async () => {
+            await setCloseAdvancedSettingsWithDelay()
+            setClosingBySave(true)
+            closeAdvancedSettings()
+          }}
```
binh-dam-ibigroup: Maybe `useCallback` and use that here.

lib/components/viewers/viewers.css
```diff
@@ -60,12 +60,14 @@
 }
 
 .otp .trip-viewer .header-text,
-.otp .route-viewer .header-text {
+.otp .route-viewer .header-text,
+.otp .advanced-settings .header-text {
   display: contents;
```
binh-dam-ibigroup: Nit: On Firefox, `display: contents` is not playing well with the vertical alignment of the "Advanced Settings" heading on desktop view (see screenshot). Vertical alignment seems fine on Chrome and Safari.

image

lib/components/mobile/batch-search-screen.tsx
```diff
@@ -22,20 +45,28 @@ interface Props {
 
 class BatchSearchScreen extends Component<Props> {
   state = {
-    planTripClicked: false
+    closeAdvancedSettingsWithDelay: false,
+    planTripClicked: false,
```
binh-dam-ibigroup, line 54: Did I get confused with a mockup? I thought the Advanced Trip Form took the entire screen on mobile view. Right now, only a tiny height is available (the height of the origin/destination and mode selector fields), so it is very awkward to use on mobile view. (see screenshot)

image

lib/components/mobile/batch-search-screen.tsx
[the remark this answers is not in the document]
binh-dam-ibigroup: Add a comment in the code reminding of that, please.

lib/components/mobile/batch-search-screen.tsx
```diff
@@ -44,32 +75,86 @@ class BatchSearchScreen extends Component<Props> {
...
+                      />
+                      <div className="switch-button-container-mobile">
+                        <SwitchButton />
+                      </div>
+                      <BatchSettings
+                        onPlanTripClick={this.handlePlanTripClick}
+                        openAdvancedSettings={
+                          () =>
+                            this.setState({ showAdvancedModeSettings: true })
+                          // eslint-disable-next-line react/jsx-curly-newline
```
binh-dam-ibigroup: Might need to `useCallback` for these, so you can also remove the eslint exception. (multiple instances)

lib/components/form/util.tsx
```diff
@@ -0,0 +1,118 @@
...
+  (setQueryParam: QueryParamChangeHandler) =>
+  (params: any): void => {
+    setQueryParam({ queryParamData: params, ...params })
+  }
+
+export const setModeButton =
+  (enabledModeButtons: string[], updateHandler: (params: any) => void) =>
+  (buttonId: string, newState: boolean): void => {
+    let newButtons
```
binh-dam-ibigroup: Write `const newButtons = newState ? ...`

lib/components/form/util.tsx
```diff
@@ -0,0 +1,118 @@
...
+    ...def,
+    Icon: function ModeButtonIcon() {
+      return <ModeIcon mode={def.iconName} />
+    }
+  })
+
+export const addCustomSettingLabels =
+  (intl: IntlShape) =>
+  (modeSetting: ModeSetting): ModeSetting => {
+    let modeLabel
```
binh-dam-ibigroup: Move `modeLabel` inside the if block, and change to `const`.

binh-dam-ibigroup (commented):
Was the mobile view of Advanced Settings supposed to be tiny?

lib/components/mobile/batch-search-screen.tsx
amy-corson-ibigroup: Weird, yes, it should expand to full screen when opened. Where are you seeing this? Just tested this branch on safari, chrome, and firefox on mobile and I can't reproduce this. Is it showing up in a certain context?

lib/components/viewers/viewers.css
amy-corson-ibigroup: Good catch! Fixed in [8017627](https://github.com/opentripplanner/otp-react-redux/pull/1249/commits/8017627278f9c492bc09360ee692667ce94f2af3)

lib/components/mobile/batch-search-screen.tsx
amy-corson-ibigroup: Since this (and other instances) is in a class component, `useCallback` isn't permitted. Is there an alternative for this?

lib/components/mobile/batch-search-screen.tsx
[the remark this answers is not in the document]
amy-corson-ibigroup: Added in [8017627](https://github.com/opentripplanner/otp-react-redux/pull/1249/commits/8017627278f9c492bc09360ee692667ce94f2af3)

lib/components/mobile/batch-search-screen.tsx
binh-dam-ibigroup: Oh, yes, if it's a component class, then just extract a class method (so that the function isn't recreated on each render).

lib/components/mobile/batch-search-screen.tsx
binh-dam-ibigroup: I see that on Firefox devtools, Firefox on Android, Chrome devtools and Chromium on Android.

lib/components/mobile/batch-search-screen.tsx
amy-corson-ibigroup: Gotcha, fixed in [0778c58](https://github.com/opentripplanner/otp-react-redux/pull/1249/commits/0778c58e3cf601a67e34fb54e3702a790b320146)

binh-dam-ibigroup (approved):
Looks much better, thank you for the changes!

lib/components/form/advanced-settings-panel.tsx
binh-dam-ibigroup, line 1: I just got this thought after looking at the location selector for the mobile view, any of the following could be done in another PR... It seems like this could be a dialog, at least in the mobile view, what do you think? That way, the Advanced Trip Options panel can be dismissed using the ESC key, and for desktop, the focus would be contained within the dialog.

PR #1250, opened 2024-07-17
Fix D/T Moving Text Chrome/Filter bug

amy-corson-ibigroup:
**Description:**

Credit to @daniel-heppner-ibigroup for discovering this solution. The latest update in Chrome is causing issues with the `filter` property impacting the `position` of elements. Switching to `opacity` creates a similar effect but with no issues without the bug!

**PR Checklist:**


josh-willis-arcadis (approved):
such a crazy bug! Thanks for fixing it!

binh-dam-ibigroup (approved):
Good find! `opacity` is more intuitive.

PR #1251, opened 2024-07-29
Add building blocks dropdown

josh-willis-arcadis:
Implements new building block dropdown component. 

Addl discussion at #1241 

**PR Checklist:**


lib/components/user/nav-login-button.css
```diff
@@ -25,6 +25,18 @@
...
+  margin: 0;
+  font-weight: 600;
+  font-size: 16px;
+}
+
+ul {
+  list-style: none;
+  padding-inline-start: 0;
+}
+
```
amy-corson-ibigroup, line 39: These are being applied to ALL `p` and `ul` elements site-wide, can you add some specificity to the selector, so it's only being applied to this menu?

lib/components/viewers/route-details.tsx
```diff
@@ -38,6 +38,13 @@ const PatternSelectButton = styled(UnstyledButton)`
   }
 `
 
+const PatternSelectDropdown = styled(Dropdown)`
+  span,
+  span.caret {
+    color: #333;
```
amy-corson-ibigroup, line 44: Nice!

amy-corson-ibigroup (changes requested):
Looking really good! The css rules are being applied globally, but once those are fixed this should be good to go.

lib/components/user/nav-login-button.css
```diff
@@ -25,6 +25,18 @@
...
+  margin: 0;
+  font-weight: 600;
+  font-size: 16px;
+}
+
+.navBarItem > ul {
+  list-style: none;
+  padding-inline-start: 0;
+}
+
```
amy-corson-ibigroup, line 39: These `>` are targeting the direct children of `.navBarItem` and neither of these are direct children 

image

daniel-heppner-ibigroup:
The percy tests caught that there's a black border around the dropdown now. Is that expected? 
image

josh-willis-arcadis:
@daniel-heppner-ibigroup @amy-corson-ibigroup I think the border was added by accident tbh. I think it looks great in the route viewer. I will remove the border on the itinerary panel to pass the percy tests.

lib/components/narrative/narrative-itineraries-header.tsx
```diff
@@ -35,6 +35,18 @@ const ItinerariesHeaderContainer = styled.div<{ showHeaderText: boolean }>`
   margin-left: ${(props) => (props.showHeaderText ? 'inherit' : 'auto')};
 `
 
+const SortResultsDropdown = styled(Dropdown)`
+  button {
+    border: none;
+    padding: 3px 7px;
+    line-height: 20px;
+  }
```
amy-corson-ibigroup, line 43: If we're going to override the default styling of the dropdown in every place we're using it, then we should change the package in otp-ui. But personally, I prefer the padding and line heights that we have set in otp-ui, so I think we can just remove the border here and leave the other dropdowns unchanged.

lib/components/viewers/route-details.tsx
```diff
@@ -149,12 +166,11 @@ class RouteDetails extends Component<Props> {
             <HeadsignSelectLabel htmlFor="headsign-selector-label">
               <FormattedMessage id="components.RouteDetails.stopsTo" />
             </HeadsignSelectLabel>
-            <SortResultsDropdown
+            <PatternSelectDropdown
```
amy-corson-ibigroup, line 169: If we decide to leave the default margins (which would mean a slightly taller dropdown) just flagging it would probably be good to throw an `align-items: center` on the parent here so that this row doesn't look uneven: 
image

amy-corson-ibigroup (changes requested):
We're repeating these a lot: 
```
line-height: 20px;
padding: 3px 7px;
  span.caret {
    margin-left: 2px;
  }
```

IF we think this is what the default dropdown should be, let's go ahead and fix it in building-blocks. But I actually think the new dropdown looks nice and is an improvement! Interested in @daniel-heppner-ibigroup opinion on this too, if we prefer to match the percy tests exactly, that's totally fine. But right now we're setting those styles in one place and immediately overriding them in another.

lib/components/viewers/route-details.tsx
daniel-heppner-ibigroup: This is the only place I really had a problem with the way the new styling looks. If we can get this vertically centered then I'm happy to approve the Percy diff

josh-willis-arcadis: Take a look a [78d52eb](https://github.com/opentripplanner/otp-react-redux/pull/1251/commits/78d52ebb5f64733579426928fae93b9e8086526b). This addresses @amy-corson-ibigroup and @daniel-heppner-ibigroup suggestions!

amy-corson-ibigroup: Looking really good, I would maybe add some `margin-bottom` to this `div` just to add a little vertical space between the selector and the "stops in this direction of travel": 
image

amy-corson-ibigroup (approved):
Looks great! Thanks so much for all the changes

lib/components/viewers/route-details.tsx
josh-willis-arcadis: added here! [78d52eb](https://github.com/opentripplanner/otp-react-redux/pull/1251/commits/78d52ebb5f64733579426928fae93b9e8086526b)

josh-willis-arcadis:
@daniel-heppner-ibigroup I think the Percy diff is good to be approved. Let me know if there are any more questions!

daniel-heppner-ibigroup:
Looks great! Thank you for your hard work here @josh-willis-arcadis

daniel-heppner-ibigroup (approved):
Nice work!

PR #1252, opened 2024-08-02
fix(apiV2): Pass batchId when calling GraphQL 'plan' endpoint

binh-dam-ibigroup:
## Description

Works in conjunction with https://github.com/ibi-group/otp-middleware/pull/246.

This PR makes the UI pass a `batchId` parameter when making a call to the GraphQL plan endpoint, so that trip requests can be grouped as a single entry in the "Recent searches" list and in usage reports.

## PR Checklist

### DIFF
--- a/lib/actions/apiV2.js
+++ b/lib/actions/apiV2.js
@@ -103,7 +103,7 @@ function createGraphQLQueryAction(
   options
 ) {
   const fetchOptions = {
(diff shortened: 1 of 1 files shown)


miles-grant-ibigroup (approved):
Wonderful!

josh-willis-arcadis (approved):
looks great!

PR #1253, opened 2024-08-05
Support GraphQL Trip Requests

binh-dam-ibigroup:
## Description

This PR adds support for parsing trip request data from OTP-middleware that includes GraphQL variables, so that they are included in the Recent Searches list.
Support for old OTP1 query params in trip requests from OTP-middleware is removed.

Note that the label for a recent search includes the modes used, however the modes will not be updated upon clicking the recent search entry (the mode selector states are not passed or saved in the trip request).

## PR Checklist


josh-willis-arcadis (approved):
Looks good!

lib/actions/user.js
```diff
@@ -156,35 +156,52 @@ export function fetchAuth0Token(auth0, intl) {
...
+    qp1.departArrive === qp2.departArrive &&
+    qp1.from?.name === qp2.from?.name &&
+    qp1.departArrive === qp2.departArrive &&
+    qp1.time === qp2.time &&
+    qp1.to?.name === qp2.to?.name &&
+    qp1.walkReluctance === qp2.walkReluctance &&
+    qp1.walkSpeed === qp2.walkSpeed &&
+    qp1.wheelchair === qp2.wheelchair
+  )
+}
```
amy-corson-ibigroup, line 188: Instead of doing this many checks, would it be possible to filter out the properties you don't want to compare (eg. modes and banned routes) and then compare using `JSON.stringify`?

PR #1255, opened 2024-08-12
fix(saved-trip-editor): Route back button to saved trips if editing

amy-corson-ibigroup:
**Description:**

Fixes a bug where, if a user accessed the "edit trip" page through a link in their email, the "back" button would send them to a broken Auth0 url. 

Now if a user is editing a trip, it will always take you back to saved trips. 

**PR Checklist:**


lib/components/user/back-link.js
```diff
@@ -12,14 +12,18 @@ const StyledButton = styled(Button)`
   padding: 0;
 `
 
+export const backButtonContent = (
```
binh-dam-ibigroup, line 15: Maybe this should be a component.

lib/components/user/monitored-trip/saved-trip-editor.tsx
```diff
@@ -52,7 +54,13 @@ const SavedTripEditor = (props: Props): JSX.Element => {
     return (
       <>
         <PageTitle title={title} />
-        <BackLink />
+        {/* If creating, back button should return to trip planner. If not, it should return
+        to saved trips, just in case the user accessed this page through their email. */}
+        {isCreating ? (
+          <BackLink />
```
binh-dam-ibigroup, line 60: If we are doing links, how about we place a `<Link to="/">` for when `isCreating` is true? (The link will preserve query params in the URL)

binh-dam-ibigroup (commented):
What do you think about using a link to "/" when creating trips?

lib/components/user/monitored-trip/saved-trip-editor.tsx
amy-corson-ibigroup: I think you're right, added in [da8783a](https://github.com/opentripplanner/otp-react-redux/pull/1255/commits/da8783aa0e80dd3ab09e6b4200fe23a61ac7d8b0)

lib/components/user/back-link.js
amy-corson-ibigroup: Good call! Fixed in [127475c](https://github.com/opentripplanner/otp-react-redux/pull/1255/commits/127475c1cf79ef983aba3a5cf13f0cf6a255eb00)

binh-dam-ibigroup (approved):
Sweet!

amy-corson-ibigroup:
@miles-grant-ibigroup @binh-dam-ibigroup  if this is looking okay after Miles reviews, and you guys wouldn't mind merging while I'm out next week, that would be amazing!

miles-grant-ibigroup (approved):
I'm tempted to rename `Content` to `Text` or similar but it's not worth blocking this whole thing

PR #1256, opened 2024-08-14
Field trip august 2024 fixes

miles-grant-ibigroup:
**Description:**
Related to the plan functionality. The plan button was causing confusion and problems so we got rid of it when field trip is open. Then we had the problem that the location changes were getting wiped away. We solved this by adding a confirm before the locations are overriden. The issue with this approach is that we will not save the updated location to the trip. Maybe that's ok?


lib/components/narrative/default/default-itinerary.js
```diff
@@ -166,7 +166,7 @@ const ITINERARY_ATTRIBUTES = [
             includeSeconds={false}
           />
           <LegIconWrapper noSpace>
-            <LegIcon leg={leg} size={5} />
+            <LegIcon leg={{ ...leg, mode: 'WALK' }} size={5} />
```
daniel-heppner-ibigroup, line 169: does legIcon even need the whole leg?

daniel-heppner-ibigroup (approved):
Looks fine to me

lib/components/narrative/default/default-itinerary.js
miles-grant-ibigroup: yeah it complains if the start/end stuff is missing

lib/components/app/call-taker-panel.js
```diff
@@ -153,10 +153,12 @@ class CallTakerPanel extends Component {
     }
   }
 
+  // eslint-disable-next-line complexity
```
josh-willis-arcadis, line 156: I've seen this warning in other places. Should be disable those as well?

josh-willis-arcadis (approved):
LGTM

lib/components/app/call-taker-panel.js
miles-grant-ibigroup: No! It's a good reminder to make our code better

PR #1257, opened 2024-08-15
fix(wizard): AT reads correct step number in account setup

amy-corson-ibigroup:
**Description:**

Fixes a bug with Safari + Voiceover that was causing Voiceover to inconsistently read the incorrect step or title name when the step was changed. Eg. The user would be on step 3 of 4 but VO would announce it as step 2 of 4. 

My best guess [based on this thread](https://github.com/nvaccess/nvda/issues/6265) is that safari was firing its nameChange/DescriptionChange event after passing the information to the screenreader in some cases. Putting `focusHeader` inside `ComponentDidUpdate` should hopefully consistently focus the header after the browser has fully fired all events.

**PR Checklist:**


binh-dam-ibigroup (approved):
Nice fix!

amy-corson-ibigroup:
@josh-willis-arcadis @binh-dam-ibigroup if this is looking okay after Josh reviews, if you guys wouldn't mind merging while I'm out next week, that would be amazing!

josh-willis-arcadis (approved):
lgtm!

PR #1259, opened 2024-08-21
only show routes from current service week

miles-grant-ibigroup:
**Description:**
Makes use of new graphql parameters to limit route and nearby results to current week

- [x] routes
- [x] nearby

blocked waiting for otp-ui to be fixed


lib/actions/apiV2.js
```diff
@@ -436,11 +437,19 @@ export const fetchNearby = (position, radius) => {
                 lon
                 code
                 gtfsId
+                stopRoutes: routes ${
+                  currentServiceWeek
+                    ? `(serviceDates: {start: "${currentServiceWeek.start}", end: "${currentServiceWeek.end}"})`
```
daniel-heppner-ibigroup: this needs to be a GraphQL variable and passed in with the other variables

lib/actions/apiV2.js
```diff
@@ -818,10 +827,20 @@ export const findRoute = (params) =>
...
+      getState().otp?.config?.routeViewer?.onlyShowCurrentServiceWeek === true
+        ? getCurrentServiceWeek()
+        : null
+
     dispatch(
       createGraphQLQueryAction(
         `{
-          routes {
+          routes ${
+            currentServiceWeek
```
daniel-heppner-ibigroup: same here, this needs to be a graphql variable and passed in with the others

lib/components/viewers/nearby/stop.tsx
```diff
@@ -70,6 +69,12 @@ const Stop = ({
       const sortedStopTimes = st.stoptimes.sort(
         (a: StopTime, b: StopTime) => fullTimestamp(a) - fullTimestamp(b)
       )
+      if (
+        stopData.nearbyRoutes &&
+        !stopData.nearbyRoutes.includes(st?.pattern?.route?.gtfsId)
+      ) {
+        return <></>
```
daniel-heppner-ibigroup, line 78: Can you add a comment(s) to better explain what's going on with this nearbyRoutes thing. It took me a while to really understand what's going on here and its relation to the service week thing

daniel-heppner-ibigroup (changes requested):
Need to change the graphql queries

lib/actions/apiV2.js
miles-grant-ibigroup: Thanks yay

miles-grant-ibigroup:
@daniel-heppner-ibigroup to fix the graphql mocks

lib/actions/apiV2.js
```diff
@@ -436,11 +438,15 @@ export const fetchNearby = (position, radius) => {
                 lon
                 code
                 gtfsId
+                stopRoutes: routes (serviceDates: $currentServiceWeek) {
```
daniel-heppner-ibigroup, line 442: why are we renaming routes to stop routes?

daniel-heppner-ibigroup (approved):
it would be even better if we didn't rename the routes in the gql query

lib/actions/apiV2.js
miles-grant-ibigroup: for OTP1 support!

amy-corson-ibigroup, line 442: I'm getting a "Unknown field argument" graphql error back when I try to view a stop

amy-corson-ibigroup (changes requested):
Looks like there might be an error in the `fetchNearby` query!

lib/actions/apiV2.js
miles-grant-ibigroup: This only works on the latest version of OTP. Are you testing against an older OTP version? I'm happy to delay merging this until newer OTP is more widespread

amy-corson-ibigroup: Oh gotcha, yes the instance of OTP I was using to test looks like the only one of ours this would break! We can discuss that internally

amy-corson-ibigroup (approved):

PR #1264, opened 2024-08-29
fix(nearby-view): Avoid blank nearby view on mobile

amy-corson-ibigroup:
**Description:**

Fixes a bug where, on mobile when user had denied geolocation, a blank page was being displayed due to lack of coordinates. This was happening because the desktop fallback (using the map center as the nearby point) required rendering a map, and mobile does not. Instead, now we pass the `initLon` and `initLat` from the config.

**PR Checklist:**


daniel-heppner-ibigroup (approved):
Looks good! Nice change. This isn't a new config item right?

amy-corson-ibigroup:
Nope! The map `initLat` and `initLon` coordinates were already there!

josh-willis-arcadis (approved):
LGTM

PR #1265, opened 2024-09-03
Deselect all submodes

josh-willis-arcadis:
**Description:**
When the user deselects all transport submode options(transit and rental), the mode should be disabled. This is reflected in the URL query params. When the mode is enabled after being disabled by deselecting all transport submode options, the defaults are used.

Blocked by https://github.com/opentripplanner/otp-ui/pull/782

**PR Checklist:**


lib/components/form/advanced-settings-panel.tsx
```diff
@@ -213,11 +223,12 @@ const AdvancedSettingsPanel = ({
         fillModeIcons
         label="test"
         modeButtons={processedModeButtons}
-        onSettingsUpdate={onSettingsUpdate(setQueryParam)}
-        onToggleModeButton={setModeButton(
-          enabledModeButtons,
-          onSettingsUpdate(setQueryParam)
+        onSettingsUpdate={onAdvancedModeSubsettingsUpdate(
```
daniel-heppner-ibigroup: What I am uncomfortable with here is that the advanced settings panel already has the mode button that corresponds with the mode setting that is being updated. I don't think we should do a bunch of work to get that mode button item again. It will require an update to the OTP UI component but I think there's a better way to go about this.

daniel-heppner-ibigroup: In fact, maybe it would be possible to do this change almost entirely within OTP UI? Could we just add a prop for `transportModesDisableCascade`? Weird name idk but something like that

daniel-heppner-ibigroup (commented):
first round of thoughts on this here. 
In general we should not be repeating work to find the mode buttons that we already have access to elsewhere in the application.

lib/components/form/advanced-settings-panel.tsx
```diff
@@ -127,7 +128,7 @@ const AdvancedSettingsPanel = ({
   modeSettingDefinitions: ModeSetting[]
   modeSettingValues: ModeSettingValues
   onPlanTripClick: () => void
-  setQueryParam: (evt: any) => void
+  setQueryParam: (evt: QueryParamChangeEvent) => void
```
daniel-heppner-ibigroup: Don't use this type here, in other places in OTP RR I believe we are just using a simple `(evt: any) => void`. If you want to make a new type and implement it for all the setQueryParams that would be awesome, but we shouldn't import it from OTP UI.

lib/components/form/util.tsx
```diff
@@ -59,13 +59,18 @@ export const onSettingsUpdate =
 export const setModeButton =
   (enabledModeButtons: string[], updateHandler: (params: any) => void) =>
   (buttonId: string, newState: boolean) => {
+    console.log('SET MODE BUTTON IS BEING CALLED. NEW STATE:::: ', newState)
```
daniel-heppner-ibigroup: delete this console log

lib/components/form/util.tsx
daniel-heppner-ibigroup, line 1: actually please just revert all changes to this file

daniel-heppner-ibigroup (changes requested):
first round, I'll finish reviewing when OTP UI is merged

lib/components/form/util.tsx
josh-willis-arcadis: good catch

lib/components/form/advanced-settings-panel.tsx
daniel-heppner-ibigroup: Please replace this type! I don't think we should be importing these internal types from OTP-UI

lib/components/form/util.tsx
daniel-heppner-ibigroup: still a change here. you can run `git checkout dev lib/components/form/util.tsx`

daniel-heppner-ibigroup (changes requested):

lib/components/form/advanced-settings-panel.tsx
```diff
@@ -169,6 +169,7 @@ const AdvancedSettingsPanel = ({
   )
 
   const processedModeSettings = processSettings(modeSettingDefinitions)
+
```
daniel-heppner-ibigroup: can you remove this new line? (it feels unimportant but it keeps the git blame cleaner)

daniel-heppner-ibigroup (approved):
awesome! this is clean!

lib/components/form/advanced-settings-panel.tsx
```diff
@@ -177,6 +178,15 @@ const AdvancedSettingsPanel = ({
     )
   )
 
+  const handleModeButtonToggle = setModeButton(
+    enabledModeButtons,
+    onSettingsUpdate(setQueryParam)
+  )
+
+  const handleAllSubmodesDisabled = (modeButton: ModeButtonDefinition) => {
```
amy-corson-ibigroup, line 185: Could we potentially add a setTimeout onto this to match the animation timing when the panel collapses? Right now you uncheck the last one, all the checkboxes check themselves, and then the panel collapses. It's looking a little more like a bug than a feature.

amy-corson-ibigroup (commented):
Curious what you both think of this `setTimeout `idea @daniel-heppner-ibigroup @josh-willis-arcadis. We can also address in a follow up PR if need be

lib/components/form/advanced-settings-panel.tsx
josh-willis-arcadis: I agree it could be better. I believe this will need to be handled in otp-ui. I can open a PR there

josh-willis-arcadis: here is that PR: https://github.com/opentripplanner/otp-ui/pull/791

amy-corson-ibigroup (approved):
Looking really good as-is, we can move the conversation about potential `setTimeout` to the OTP-UI PR! Thank you so much for your work on this!! :) Also just flagging daniel's note about the +1 line on 172.

PR #1269, opened 2024-09-05
add option to use arrival time in nearby view

daniel-heppner-ibigroup:
**Description:**
Adds configuration option to use the arrival time rather than departure time in Nearby View and other places where the patternrow is used.

### DIFF
--- a/lib/util/viewer.js
+++ b/lib/util/viewer.js
@@ -18,13 +18,16 @@ export function getSecondsUntilDeparture(
   useSchedule,
   useArrivalTime
 ) {
-  const time = useSchedule
-    ? useArrivalTime
+  let time
+  if (useSchedule) {
+    time = useArrivalTime
       ? stopTime.scheduledArrival
       : stopTime.scheduledDeparture
-    : useArrivalTime
-    ? stopTime.realtimeArrival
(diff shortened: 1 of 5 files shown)


lib/util/viewer.js
```diff
@@ -13,12 +13,20 @@ import { isBlank } from './ui'
...
+  useSchedule,
+  useArrivalTime
+) {
+  const time = useSchedule
+    ? useArrivalTime
+      ? stopTime.scheduledArrival
+      : stopTime.scheduledDeparture
+    : useArrivalTime
+    ? stopTime.realtimeArrival
     : stopTime.realtimeDeparture
```
miles-grant-ibigroup: Oy vey

miles-grant-ibigroup (approved):
Will leave it up to the second reviewer to decide if that ternary is alright

lib/util/viewer.js
amy-corson-ibigroup: I guess if it works I don't mind the long ternary, but if we could add a comment explaining the logic, just so it's not a headache to parse in the future, that'd be great!!

amy-corson-ibigroup (approved):
Beautiful new config item sorry for long delay in reviewing

PR #1270, opened 2024-09-06
set `viewedRouteStops` when route viewer is active

miles-grant-ibigroup:
**Description:**
Blocked waiting for https://github.com/opentripplanner/otp-ui/pull/772

**PR Checklist:**


lib/components/map/default-map.tsx
```diff
@@ -427,6 +429,19 @@ class DefaultMap extends Component {
...
+  const viewedRoute = state.otp?.ui?.viewedRoute?.routeId
+  const nearbyViewerActive =
+    state.otp.ui.mainPanelContent === MainPanelContent.NEARBY_VIEW
+  const viewedRouteStops =
+    viewedRoute && !nearbyViewerActive
+      ? Object.entries(
+          state.otp?.transitIndex?.routes?.[viewedRoute]?.patterns || {}
+        ).reduce((acc, cur) => {
+          return Array.from(
+            new Set([...cur?.[1]?.stops.map((s) => s.id), ...acc])
```
daniel-heppner-ibigroup: I think this can be cleaned up. Remove the Array.from and Set from this and do it outside the reducer so it doesn't have to be repeated on each loop. Also, maybe we can move this to a function above? I found it quite difficult to read what was happening here.

daniel-heppner-ibigroup: Alternatively just break out more variables. Like put the Object.entries into a variable.

daniel-heppner-ibigroup (changes requested):
Just clean up the state mapper. I also noticed that sometimes a handful of stops just don't show up. Is that an OTP thing?

miles-grant-ibigroup:
Blocked by https://github.com/opentripplanner/otp-react-redux/pull/1270

daniel-heppner-ibigroup (approved):
Nice thank you!

amy-corson-ibigroup (approved):
Looks great!!

PR #1277, opened 2024-09-17
Skeletonize Operator Logo Loading

miles-grant-ibigroup:
**Description:**
Uses `Skeleton` instead of `InlineLoading` which results in a less abrupt flash.

Thoughts? @amy-corson-ibigroup

### DIFF
--- a/lib/components/util/transit-operator-icons.tsx
+++ b/lib/components/util/transit-operator-icons.tsx
@@ -70,7 +70,7 @@ const TransitOperatorLogos = ({
   return (
     <>
       {loading ? (
-        <Skeleton style={{ marginRight: '0.5ch' }} width={20} />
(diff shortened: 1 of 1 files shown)


amy-corson-ibigroup (approved):
Much cleaner! Can we add a `height` too so it's square? But yes I love this

lib/components/util/transit-operator-icons.tsx
```diff
@@ -71,10 +70,7 @@ const TransitOperatorLogos = ({
   return (
     <>
       {loading ? (
-        <>
-          <InlineLoading />
-          <span style={{ marginRight: '0.5ch' }} />
-        </>
+        <Skeleton style={{ marginRight: '0.5ch' }} width={20} />
```
amy-corson-ibigroup:
```suggestion
        <Skeleton height={20} style={{ marginRight: '0.5ch' }} width={20} />
```

PR #1279, opened 2024-09-19
Allow for Nearby View String Overrides

miles-grant-ibigroup:
**Description:**
Allows for configurations to override a string that was previously just sent directly from api output to UI.


daniel-heppner-ibigroup (approved):
huge

amy-corson-ibigroup (approved):
Beautiful!

PR #1282, opened 2024-10-08
allow multiple transport modes to be specified in a single transportmode

daniel-heppner-ibigroup:
**Description:**
We need to allow several modes to be grouped together under a single icon. This will allow combining rail modes (like TRAM, RAIL, SUBWAY) together or bus modes (BUS, TROLLEYBUS) together under a single checkbox. 
Support for this already existed in OTP UI so we are just enabling support for it in OTP RR with this change.


example-config.yml
```diff
@@ -268,7 +268,11 @@ modes:
       iconName: wheelchair
       type: CHECKBOX # Possible options: CHECKBOX, SUBMODE, SLIDER, DROPDOWN
   transitModes:
-    - mode: BUS
+      # Mode can be a string or an array of strings.
+    - mode: ["TROLLEYBUS", "BUS"]
+      # When mode is an array an overrideMode must be provided.
+      # This specifies the displayed mode used for icon and URL param.
+      overrideMode: BUS
```
miles-grant-ibigroup, line 275: This key title is confusing. But I can't think of a better one

lib/reducers/create-otp-reducer.js
```diff
@@ -135,10 +135,18 @@ export function getInitialState(userDefinedConfig) {
 
   const transitModeSettings = config?.modes?.transitModes.map((transitMode) => {
     const { mode, overrideMode } = transitMode
+    if (Array.isArray(mode) && !overrideMode) {
+      console.warn(
```
miles-grant-ibigroup, line 139: When this warning fires the entire app does still crash but I guess the warning is there so that's good enough

miles-grant-ibigroup (approved):
Works great!

example-config.yml
daniel-heppner-ibigroup: Was trying not to make it a breaking change.

miles-grant-ibigroup: we could've done a thing where we just use the first item in the array as the override but this is fine too!

daniel-heppner-ibigroup: that's not a bad idea! we'll see what second reviewer thinks

binh-dam-ibigroup: The name can be confusing at first. Maybe change `overrideMode` to `summaryMode` or `displayedMode` but I don't feel strongly about it. It looks like branded names are already supported, so that should be fine.

binh-dam-ibigroup (approved):

example-config.yml
daniel-heppner-ibigroup: I wanted to keep it so we didn't change the config at all. I agree that the name choice wasn't great but I don't think it's worth updating now.

PR #1283, opened 2024-10-09
respect mode text color

miles-grant-ibigroup:
**Description:**
Makes sure mode based text color overrides are used on the map

**PR Checklist:**


lib/actions/apiV2.js
```diff
@@ -1144,6 +1145,16 @@ export function routingQuery(searchId = null, updateSearchInReducer) {
                               config.transitOperators
                             ),
                             { color: leg?.route?.color, mode: leg.mode }
+                          ).split('#')?.[1],
+                          textColor: getRouteTextColorBasedOnSettings(
```
daniel-heppner-ibigroup, line 1155: put this in a const at the top of the map. also refactor `getRouteOperator` into its own variable so it isn't repeated.

lib/util/viewer.js
```diff
@@ -353,6 +353,10 @@ export function getRouteColorBasedOnSettings(operator = {}, route = {}) {
       return backgroundColor
   }
 }
+export function getRouteTextColorBasedOnSettings(operator = {}, route = {}) {
```
daniel-heppner-ibigroup, line 358: delete this change and revert the file

daniel-heppner-ibigroup (changes requested):

daniel-heppner-ibigroup (approved):
alright, this is looking good now. thanks for the fix

amy-corson-ibigroup (approved):
Beautiful colors!

PR #1284, opened 2024-10-11
make use of new stops+stations layer

miles-grant-ibigroup:
**Description:**
Waiting for otp-ui

**PR Checklist:**


amy-corson-ibigroup (approved):
Beautiful stations! I don't think this is blocking but just flagging that hovering the corresponding card in the nearby view isn't hilighting the new station circle, which maybe something we want to fix in the future.

josh-willis-arcadis (approved):
lovely!

PR #1288, opened 2024-10-15
Overlay Popup Behavior Correction

miles-grant-ibigroup:
**Description:**
The jumping is because of transitive. I couldn't think of a good way to stop transitive from jumping and then thought about what the desired behavior would be without the jumping (just randomly flashing colors) and it also feels really weird so I think the solution is to just disable hover effects altogether. 

**PR Checklist:**


lib/components/map/itinerary-summary-overlay.tsx
```diff
@@ -152,14 +152,18 @@ const ItinerarySummaryOverlay = ({
...
 
   if (!itins || !visible) return <></>
-  const mergedItins: ItinWithGeometry[] = addTrueIndex(
-    doMergeItineraries(itins).mergedItineraries.map(addItinLineString)
+  const indexedItins: ItinWithGeometry[] = addTrueIndex(
+    itins.map(addItinLineString)
+  )
+  const mergedItins = doMergeItineraries(indexedItins).mergedItineraries
+
+  const midPoints = mergedItins.reduce(
```
daniel-heppner-ibigroup: this has an any type. I think `reduce` needs to have a type and `mergedItins` needs to be typed to not get errors.
Either type `narrative-itineraries.js` and `doMergeItineraries()` or assign a type to the variable here.

daniel-heppner-ibigroup (changes requested):
typescript :(

lib/components/map/itinerary-summary-overlay.tsx
daniel-heppner-ibigroup: it still has an any type
image

daniel-heppner-ibigroup (approved):
fine now because I fixed it

josh-willis-arcadis (approved):
Changes look good. Looks like percy tests are failing visual tests? Not sure what is happening there.

miles-grant-ibigroup:
> Changes look good. Looks like percy tests are failing visual tests? Not sure what is happening there.

Daylight savings!

PR #1290, opened 2024-10-16
Consolidate find stop times for stop query

amy-corson-ibigroup:
**Description:**
Removes the shorter query for operators, as it was causing issues with the schedule in the nearby view.

**PR Checklist:**


josh-willis-arcadis (approved):
LGTM!

josh-willis-arcadis:
looks like snapshots need to be cleaned up

miles-grant-ibigroup (changes requested):
Still seeing a bug. If I go to the schedule view for a stop, then click on a stop popup on the map, I see the weird "Loading Stop...". 

We might need a new approach to getting these icons?

lib/components/map/default-map.tsx
```diff
@@ -86,6 +86,8 @@ function getLayerName(overlay, config, intl) {
       return intl.formatMessage({ id: 'components.MapLayers.streets' })
     case 'Satellite':
       return intl.formatMessage({ id: 'components.MapLayers.satellite' })
+    case 'Stops':
```
miles-grant-ibigroup, line 89: Why is this here?

miles-grant-ibigroup (approved):
I'm so confused how/why this works but it totally does! Yay

lib/components/map/default-map.tsx
amy-corson-ibigroup: Because the otpUi-stopsandstations layer has "Stops" as the name!
image

PR #1291, opened 2024-10-16
feat(advanced-trip-form): Add autoPlan to advanced trip form

amy-corson-ibigroup:
**Description:**

Add autoPlan to advanced trip form. If autoPlan is on, and there are validation errors in the location field, do not call the plan trip function.

**PR Checklist:**


daniel-heppner-ibigroup (approved):
I like the approach here and I think it works nicely!

josh-willis-arcadis (approved):
LGTM!

PR #1292, opened 2024-10-17
Custom GraphQL and Mobility Profile Routing

binh-dam-ibigroup:
## Description

This PR sets up craco to accept custom plan GraphQL files (or use a default one).
Custom plan query files are passed through the custom `config.js` and placed in the redux state at startup.
For configuration-enabled mobility profiles, that new parameter is added to the variables passed to the GraphQL `plan` query. 

To test, ensure that planning trips with mobility profiles produce different results based on the profiles, and that on configs without mobility profiles, planning trips is not affected.

**PR Checklist:**


josh-willis-arcadis (approved):
changes are working with multiple configs with and without mobility profile.

lib/main.js
```diff
@@ -34,6 +34,18 @@ import(CSS)
...
+// defaults to ./config.js
+// defined in webpack config:
+// The JS_CONFIG variable is passed to this file by webpack's `DefinePlugin` that replaces the variable
+// with its content at compile time (like C's `#define` preprocessor directive).
+// eslint-disable-next-line no-undef
+const jsConfig = require(JS_CONFIG).configure(otpConfig)
+
+// Plug the plan query into the config (if available)
+otpConfig.api.planQuery = jsConfig.planQuery
+
```
miles-grant-ibigroup: Do we still need this? Why not just put it in the js config?

craco.config.js
```diff
@@ -65,6 +67,13 @@ module.exports = {
       }
       addBeforeLoader(webpackConfig, loaderByName('file-loader'), yamlLoader)
 
+      // Support YAML
```
miles-grant-ibigroup: Comment needs updating

lib/actions/apiV2.js
```diff
@@ -1046,6 +1048,9 @@ export function routingQuery(searchId = null, updateSearchInReducer) {
       ...currentQuery,
       numItineraries: numItineraries || getDefaultNumItineraries(config)
     }
+    if (config.mobilityProfile) {
+      baseQuery.mobilityProfile = loggedInUser?.mobilityProfile?.mobilityMode
+    }
```
miles-grant-ibigroup, line 1054: Do we have a GMAP specific branch for otp-rr? Do we want to bring this into dev?

miles-grant-ibigroup (changes requested):
Overall works, even if it's a bit hacky. Just want to double check a few things!

lib/actions/apiV2.js
binh-dam-ibigroup: Not yet, although try to move specific stuff to config would be best.

lib/main.js
binh-dam-ibigroup: This puts the custom query into the redux state, so it can be retrieved from actions.

craco.config.js
binh-dam-ibigroup: Thanks, updated in https://github.com/opentripplanner/otp-react-redux/pull/1292/commits/37260077aab004d45b0ced9dbef13c10226dbb8c

lib/main.js
binh-dam-ibigroup: I updated the comment in https://github.com/opentripplanner/otp-react-redux/pull/1292/commits/e60f6942111a56e4be995cdb0349a5522f30d94a.

miles-grant-ibigroup (approved):
A bit nervous about this but it seems to work alright...

PR #1296, opened 2024-10-28
fix: don't zoom when clicking itinerary

daniel-heppner-ibigroup:
**Description:**
When clicking on an itinerary, the app would re-zoom to the from/to rather than fitting the whole itinerary in view. This fixes that with a simple one liner.


binh-dam-ibigroup (approved):
Nice fix

josh-willis-arcadis (approved):
LGTM

PR #1297, opened 2024-10-30
Support naming and keying transit vehicle overlay

miles-grant-ibigroup:
**Description:**
Previously there was no way to name/key the transit vehicle overlay. This adds a new way to do that without needing any new explicit configuration types!


lib/components/map/default-map.tsx
```diff
@@ -341,7 +345,13 @@ class DefaultMap extends Component {
           {/* The default overlays */}
           <EndpointsOverlay />
           <RouteViewerOverlay />
-          <TransitVehicleOverlay ModeIcon={ModeIcon} />
+          <TransitVehicleOverlay
+            id={routeBasedTransitVehicleOverlayNameOverride || undefined}
```
daniel-heppner-ibigroup: maybe move the `|| undefined` into the const so you don't have to repeat it on each of these!

daniel-heppner-ibigroup (approved):
LGTM. just one suggestion

amy-corson-ibigroup (approved):
Looks good!

amy-corson-ibigroup:
Only thing is that percy is failing but I'm guessing you already noticed that!!

PR #1300, opened 2024-11-06
Small changes and header css

amy-corson-ibigroup:
**Description:**

- Add margin to loading spinner to avoid clipping.
- Fix small regression on headsign selector in pattern viewer, to avoid weird spacing on buttons.
- Clean up navbar css to avoid clipping on all screen sizes.
- Sort css
- Fix cut-off in pattern viewer stops container

**PR Checklist:**


lib/components/app/app.css
```diff
@@ -47,36 +47,266 @@
   }
 }
 
+/* Navbar settings */
+
+.navbar .navbar-header {
```
binh-dam-ibigroup, line 52: One of these settings alters the vertical positioning of the agency logo in desktop wide view (See screnshot), that means we need CSS to fix that.

image

binh-dam-ibigroup (changes requested):
I like the cleanup! Could you address the agency logo situation please?

lib/components/app/app.css
amy-corson-ibigroup: A lot of the existing logo positioning css you're seeing here is coming from the branding logo css in the configuration. This PR should ideally remove the need for that, so I've opened an internal PR to adjust these margins and tagged you there as well!

amy-corson-ibigroup:
Added to this PR: tried a new approach to the height of the stops `ol` in the pattern viewer, as the current approach is resulting in the last stop being cut off on most screen sizes.

binh-dam-ibigroup (approved):

miles-grant-ibigroup (commented):
Code looks good! I'm noticing the navbar icons are shifted down a bit, and are no longer perfectly centered. Otherwise though I'm seeing impressively few regressions. The header icon issue is a shame but it seems like there's been work done there to resolve that. 

That being said could I ask that this PR be put on hold until that other PR is ready to merge and that they're then merged at the same time?

binh-dam-ibigroup:
> That being said could I ask that this PR be put on hold until that other PR is ready to merge and that they're then merged at the same time?

Any specific reason to delay if this PR is ready?

miles-grant-ibigroup (changes requested):
Screenshot 2024-11-26 at 11 19 41 AM
This is what the default config looks like. The left-padding is too strong and something is wrong with the centering!

amy-corson-ibigroup:
Shortened the gap which should fix the padding issue! The centering was because of a `padding-top` in the example css, which I removed

miles-grant-ibigroup (approved):
I'm scared but I'm also excited

PR #1301, opened 2024-11-07
nearby: use stoptime headsigns if applicable

miles-grant-ibigroup:
It seems too easy... Did I miss some edge case?


amy-corson-ibigroup (approved):

miles-grant-ibigroup:
@daniel-heppner-ibigroup do you have any idea why the snapshots are just ever so slightly off? What's going on here?

daniel-heppner-ibigroup (changes requested):
I think what's happening is the duplicate checking code that follows the line you changed in stop.tsx is no longer matching a duplicate. Double check that and make sure it's okay, or change the duplicate checking code to maybe not use the headsign since it's now not consistent for a pattern?

miles-grant-ibigroup:
> I think what's happening is the duplicate checking code that follows the line you changed in stop.tsx is no longer matching a duplicate. Double check that and make sure it's okay, or change the duplicate checking code to maybe not use the headsign since it's now not consistent for a pattern?

Good catch thanks!

daniel-heppner-ibigroup (approved):
snapshot update actually looks really good! At that stop the headsign doesn't show East Green Lake or Roosevelt Station.

PR #1303, opened 2024-11-08
Transit Overlay: Restore missing OTP1 fields 

miles-grant-ibigroup:
There was a lastUpdated field that was removed at some point. But now it's back!


daniel-heppner-ibigroup (commented):
Are we maintaining support for OTP1 or are we translating an OTP2 vehicle into something that can be used with our code that still is looking for OTP1 props?

miles-grant-ibigroup:
> Are we maintaining support for OTP1 or are we translating an OTP2 vehicle into something that can be used with our code that still is looking for OTP1 props?

OTP1 transit vehicle overlay uses OTP1 props to maintain OTP1 support and this PR makes sure that that overlay works correctly

daniel-heppner-ibigroup (approved):
sounds good to me

josh-willis-arcadis (approved):
LGTM!

PR #1302, opened 2024-11-08
feat(trusted-companions): Plan trip with mobility profile for self or on behalf of trusted companion

josh-willis-arcadis:
**Description:**
Allows a user to plan a trip according to a mobility profile for their self or on behalf of a trusted companion. 

**PR Checklist:**
- [ ] Does the code follow accessibility standards (WCAG 2.1 AA Compliant)?
- [ ] Are all languages supported (Internationalization/Localization)?
- [ ] Are appropriate Typescript types implemented?

PR #1304, opened 2024-11-21
Trip sharing

binh-dam-ibigroup:
## Description

Works with https://github.com/ibi-group/otp-middleware/pull/271, needs #1302.

This PR adds trip sharing in monitored trips: A user saving a trip (whether for themselves or a dependent) can assign confirmed companions as trip companions or observers. Someone assigned as a primary traveler, companion, or dependent has read-only access to the trip (they can view trip details but cannot make changes).

## PR Checklist
- [?] Does the code follow accessibility standards (WCAG 2.1 AA Compliant)? => May need feedback about the use of disabled inputs.

Screenshot:

read-only trip companions


josh-willis-arcadis (approved):
Trip notifications are working for observers and dependents!

i18n/en-US.yml
```diff
@@ -639,6 +639,10 @@ components:
     tripNotAvailableOnDay: Trip not available on {repeatedDay}
     unsavedChangesExistingTrip: You haven't saved your trip yet. If you leave, changes will be lost.
     unsavedChangesNewTrip: You haven't saved your new trip yet. If you leave, it will be lost.
+  TripCompanionsPane:
```
amy-corson-ibigroup, line 645: I don't think the issue stems from this PR but I am seeing an issue with i18n strings in the wizard when it comes to the travel companion field 
image

lib/components/user/monitored-trip/saved-trip-editor.tsx
amy-corson-ibigroup, line 1: Thoughts on adding an informational alert on this page if it's `isReadOnly` letting users know why everything is disabled? We can use the `Alert` component from building blocks. Happy to suggest some language if this is the direction we want to go in

lib/components/user/monitored-trip/trip-basics-pane.tsx
```diff
@@ -138,16 +139,23 @@ const RenderAvailableDays = ({
...
   isCreating: boolean
+  isReadOnly: boolean
   monitoredTrip: MonitoredTrip
 }) => {
   const intl = useIntl()
   const baseColor = getBaseColor()
+
+  if (isReadOnly) {
+    return <MonitoredDays days={dayFieldsToArray(monitoredTrip)} />
+  }
```
amy-corson-ibigroup, line 157: Smart!

amy-corson-ibigroup (approved):
Instead of passing the `disabled` prop so many times, could we use a `fieldset` and pass it once? https://www.w3schools.com/tags/att_fieldset_disabled.asp

Otherwise code looks good! Just a few suggestions.

lib/components/user/monitored-trip/saved-trip-editor.tsx
binh-dam-ibigroup: I do like the idea, looking at a banner that says "<user> has added you to this trip. Therefore, you cannot make changes.". The Alert component in building-blocks is not quite usable for this purpose without restyling it (tricky to import, excess padding/margins even without the icon), so I'll use the react-bootstrap one for the time being.

binh-dam-ibigroup:
> Instead of passing the `disabled` prop so many times, could we use a `fieldset` and pass it once? https://www.w3schools.com/tags/att_fieldset_disabled.asp

Good catch! Updated in https://github.com/opentripplanner/otp-react-redux/pull/1304/commits/bcba92673e222da4fa013da03ac928204722518a.

lib/components/user/monitored-trip/saved-trip-editor.tsx
binh-dam-ibigroup: @amy-corson-ibigroup Let me know what you think of https://github.com/opentripplanner/otp-react-redux/pull/1304/commits/5cc84f0c028b92c4a96d04d85b457f7d972a4d30.

i18n/en-US.yml
binh-dam-ibigroup: It is indeed a separate issue, see #1331.

amy-corson-ibigroup:
Looks great!! Only nitpick is that the `alert` and its parent `div` both have `margin-bottom`s that are creating a pretty significant gap, I would maybe remove one instance so the spacing is even 
image

binh-dam-ibigroup:
> Looks great!! Only nitpick is that the `alert` and its parent `div` both have `margin-bottom`s that are creating a pretty significant gap, I would maybe remove one instance so the spacing is even

Good idea, addressed in https://github.com/opentripplanner/otp-react-redux/pull/1304/commits/55ed3b7b3d2078d4b4d04a25321067f689924fb4.

PR #1308, opened 2024-11-25
Only call stopTimes if we don't already have the data

amy-corson-ibigroup:
**Description:**
Fixes a bug where interaction with `setViewedRoute` was causing the `findStopTimesForStop` and subsequent response to fire forever. This should only fire when the stop information isn't already there. Also adds a `forceFetch` param for the stop schedule viewer, so that if you're trying to look at different schedule days, it will update the info with the new day.

**PR Checklist:**

### DIFF
--- a/lib/actions/apiV2.js
+++ b/lib/actions/apiV2.js
@@ -506,12 +506,9 @@ export const fetchNearby = (position, radius, currentServiceWeek) => {
 export const findStopTimesForStop = (params) =>
   function (dispatch, getState) {
     // If the stop is already in the store, don't fetch it again, unless we are forcing a refetch
-    if (
-      !params.forceFetch &&
-      getState().otp.transitIndex.stops[params.stopId]
-    ) {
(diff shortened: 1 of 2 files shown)


lib/actions/apiV2.js
```diff
@@ -505,6 +505,13 @@ export const fetchNearby = (position, radius, currentServiceWeek) => {
 
 export const findStopTimesForStop = (params) =>
   function (dispatch, getState) {
+    // If the stop is already in the store, don't fetch it again, unless we are forcing a refetch
+    if (
+      !params.forceFetch &&
+      getState().otp.transitIndex.stops[params.stopId]
+    ) {
+      return
+    }
```
miles-grant-ibigroup: Can we one-line this if eslint lets us? Getting rid of the brackets might help.

miles-grant-ibigroup (approved):
Wonderful work!

josh-willis-arcadis (approved):
LGTM!

PR #1309, opened 2024-11-26
replace hardcoded 'node_modules' path

ch-routerank:
**Description:**
This PR aims to replace the hard coded `node_modules` path from the i18n yml require line. This is helpful/necessary if the code isn't imported from the same relative path compared to the node_modules folder. Prefixing the include string with `@opentripplanner` make the lazy load mechanism find the files.

**PR Checklist:**

No text, logic or UI changes from this change.


miles-grant-ibigroup (approved):
Works great! Thanks so much for these changes

binh-dam-ibigroup (approved):
Seems to work, thank you!

miles-grant-ibigroup:
Tests will fail because it's an external PR but I've run them locally and all seems to work! If it all melts down we can revert

PR #1310, opened 2024-11-26
fix(base-renderer): remove OTP1 compatability conversion

miles-grant-ibigroup:
The alert view was accounting for the OTP1 compatibility itinerary-body used to maintain. That compatibility has now been removed. As a result, we now no longer need to convert data to OTP1 format


amy-corson-ibigroup (approved):

binh-dam-ibigroup (approved):
Thanks for this fix, glad to start removing legacy OTP1 stuff!

PR #1312, opened 2024-12-02
feat: add network connection status to ui and redux

amy-corson-ibigroup:
**Description:**
Creates a UI status bar and an invisible status region for screen readers that announces when the browser connection has been lost. 

**PR Checklist:**

| Web | Mobile |
|--------|-------|
|image|image|

### DIFF
--- a/lib/components/mobile/navigation-bar.js
+++ b/lib/components/mobile/navigation-bar.js
@@ -9,6 +9,7 @@ import * as uiActions from '../../actions/ui'
 import { accountLinks, getAuth0Config } from '../../util/auth'
 import { ComponentContext } from '../../util/contexts'
 import { injectIntl } from 'react-intl'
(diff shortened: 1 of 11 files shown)


lib/components/mobile/navigation-bar.js
```diff
@@ -111,6 +114,9 @@ class MobileNavigationBar extends Component {
             )}
           </MobileBar>
         </Navbar>
+        <NetworkConnectionBanner
```
josh-willis-arcadis, line 117: I think this is causing the percy tests to fail on mobile nearby view? not sure what else would cause the issues percy has.

josh-willis-arcadis (approved):
Everything is behaving as expected! Just need fix failing percy tests

binh-dam-ibigroup (approved):
Looks good. Not sure what CSS change made the Percy test fail, I'd be ok to bypass/override it.

PR #1314, opened 2024-12-03
fix(trip-basics-pane): remove ability to select days that the trip is…

josh-willis-arcadis:
**Description:**
Expected: Editing an existing trip you should not be  able to select days/times that the trip does not run.

Actual: Editing an existing trip you can select days/times that the trip does not run and send notification on days when it is not selected

**PR Checklist:**


amy-corson-ibigroup (approved):
Looks good to me!

daniel-heppner-ibigroup (approved):
lgtm

PR #1316, opened 2024-12-04
remove buggy useEffect usage

daniel-heppner-ibigroup:
**Description:**
These useEffect usages were causing bugs. Since useEffect calls its callback any time one of the dependencies change, it can be easy to accidentally introduce unexpected bugs. In this case, the query params changing was causing the dateTime effect to be called, which was setting the itinerary sort when external parameters were changing. In this case, we noticed changing the sort order from the itinerary header was causing the effect to run, immediately overriding the setting. 

This is actually a good lesson in why useEffect is dangerous!


miles-grant-ibigroup (approved):
All is working! Thanks for the cleanup

lib/components/form/call-taker/date-time-picker.tsx
```diff
@@ -258,7 +273,7 @@ const DateTimeOptions = ({
         <input
           className="datetime-slim"
           onChange={(e) => {
-            setTime(e.target.value)
+            handleTimeChange(e.target.value)
```
binh-dam-ibigroup: Could you set up a callback for the entire `onChange` event handler also? That way, the event handler is not recreated on each render.

lib/components/form/call-taker/date-time-picker.tsx
```diff
@@ -121,69 +151,32 @@ const DateTimeOptions = ({
...
 
   const timeRef = useRef(null)
 
   const intl = useIntl()
 
-  /**
-   * Parse a time input expressed in the agency time zone.
-   * @returns A date if the parsing succeeded, or null.
-   */
-  const parseInputAsTime = (
```
binh-dam-ibigroup, line 134: Good move.

lib/components/form/call-taker/date-time-picker.tsx
```diff
@@ -121,69 +151,32 @@ const DateTimeOptions = ({
...
+  const dateTime = parseInputAsTime(homeTimezone, time, date)
 
   // Update state when external state is updated
   useEffect(() => {
     if (initialDate !== date) setDate(initialDate)
     if (initialTime !== time) {
-      setTime(initialTime)
+      handleTimeChange(initialTime || '')
     }
   }, [initialTime, initialDate])
```
binh-dam-ibigroup, line 174: eslint is highlighting missing dependencies here: `date`, `time`, and `handleTimeChange`, could you add them to the dependency array as well.

lib/components/form/call-taker/date-time-picker.tsx
```diff
@@ -121,69 +151,32 @@ const DateTimeOptions = ({
...
-          'Invalid Time'
-      )
-    }
-  }, [time])
-
   useEffect(() => {
     if (initialDepartArrive && departArrive !== initialDepartArrive) {
       setDepartArrive(initialDepartArrive)
     }
   }, [initialDepartArrive])
```
binh-dam-ibigroup, line 182: While we are in fixing dependency arrays, could you adding `departArrive` in here too per eslint.

lib/components/form/call-taker/date-time-picker.tsx
```diff
@@ -197,42 +190,64 @@ const DateTimeOptions = ({
...
+            ...sort,
+            type: DepartArriveTypeMap[newValue]
+          }
         })
-      )
-    }
-  }, [departArrive, setTime, setDate, homeTimezone])
+      }
+    },
+    [syncSortWithDepartArrive, sort, importedUpdateItineraryFilter]
```
binh-dam-ibigroup, line 228: Here, eslint is suggesting to add `dateTime`, `handleTimeChange`, `homeTimezone` to the dependency array.

binh-dam-ibigroup (changes requested):
Could you `useCallback` for the handler of the `<input> onChange` event, and see if adding relevant variables to the dependency arrays in the hooks?

lib/components/form/call-taker/date-time-picker.tsx
daniel-heppner-ibigroup: thanks!

lib/components/form/call-taker/date-time-picker.tsx
daniel-heppner-ibigroup: the problem is these useEffect hooks are propagating updates from the store into the component, adding those deps would cause a loop. Otherwise I would have just replaced them with callbacks

lib/components/form/call-taker/date-time-picker.tsx
daniel-heppner-ibigroup: same as above, causes a loop

lib/components/form/call-taker/date-time-picker.tsx
daniel-heppner-ibigroup: Good note, added them.

lib/components/form/call-taker/date-time-picker.tsx
binh-dam-ibigroup: @daniel-heppner-ibigroup  Okay, could you add a comment about the undesired side effect?

daniel-heppner-ibigroup: There is one in [caf809c](https://github.com/opentripplanner/otp-react-redux/pull/1316/commits/caf809ca655e5b24f7c9705ae66de21a3b6426bb)

binh-dam-ibigroup (approved):
Let's give it a shot, thanks for the changes.

PR #1317, opened 2024-12-05
correctly make use of scoped vehicle ids

miles-grant-ibigroup:
Our behavior wasn't working because OTP `vehicleId` is now scoped. This adds support for a scoped `vehicleId` and therefore restores the correct behavior of the transit vehicle overlay.


daniel-heppner-ibigroup (approved):
LGTM

josh-willis-arcadis (approved):
looks great!

PR #1319, opened 2024-12-06
fix crash when applicableModeSettings is null

daniel-heppner-ibigroup:
**Description:**
Since this isn't a typescript file we didn't detect this sooner, but applicableModeSetting can be null which causes a crash. We fix that here with the trusty null coalescing operator.


amy-corson-ibigroup (approved):

miles-grant-ibigroup (approved):
Getting deja vu somehow

PR #1320, opened 2024-12-09
Nearby View: Add Location Field Search

miles-grant-ibigroup:
Adds a location field search field to the nearby view. There are still some styling concerns, and the current location issue is annoying as well. Would be nice to get that fixed.

Should we have it try to render a string on first load? If we did that then the placeholder text would never appear and that's why I'm against it.

Should we add using this thing to the Percy tests?

Screenshot 2024-12-09 at 1 00 40 PM


lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -286,6 +306,32 @@ function NearbyView({
       >
         {/* This is used to scroll to top */}
         <div aria-hidden ref={firstItemRef} />
+        <LocationField
```
amy-corson-ibigroup, line 334: you're missing a `currentPosition` prop here that's making it impossible to use current location

lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -139,6 +149,12 @@ function NearbyView({
     [nearbyViewCoords, currentPosition, map]
   )
 
+  const reverseCoords = (coords) => {
```
amy-corson-ibigroup: Can we add a `useEffect` that checks if the `currentPosition` coords are the same as `finalNearbyCoords` and then runs `reverseCoords`? Right now if I use current location, there's no feedback that my location was picked up correctly

lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -286,6 +306,32 @@ function NearbyView({
...
+            id: 'components.NearbyView.searchNearby'
+          })}
+          location={{
+            // Provide a 0 default in case the nearby view coords are null
+            lat: 0,
+            lon: 0,
+            ...nearbyViewCoords,
+            name: reversedPoint
+          }}
+          LocationIconComponent={() => (
```
amy-corson-ibigroup, line 350: Hmm this isn't showing up on chrome or safari
image

amy-corson-ibigroup: Also, when it does show up it feels v close to the placeholder text? Probably better suited to a OTP-UI PR, but I guess just flagging it here

image

lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -286,6 +306,32 @@ function NearbyView({
...
+          locationType="to"
+          onLocationSelected={(selection) => {
+            const { location } = selection
+            setViewedNearbyCoords(location)
+            map && zoomToPlace(map, location)
+            setReversedPoint(location.name || '')
+          }}
+          sortByDistance
+        />
         {loading && (
```
amy-corson-ibigroup: Loading spinner is showing up on the location field which is a little misleading? Could we position it down to the results container

image

amy-corson-ibigroup (changes requested):
This looks great!!! Found some opportunities for cleanup, also looks like there might be a few type errors

lib/components/viewers/nearby/nearby-view.tsx
miles-grant-ibigroup: Yeah these might both be otp-ui issues!

miles-grant-ibigroup:
These were some beautifully excellent catches thank you! All resolved

lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -328,15 +329,15 @@ function NearbyView({
...
+            }
           }}
           sortByDistance
         />
-        {loading && (
-          <FloatingLoadingIndicator>
-            <Loading extraSmall />
-          </FloatingLoadingIndicator>
-        )}
+        {loading && <Loading extraSmall />}
```
amy-corson-ibigroup, line 366: Nice

amy-corson-ibigroup (approved):
Looks great! Thanks for the changes (and sorry for how long it took me to look at them)

i18n/fr.yml
```diff
@@ -401,6 +401,7 @@ components:
     headsign: "{destination}"
     nearbyListIntro: Liste de {count} entités à proximité.
     nothingNearby: Aucun lieu à proximité.
+    searchNearby: Search nearby...
```
josh-willis-arcadis, line 404: Will this get flagged for wrong translation on weblate? I assume this was put here to pass the CI tests.

josh-willis-arcadis (approved):
LGTM!

i18n/fr.yml
miles-grant-ibigroup: Fingers crossed!

PR #1323, opened 2024-12-13
Bundle diet

binh-dam-ibigroup:
## Description

This PR reduces bundle size with an OTP-UI itinerary-body package that doesn't require the full fa-solid library.
After deduplication, production bundle size should be ~8.5MB (~1MB zipped), down from ~12MB.

## PR Checklist

Before | After screenshot:

before and after screenshot of bundle analyzer results


daniel-heppner-ibigroup (approved):
Excellent

josh-willis-arcadis (approved):
Good catch!

PR #1324, opened 2024-12-13
Prevent constant updating in nearby view

amy-corson-ibigroup:
**Description:**

Fixes constant rerendering of the nearby view by only updating `currentPosition` if the user moves .0001 lat/long away from their initial position. 

Is this the most efficient way to do this? Should this instead be set in the nearby view? Should we potentially calc the km distance between the two points instead, to make it more geographically consistent? Open to a lot of feedback on this 

**PR Checklist:**

### DIFF
--- a/lib/components/app/responsive-webapp.js
+++ b/lib/components/app/responsive-webapp.js
@@ -137,21 +137,24 @@ class ResponsiveWebapp extends Component {
       }
     }
 
-    navigator.geolocation.watchPosition(
-      // On success
-      (position) => {
-        const shouldUpdate = this.positionShouldUpdate(position)
-        if (shouldUpdate) {
-          receivedPositionResponse({ position })
-        }
-      },
-      // On error
-      (error) => {
-        console.log('error in watchPosition', error)
-      },
-      // Options
-      { enableHighAccuracy: true }
-    )
+    if (isMobile()) {
+      navigator.geolocation.watchPosition(
+        // On success
+        (position) => {
+          const shouldUpdate = this.positionShouldUpdate(position)
+          if (shouldUpdate) {
+            receivedPositionResponse({ position })
+          }
+        },
+        // On error
+        (error) => {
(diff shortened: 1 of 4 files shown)


lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -246,6 +247,8 @@ function NearbyView({
       </li>
     ))
 
+  console.log(nearbyItemList)
```
daniel-heppner-ibigroup: remove this!

daniel-heppner-ibigroup (changes requested):
- Update snapshots
- Fix percy tests
- Maybe add a comment that talks about approximately what the distance of .001 lat/lon is. I think it depends on location

Maybe we should just not update the position based on currentPosition changes?

lib/components/viewers/nearby/nearby-view.tsx
amy-corson-ibigroup: Oops!

amy-corson-ibigroup:
Changes to the percy snapshots are being caused by changes to the `h3`! Before the invisible header was taking up space, now it's not

daniel-heppner-ibigroup (approved):
Thanks for the fixes!

lib/components/app/responsive-webapp.js
```diff
@@ -119,6 +138,25 @@ class ResponsiveWebapp extends Component {
       }
     }
 
+    // Watch for position changing on mobile
+    if (isMobile()) {
+      navigator.geolocation.watchPosition(
+        // On success
+        (position) => {
+          const shouldUpdate = this.positionShouldUpdate(position)
+          if (shouldUpdate) {
```
miles-grant-ibigroup, line 147: You can inline this. It's only used once

miles-grant-ibigroup (approved):
Looks and works great! Wonderful work. Just one minor nitpick

PR #1329, opened 2024-12-19
strict filtering use combination modes

daniel-heppner-ibigroup:
**Description:**
We can't rely on `activeModes` because it misses addTransportModes that are added by mode settings. In one case, we added scooter and bike via the advanced settings/mode settings. These weren't included in the activeModes list. This change uses the post-processed combinations to determine what the active modes are for strict filtering.


binh-dam-ibigroup (approved):

miles-grant-ibigroup (approved):
Still nervous about this fix but I've tested it a bit and everything seems to be working

PR #1332, opened 2024-12-20
More suspense

binh-dam-ibigroup:
## Description

Quick attempt to lazy load print routes (and move printable-itinerary out of the main bundle) (see screenshot)
Uncompressed bundle size reduction about 500-600kB

image

## PR Checklist

### DIFF
--- a/lib/components/util/with-suspense.tsx
+++ b/lib/components/util/with-suspense.tsx
@@ -1,0 +1,16 @@
+import React, { ComponentType, Suspense } from 'react'
+
+/**
+ * Wraps a component with suspense, assuming the wrapped component is lazily loaded.
+ */
+export default function withSuspense(
+  WrappedComponent: ComponentType
+): ComponentType {
+  const suspensedComponent = (props: any): JSX.Element => (
+    <Suspense fallback={<span />}>
+      <WrappedComponent {...props} />
+    </Suspense>
+  )
+
+  return suspensedComponent
+}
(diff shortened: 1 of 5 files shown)


lib/components/util/with-suspense.tsx
```diff
@@ -0,0 +1,16 @@
+import React, { ComponentType, Suspense } from 'react'
+
+/**
+ * Wraps a component with suspense, assuming the wrapped component is lazily loaded.
+ */
+export default function withSuspense(
+  WrappedComponent: ComponentType
+): ComponentType {
+  const suspensedComponent = (props: any): JSX.Element => (
```
daniel-heppner-ibigroup: would `unknown` maybe be better here? Not sure. Or maybe you could even use a generic to use the props type from the wrapped component?

daniel-heppner-ibigroup (approved):
I thought we already got the pdf export library out of the bundle! Anyway, maybe we can make the types on your HOC more robust?

binh-dam-ibigroup:
> I thought we already got the pdf export library out of the bundle!

This PR deals with the OTP-UI printable-itinerary package.

lib/components/util/with-suspense.tsx
binh-dam-ibigroup: Great idea! Updated in https://github.com/opentripplanner/otp-react-redux/pull/1332/commits/f2af8414da633b2dadea7c1b53dabd2fff44ca7c

miles-grant-ibigroup (approved):
Seeing some flashing on the account page when first clicking around (which is why I initially didn't suspend those components) but I think this tradeoff is worth it for the smaller bundle size!

binh-dam-ibigroup:
> Seeing some flashing on the account page when first clicking around

For that, eventually we'd have to lazy-load the portion that is below the header and subheader.

PR #1333, opened 2024-12-23
Support Floating Vehicle Bunching

miles-grant-ibigroup:
**Description:**
Screenshot 2024-12-23 at 2 44 09 PM

**PR Checklist:**


lib/components/viewers/nearby/vehicle-rent.tsx
```diff
@@ -119,6 +120,18 @@ const Vehicle = ({
         )}
         {fromToSlot}
       </CardBody>
+      {vehicle.additionalCount > 0 && (
+        <CardFooter>
+          <span>
+            <FormattedMessage
+              id="components.NearbyView.additionalVehicles"
```
binh-dam-ibigroup, line 145: Was this layout more advantageous than placing this text underneath the distance label (see illustration)?
image

lib/components/viewers/nearby/styled.tsx
```diff
@@ -108,3 +109,30 @@ export const Scrollable = styled.div`
...
+  font-size: 15px;
+  padding: 5px 12px;
+  margin-top: -6px;
+
+  span {
+    padding-top: 5px;
+    display: block;
+  }
+
+  &::before {
```
binh-dam-ibigroup, line 127: Is this for the margin between the footer and the body of the card, and can it be done using `margin-top` of `CardFooter` itself instead?

lib/components/viewers/nearby/vehicle-rent.tsx
```diff
@@ -119,6 +120,18 @@ const Vehicle = ({
         )}
         {fromToSlot}
       </CardBody>
+      {vehicle.additionalCount > 0 && (
+        <CardFooter>
+          <span>
```
binh-dam-ibigroup, line 143: Is the span needed?

binh-dam-ibigroup (approved):
Looks good. See if you can simplify the CSS. My comment about putting the text below the distance is not blocking.

lib/components/viewers/nearby/vehicle-rent.tsx
miles-grant-ibigroup: I thought about this but it gets busy especially in the case of a 2-line title

lib/components/viewers/nearby/vehicle-rent.tsx
daniel-heppner-ibigroup: The span is keeping the text separate from the &::before

lib/components/viewers/nearby/styled.tsx
daniel-heppner-ibigroup: This is used to create the border radius effect on the body of the card, so it looks like the footer is peeking out from below it. 
image

amy-corson-ibigroup (commented):
Hiding all these vehicles at once definitely makes the UI cleaner, and I like the design of the footer, but I'm wondering if we could expand/collapse the list of vehicles instead of just filtering them all out? What is the use of knowing there are 40+ scooters nearby if I can't plan a trip using one?

miles-grant-ibigroup:
> Hiding all these vehicles at once definitely makes the UI cleaner, and I like the design of the footer, but I'm wondering if we could expand/collapse the list of vehicles instead of just filtering them all out? What is the use of knowing there are 40+ scooters nearby if I can't plan a trip using one?

We'll have to think of something clever here... I think a giant list of scooters will never be helpful. Now that we are keeping all the scooters on the map a user could use the map to do that?

What user is trying to select one of 40 scooters in a 15 meter radius to travel from?

amy-corson-ibigroup:
Won't the radius be the same one passed from the config? I'm pretty sure that radius is 2000m by default. This solution might work if you're looking for scooters within eyesight of one another, but not if you're looking at two scooters over a mile apart. For example: 
image
Plus, mobile users cannot see a map in nearby view. The only way to see those entities on the map is by using the from/to buttons on the card. This PR essentially makes it so mobile users can see the scooter closest to them in nearby view and no other scooters.

miles-grant-ibigroup:
> Won't the radius be the same one passed from the config? I'm pretty sure that radius is 2000m by default. This solution might work if you're looking for scooters within eyesight of one another, but not if you're looking at two scooters over a mile apart. 

Ah good point. I think we changed our implementation at some point because it still looked too busy with multiple scooter cards.

> Plus, mobile users cannot see a map in nearby view. The only way to see those entities on the map is by using the from/to buttons on the card. This PR essentially makes it so mobile users can see the scooter closest to them in nearby view and no other scooters.

That's right, but what is the use case for a mobile user wanting to find a scooter that is not the one that is closest to them?

miles-grant-ibigroup:
@amy-corson-ibigroup I've added more limited bunching and a link to rent the scooters (if applicable)

amy-corson-ibigroup:
> That's right, but what is the use case for a mobile user wanting to find a scooter that is not the one that is closest to them?

One example would be if you are traveling with a friend. I've added new user stories to the tech call so we can discuss. Even with the updated radius I still do not think this is the correct approach for this issue.

miles-grant-ibigroup:
We will re-do this elsewhere

PR #1336, opened 2025-01-07
Pattern viewer configuration and styling

amy-corson-ibigroup:
**Description:**
- Adds config item `sortRoutePatternsByVehicles` which makes sorting the route patterns by vehicle count in the dropdown optional.
- Fixes some minor styling issues for the pattern viewer select dropdown, to prevent the headsigns from clipping.

**PR Checklist:**

| Before | After |
|--------|-------|
|image|image|
|image|image
|


lib/components/viewers/route-details.tsx
```diff
@@ -87,33 +93,44 @@ class RouteDetails extends Component<Props> {
...
-      if (aVehicleCount === bVehicleCount) {
-        return b.geometryLength - a.geometryLength
-      }
-      return bVehicleCount - aVehicleCount
-    })
+    )
+
+    const headsigns = sortPatternsByVehicleCount
+      ? unsortedHeadsigns.sort((a, b) => {
+          // sort by number of vehicles on that pattern
```
miles-grant-ibigroup: We can avoid adding the new variable and the ternary if you instead add an `if(!sortPatternsByVehicleCount) return 0` at the top of the original sort method

lib/components/viewers/route-details.tsx
```diff
@@ -231,11 +248,22 @@ class RouteDetails extends Component<Props> {
   }
 }
 
+const mapStateToProps = (state: any) => {
+  const { sortRoutePatternsByVehicleCount } = state.otp.config.routeViewer
```
miles-grant-ibigroup:
```suggestion
  const { sortRoutePatternsByVehicleCount } = state.otp.config?.routeViewer
```
route viewer is not guaranteed to be present!

miles-grant-ibigroup (changes requested):
We can make this much smaller!

lib/components/viewers/route-details.tsx
amy-corson-ibigroup: Isn't it more efficient to avoid iterating through the `sort` if we don't have to?

miles-grant-ibigroup: If we always return 0 it's still O(n) and the code is much cleaner

miles-grant-ibigroup (approved):
Wonderful!

lib/components/viewers/route-details.tsx
```diff
@@ -231,11 +245,23 @@ class RouteDetails extends Component<Props> {
   }
 }
 
+const mapStateToProps = (state: any) => {
```
binh-dam-ibigroup: Use `state: AppReduxState` (import accordingly) and that will help reduce the `?` on line 250 below.

lib/components/viewers/route-details.tsx
```diff
@@ -231,11 +245,23 @@ class RouteDetails extends Component<Props> {
...
+  const sortRoutePatternsByVehicleCount =
+    state.otp?.config?.routeViewer?.sortRoutePatternsByVehicleCount
```
binh-dam-ibigroup: You can remove `?` after `otp` and `config` after using the state type.

binh-dam-ibigroup (approved):
One type tweak, please, and it should be good to go.

PR #1338, opened 2025-01-15
Update Query Time if "Plan Trip" clicked

amy-corson-ibigroup:
**Description:**
Previously, the `updateQueryTimeIfLeavingNow` function was being called incorrectly in a way that wasn't connected to redux. This ensures the time is updated on every "plan trip" click. 

**PR Checklist:**


miles-grant-ibigroup (approved):
Changes make sense. Part of me wants to see this moved to the plan request submission logic to avoid having to drill it everywhere but things seem to work so good enough for now!

josh-willis-arcadis (approved):
LGTM!

PR #1345, opened 2025-01-27
handle failed connection error, remove otp1 errors

daniel-heppner-ibigroup:
**Description:**
Adds support for showing an error when there's a failed connection. It also removes the old errors prop which seemed to just be for OTP1 things. Does this cause any issues? I don't think we used those errors anymore.


lib/components/narrative/narrative-itineraries-header.tsx
```diff
@@ -175,13 +165,6 @@ export default function NarrativeItinerariesHeader({
               >
                 {pending ? searching : itinerariesFound}
               </h1>
-              {errors.length > 0 && (
-                <IssueButton onClick={onToggleShowErrors}>
```
miles-grant-ibigroup, line 179: We don't need `onToggleShowErrors` anymore

miles-grant-ibigroup: There should be an entire error view that can go too

lib/components/narrative/narrative-itineraries.js
```diff
@@ -556,8 +536,24 @@ const reduceErrorsFromResponse = (acc, cur) => {
...
+  const itineraries = getActiveItineraries(state)
+  const search = getActiveSearch(state)
+  if (errors && itineraries.length === 0 && !pending) {
+    errors.NO_TRANSIT_CONNECTION = new Set()
+  }
+  if (search?.response?.some((res) => res.error)) {
+    errors.SYSTEM_ERROR = new Set()
+  }
 
+  return errors
```
miles-grant-ibigroup: We have put in a ton of effort to explicitly avoid doing this. The errors should be generated within the datastore, not in the view handler

miles-grant-ibigroup (changes requested):
Could I request to break this into 2 PRs? I like the conversion of the batch ui to use the new error viewer, but the "default" error stuff is likely going to break a lot of the previous approaches used to generate "default" errors. It also goes against the conventions we established around where errors should be generated. Not in the render views!

Also to completely remove the batch ui error display please remove the rest of the react code and also the i18n keys

lib/components/narrative/metro/metro-error-renderer.tsx
```diff
@@ -83,11 +83,11 @@ const ErrorRenderer = ({
           return (
             <Container key={error}>
               <Icon Icon={ExclamationCircle} size="3x" />
-              <h3>
+              <h2>
```
miles-grant-ibigroup, line 87: Why did we bump this up? I think h3 is cleaner

lib/util/state.js
```diff
@@ -119,16 +119,32 @@ export const getActiveItineraries = createSelector(
...
+    return acc
+  }, {})
   const itineraries = getActiveItineraries(state)
-  if (errors && itineraries.length === 0 && !pending) {
-    errors.NO_TRANSIT_CONNECTION = new Set()
+  const additionalErrors = {}
+  if (itineraries.length === 0 && !activeSearch?.pending) {
+    additionalErrors.NO_TRANSIT_CONNECTION = new Set()
+  }
+  if (activeSearch?.response?.some((res) => res.error)) {
```
miles-grant-ibigroup, line 143: Can we use `getActiveItineraries` here instead?

lib/util/state.js
```diff
@@ -119,16 +119,32 @@ export const getActiveItineraries = createSelector(
...
+    additionalErrors.SYSTEM_ERROR = new Set()
   }
-  return errors
+
+  return { ...routingErrors, ...additionalErrors }
```
miles-grant-ibigroup: I like your thinking here but doesn't this open the door to duplicates?

miles-grant-ibigroup (changes requested):
A few small issues to work out

lib/components/narrative/metro/metro-error-renderer.tsx
daniel-heppner-ibigroup: a11y issues because the headers were not sequential. I think we never displayed an error in the a11y tests before and now we do?

lib/util/state.js
daniel-heppner-ibigroup: No duplicates because it's an object with keys!

lib/util/state.js
daniel-heppner-ibigroup: we use it above but here we need the search to get the error value

lib/util/state.js
miles-grant-ibigroup: Whoops duh!

lib/components/narrative/metro/metro-error-renderer.tsx
miles-grant-ibigroup: Can we style it back down then? It's way too big as it stands

miles-grant-ibigroup (changes requested):
Will be ok once we get the h2 size right. We should add error display to the Percy tests

miles-grant-ibigroup (approved):
A bit nervous about all these structural changes but let's merge and hope for the best. Approval conditional on fixing those new Percy tests. It's a good idea

lib/components/narrative/metro/metro-error-renderer.tsx
```diff
@@ -23,9 +23,10 @@ const Container = styled.li`
   margin: 0;
   padding: 0 1em;
 
-  h3 {
+  h2 {
     grid-column: 2;
     grid-row: 1;
+    font-size: 24px;
```
binh-dam-ibigroup: Sort CSS

i18n/en-US.yml
```diff
@@ -355,7 +355,6 @@ components:
     itinerariesFound: >-
       {itineraryNum, plural, one {# itinerary found} other {# itineraries found}
       }
-    numIssues: "{issueNum, plural, one {# issue} other {# issues} }"
```
binh-dam-ibigroup, line 358: Can you remove this entry in the other languages too?

lib/components/narrative/narrative-itineraries-header.tsx
binh-dam-ibigroup: Remove `onToggleShowErrors` from the props.

lib/components/narrative/narrative-itineraries.js
```diff
@@ -448,7 +444,6 @@ class NarrativeItineraries extends Component {
         <NarrativeItinerariesHeader
           customBatchUiBackground={customBatchUiBackground}
           enabledSortModes={enabledSortModes}
-          errors={errors}
```
binh-dam-ibigroup, line 451: Also remove:
- prop `onToggleShowErrors`
- the corresponding component state `showErrors`, and propagate as needed.
- prop `showingErrors` (remove prop from `NarrativeItinerariesHeader` too)

lib/util/state.js
```diff
@@ -119,16 +119,32 @@ export const getActiveItineraries = createSelector(
...
 
-/**
- * If there are no itineraries, generate an error
- * so we're not left with an empty results container.
- */
-export const createNoItinerariesErr = (state, errors, pending) => {
+export function getActiveSearchErrors(state) {
+  const activeSearch = getActiveSearch(state)
+  if (!activeSearch) return {}
+  const routingErrors = activeSearch?.response?.reduce((acc, cur) => {
```
binh-dam-ibigroup: Maybe rename the top-level `routingErrors` to something else to not confuse with the inner loop veriable below.

lib/components/narrative/narrative-itineraries.js
```diff
@@ -478,86 +473,57 @@ class NarrativeItineraries extends Component {
               itinerariesLength={mergedItineraries.length}
             />
           )}
-          {showingErrors || mergedItineraries.length === 0 ? (
-            <NarrativeItinerariesErrors
```
binh-dam-ibigroup, line 482: Delete corresponding component and import.

lib/components/narrative/narrative-itineraries.js
```diff
@@ -385,8 +383,6 @@ class NarrativeItineraries extends Component {
       activeSearch,
       customBatchUiBackground,
       enabledSortModes,
-      errorMessages,
```
binh-dam-ibigroup, line 388: Remove other instances of `errorMessages` in this file.

binh-dam-ibigroup (changes requested):
See if you can remove the unused props and components as indicated?

lib/components/narrative/narrative-itineraries.js
daniel-heppner-ibigroup: good catch.

binh-dam-ibigroup (approved):
Thanks for the cleanup!

PR #1346, opened 2025-01-30
correct `showScheduleDeviation`, enable flex

miles-grant-ibigroup:
`showScheduleDeviation` was incorrectly configured to not pull the config from the `itinerary` settings object. Additionally there was no flex support. This is now corrected.

### DIFF
--- a/lib/util/config-types.ts
+++ b/lib/util/config-types.ts
@@ -301,6 +301,8 @@ export interface ItineraryConfig {
   showLegDurations?: boolean
   showPlanFirstLastButtons?: boolean
   showRouteFares?: boolean
+  /** Whether to show the x minutes late/early in the itinerary body */
+  showScheduleDeviation?: boolean
   sortModes?: ItinerarySortOption[]
   syncSortWithDepartArrive?: boolean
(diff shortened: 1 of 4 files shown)


lib/util/config-types.ts
```diff
@@ -301,6 +301,8 @@ export interface ItineraryConfig {
   showLegDurations?: boolean
   showPlanFirstLastButtons?: boolean
   showRouteFares?: boolean
+  /** Whether to show the x minutes late/early in the itinerary body */
```
amy-corson-ibigroup, line 304: Weird that `AppConfig` has these comments and `ItineraryConfig` doesn't, but I guess having this one be the only item with a comment doesn't really matter?

amy-corson-ibigroup (approved):
While we're in here, can we change the example config to indicate the default for `showScheduleDeviation` is `false`

lib/util/config-types.ts
miles-grant-ibigroup: Yeah it's the first of many in this file!

daniel-heppner-ibigroup (approved):
Looks good

PR #1347, opened 2025-02-03
Lazily-load company icons

binh-dam-ibigroup:
## Description

This PR creates new shared components to lazily-load and render company icons (reduces initial bundle size by 100kB/50kB compressed).

**PR Checklist:**


lib/components/viewers/nearby/vehicle-rent.tsx
```diff
@@ -106,7 +95,15 @@ const Vehicle = ({
       <CardHeader>
         <CardTitle>
           <IconWithText
-            icon={<StationIcon companyLabel={companyLabel} vehicle={vehicle} />}
+            icon={
+              <CompanyIcon
+                company={vehicle.network}
+                fallbackContent={<span>{getVehicleIcon(formFactor)}</span>}
```
daniel-heppner-ibigroup: what's the span for?

daniel-heppner-ibigroup (approved):
This works really well! Nice work. I wonder if there's a way to test that we don't accidentally import the icons again in the future.

lib/components/viewers/nearby/vehicle-rent.tsx
binh-dam-ibigroup: It was there when you wrote this component originally! Seems that it is not needed, so I removed it in https://github.com/opentripplanner/otp-react-redux/pull/1347/commits/31a4c8ded0be9772d5e25b0a6cdbb28ec10f2dc6.

miles-grant-ibigroup (approved):
Code looks clean, and everything still works! However, I'm noticing a bit of a slowdown when opening itineraries, although that may be a coincidence? Still might be something to keep an eye on

PR #1351, opened 2025-02-12
Add character limit of 30 to place editor

amy-corson-ibigroup:
**Description:**
Adds a character limit to the place name editor, and validation to yup. Also adds an aria alert to AT users to let them know how far over the character limit they are. This allows for a slightly cleaner UX experience on smaller screen sizes in the location field, as previous there was no upper limit to how large a place name would be and the name was cut off around 30 characters anyway.

Should this be added to saved trip name as well for consistency?

**PR Checklist:**

| Without error | With error |
|--------|-------|
|image|image|


lib/components/user/places/place-editor.tsx
```diff
@@ -135,14 +137,39 @@ class PlaceEditor extends Component<Props> {
...
-                <FormattedMessage id="components.PlaceEditor.namePrompt" />
+              <ControlLabel
+                htmlFor="name"
+                style={{
+                  display: 'flex',
+                  flexDirection: 'row',
+                  justifyContent: 'space-between'
+                }}
+              >
+                <>
```
daniel-heppner-ibigroup: Is this fragment needed?

lib/components/user/places/place-editor.tsx
```diff
@@ -135,14 +137,39 @@ class PlaceEditor extends Component<Props> {
...
+                  <FormattedMessage id="components.PlaceEditor.namePrompt" />
+                </>
+                <div style={{ fontWeight: 300 }}>
```
daniel-heppner-ibigroup: div isn't allowed in label I don't think. I think maybe the actual label should just be prompt and the remaining character should be outside the label

daniel-heppner-ibigroup (approved):
Please fix the HTML issues and then you're good to go!

lib/components/user/places/place-editor.tsx
amy-corson-ibigroup: Good catch thank you!

lib/components/user/places/favorite-place-screen.js
```diff
@@ -123,7 +123,7 @@ class FavoritePlaceScreen extends Component {
     if (place?.name === null) {
       place.name = ''
     }
-
+
```
binh-dam-ibigroup: This whitespace change can be reverted.

lib/components/user/places/place-editor.tsx
binh-dam-ibigroup: Do we need this fragment?

lib/components/user/places/place-editor.tsx
binh-dam-ibigroup: This comment was not addressed. Worst case, replace the div with a span.

binh-dam-ibigroup (commented):
Did you need to address https://github.com/opentripplanner/otp-react-redux/pull/1351/files#r1968507399?

lib/components/user/places/place-editor.tsx
amy-corson-ibigroup: Whoops, perhaps my local commit didn't get pushed? This should be fixed in 4b8fd0247a0843f3ec30fe4c0a50869a5d3bcc6d

lib/components/user/places/place-editor.tsx
amy-corson-ibigroup: I don't think so! Thanks

lib/components/user/places/place-editor.tsx
```diff
@@ -144,6 +149,26 @@ class PlaceEditor extends Component<Props> {
               <ControlLabel htmlFor="name">
                 <FormattedMessage id="components.PlaceEditor.namePrompt" />
               </ControlLabel>
+              <div
```
binh-dam-ibigroup, line 152: Consider adding `aria-hidden` to this block, so that the exceeded character limit isn't announced twice, once here and once in the help block below.

binh-dam-ibigroup (approved):

lib/components/user/places/place-editor.tsx
amy-corson-ibigroup: Good idea, thank you!

PR #1353, opened 2025-02-18
Add haptic feedback option

binh-dam-ibigroup:
**Description:**

## PR Checklist

| Before | After |
|--------|-------|
| before screenshot | after screenshot  |


amy-corson-ibigroup (approved):
Not blocking, but can you elaborate in the description the reason for the change? It seems unconventional to control mobile phone haptics through the web app, I think you'd usually see this option in the mobile app settings. We might also consider making this option configurable?

binh-dam-ibigroup:
> Not blocking, but can you elaborate in the description the reason for the change?

Mobile app folks decided without coordinating with us that "haptic" would reside with other notification options, and they are on crunch time now to implement that.

> It seems unconventional to control mobile phone haptics through the web app, I think you'd usually see this option in the mobile app settings. We might also consider making this option configurable?

Another approach would be to simply "hide" the check box for "haptic" notifications. What do you think? We still need to render it, otherwise, saving a user profile would result in the "haptic" option becoming unchecked on the mobile app.

miles-grant-ibigroup (changes requested):
I think I'd prefer rendering the option but making it `display: none`. Otherwise someone who doesn't use the mobile app will be very confused!

binh-dam-ibigroup:
> I think I'd prefer rendering the option but making it display: none

@miles-grant-ibigroup It turns out that formik is doing some magic to keep "haptic" in the user's notification channels, even when saving user settings using the dev branch, so this PR is not needed.

PR #1356, opened 2025-02-25
Improve recent searches

binh-dam-ibigroup:
## Description

This PR remove duplicate (i.e. same-text) searches from the recents list in the location field and from the recent trip searches.
In addition, the last picked search from location field or recent trip searches will be placed at the top of the list.

Behind the scenes, a performance improvement is done to avoid downloading twice the entire set recent searches of the logged-in user.

## PR Checklist


amy-corson-ibigroup (approved):
Very clean! Thank you for the changes and the added tests.

daniel-heppner-ibigroup (approved):
Works great! Clean code changes, and thanks for the tests

PR #1357, opened 2025-02-27
take stoptime headsign into account when filtering nearby stop times

miles-grant-ibigroup:
**Description:**
Previously we weren't taking this into account. As a result we were merging departures that went on different deviations. 

**PR Checklist:**


daniel-heppner-ibigroup (approved):
thanks for the fix

binh-dam-ibigroup (approved):

PR #1360, opened 2025-03-07
Update OTP-UI March 2025

miles-grant-ibigroup:
Lots of updates


amy-corson-ibigroup (approved):
Looks good! It'd be nice for these package updates if we could include or links to the individual OTP-UI releases so it's a little easier to see exactly what we're updating?

amy-corson-ibigroup:
Also looks like you gotta update snapshots

josh-willis-arcadis (approved):

PR #1361, opened 2025-03-07
feat: reverse geocode the current location

daniel-heppner-ibigroup:
**Description:**
This PR reverse geocodes the user's current location when selected from the location field.


lib/actions/map.js
```diff
@@ -74,15 +74,49 @@ export function setLocation(payload) {
...
+        if (geocodedLocation.name) {
+          // Create a new location object with the geocoded name but keep everything else the same
+          const updatedLocation = {
+            ...location,
+            name: geocodedLocation.name
+          }
+
+          // Update the location with the geocoded name
+          dispatch(
+            settingLocation({
```
amy-corson-ibigroup: Instead of calling reverse geocoder in this function, just replace `settingLocation` with the `setLocation` call above and add `reverseGeocode: true` to the payload

amy-corson-ibigroup (changes requested):
Works well, but the `setLocation` function on line 41 of this file already has a method to reverse geocode! Pretty sure we can just use that instead?

lib/actions/map.js
daniel-heppner-ibigroup: Ooo excellent suggestion, I forgot that was in there. I'm impressed you realized that! great awareness

amy-corson-ibigroup (approved):
Looks good! Thanks for the changes!!

josh-willis-arcadis (approved):
LGTM!

PR #1369, opened 2025-03-20
Fix Recent Trips Actions

binh-dam-ibigroup:
## Description

This PR addresses incorrect behavior when clicking a recent search:
- The modes will be the current selection of the mode selector
- The date/time will be displayed correctly (instead of e.g. 'Yesterday, Leave Now')

**PR Checklist:**


lib/components/form/user-settings.js
```diff
@@ -337,7 +337,13 @@ const TripRequest = ({
       // In previous versions of this code, tripRequest.id was passed to setQueryParam,
       // however that causes an extra entry to be added to the recent searches on click,
       // in addition to the entry added when clicking the "Plan" (magnifying glass) button.
-      setQueryParam(query)
+      // Also, pass query params without the mode, otherwise the selection from the mode selector will not apply.
+      const { modes, ...queryWithoutModes } = query
+      // If depart/arrive is 'NOW', change that to 'DEPART' so that the correct time is shown in the date button.
+      if (queryWithoutModes.departArrive === 'NOW') {
+        queryWithoutModes.departArrive = 'DEPART'
+      }
```
amy-corson-ibigroup: Just an idea, but instead of populating the d/t selector to reflect the time the trip was originally planned, could we set the d/t selector by default to "today/leave now"? Or even avoid changing the d/t selector? I think in most cases the user is going to be trying to re-plan the trip for a time in the future versus look at results for a time that's in the past

amy-corson-ibigroup (approved):
Looks really good, just an idea to make the user experience a little smoother. Thank you for the changes!

lib/components/form/user-settings.js
miles-grant-ibigroup: I agree. User is most likely trying to plan the same trip again but now. Should we have it set to "leave now" instead?

miles-grant-ibigroup: That might be out of the scope of this PR though?

miles-grant-ibigroup (approved):
Code looks good, but agree with Amy that we should maybe review this behavior

lib/components/form/user-settings.js
binh-dam-ibigroup: https://github.com/opentripplanner/otp-react-redux/pull/1369/commits/826d86038e63b080195310a726b15028a6365c6d leaves the settings of the date time selector as they were prior to selecting the recent search entry, let me know what you think @amy-corson-ibigroup  and @miles-grant-ibigroup

miles-grant-ibigroup (approved):
I think we're good to go!

amy-corson-ibigroup (approved):
Looks great, thank you for the changes!

Issue #1359, opened 2025-03-06
FETCH_NEARBY_ERROR


nukeador:
After pull the latest git version of the code, when I get into a stop I get this error
```

Uncaught TypeError: name is undefined
    getCompanyIcon webpack-internal:///./node_modules/@opentripplanner/icons/lib/companies/index.js:33
    CompanyIcon webpack-internal:///./lib/components/util/company-icon-internal.tsx:19
    React 12
    unstable_runWithPriority webpack-internal:///./node_modules/scheduler/cjs/scheduler.development.js:653
    React 8
    unstable_runWithPriority webpack-internal:///./node_modules/scheduler/cjs/scheduler.development.js:653
    React 16
    unstable_runWithPriority webpack-internal:///./node_modules/scheduler/cjs/scheduler.development.js:653
    React 4
    Redux 5
    routerMiddleware webpack-internal:///./node_modules/connected-react-router/esm/middleware.js:28
    Redux 2
    _callee2$ webpack-internal:///./lib/actions/ui.js:502
    Babel 3
    asyncGeneratorStep webpack-internal:///./lib/actions/ui.js:48
    _next webpack-internal:///./lib/actions/ui.js:49
    promise callback*asyncGeneratorStep webpack-internal:///./lib/actions/ui.js:48
    _next webpack-internal:///./lib/actions/ui.js:49
    _asyncToGenerator webpack-internal:///./lib/actions/ui.js:49
    _asyncToGenerator webpack-internal:///./lib/actions/ui.js:49
    setLocale webpack-internal:///./lib/actions/ui.js:522
    Redux 2
    _initializeOrUpdateLocale webpack-internal:///./lib/components/app/responsive-webapp.js:394
    RouterWrapperWithAuth0 webpack-internal:///./lib/components/app/responsive-webapp.js:372
    React 13
    <anonymous> webpack-internal:///./lib/main.js:81
    <anonymous> webpack-internal:///./lib/main.js:118
    js http://localhost:9966/bundle.js:3242
    __webpack_require__ http://localhost:9966/bundle.js:836
    fn http://localhost:9966/bundle.js:130
    1 http://localhost:9966/bundle.js:60153
    __webpack_require__ http://localhost:9966/bundle.js:836
    <anonymous> http://localhost:9966/bundle.js:1011
    <anonymous> http://localhost:9966/bundle.js:1014
index.js:33

```

miles-grant-ibigroup:
[...]

PR #1370, opened 2025-03-20
fix: prevent "name is undefined" error in CompanyIcon component

nukeador:
I made three key changes to fix the "Uncaught TypeError: name is undefined" error reported in #1359 

In company-icon-internal.tsx:

Added a check for company existence before calling getCompanyIcon(company)

Changed from: const CompanyIcon = getCompanyIcon ? getCompanyIcon(company) : null
To: const CompanyIcon = getCompanyIcon && company ? getCompanyIcon(company) : null

In vehicle-rent.tsx:

Added a fallback empty string for the company prop
Changed from: company={vehicle.network}
To: company={vehicle.network || ''}

In rental-station.tsx:
Added a fallback empty string for the company prop
Changed from: company={company}
To: company={company || ''}

The error was occurring because the getCompanyIcon function in the @opentripplanner/icons package was trying to access a property of name when company was undefined or null. These changes ensure that we only call getCompanyIcon when company has a value, and we provide empty string fallbacks when passing company props.

These changes prevent the error by ensuring we never pass undefined values to functions that expect strings.


miles-grant-ibigroup (approved):
These changes look good! The blank fallback strings give me a bit of pause but I'll let @binh-dam-ibigroup think on those

binh-dam-ibigroup (approved):
I don't think the fixes interfere with anything. Thanks for making them!

PR #1371, opened 2025-03-26
add stations support to nearby view

daniel-heppner-ibigroup:
**Description:**
Adds support for Stations with sub-stops in nearby view. This version is based on dev instead of the scooter merging.


lib/components/viewers/nearby/stop.tsx
```diff
@@ -19,12 +20,21 @@ const { getUserTimezone } = coreUtils.time
 export const fullTimestamp = (stoptime: StopTime) =>
   (stoptime.serviceDay || 0) + (stoptime.realtimeDeparture || 0)
 
+// Style for child stop headers
+const ChildStopHeader = styled.div`
+  background-color: #f5f5f5;
```
miles-grant-ibigroup: Why is the color gray? Can it be transparent/white instead?

lib/components/viewers/nearby/stop.tsx
```diff
@@ -19,12 +20,21 @@ const { getUserTimezone } = coreUtils.time
...
+  border-radius: 4px 4px 0 0;
+  font-size: 14px;
+  font-weight: bold;
```
miles-grant-ibigroup: This doesn't look great. Can we do hierarchy with just font size?

lib/components/viewers/nearby/stop.tsx
```diff
@@ -19,12 +20,21 @@ const { getUserTimezone } = coreUtils.time
...
+  padding: 10px 15px;
```
miles-grant-ibigroup: Can we have the padding match the units we're using elsewhere? I think we're on units of 8

lib/components/viewers/nearby/stop.tsx
```diff
@@ -95,20 +105,40 @@ const Stop = ({
         alwaysShowLongName={nearbyViewConfig?.alwaysShowLongName}
         homeTimezone={homeTimezone}
         key={index}
+        noRoundedBottom={noRoundedBottom}
```
miles-grant-ibigroup: Any way we can avoid a prop and use css instead?

lib/components/viewers/nearby/stop.tsx
```diff
@@ -95,20 +105,40 @@ const Stop = ({
...
+  )
+
   const inHomeTimezone = homeTimezone && homeTimezone === getUserTimezone()
   const timezoneWarning = !inHomeTimezone && (
     <StyledAlert>
       <TimezoneWarning homeTimezone={homeTimezone} />
     </StyledAlert>
   )
 
+  console.log(patternRows)
```
miles-grant-ibigroup: oopsie

lib/components/viewers/styled.ts
```diff
@@ -145,8 +148,10 @@ export const PatternRowItem = styled.li<{ roundedTop?: boolean }>`
   }
 
   &:last-of-type .header.stop-view {
```
miles-grant-ibigroup: Can we not take advantage of selectors instead of adding a new prop? Would make a lot of this code much much cleaner

miles-grant-ibigroup (changes requested):
Only a few small cleanups needed! Mostly in getting rid of that new prop and making the new headers less bold. Can we have it match the From Here/To Here? 

Also "view schedule" doesn't work

daniel-heppner-ibigroup:
I think I made some big improvements in the latest commit, let me know what you think. I'm still not sure about the styling for the headers.

lib/components/viewers/nearby/stop-card-header.tsx
```diff
@@ -78,14 +78,16 @@ const StopCardHeader = ({
...
-            style={{ color: 'inherit', fontSize: 'small' }}
-            to={actionPath}
-            toParams={actionParams}
-          >
-            <IconWithText Icon={actionIcon}>{actionText}</IconWithText>
-          </Link>
+          {actionPath && actionText ? (
+            <Link
+              className="pull-right"
+              style={{ color: 'inherit', fontSize: 'small' }}
```
miles-grant-ibigroup: Any way we can get some hover underlines on this?

lib/components/viewers/pattern-row.tsx
```diff
@@ -25,6 +25,7 @@ import StopTimeCell from './stop-time-cell'
 type Props = {
   alwaysShowLongName?: boolean
   homeTimezone?: string
+  noRoundedBottom?: boolean
```
miles-grant-ibigroup: I think this can be dropped now

lib/components/viewers/pattern-row.tsx
```diff
@@ -59,11 +60,11 @@ const renderDay = (homeTimezone: string, day: number): JSX.Element => {
  * Represents a single pattern row for displaying arrival times in the stop
  * viewer.
  */
+// eslint-disable-next-line complexity
```
miles-grant-ibigroup, line 61: I think this can be dropped now

miles-grant-ibigroup (changes requested):
Great work with the CSS! I'm in agreement something is off with the headers now. Maybe same font size but higher font weight? Otherwise code looks pretty good and working well for me. Will leave it up to the second reviewer to nitpick spacings and paddings

lib/components/viewers/nearby/stop-card-header.tsx
daniel-heppner-ibigroup: image
??

daniel-heppner-ibigroup: I moved it to a styled component anyway

lib/components/viewers/pattern-row.tsx
daniel-heppner-ibigroup: still needed

lib/components/viewers/nearby/stop-card-header.tsx
miles-grant-ibigroup: Probably a safari issue

lib/components/viewers/pattern-row.tsx
miles-grant-ibigroup: It's still here

lib/components/viewers/pattern-row.tsx
miles-grant-ibigroup: Are you sure

miles-grant-ibigroup (approved):
Beautiful! Still a few rough edges but happy for this to go to second review. Snapshots seem messed up you might want to clear your jest cache.

lib/components/viewers/pattern-row.tsx
daniel-heppner-ibigroup: image

lib/components/viewers/nearby/stop.tsx
```diff
@@ -3,28 +3,44 @@ import { connect } from 'react-redux'
...
-import { Card, PatternRowContainer, StyledAlert } from './styled'
 import StopCardHeader from './stop-card-header'
 
 const { getUserTimezone } = coreUtils.time
 
 export const fullTimestamp = (stoptime: StopTime) =>
   (stoptime.serviceDay || 0) + (stoptime.realtimeDeparture || 0)
 
+// Style for child stop headers
+const ChildStopHeader = styled.div`
```
amy-corson-ibigroup, line 26: Can we add an `align-items: center` here? The header and link are looking a little misaligned 
image

lib/components/viewers/nearby/stop.tsx
```diff
@@ -3,28 +3,44 @@ import { connect } from 'react-redux'
...
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  padding: 8px 16px;
```
amy-corson-ibigroup: I think `10px 16px` could make these labels look a little less cramped!

lib/components/viewers/nearby/stop.tsx
```diff
@@ -96,35 +111,90 @@ const Stop = ({
...
-          <FormattedMessage id="components.StopViewer.viewSchedule" />
+          !isParentStop ? (
+            <FormattedMessage id="components.StopViewer.viewSchedule" />
+          ) : undefined
         }
         fromToSlot={fromToSlot}
         stopData={stopData}
       />
       <div>
         <div>{timezoneWarning}</div>
```
amy-corson-ibigroup: Could we add a `margin-bottom: 1rem` on the departure timezone notice? I think it could help the overall spacing (not just for stations) because it matches the margin between it and the from/to picker.
| Station | Stops|
|--------|-------|
|image|image|

amy-corson-ibigroup (changes requested):
This looks great, thank you so much for the changes! How would we feel about upping the top/bottom padding on the label, increasing the margin between the first label and the "departure times" notice, and adding either a border or a background color? I think the first stop label is bleeding into the station card, adding a little separation could make it a little clearer: 
| Now | Border | Background | 
|--------|-------|-------|
|image|image|image

(Requesting changes for the convo but nothing technically blocking here, so if we decide to keep as is I'll approve)

miles-grant-ibigroup:
> This looks great, thank you so much for the changes! How would we feel about upping the top/bottom padding on the label, increasing the margin between the first label and the "departure times" notice, and adding either a border or a background color? I think the first stop label is bleeding into the station card, adding a little separation could make it a little clearer:
> 
> Now	Border	Background
> image	image	image
> (Requesting changes for the convo but nothing technically blocking here, so if we decide to keep as is I'll approve)

I love the bigger border and the centering! The gray background I have some reservations about (maybe something transparent instead?) but I think that should wait for a future PR!

lib/components/viewers/nearby/stop.tsx
daniel-heppner-ibigroup: I feel like that is too much whitespace tbh, not a fan. Especially because it's repeated on each card, it is a lot of extra space for something that isn't that important.

daniel-heppner-ibigroup:
I liked the border proposal more too, so I added a thin border to the top. Let me know if it's good! 
image

amy-corson-ibigroup (approved):
Looks great! Thanks for the changes!!

lib/components/viewers/nearby/stop.tsx
```diff
@@ -25,14 +25,19 @@ export const fullTimestamp = (stoptime: StopTime) =>
...
-  padding: 8px 16px;
+  padding: 10px 16px;
 
   .child-stop-name {
     font-size: 16px;
     font-weight: 600;
   }
+
+  &:first-of-type {
+    border-top: 1px solid #999999;
```
miles-grant-ibigroup: Could we use the nearby view background color instead?

miles-grant-ibigroup (changes requested):
I'm happy with the border compromise! Just one small color note

miles-grant-ibigroup (approved):
What do you think about decreasing the transparency of the line?

PR #1372, opened 2025-03-31
Avoid Unnecessary Flex Requests

miles-grant-ibigroup:
**Description:**
Previously we were generating upwards of 6 flex requests per request. It turns out we didn't have to do that. An OTP-UI patch would be a cleaner way to do this, but since we will soon be moving to PlanConnection, I think this quicker solution makes more sense!

### DIFF
--- a/lib/actions/api-utils.ts
+++ b/lib/actions/api-utils.ts
@@ -1,0 +1,4 @@
+import { TransportMode } from '@opentripplanner/types'
+
+export const countFlexModes = (modes: TransportMode[]): number =>
+  modes.filter((m) => m.mode === 'FLEX').length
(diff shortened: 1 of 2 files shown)


daniel-heppner-ibigroup (changes requested):
Why not just add this to the combination generator in OTP UI. I don't like splitting the logic across both repositories. Is this really a quicker solution? this exact code would work fine in OTP UI I think

miles-grant-ibigroup:
> Why not just add this to the combination generator in OTP UI. I don't like splitting the logic across both repositories. Is this really a quicker solution? this exact code would work fine in OTP UI I think

We have dependents who use core-utils as is at the moment and I don't know if we want to make a breaking change when we know the move to PlanConnection is imminent

daniel-heppner-ibigroup (approved):
okay, I'm down to keep this here given the other considerations.

amy-corson-ibigroup (approved):
LGTM

PR #1374, opened 2025-04-16
Nearby View Location Field Completeness

miles-grant-ibigroup:
We couldn't use the connected location field in the nearby view since it does some funky stuff with the routing. So we import the props we need manually. Really otp-ui should be doing more of this `geocoderConfig` processing...

I also cleaned up some other props to move them to the redux to avoid prop drilling


amy-corson-ibigroup (approved):
LGTM! Thank you

lib/components/mobile/welcome-screen.tsx
```diff
@@ -62,11 +57,7 @@ const mapDispatchToProps = {
 }
 
 const mapStateToProps = (state: any) => {
```
daniel-heppner-ibigroup, line 59: just remove this function

daniel-heppner-ibigroup (approved):
cool, thanks for the clean up!

PR #1376, opened 2025-04-22
Check fares are equal before merging itins

amy-corson-ibigroup:
**Description:**
Previously, we would merge two itineraries even if the default fares were different, which could be misleading to the user since the itin preview implied the details between these itineraries were the same. This makes sure the fares are equal (at least based on the default fare type) before merging itineraries.

**PR Checklist:**

| Before | After |
|--------|-------|
|image|image|


daniel-heppner-ibigroup (approved):
This code looks good but I don't understand why we're doing this. When are we seeing a different fare for otherwise similar itineraries?

amy-corson-ibigroup:
Updated the PR description to explain the reasoning for the change!

binh-dam-ibigroup (approved):

PR #1377, opened 2025-04-22
Avoid fetch error if a user hasn't finished setting up their account

amy-corson-ibigroup:
**Description:**
Previously if a user had created an account but had not yet made it all the way to the end of the account setup wizard, all otp requests would return errors. This makes sure we're only associating the search with an account if the account has been fully set up.

**PR Checklist:**

### DIFF
--- a/lib/actions/api.js
+++ b/lib/actions/api.js
@@ -82,7 +82,7 @@ function getOtpFetchOptions(state, includeToken = false) {
   if (isOtpServerSameAsMiddleware) {
     // Use access token only if the user has completed entire account setup
     // (Otherwise every request returns an error)
-    if (includeToken && state.user.loggedInUser.hasConsentedToTerms) {
+    if (includeToken && state.user?.loggedInUser?.hasConsentedToTerms) {
       const { accessToken, loggedInUser } = state.user
       if (accessToken && loggedInUser) {
         token = accessToken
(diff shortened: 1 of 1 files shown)


lib/actions/api.js
```diff
@@ -80,7 +80,9 @@ function getOtpFetchOptions(state, includeToken = false) {
 
   const isOtpServerSameAsMiddleware = apiBaseUrl === api.host
   if (isOtpServerSameAsMiddleware) {
-    if (includeToken && state.user) {
+    // Use access token only if the user has completed entire account setup
+    // (Otherwise every request returns an error)
+    if (includeToken && state.user?.loggedInUser?.hasConsentedToTerms) {
       const { accessToken, loggedInUser } = state.user
       if (accessToken && loggedInUser) {
```
binh-dam-ibigroup: Another way to avoid redundant checks is to say`if (accessToken && loggedInUser?.hasConsentedToTerms)` on this line.

binh-dam-ibigroup (approved):

lib/actions/api.js
amy-corson-ibigroup: You mean instead of the change on line 85?

binh-dam-ibigroup: Yes.

amy-corson-ibigroup: I see, much cleaner! Thank you

josh-willis-arcadis (approved):
LGTM!

PR #1378, opened 2025-04-23
Gracefully handle pattern being wrong in pattern viewer

amy-corson-ibigroup:
**Description:**

Previously, if the pattern in the URL didn't match a valid pattern id, we were showing a completely blank pattern page. 

This was particularly tough when a user tried to open the pattern viewer with no internet connection, the URL would populate the `patternId` in the URL as "undefined". When the internet connection restored, because the URL had "undefined" there was no way for the page to recover gracefully. 

Now, if the `patternId` does not exist on the route, we instead reload the page with a valid pattern.

Open to feedback on this. Maybe we only reroute explicitly if the `patternId` is set to the string "undefined"? I cannot imagine a lot of instances where a user is typing the pattern into the URL to find the correct page, but if those cases exist, maybe it's better to let the user know something went wrong rather than automatically redirect them?

**PR Checklist:**

### DIFF
--- a/lib/components/viewers/pattern-viewer.tsx
+++ b/lib/components/viewers/pattern-viewer.tsx
@@ -49,12 +49,19 @@ const PatternViewer = ({
 }: Props) => {
   const intl = useIntl()
 
+  // @ts-expect-error TODO: add type to ComponentContext
+  const { ModeIcon, RouteRenderer } = useContext(ComponentContext)
+
+  const routePatternKeys = route?.patterns && Object.keys(route?.patterns)
+  const patternId = viewedRoute?.patternId
+  const routeId = viewedRoute?.routeId || null
+
   /**
    * If we're viewing a pattern's stops, route to main route viewer.
    */
   const _backClicked = useCallback(() => {
     // The if test is for typescript checks.
-    if (viewedRoute?.routeId) {
+    if (viewedRoute && routeId) {
       setViewedRoute({
         ...viewedRoute,
         patternId: undefined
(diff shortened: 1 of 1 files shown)


lib/components/viewers/pattern-viewer.tsx
```diff
@@ -67,6 +67,23 @@ const PatternViewer = ({
...
+  // (ex. the URL was /route/123/undefined)
+  if (
+    viewedRoute?.patternId &&
+    !routePatternKeys?.includes(viewedRoute?.patternId) &&
+    routePatternKeys &&
+    routePatternKeys.length > 0
+  ) {
+    // Set the patternId to the first pattern in the route (this will reload the page).
+    setViewedRoute({
+      patternId: routePatternKeys[0],
```
binh-dam-ibigroup, line 79: One thing I had been thinking for a while is to turn these arrow buttons into links to the first pattern that normally appears when you open the pattern viewer (I think it is not necessarily the pattern at index zero). It's probably a bit more involved but would also clean up the code and the navigation. What do you think?

lib/components/viewers/pattern-viewer.tsx
```diff
@@ -67,6 +67,23 @@ const PatternViewer = ({
...
 
+  const routePatternKeys = route?.patterns && Object.keys(route?.patterns)
+
+  // If the patternId does not exist in the route, course correct back to a valid pattern.
+  // (ex. the URL was /route/123/undefined)
+  if (
+    viewedRoute?.patternId &&
+    !routePatternKeys?.includes(viewedRoute?.patternId) &&
+    routePatternKeys &&
+    routePatternKeys.length > 0
```
binh-dam-ibigroup: These two checks are redundant with line 76.

lib/components/viewers/pattern-viewer.tsx
```diff
@@ -67,6 +67,23 @@ const PatternViewer = ({
   // @ts-expect-error TODO: add type to ComponentContext
   const { ModeIcon, RouteRenderer } = useContext(ComponentContext)
 
+  const routePatternKeys = route?.patterns && Object.keys(route?.patterns)
+
+  // If the patternId does not exist in the route, course correct back to a valid pattern.
+  // (ex. the URL was /route/123/undefined)
+  if (
+    viewedRoute?.patternId &&
```
binh-dam-ibigroup: Extract a var for `patternId`.

lib/components/viewers/pattern-viewer.tsx
```diff
@@ -67,6 +67,23 @@ const PatternViewer = ({
...
+    !routePatternKeys?.includes(viewedRoute?.patternId) &&
+    routePatternKeys &&
+    routePatternKeys.length > 0
+  ) {
+    // Set the patternId to the first pattern in the route (this will reload the page).
+    setViewedRoute({
+      patternId: routePatternKeys[0],
+      routeId: viewedRoute?.routeId
```
binh-dam-ibigroup: Extract a var for `routeId`.

binh-dam-ibigroup (approved):
Fix is good for now, but I would like to turn these arrow buttons into links eventually.

lib/components/viewers/pattern-viewer.tsx
```diff
@@ -136,7 +153,7 @@ const PatternViewer = ({
     )
   }
 
-  return null
+  return <p>oops we cannot find this pattern.</p>
```
binh-dam-ibigroup: Nit - intl

amy-corson-ibigroup: 🫣 Left this in by mistake haha thanks for catching! If we redirect to a real pattern, there shouldn't be any reason for this

lib/components/viewers/pattern-viewer.tsx
amy-corson-ibigroup: Taking away the check for `routePatternKeys` is still throwing a typescript error but I got rid of line 78! thanks!

lib/components/viewers/pattern-viewer.tsx
amy-corson-ibigroup: Can you clarify what buttons you mean? I thought at first you meant these, but I checked and they're already links
image

binh-dam-ibigroup: Okay, it must have been a while ago so I forgot!... I recall there is complicated piece of code to order the patterns but that is done u[on fetching the info for a route.

josh-willis-arcadis (approved):
LGTM

PR #1384, opened 2025-04-30
Switch formatted currency number to intl string so we can override

amy-corson-ibigroup:
**Description:**

This passes the formatted fare to an intl string so that we can override the string in custom configs if necessary

**PR Checklist:**


miles-grant-ibigroup (approved):
So annoying but glad it works

daniel-heppner-ibigroup (approved):
lgtm. what's with all the changes to the language files?

PR #1385, opened 2025-04-30
Add geocoderResultsOrder to save place field

amy-corson-ibigroup:
**Description:**
Pass geocoderResultsOrder to places location field

**PR Checklist:**


lib/components/user/places/place-editor.tsx
```diff
@@ -251,6 +252,7 @@ class PlaceEditor extends Component<Props> {
 
             <PlaceLocationField
               className="form-control"
+              geocoderResultsOrder={geocoderResultsOrder}
```
miles-grant-ibigroup, line 258: `geocoderResultsOrder` is already included in `geocoderConfig` which is already passed in! Can't we just pull it from there instead of adding a new prop?

miles-grant-ibigroup (changes requested):
I see a potential cleanup!

lib/components/user/places/place-editor.tsx
amy-corson-ibigroup: Good catch! Updated in [0045ca5](https://github.com/opentripplanner/otp-react-redux/pull/1385/commits/0045ca5870c7f895a65e7d26f5cc5676522f511e)

lib/components/user/places/place-editor.tsx
```diff
@@ -36,7 +36,9 @@ import { PlaceLocationField } from './place-location-field'
 
 type Props = WrappedComponentProps &
   FormikProps<UserSavedLocation> & {
-    geocoderConfig: GeocoderConfig
+    geocoderConfig: GeocoderConfig & {
+      geocoderResultsOrder?: Array<string>
+    }
```
miles-grant-ibigroup, line 41: There might be a type for this elsewhere! But that's a nit

miles-grant-ibigroup (approved):
Nitpick but good enough for now!

lib/components/user/places/place-editor.tsx
amy-corson-ibigroup: I looked, not so far, we have to add it to coreUtils :(

josh-willis-arcadis (approved):
LGTM

PR #1386, opened 2025-05-05
fix: place editor reverse geocode current location

daniel-heppner-ibigroup:
**Description:**
This fixes a bug in the place editor where it would not reverse geocode the current position after the first time it was requested. It would probably be good to overhaul the way the locations are handled to be more centralized, because now it feels a bit repetitive since we do this reverse geocode thing in 3 different spots, but they're all just different enough that it wouldn't really be easy to consolidate. 

**PR Checklist:**


miles-grant-ibigroup (approved):
Seems to work!

amy-corson-ibigroup (approved):

PR #1388, opened 2025-05-07
New saved place UI

amy-corson-ibigroup:
**Description:**
- Update the Saved Places UI to more closely match the saved trips UI
- Remove delete button and move delete function into edit page
- Add toast to confirm place has been deleted (provides status update to AT users)

**PR Checklist:**

| Before | After |
|--------|-------|
|image|image|


lib/components/user/places/favorite-place-list.js
```diff
@@ -12,9 +12,10 @@ import {
   getPlaceMainText
 } from '../../../util/user'
 import { isBlank } from '../../../util/ui'
+import { Plus } from '@styled-icons/fa-solid'
```
miles-grant-ibigroup:
```suggestion
import { Plus } from '@styled-icons/fa-solid/Plus'
```
Help! Our bundle size!

lib/components/user/places/favorite-place-screen.js
```diff
@@ -12,6 +12,7 @@ import { injectIntl, FormattedMessage } from 'react-intl'
 import { connect } from 'react-redux'
 import styled from 'styled-components'
 import * as yup from 'yup'
+import { Trash } from '@styled-icons/fa-solid'
```
miles-grant-ibigroup:
```suggestion
import { Trash } from '@styled-icons/fa-solid/Trash'
```
aaaaaaa

lib/components/util/toasts.tsx
```diff
@@ -31,14 +31,15 @@ export function toastSuccess(
...
   place: UserSavedLocation,
-  intl: IntlShape
+  intl: IntlShape,
+  change: 'Remembered' | 'Deleted'
 ): void {
   toastSuccess(
     getPlaceMainText(place, intl),
     intl.formatMessage({
-      id: 'actions.user.placeRemembered'
+      id: `actions.user.place${change}`
```
miles-grant-ibigroup: I remember there was discussion a while ago about avoiding dynamic i18n string generation unless we really couldn't avoid it. Maybe a compromise could be 
```suggestion
      id: `actions.user.place.${change}`
```

miles-grant-ibigroup (changes requested):
Looks much cleaner!

lib/components/util/toasts.tsx
amy-corson-ibigroup: Yes, you're right I don't know why I always forget this! I added a less invasive conditional, lmk if that's still too much

lib/components/util/toasts.tsx
```diff
@@ -31,14 +31,18 @@ export function toastSuccess(
 /**
  * Helper that will display a toast notification when a place is saved.
  */
-export function toastOnPlaceSaved(
+export function toastOnPlaceChanged(
   place: UserSavedLocation,
-  intl: IntlShape
+  intl: IntlShape,
+  change: 'Remembered' | 'Deleted'
```
miles-grant-ibigroup, line 37: Nitpick: Can we change this to an enum? Magic strings are annoying

miles-grant-ibigroup (approved):
One potential fix, but I'm happy to just add a comment about it! Second reviewer can decide

lib/components/util/toasts.tsx
amy-corson-ibigroup: I'm actually going to change the way we handle toasts in imminent follow up PR 🙃

josh-willis-arcadis (changes requested):
The 'Delete this place' button does not show up on mobile. Once that is working I think it should be good to go!

amy-corson-ibigroup:
Wow good catch @josh-willis-arcadis I didn't realize the button layout was so different on mobile. I updated so that the mobile layout is the same as the web.

josh-willis-arcadis (approved):
Looks great now! Ty for the changes. Looks like CI is failing due to unused strings?

PR #1389, opened 2025-05-09
remove time spent active

josh-willis-arcadis:
**Description:**

Removes the time spent active information in the trip preview layout

**PR Checklist:**

### DIFF
--- a/lib/components/app/trip-preview-layout-base.tsx
+++ b/lib/components/app/trip-preview-layout-base.tsx
@@ -74,6 +74,7 @@ class TripPreviewLayoutBase extends Component<Props, State> {
       title
     } = this.props
     const { LegIcon } = this.context
+    const displayTimeActive = config?.itinerary?.showTimeActive
 
     return (
       <div className="otp print-layout">
@@ -123,7 +124,7 @@ class TripPreviewLayoutBase extends Component<Props, State> {
             />
             <TripDetails
               className="percy-hide"
-              displayTimeActive={false}
+              displayTimeActive={displayTimeActive}
               itinerary={itinerary}
             />
           </>
--- a/lib/components/app/trip-preview-layout-base.tsx
+++ b/lib/components/app/trip-preview-layout-base.tsx
@@ -121,7 +121,11 @@ class TripPreviewLayoutBase extends Component<Props, State> {
               itinerary={itinerary}
               LegIcon={LegIcon}
             />
-            <TripDetails className="percy-hide" itinerary={itinerary} />
+            <TripDetails
+              className="percy-hide"
+              displayTimeActive={false}
+              itinerary={itinerary}
+            />
           </>
         )}
       </div>
--- a/example-config.yml
+++ b/example-config.yml
@@ -372,6 +372,8 @@ itinerary:
   showAlertEffectiveDateTimeText: true
   # Show all walking legs regardless of distance
   showAllWalkLegs: false
+  # show time spent active in an itinerary
+  showTimeActive: true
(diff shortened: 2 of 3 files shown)


lib/components/app/trip-preview-layout-base.tsx
```diff
@@ -121,7 +121,11 @@ class TripPreviewLayoutBase extends Component<Props, State> {
               itinerary={itinerary}
               LegIcon={LegIcon}
             />
-            <TripDetails className="percy-hide" itinerary={itinerary} />
+            <TripDetails
+              className="percy-hide"
+              displayTimeActive={false}
```
miles-grant-ibigroup: This needs to be configurable/displayed by default!

miles-grant-ibigroup (changes requested):

lib/components/app/trip-preview-layout-base.tsx
josh-willis-arcadis: good catch. I forgot to pull the config value.

miles-grant-ibigroup (changes requested):
Sorry to be a bother but since this is such a simple feature might it be better to disable this via css? I think a prop to hide a single row is clutter. Could we not just patch otp-ui to add a class name?

miles-grant-ibigroup (changes requested):
Dupe of https://github.com/opentripplanner/otp-react-redux/pull/1394?

josh-willis-arcadis:
Definitely a dupe of https://github.com/opentripplanner/otp-react-redux/pull/1394. Which one do we like better? @miles-grant-ibigroup @daniel-heppner-ibigroup

daniel-heppner-ibigroup:
Sorry but I like mine better because it passes in the config items with the other config stuff in the connect function. However I'm kinda with Miles that we should just do this with a CSS class.

josh-willis-arcadis:
@daniel-heppner-ibigroup You have a cleaner solution so lets use your PR. The `TripDetails` component already supports the prop `displayTimeActive` so I think we should just leave it as is.

danielhep:
I kinda feel like if we're going to have a config item for this, we should just go all the way and make it an array where you can specify all the items that get displayed in Trip Details. Like tripDetailsItems: ['fare', 'co2', 'timeActive', 'duration']

josh-willis-arcadis:
@daniel-heppner-ibigroup that is a better long term solution. Can we use this as a stopgap for now and clean up the props/trip-details component in a future PR?

miles-grant-ibigroup:
> @daniel-heppner-ibigroup that is a better long term solution. Can we use this as a stopgap for now and clean up the props/trip-details component in a future PR?

Can we do the CSS approach instead? I really want to stop using these props

josh-willis-arcadis:
yeah I will implement css. like this? @miles-grant-ibigroup 
```
            <TripDetails
              className={`percy-hide${
                displayTimeActive ? ' show-time-active' : ''
              }`}
              itinerary={itinerary}
            />
```

miles-grant-ibigroup:
> yeah I will implement css. like this? @miles-grant-ibigroup
> 
> ```
>             <TripDetails
>               className={`percy-hide${
>                 displayTimeActive ? ' show-time-active' : ''
>               }`}
>               itinerary={itinerary}
>             />
> ```

Eeek! No! I meant more adding a `time-active` class to the time active row and removing the `displayTimeActive` prop entirely

josh-willis-arcadis:
closing in favor of https://github.com/opentripplanner/otp-ui/pull/841

PR #1390, opened 2025-05-12
fix: support displaying endpoints overlay popup without persistence

miles-grant-ibigroup:
Previously persistence had to be enabled to show the endpoint overlay popups. This PR corrects this by adding a new `forceDisplayEndpointsPopup` config item. Please note that the "save as home" and "save as work" won't do anything without an enhancement to otp-ui.

### DIFF
--- a/example-config.yml
+++ b/example-config.yml
@@ -122,6 +122,9 @@ map:
   initLon: -122.682
   # autoFlyOnTripFormUpdate: false
   # navigationControlPosition: "bottom-right"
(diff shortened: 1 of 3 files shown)


amy-corson-ibigroup (approved):
It feels a little weird to have the "save as home/work" buttons do nothing, but i guess we can address that in a future PR

josh-willis-arcadis (approved):
LGTM

PR #1396, opened 2025-05-15
Build with Vite

binh-dam-ibigroup:
## Description

This PR sets Vite as the new build system.

We place customization files (CSS, JS, GraphQL) in the /tmp folder so that imports can use relative paths within the repo root. index.html is copied to the repo root for Vite to pick up, and gets some pre-processing (inject the main script and remove HTML comments).

Other things that we used to configure explicitly that are now included out-of-the-box:
- lazy imports of YML files using variables (thanks to Vite's "glob import" feature)
- hot reloading
- minification
- preview mode (used for Percy/a11y tests)
- faster builds and simpler config than with Webpack
- `yarn start` works out of the box, even with no environment variables passed (it will use the example subfolder).

Things that were dropped:
- Old dependencies (some have been upgraded to play well with Vite)
- Circular dependencies - There is a [circular dependency plugin](https://www.npmjs.com/package/vite-plugin-circular-dependency) but I have not been able to make it work at this time, and it doesn't seem to be able to return a count of the circular dependencies found.
- `require` syntax
- trimet-mod-otp configs no longer work.

Things that don't change:
- Babel and react-scripts are very much still here to run jest without making too many changes. Changes to jest should be handled separate from this PR.
- `mastarm deploy` seems to still be working (all content in dist is flat, no subfolders)
- No visual changes
- No changes to custom configs (trimet-mod-otp configs no longer work).

**PR Checklist:**


lib/index.tpl.html
```diff
@@ -16,9 +14,5 @@
   </head>
   <body>
     <div id="main"></div>
-    <script>
```
binh-dam-ibigroup, line 19: Vite will inject hot reloading code similar to this content in development mode (`yarn start`), so this is no longer needed.

lib/index.tpl.html
```diff
@@ -1,8 +1,6 @@
 <!doctype html>
-<!--
-  Initialize the default language to English to meet a11y requirements,
-  it will be subsequently updated by the setLocale action.
--->
+<!-- Initialize the default language to English to meet a11y requirements. -->
```
binh-dam-ibigroup, line 2: I have added a hook to strip out single line comments, so this text can be removed at build time.

lib/actions/ui.js
```diff
@@ -18,6 +17,7 @@ import {
   ItineraryView
 } from '../util/ui'
 import { getModesForActiveAgencyFilter, getUiUrlParams } from '../util/state'
+import { loadLocaleData } from '../util/i18n-loader'
```
daniel-heppner-ibigroup, line 20: Do you remember what the purpose of moving this into its own file was? Just curious

lib/index.tpl.html
daniel-heppner-ibigroup: Thanks for the explanation!

vite.config.js
```diff
@@ -0,0 +1,109 @@
...
+  }
+}
+
+// Empty tmp folder before copying stuff there.
+fs.emptyDirSync('./tmp')
+
+// index.html is placed at the root of the repo for Vite to pick up.
+customFile('HTML_FILE', './index.html', './lib/index.tpl.html')
+customFile('CUSTOM_CSS', './tmp/custom-styles.css', './example/example.css')
+customFile('YAML_CONFIG', './tmp/config.yml', './example/example-config.yml')
```
binh-dam-ibigroup, line 42: Miles says hot reloading from the original customizations files is important.

miles-grant-ibigroup: I think @daniel-heppner-ibigroup's suggestion of doing a symlink instead of a copy is a good idea!

binh-dam-ibigroup: @miles-grant-ibigroup  and @daniel-heppner-ibigroup The old craco.config.js does copy the JS configs (and custom graphql) files. We can either: keep copying those files, so that no changes in the imports in those files are needed, or go ahead with symlinks, but that requires updating the imports of config.js that use OTP-RR relative paths. What do you think?

binh-dam-ibigroup: @miles-grant-ibigroup  and @daniel-heppner-ibigroup  To avoid modifying custom files, I added a watch so it copies edited custom files over for hot reloading (https://github.com/opentripplanner/otp-react-redux/pull/1396/commits/e2d2774a213461e1fb9ccdedf72943675c1ca40e).

danielhep: Nice fix. Seems just as good as a symlink

daniel-heppner-ibigroup: It's so fast!

daniel-heppner-ibigroup (approved):
This is working really well. I am glad we resolved the config hot reloading issue so easily. I'm certainly impressed!

lib/actions/ui.js
binh-dam-ibigroup: Yes, it is to isolate the parts that use the Vite-specific glob imports. This file is mocked for tests (tests continue to use Babel).

miles-grant-ibigroup (approved):
Have tested this in almost every configuration I can think of and things look good! I'd prefer if a third reviewer could test with everything they can think of and then I'd say this is ready for merge

alec-georgoff (commented):
I'm still working on testing everything, but I noticed that vite complains if we use Node 21 (which is set in `.nvmrc`). We should update `.nvmrc` to one of the supported versions:

image

Any thoughts on which one to use?

binh-dam-ibigroup:
> I'm still working on testing everything, but I noticed that vite complains if we use Node 21 (which is set in `.nvmrc`). We should update `.nvmrc` to one of the supported versions:
> image
> 
> Any thoughts on which one to use?

Good point @alec-georgoff, I'd recommend Node 22. The GitHub CI files in this PR already reference that version.

alec-georgoff (commented):
(removed)

alec-georgoff:
> Uh oh...I was trying to build SEPTA and ran into errors. @danielhep confirmed he saw the same error on his computer.
> 
> image

Following up with a little bit more info:

## Missing `planGraphQLFile` in `deploy-config`

It was discovered that `./run.js septa-metro` was not working on the `dev` branch or on this PR's branch. After investigation, the cause was determined to be the absence of this line in `septa-metro/deploy-config.yml`:

```
planGraphQLFile: ~/git/configurations/otp/septa/planQuery-with-blockId.graphql
```

Adding that line into the config allows `septa-metro` to run on the `dev` branch.

## Vite issues

However, even with the missing line from above, we continue to see the error on this PR's branch. More investigation is required.

alec-georgoff (approved):

miles-grant-ibigroup:
New potential issue: the transitive overlay images aren't being exported correctly when doing a build! But they work in dev mode

PR #1397, opened 2025-05-20
Reload App on Locale Change

josh-willis-arcadis:
**Description:**
Fixes an issue where the `GeolocateControl` and `NavigationControl` components in default-map were not localizing. reloading the page ensures those components have the correct translations. 

**PR Checklist:**


lib/actions/ui.js
```diff
@@ -432,8 +432,9 @@ export function showMobileSearchScreen() {
  * If the specified locale is null, fall back to the defaultLocale
  * set in the configuration.
  * Also update the lang attribute on the root <html> element for accessibility.
+ * If the locale is updated, reload the page to apply the new locale.
  */
-export function setLocale(locale) {
+export function setLocale(locale, isUpdatingLocale = false) {
```
daniel-heppner-ibigroup, line 444: In what cases would this be called where isUpdatingLocale is set to false? It's confusing to me because to me a function called `setLocale` has the sole responsibility of updating the locale. 

I'd be fine with a clearer comment here.

What are the cases where we want to setLocale without causing a reload?

daniel-heppner-ibigroup (changes requested):
Can you clarify things a bit for me and maybe improve the comments a bit? The functionality works well, exactly as expected. I just want to know what the point of the new argument is for setLocale

lib/actions/ui.js
josh-willis-arcadis: This function is called on the intial page load to set the locale, not update it. I added `isUpdatingLocale` to prevent a reloading loop on the initial page render. `isUpdatingLocale` should only be set to `true` when a user updates the locale using the locale selector, all other use cases should be `false`. @daniel-heppner-ibigroup

daniel-heppner-ibigroup (approved):
Sounds good!

amy-corson-ibigroup (approved):

PR #1399, opened 2025-05-21
Remove redundant text from delete button

josh-willis-arcadis:
**Description:**
Removes redunant text from the delete button on the saved trip editor. A future PR will be needed to clean up the translation files to remove redunant text in buttons throughout the app.

**PR Checklist:**

### DIFF
--- a/i18n/en-US.yml
+++ b/i18n/en-US.yml
@@ -533,7 +533,7 @@ components:
     signInText: Log in to save trip
     signInTooltip: Please log in to save trip.
   SavedTripEditor:
-    deleteSavedTrip: Delete saved trip
+    deleteSavedTrip: Delete
     editSavedTrip: Edit saved trip
     readOnlyBanner: >-
       {creator} created and added you to this trip. Therefore, you cannot make


amy-corson-ibigroup (approved):
Looks good! Long term we should maybe consider creating one "delete" string under `common.forms` that we use for all of these, but I'll defer to Binh!!

binh-dam-ibigroup:
> Looks good! Long term we should maybe consider creating one "delete" string under `common.forms` that we use for all of these, but I'll defer to Binh!!

I thought we had such a string! If you feel like doing it in this PR, go ahead.

binh-dam-ibigroup (approved):

binh-dam-ibigroup:
> > Looks good! Long term we should maybe consider creating one "delete" string under `common.forms` that we use for all of these, but I'll defer to Binh!!
> 
> I thought we had such a string! If you feel like doing it in this PR, go ahead.

Feel free to update the other languages too.

josh-willis-arcadis:
A ticket is being created to update buttons on other forms. I will clean up strings and languages in that PR. @binh-dam-ibigroup

PR #1400, opened 2025-05-22
Use `toast.promise` to correctly announce status updates

amy-corson-ibigroup:
**Description:**
Previously, Toasts were not announcing updates to AT. This is because the library is ineffectively rendering content inside its status regions. You can read more about that here: https://github.com/timolins/react-hot-toast/issues/335#issuecomment-2558193154 Essentially, status regions only announce when the content is updated, and react-hot-toast does not account for this.

A workaround for this is to utilize the `toast.promise` which takes a promise and returns a toast based on the status. This means the content changes from `loading` to `success` which means it will announce the success (or failure) to AT users. In most cases this is a little cleaner than what we were doing before. In one case you'll notice we have to fake a promise to get the same effect. 

Let me know what you think, I am very open to feedback on this PR. Would also really appreciate any AT testing.

**PR Checklist:**


lib/actions/user.js
```diff
@@ -496,30 +496,29 @@ export function createOrUpdateUserMonitoredTrip(
...
-                id: 'components.SavedTripEditor.saveNewTrip'
-              })
-            : intl.formatMessage({
-                id: 'components.SavedTripEditor.editSavedTrip'
-              }),
-          intl.formatMessage({ id: 'actions.user.preferencesSaved' })
-        )
-      }
-
+      console.log(status, data)
```
binh-dam-ibigroup: Remove `console.log` statement.

lib/components/user/wizard.tsx
```diff
@@ -95,9 +95,14 @@ class Wizard extends Component<Props> {
     } else {
       // Display a toast to acknowledge saved changes
       // (although in reality, changes quietly took effect in previous screens).
-      toastSuccess(
-        title,
-        intl.formatMessage({ id: 'actions.user.preferencesSaved' })
+      toastPromise(
+        // Fake promise to make sure toast is announced to screenreaders. See https://github.com/opentripplanner/otp-react-redux/pull/1400 for details.
+        new Promise((resolve) => setTimeout(resolve, 100)),
```
binh-dam-ibigroup: You could probably resolve the promise immediately without a timer, e.g. 
```suggestion
        new Promise((resolve) => resolve()),
```

lib/components/user/user-account-screen.tsx
```diff
@@ -106,36 +106,19 @@ class UserAccountScreen extends Component<Props> {
...
-      }
-    } catch {
-      // Remove any toasts before showing alert.
-      toast.remove()
-      alert(
-        intl.formatMessage({
-          id: 'components.UserAccountScreen.errorUpdatingProfile'
-        })
-      )
+      await toastPromise(submitForm(), submitSuccessMessage, intl)
```
binh-dam-ibigroup: In "wizard" mode (e.g. when creating accounts), don't show toasts in response to individual field changes. Otherwise, each change will be toasted/announced in addition to the final toast when the user clicks "Finish".

binh-dam-ibigroup (changes requested):
Approach is good, however please remove toast when editing in wizard mode so that a single acknowledgement is done upon clicking "Finish".

lib/components/user/wizard.tsx
amy-corson-ibigroup: TS needs an argument for `resolve` so I added `null`! Thanks

lib/components/user/user-account-screen.tsx
amy-corson-ibigroup: Thanks for figuring out the issue! Fixed in [49656f1](https://github.com/opentripplanner/otp-react-redux/pull/1400/commits/49656f11898346dfeb458ae09e293f7f5930ff53)

binh-dam-ibigroup (approved):
Nice

binh-dam-ibigroup:
@amy-corson-ibigroup Could you address this error message during code checks:
```
Message 'components.UserAccountScreen.updating' from locale en-US is not used in code.
Message 'components.UserAccountScreen.updating' from locale fr is not used in code.
```

josh-willis-arcadis (approved):

PR #1402, opened 2025-05-28
Tweaks to nearby view structure and styling

amy-corson-ibigroup:
**Description:**
- Prevents nearby view search from stretching mobile viewports
- Prevents nearby view search from hiding results behind map
- Removes nearby view components that are not `li` from the `ol`

**PR Checklist:**

| Before | After |
|--------|-------|
|image|image|


lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -327,12 +327,7 @@ function NearbyView({
           />
         </InvisibleA11yLabel>
       )}
-      <NearbySidebarContainer
-        className="base-color-bg"
-        style={{ marginBottom: 0 }}
-      >
-        {/* This is used to scroll to top */}
-        <div aria-hidden ref={firstItemRef} />
+      <div style={{ padding: '1em' }}>
```
miles-grant-ibigroup, line 330: This now means the location field doesn't scroll with the rest of the items. This is causing problems in configs that hide the location field but also is a behavior that I've seen explicitly asked _not _ to have

miles-grant-ibigroup: It's also causing this issue when jumping to a specific stop id Screenshot 2025-05-28 at 2 45 54 PM

lib/components/viewers/nearby/styled.tsx
```diff
@@ -20,15 +20,13 @@ export const NearbySidebarContainer = styled.ol`
   gap: 1em;
   padding: 0 1em;
   list-style: none;
+  overflow-y: scroll;
+  height: 90%;
```
miles-grant-ibigroup, line 24: This feels like a bit of a hack to deal with the location field. Could we not instead just wrap the location field in an `li`? Or move the `ol` styles to a div and then have an `ol` that has no styles?

lib/components/viewers/viewers.css
```diff
@@ -259,7 +259,7 @@
   color: #111;
   display: flex !important;
   height: 50px;
-  margin-bottom: -5px !important;
+  margin-bottom: 0px !important;
```
miles-grant-ibigroup, line 262: This is causing the clear button to become misaligned Screenshot 2025-05-28 at 2 47 29 PM

miles-grant-ibigroup (changes requested):
Some regressions to deal with

amy-corson-ibigroup:
The `overflow-y: scroll` on the container is what's causing the issue on the location field menu (see "before" image above). I'm not seeing a clean workaround for this, but let me know if I'm missing something.

Not to mention having a scrollable container (location results menu) inside another scrollable container (nearby view results) sounds like it could get very messy, especially on mobile. What is the concern with pinning the location field to the top?

miles-grant-ibigroup:
> The `overflow-y: scroll` on the container is what's causing the issue on the location field menu (see "before" image above). I'm not seeing a clean workaround for this, but let me know if I'm missing something.
> 
Is the solution I proposed not tenable?

> Not to mention having a scrollable container (location results menu) inside another scrollable container (nearby view results) sounds like it could get very messy, especially on mobile. What is the concern with pinning the location field to the top?

If we can get it to work cleanly then I have no problem with it, but the percy tests show quite well that the clipping is a real issue

I believe the solution I proposed should solve these problems, even though it's an extra layer which admittedly does suck

amy-corson-ibigroup:
Sure, it's no problem to fix the clipping on highlighted items. Done in [9d0d948](https://github.com/opentripplanner/otp-react-redux/pull/1402/commits/9d0d948405cf3f6b3bda29c9fbe8545489bf9966)

> Could we not instead just wrap the location field in an li?

No, it doesn't belong in the list of nearby view results.

> Or move the ol styles to a div and then have an ol that has no styles?

The `ol` needs to scroll and to scroll it needs a set height.

> This now means the location field doesn't scroll with the rest of the items. This is causing problems in configs that hide 
the location field but also is a behavior that I've seen explicitly asked _not _ to have

- So are there concerns with having the location field pinned to the top? Or no
- I'm not seeing a config parameter that hides the nearby location field? Can you elaborate

miles-grant-ibigroup:
There are some configs that hide the location field using css. I guess my concerns with the location field sticky at the top are that it can be confusing to see the cards cut off the way they are now. Is there really no way to include the location field in the list? What if we have a scrollable div with a non-scrollable ol within it

amy-corson-ibigroup:
Can you please elaborate on what you mean by "cards cut off"? The only example I've seen is the clipping on the highlighted stop card and that's been fixed. 

I can add a `className` to the location field padding container for CSS configuration?

I don't think the location field can be inside a scrollable container, because the location field result menu technically counts as overflow. Anything with an `overflow` attribute is going to clip the menu. If there are people that really want the location field to scroll with the nearby items, that can be adjusted in CSS by making the parent container `overflow-y: scroll` and overriding the `overflow-y` in the `ol`, but it will clip the results so I don't think it should be the default.

miles-grant-ibigroup:
> Can you please elaborate on what you mean by "cards cut off"? The only example I've seen is the clipping on the highlighted stop card and that's been fixed.
> 
If you scroll the card list there is a sharp cut between the top of the now cut-off cards and the padding that contains the location field.

> I can add a `className` to the location field padding container for CSS configuration?
> 
This is not the issue! The problem is that there is now a weird unscrollable section at the top, since only the ol scrolls. Can't we have the div scroll and then leave the ol unscrollable?

> I don't think the location field can be inside a scrollable container, because the location field result menu technically counts as overflow. Anything with an `overflow` attribute is going to clip the menu. If there are people that really want the location field to scroll with the nearby items, that can be adjusted in CSS by making the parent container `overflow-y: scroll` and overriding the `overflow-y` in the `ol`, but it will clip the results so I don't think it should be the default.
We should take this to a larger discussion

miles-grant-ibigroup (approved):
Looks good after discussion!

josh-willis-arcadis (approved):
LGTM! I think tests need to be updated

PR #1405, opened 2025-05-29
June 2025 Field Trip Updates

miles-grant-ibigroup:
- [x] field trip planned trips are always in the future
- [x] field trip report date is next business day
- [x] add now button
- [x] calltaker mode icons are more highlighted on dark backgrounds

I'm a little worried about the now button. It can set a weird time in some cases, and does it conflict with the dropdown now? Is it too confusing?


lib/components/form/call-taker/date-time-picker.tsx
```diff
@@ -201,8 +201,11 @@ const DateTimeOptions = ({
   }, [dateTime, departArrive, homeTimezone, setQueryParam])
 
   const handleDepartArriveChange = useCallback(
-    (e: ChangeEvent<HTMLSelectElement>) => {
-      const newValue = e.target.value as DepartArriveValue
+    (e: ChangeEvent<HTMLSelectElement> | string) => {
```
daniel-heppner-ibigroup: use DepartArriveValue here instead of string

daniel-heppner-ibigroup (approved):
The now button seems to be working fine, I wasn't able to see any glitches

daniel-heppner-ibigroup:
The button does look weird next to the dropdown. But my solution would be to remove it entirely, however given that this PR is adding it, I'm guessing that it was specifically asked for? Wouldn't be my choice

miles-grant-ibigroup:
> The button does look weird next to the dropdown. But my solution would be to remove it entirely, however given that this PR is adding it, I'm guessing that it was specifically asked for? Wouldn't be my choice

Yes this is why it's hidden by default

amy-corson-ibigroup (approved):

PR #1406, opened 2025-05-29
Add alternate route names to itin description

amy-corson-ibigroup:
**Description:**
Adds the alternative route to the itinerary description in the case that itineraries are merged

**PR Checklist:**

| UI | AT |
|--------|-------|
|image|image|


lib/components/narrative/metro/attribute-utils.tsx
```diff
@@ -51,16 +51,25 @@ export const removeInsignificantWalkLegs = (leg: Leg): boolean =>
...
 ): React.ReactNode => {
   return intl.formatList(
     itinerary.legs
-      .map((leg: Leg) => {
-        return leg.routeShortName
+      .map((leg: Leg & { alternateRoutes?: { [id: string]: Leg } }) => {
+        const alternateRouteNames =
+          leg.alternateRoutes &&
+          Object.entries(leg.alternateRoutes).map(
+            (route) => route[1].routeShortName
```
miles-grant-ibigroup: Route short name is sometimes not specified! This should be an optional

miles-grant-ibigroup (changes requested):
Should we extract the alternateRouteNames extraction code to a new method?

lib/components/narrative/metro/attribute-utils.tsx
amy-corson-ibigroup: Added a filter to this list to match the one on line 74

miles-grant-ibigroup: This is good but not what I was talking about! `route[1]` might not contain `routeShortName` which is the issue

amy-corson-ibigroup: Okay, added a fallback to this line

lib/components/narrative/metro/attribute-utils.tsx
```diff
@@ -49,18 +49,32 @@ export const removeInsignificantWalkLegs = (leg: Leg): boolean =>
...
 ): React.ReactNode => {
   return intl.formatList(
     itinerary.legs
-      .map((leg: Leg) => {
-        return leg.routeShortName
+      .map((leg: Leg & { alternateRoutes?: { [id: string]: Leg } }) => {
+        const alternateRouteNames =
+          leg.alternateRoutes &&
+          getAlternateRoutesFromLeg(leg)
+            .map((route) => route[1].routeShortName || undefined)
```
miles-grant-ibigroup: This line makes me feel uncomfortable. Without optionals this feels like a disaster waiting to happen... Can we add some optionals here instead of the `|| undefined`?

miles-grant-ibigroup (approved):
Works, but could possibly be cleaned up a bit more

lib/components/narrative/metro/attribute-utils.tsx
amy-corson-ibigroup: That's fair, I was keeping the function consistent, but I agree with you it doesn't make sense to have routes without short names not represented in the list. I added a fallback for the alternate routes but also the default route below.

miles-grant-ibigroup: Great! We should maybe throw a comment in here saying that this could all be cleaned up in the future... Probably when we implement `nextLegs`!

josh-willis-arcadis (approved):
LGTM

PR #1408, opened 2025-06-03
Indicate to AT that element is currently selected

amy-corson-ibigroup:
**Description:**
Adds a "(currently selected)" string to active links so AT users know which route is active.
Also adds it to locale selector and sort option since `aria-selected` is being a little inconsistent on those components.

**PR Checklist:**


daniel-heppner-ibigroup (approved):
Nice!

lib/components/viewers/route-row.tsx
```diff
@@ -205,6 +206,13 @@ export class RouteRow extends PureComponent<Props> {
               </ModeIconElement>
             )}
             <RouteName route={route} RouteRenderer={RouteRenderer} />
+            {isActive && (
+              <InvisibleA11yLabel>
+                (
+                <FormattedMessage id="components.RouteViewer.currentlySelected" />
+                )
```
miles-grant-ibigroup: We've been told in the past that brackets have to be localized.

miles-grant-ibigroup (changes requested):
Good change! But one small problem

amy-corson-ibigroup:
I updated this PR to also add this string to the locale selector and the sorting dropdown! @miles-grant-ibigroup @daniel-heppner-ibigroup

lib/components/narrative/narrative-itineraries-header.tsx
```diff
@@ -192,6 +192,11 @@ export default function NarrativeItinerariesHeader({
                     role="option"
                   >
                     {sortOption.text}
+                    {sortText === sortOption.text && (
+                      <InvisibleA11yLabel>
+                        <FormattedMessage id="coommon.currentlySelected" />
```
miles-grant-ibigroup: Typo!

i18n/fr.yml
```diff
@@ -76,6 +76,7 @@ actions:
       réessayez.
 common:
   coordinates: "{lat}; {lon}"
+  currentlySelected: (actuellement sélectionné)
```
miles-grant-ibigroup: Do we need a hardcoded space here to prevent the words from squishing together?

miles-grant-ibigroup (changes requested):
Looks good! Just a few small issues

i18n/fr.yml
amy-corson-ibigroup: might as well!

amy-corson-ibigroup: you 🤝 codespell

lib/components/narrative/narrative-itineraries-header.tsx
```diff
@@ -192,6 +192,11 @@ export default function NarrativeItinerariesHeader({
                     role="option"
                   >
                     {sortOption.text}
+                    {sortText === sortOption.text && (
+                      <InvisibleA11yLabel>
+                        <FormattedMessage id="commmon.currentlySelected" />
```
miles-grant-ibigroup: Oh no!

miles-grant-ibigroup (approved):
One more typo and then this great fix is ready to go!

lib/components/narrative/narrative-itineraries-header.tsx
amy-corson-ibigroup: cant spell this to save my life huh

PR #1409, opened 2025-06-04
Don't allow geocoded locations to be user unfriendly

miles-grant-ibigroup:
When scooters/bikes have no code, we render a strange string that ends with a colon. This is untenable. The quick solution here is to dis-allow this behavior. 

I'd like to add a configuration option where this can be configurable to further customize forced reverse behavior, but I don't see how this will be possible without structural changes in otp-ui


amy-corson-ibigroup (approved):
This works but it's taking almost a full second to load when I test... Not blocking for this but we may need to eventually look at loading UI states for location field / geocoder

alec-georgoff:
Seems reasonable to me! Small nit but if you could add screenshots of the before/after behavior it would be helpful to future developers who may need to look back on this change.

alec-georgoff (approved):

PR #1411, opened 2025-06-05
feat(itinerary): add ability to sort itineraries by fare

josh-willis-arcadis:
Adds logic to sort itineraries by only transit fare. The difference between sorting by fare and cost is that itineraries with driving or biking have increased cost than just transit itineraries. 

**PR Checklist:**


lib/util/itinerary.tsx
```diff
@@ -453,10 +454,12 @@ export function addSortingCosts<T extends Itinerary>(
     totalFareResult === null ? Number.MAX_VALUE : totalFareResult
 
   const rank = calculateItineraryCost(itinerary, config)
+  const transitFare = getFare(itinerary).transitFare || 0
```
daniel-heppner-ibigroup: what happens if you just let this stay as `undefined` instead setting it to zero? does that mess up the sort?

lib/util/itinerary.tsx
```diff
@@ -48,6 +48,7 @@ export interface ItineraryWithCO2Info extends Itinerary {
 export interface ItineraryWithSortingCosts extends Itinerary {
   rank: number
   totalFare: number
+  transitFare?: number
```
daniel-heppner-ibigroup, line 51: I noticed this is an optional, which leads me to my next question. I think either you should remove the optional here or remove the `|| 0` below

daniel-heppner-ibigroup (approved):
Looks good! Simple change

lib/util/itinerary.tsx
josh-willis-arcadis: when the value is zero or `undefined` the sort does not change.

lib/util/itinerary.tsx
josh-willis-arcadis: I'll remove the optional, as both values of zero and `undefined` do not change the sort order

lib/util/itinerary.tsx
josh-willis-arcadis: I guess we can remove `|| 0` as well for cleaner code. I'll let second reviewer decide.

lib/util/itinerary.tsx
```diff
@@ -453,10 +454,12 @@ export function addSortingCosts<T extends Itinerary>(
     totalFareResult === null ? Number.MAX_VALUE : totalFareResult
 
   const rank = calculateItineraryCost(itinerary, config)
+  const transitFare = 0 // getFare(itinerary).transitFare || 0
```
amy-corson-ibigroup: I think you may need to revert this comment

amy-corson-ibigroup (changes requested):
Blocking because the logic to actually return the fare is commented out.

Wondering also if the user facing label "fare" should be changed to "transit fare"? I'm worried about people assuming we're incorporating rental bike price into the total fare for micromobility trips

amy-corson-ibigroup:
There's also a distinction between a $0 fare and "no fare information". I don't think "no fare info" should be weighted the exact same as a free trip? Feels like if you sort by cheapest fare you should see free trips before you see n/a fares. Idk @daniel-heppner-ibigroup what are your thoughts on this

lib/util/itinerary.tsx
amy-corson-ibigroup: Maybe if we get an `undefined` fare we return .01 or something? So that the confirmed free trips are sorted above fares where we don't actually know.

lib/util/itinerary.tsx
josh-willis-arcadis: oops!

lib/util/itinerary.tsx
josh-willis-arcadis: if a fare is `undefined` would that be a data issue? Should we try to handle data issues?

amy-corson-ibigroup: If agencies don't provide us fares we display "no fare available" and i think that's fine to do, I just don't know how it fits into this PR

amy-corson-ibigroup (approved):
I'm going to approve so I'm not blocking while on vacation, but would still love to discuss the difference in sorting between undefined fares and $0 fares before we merge!

lib/util/state.js
```diff
@@ -50,6 +50,8 @@ export function sortItineraries(type, direction, a, b, config) {
       return dirFactor * (a.duration - b.duration)
     case 'COST':
       return dirFactor * (a.totalFare - b.totalFare)
+    case 'FARE':
+      return dirFactor * (a.transitFare - b.transitFare)
```
daniel-heppner-ibigroup: I think no fare information itineraries should be sorted opposite a low fare. So remove the || 0 above and then change this so that it sorts undefined as if it's a large fare.
since:
if a>b then a-b>0
so:
if a is undefined, let's pretend a>b so return 1
if b is undefined, let's pretend a<b so return -1
if both are undefined, return 0.

not sure what the cleanest code is to do this. maybe make a little function to do it? or maybe there's a clean logic way of doing it

daniel-heppner-ibigroup (changes requested):

lib/util/state.js
josh-willis-arcadis: I think that is a good way to handle undefined values. Addressed in [c2ede03](https://github.com/opentripplanner/otp-react-redux/pull/1411/commits/c2ede03f1490bb72116ae7aa8d4e55a6e9d7a67e)

lib/util/itinerary.tsx
josh-willis-arcadis: I swapped back to optional to better handle undefined fare values.

lib/util/state.js
```diff
@@ -33,14 +33,15 @@ export function getActiveSearch(state) {
...
-  return a - b
+function compareItineraryFare(a, b, direction = 'ASC') {
+  const aMissing = a === undefined || a === null
+  const bMissing = b === undefined || b === null
+  // Always sort missing values (null/undefined) to the end of the array.
+  if (aMissing && bMissing) return 0
+  if (aMissing) return 1
+  if (bMissing) return -1
+  // If both fares are defined, compare them according to direction
+  return direction === 'ASC' ? a - b : b - a
```
daniel-heppner-ibigroup, line 44: hahah so the old version didn't work? Good thing you wrote the test!!

__tests__/util/state.js
```diff
@@ -99,4 +100,66 @@ describe('util > state', () => {
...
+      const sorted = [...itineraries].sort((a, b) =>
+        sortItineraries('FARE', 'DESC', a, b)
+      )
+      // Fares: 200, 100, 50, undefined, null
+      expect(sorted.map((i) => i.transitFare)).toEqual([
+        200,
+        100,
+        50,
+        undefined,
+        null
```
daniel-heppner-ibigroup, line 141: These look good, but can you also add a $0 fare to make sure that it doesn't get accidentally lumped in with null/undefined?

daniel-heppner-ibigroup (approved):
Looks good! Please just make one more little change to the tests then it's good to go. Why did multiplying by -1 not work in the dirFactor thing?

lib/util/state.js
josh-willis-arcadis: Old version didn't handle the direction of undefined/null values. Good thinking on a unit test!

__tests__/util/state.js
josh-willis-arcadis: addressed in [52e6b1e](https://github.com/opentripplanner/otp-react-redux/pull/1411/commits/52e6b1ebd0b868e9dea338c1b66bd8f6d5bd9af9)

PR #1412, opened 2025-06-09
Pass auth0 correct chinese locale code

amy-corson-ibigroup:
**Description:**
Passes the correct locale code for Chinese to Auth0 

Opted not to use a `switch` after reading [this](https://www.oreilly.com/library/view/high-performance-javascript/9781449382308/ch04.html#if-else_versus_switch:~:text=The%20prevailing%20theory%20on%20using%20if%2Delse%20versus%20switch%20is%20based%20on%20the%20number%20of%20conditions%20being%20tested%3A%20the%20larger%20the%20number%20of%20conditions%2C%20the%20more%20inclined%20you%20are%20to%20use%20a%20switch%20instead%20of%20if%2Delse) but I'm open to it

**PR Checklist:**

### DIFF
--- a/lib/components/user/nav-login-button-auth0.tsx
+++ b/lib/components/user/nav-login-button-auth0.tsx
@@ -32,8 +32,7 @@ const NavLoginButtonAuth0 = ({
   let auth0Locale = locale
   if (locale === 'zh-Hant') {
     auth0Locale = 'zh-TW'
-  }
-  if (locale === 'zh-Hans') {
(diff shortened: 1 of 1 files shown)


daniel-heppner-ibigroup (approved):
This looks fine. I was thinking maybe it should be in a function at the top of the file that returns the auth0locale from a switch statement, but with only two things I think this is fine. 

edit: can you make it an else if instead of an if? Slightly better perf

binh-dam-ibigroup (approved):

PR #1413, opened 2025-06-10
Add an config error to warn about rental vehicles

amy-corson-ibigroup:
**Description:**

Add a conditional error message if rental is enabled. Used in cases where there are free-floating vehicles and we may want to warn users that conditions are subject to change.

**PR Checklist:**


lib/actions/apiV2.js
```diff
@@ -1224,6 +1225,14 @@ export function routingQuery(searchId = null, updateSearchInReducer) {
                     })
                   }
                 }
+                if (
+                  displayMicromobilityRentalError &&
+                  query.variables.modeButtons.includes('rent')
```
daniel-heppner-ibigroup: Should we display this anytime the user selected rent or only if the response includes any rental itineraries?

lib/actions/apiV2.js
```diff
@@ -1209,7 +1209,8 @@ export function routingQuery(searchId = null, updateSearchInReducer) {
               ) {
                 // Add accessibility error, if it is turned on
                 const state = getState()
-                const { displayA11yError } = state.otp.config?.itinerary
+                const { displayA11yError, displayMicromobilityRentalError } =
+                  state.otp.config?.itinerary
```
daniel-heppner-ibigroup, line 1214: I see why you called this error, since we use the error renderer to show it, but I think it would be better to call it a warning. From the perspective of someone writing a config file for OTP-RR who doesn't know the implementation details this feels more like a warning than an error, right?

daniel-heppner-ibigroup (changes requested):
Can we change the name of the config item?

lib/actions/apiV2.js
amy-corson-ibigroup: Hmm I changed this in [ce63aa0](https://github.com/opentripplanner/otp-react-redux/pull/1413/commits/ce63aa0e2197f11b3e06ce041b1e328dc67e014e), but curious if the 2nd reviewer has an opinion. I feel like it might be useful to keep the naming conventions for all the "errors" the same, so that it's clear this is being displayed by the error renderer, rather than some other component?

daniel-heppner-ibigroup: If we renamed the error renderer to Alert or something maybe that would be better. I just don't think config items should be dictated by the name of the component that renders them, that's an implementation detail that isn't important to the end user.

daniel-heppner-ibigroup (approved):

lib/actions/apiV2.js
amy-corson-ibigroup: I thought I responded to this one, sorry! Decided to do it this way because 1. It was less taxing than filtering through all the itineraries and 2. If your search includes rental, it seems like relevant information to have even if the rental trips don't show up in every search? But it's not a big deal if we want to switch

lib/actions/apiV2.js
miles-grant-ibigroup: Yeah can we make this consistent with the `displayA11yError`? Both in terms of labeling, implementation, and even the comment in the example-config.yml

miles-grant-ibigroup (changes requested):
This should be consistent with the `displayA11yError` including the actual error ID itself

amy-corson-ibigroup:
Okay I have changed the name back, update the config, applied the error only when dealing with actual itineraries with rental legs, and `getActiveItineraries(getState())` wasn't the appropriate check for this section (it was leaving out RENT+WALK itineraries) so i fixed that too.

lib/actions/apiV2.js
```diff
@@ -1203,27 +1203,36 @@ export function routingQuery(searchId = null, updateSearchInReducer) {
...
-                    }
                     response.data.plan.routingErrors.push({
                       code: 'OTP_RR_A11Y_ROUTING_ENABLED'
                     })
                   }
                 }
+                if (displayMicromobilityRentalWarning) {
+                  if (
+                    withCollapsedShortNames.find((itin) =>
+                      itin.legs.find((leg) => leg.rentedBike)
```
miles-grant-ibigroup, line 1229: Can we add a comment that this is a deprecated field and we should change it?

miles-grant-ibigroup (approved):
Looks good! Thanks for the changes! I just have one small nitpick. If we want to be really thorough we should actually fix it but that would require a core-utils change so I'm happy to just add a comment

lib/actions/apiV2.js
amy-corson-ibigroup: I'm not seeing anything to show this is deprecated? It's not noted here whereas other fields are: https://docs.opentripplanner.org/api/dev-2.x/graphql-gtfs/types/Leg

PR #1416, opened 2025-06-12
Nearby View: Show Child stop codes

miles-grant-ibigroup:
Not sure about the styling...

Also contains a fix to not block the entire UI when there is an error, but still valid data.


daniel-heppner-ibigroup (approved):
image
Looks good to me. Once this is merged we can fix the error handling in my PR.

alec-georgoff (approved):

PR #1417, opened 2025-06-12
add loading message and clarify itinerary instructions

amy-corson-ibigroup:
**Description:**
- Add a loading status when itinerary results are loading
- Add instructions about how to open the itinerary for screenreaders

**PR Checklist:**


binh-dam-ibigroup (approved):

i18n/en-US.yml
```diff
@@ -360,7 +360,9 @@ components:
     title: Define Your Mobility Profile
   NarrativeItinerariesHeader:
     changeSortDir: Change sort direction
-    howToFindResults: To view results, see the Itineraries Found heading below.
+    howToFindResults: >-
+      To view results, see the Itineraries Found heading below. Select a time
+      button to open itinerary details.
```
miles-grant-ibigroup: Why is this split across 2 lines?

lib/components/app/batch-routing-panel.tsx
```diff
@@ -217,6 +217,15 @@ class BatchRoutingPanel extends Component<Props> {
                       overflowY: 'hidden'
                     }}
                   >
+                    <InvisibleA11yLabel
+                      aria-live="assertive"
+                      as="div"
+                      role="alert"
+                    >
+                      {activeSearch?.pending > 0 && (
```
miles-grant-ibigroup, line 232: we want the label even if the message is hidden?

miles-grant-ibigroup (approved):
Looks good! Just a few small things to check

lib/components/app/batch-routing-panel.tsx
amy-corson-ibigroup: Yes, status regions only announce if the content changes, so the status label needs to be there before we introduce the loading message

i18n/en-US.yml
amy-corson-ibigroup: Just trying to do the linter's job i guess!

PR #1418, opened 2025-06-12
pass intl object to trip status component

josh-willis-arcadis:
The pause/resume commands were not working due to the intl object being null. This is ensures the intl object is always present.

**PR Checklist:**


amy-corson-ibigroup (changes requested):
We're using `InjectIntl` in the `TripStatus` component, so this prop is redundant. This should help: https://stackoverflow.com/questions/36717880/how-to-use-react-intl-2-with-redux

lib/components/user/monitored-trip/trip-basics-pane.tsx
amy-corson-ibigroup, line 1: You don't really need this in the trip basics pane, it's only necessary whenever we're passing `intl` using `ownProps`

amy-corson-ibigroup (approved):
Works! You don't really need the change in trip-basics-pane but i dont know if it matters.

lib/components/user/monitored-trip/trip-basics-pane.tsx
josh-willis-arcadis: addressed in [b43d931](https://github.com/opentripplanner/otp-react-redux/pull/1418/commits/b43d9318d122f12022434c2be7e118b7044d48a7)

daniel-heppner-ibigroup (approved):
I did not realize the order of these matters. huh!

PR #1419, opened 2025-06-12
Add error display to nearby view

daniel-heppner-ibigroup:
**Description:**
This PR reuses the metro error component to display an error in the nearby view in case of some error in fetching the data. 

**PR Checklist:**

image


lib/actions/api.js
```diff
@@ -308,7 +308,7 @@ export const parkAndRideResponse = createAction('PARK_AND_RIDE_RESPONSE')
 export function parkAndRideQuery(
   params,
   responseAction = parkAndRideResponse,
-  errorAction = parkAndRideResponse,
+  errorAction = parkAndRideError,
```
miles-grant-ibigroup: Was this intentional? What is the benefit of doing this? I thought we switched it to response when we made it more resilient to errors

lib/components/narrative/metro/metro-error-renderer.tsx
```diff
@@ -21,7 +21,7 @@ const Container = styled.li`
   grid-template-rows: 1fr max-content;
   list-style-type: none;
   margin: 0;
-  padding: 0 1em;
+  padding: 1em;
```
miles-grant-ibigroup: Why was this change made? It's having an impact outside the nearby view

lib/components/narrative/metro/metro-error-renderer.tsx
```diff
@@ -40,8 +40,33 @@ const Container = styled.li`
     grid-row: 2;
     padding-bottom: 10px;
   }
+
+  p:not(h2 ~ p) {
```
miles-grant-ibigroup: Can you add a comment about what this is selecting?

miles-grant-ibigroup (changes requested):
Some unintended side effects!

lib/components/narrative/metro/metro-error-renderer.tsx
daniel-heppner-ibigroup: It was dependent on the header existing to make the component look right. With no header the icon is the tallest element, and with no padding it just went right to the edge. I'd rather have the padding on the wrapper component instead of on the children.

lib/components/narrative/metro/metro-error-renderer.tsx
daniel-heppner-ibigroup: good idea

lib/actions/api.js
daniel-heppner-ibigroup: The blame on this line shows the last change was 4 years ago. I am not sure what the reasoning was, but it caused the app the crash if there was no network connection.

daniel-heppner-ibigroup: well, I undid it and it seems to be fine. I guess the crash was something else.

lib/components/narrative/metro/metro-error-renderer.tsx
miles-grant-ibigroup: I'm not sure I like that we're changing unrelated styles. Could we extract the shared styles and then maintain the current nearby view display? Same goes for the support for an error with no header

daniel-heppner-ibigroup: found another way to get the padding I needed

lib/components/narrative/metro/metro-error-renderer.tsx
```diff
@@ -38,10 +38,38 @@ const Container = styled.li`
   p {
     grid-column: 2;
     grid-row: 2;
-    padding-bottom: 10px;
```
miles-grant-ibigroup, line 41: Sorry to nitpick so much, but I'd really like to avoid any changes to non-nearby view errors. Left is this PR, right is current. Screenshot 2025-07-08 at 1 05 13 PM

lib/components/narrative/metro/metro-error-renderer.tsx
```diff
@@ -38,10 +38,38 @@ const Container = styled.li`
...
+  iconSize = '3x'
+}: {
+  body?: React.ReactNode
+  header?: React.ReactNode
+  icon?: React.ElementType
+  iconSize?: string
+}): JSX.Element => (
+  <Container>
+    <Icon Icon={icon} size={iconSize} />
+    {header && <h2>{header}</h2>}
```
miles-grant-ibigroup: I'm not sure how I feel about an optional header. Can we maybe use a defaultMessage to just say "Warning" or "Alert"?

miles-grant-ibigroup (changes requested):
A few outstanding issues

daniel-heppner-ibigroup:
Okay, I went ahead and undid my changes that allowed for no title error renderers. Does simplify things.

miles-grant-ibigroup (approved):
Great work! Good component re-use. Just needs french strings and then I think we're good to go. Please don't manually approve percy tests -- this should be a 0-pixel-change PR

lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -62,6 +63,7 @@ type Props = {
   // Todo: type nearby results
   nearby: any
   nearbyViewCoords?: LatLonObj
+  nearbyViewError?: any
```
alec-georgoff, line 66: How much of a lift would it be to type this? I'm hesitant to add new `any` typings when we already have so many to fix

daniel-heppner-ibigroup: I looked into this and I don't think it's in the scope for this PR. This isn't really a "new" `any` since this data structure already existed in Redux. Creating a type for this would be great but it's kind of a big task and would involve typing a bunch of error values across the app, since it is just the generic error you get when `fetch` throws.

alec-georgoff (approved):

daniel-heppner-ibigroup:
@binh-dam-ibigroup help with translations please. :)

binh-dam-ibigroup:
@daniel-heppner-ibigroup FYI there are conflicts.

PR #1421, opened 2025-06-13
changed fixed with to min width

amy-corson-ibigroup:
**Description:**
The fixed width was not handling longer translations gracefully

**PR Checklist:**

| Before | After |
|--------|-------|
|image|image|


miles-grant-ibigroup (approved):
I have a memory that we changed it from min width to width at some point but I can't figure out why we possibly did that... I'm sure it'll be fine! We can always override

josh-willis-arcadis (approved):

PR #1422, opened 2025-06-19
Improve toast annoucnements for AT

amy-corson-ibigroup:
**Description:**
Adds "alert" to default settings to more consistently force toast to announce to AT.

Previously "create account complete" announcement upon exiting the wizard was not being announced because the URL/Page title changing was higher priority. Ensuring both announcements were read properly was requiring a complex set of fake `Promises`,  `setTimeouts` and aria settings. That toast is not actually needed, as all the changes were already saved. Instead, replace the toast with the a message on the "Account Complete" screen that informs users changes were saved and that clicking the "finish" button will redirect them.

**PR Checklist:**


lib/components/user/wizard.tsx
```diff
@@ -93,17 +93,6 @@ class Wizard extends Component<Props> {
         this._routeTo(nextId)
       }
     } else {
-      // Display a toast to acknowledge saved changes
-      // (although in reality, changes quietly took effect in previous screens).
-      toastPromise(
```
binh-dam-ibigroup, line 98: Also remove the imports of `'../util/toasts'` that are now unused.

i18n/en-US.yml
```diff
@@ -169,7 +169,9 @@ components:
   A11yPrefs:
     accessibilityRoutingByDefault: Prefer accessible trips by default
   AccountSetupFinishPane:
-    message: You are ready to start planning your trips.
+    message: >-
+      Your changes have been saved. You are ready to start planning trips. Click
```
binh-dam-ibigroup, line 173: That's a clever change!

binh-dam-ibigroup (approved):

lib/components/user/wizard.tsx
amy-corson-ibigroup: Good call, thank you!

josh-willis-arcadis (approved):
LGTM

PR #1431, opened 2025-06-25
Add realtime icon to itinerary

alec-georgoff:
**Description:**

This PR implements the `realtime-icon` class to the rendered time in the itinerary when the vehicle has realtime information.

**PR Checklist:**


lib/components/viewers/realtime-status-label.tsx
```diff
@@ -62,6 +62,7 @@ const STATUS = {
  * will be rendered above the status. Also, this can optionally be rendered with
  * a background color for a label-like presentation.
  */
+// eslint-disable-next-line complexity
```
alec-georgoff, line 65: Honestly, I don't have a full understanding of this rule. [Documentation here](https://eslint.org/docs/latest/rules/complexity). It was triggered by the addition of the ternary operation when applying the classname to the time div.

lib/components/viewers/realtime-status-label.tsx
```diff
@@ -106,12 +107,12 @@ const RealtimeStatusLabel = ({
         <TimeStruck aria-hidden>
           <FormattedTime timeStyle="short" value={originalTime} />
         </TimeStruck>
-        <div>
+        <div className={isRealtime ? 'realtime-icon realtime' : ''}>
```
miles-grant-ibigroup: Adding the realtime class is adding the realtime icon. I think this is fine, but we need to account for the new positioning.
Screenshot 2025-06-26 at 10 57 15 AM

Otherwise, we should avoid using the `realtime` class

miles-grant-ibigroup (changes requested):
Small possibly unintended side effect!

lib/components/viewers/realtime-status-label.tsx
alec-georgoff: Removed the usage of `realtime`! Found a way to accomplish what we wanted without it

miles-grant-ibigroup (approved):
Great work! Very clean and works great

PR #1433, opened 2025-06-25
give map popup the agency names so it can render them in the stop popup

daniel-heppner-ibigroup:
**Description:**
We needed to render the agency name for a stop in the stop popup. Sadly this information is not easy to get, it requires another network request. This change provides a callback for the tile overlay to request information about a stop ID, and then gives it a map of all the stops from the transit index along with their agency object. Note that because the stops can have multiple agencies, we just take the first one. Feel free to suggest a different behavior for shared stops.


daniel-heppner-ibigroup:
Blocked by https://github.com/opentripplanner/otp-ui/pull/857

daniel-heppner-ibigroup:
This is blocked as in we can't merge it until we have the new version numbers, but it can be reviewed!

miles-grant-ibigroup:
> This is blocked as in we can't merge it until we have the new version numbers, but it can be reviewed!

otp-ui is ready for merge! I'll wait until this PR has the updated package version and merge conflicts fixed

miles-grant-ibigroup:
Is this ready for review? If so, can you update package versions and fix merge conflicts?

daniel-heppner-ibigroup:
done!

lib/components/map/default-map.tsx
```diff
@@ -173,6 +175,13 @@ class DefaultMap extends Component {
     return <TransitOperatorIcons stopId={stopId} />
   }
 
+  getAgencyFromStopId = async (stopId: string) => {
+    this.props.findStopTimesForStop({
+      date: getCurrentDate(),
+      stopId
+    })
+  }
+
```
miles-grant-ibigroup: What's this for?

lib/components/map/default-map.tsx
```diff
@@ -264,6 +274,9 @@ class DefaultMap extends Component {
       lat: null,
       lon: null
     })
+
+    // Fetch feeds in the background
+    this.props.findFeeds()
```
miles-grant-ibigroup, line 297: We fetch feeds every time the query changes? Is there no better time? If we do it like this can we at least cache the result

lib/components/map/default-map.tsx
```diff
@@ -429,6 +443,7 @@ class DefaultMap extends Component {
                 // This must be a method that returns an array of JSX
                 // as the base-map requires that every toggleable layer
                 // is its own component, and not a subcomponent of another component
+                console.log('feeds', feeds)
```
miles-grant-ibigroup: uh oh!

lib/components/map/default-map.tsx
```diff
@@ -462,6 +478,7 @@ class DefaultMap extends Component {
 const mapStateToProps = (state) => {
   const activeSearch = getActiveSearch(state)
   const viewedRoute = state.otp?.ui?.viewedRoute?.routeId
+  const stops = state.otp.transitIndex.stops
```
miles-grant-ibigroup, line 508: What if transitIndex is undefined

lib/components/map/default-map.tsx
```diff
@@ -486,6 +503,7 @@ const mapStateToProps = (state) => {
     bikeRentalStations: state.otp.overlay.bikeRental.stations,
     carRentalStations: state.otp.overlay.carRental.stations,
     config: state.otp.config,
+    feeds: state.otp.transitIndex.feeds,
```
miles-grant-ibigroup, line 534: What if transit index is undefined

miles-grant-ibigroup (changes requested):
A few questions

lib/components/map/default-map.tsx
daniel-heppner-ibigroup: oops! no longer needed

lib/components/map/default-map.tsx
daniel-heppner-ibigroup: doesn't componentDidMount just run once?

lib/components/map/default-map.tsx
daniel-heppner-ibigroup: it won't be, it's defined in the initial state in the reducer.

lib/components/map/default-map.tsx
miles-grant-ibigroup: My bad! i thought this was part of a different method.

miles-grant-ibigroup (changes requested):
This PR is making the stop code appear twice. Is there any way to make this optional? Screenshot 2025-08-08 at 3 46 24 PM

miles-grant-ibigroup:
Also seeing some instances of the name and the icon not matching
Screenshot 2025-08-08 at 3 47 27 PM

miles-grant-ibigroup (approved):
Works well and code looks good! STill seeing some funky behavior when clickign on stations but I guess that's ok?

binh-dam-ibigroup:
I am not seeing the updated text using other languages (e.g. Spanish). Any ideas? I'll try a to find the cause of that.

binh-dam-ibigroup (approved):
Necessary changes for the scope of this PR seem fine. Out of scope for this PR: Better i18n handling for stop id formatting in other languages.

PR #1434, opened 2025-06-30
update trip details

daniel-heppner-ibigroup:
**Description:**
Updates trip-details to the new version that provides classnames and slightly refactored the props.


miles-grant-ibigroup (changes requested):
Not sure I see the reasoning for this... Seems to conflict with https://github.com/opentripplanner/otp-ui/pull/855

miles-grant-ibigroup (approved):
Aha I see now that this is separate. Approved if tests pass

package.json
```diff
@@ -247,5 +247,6 @@
     "snapshot": {
       "percy-css": ".percy-hide { opacity: 0!important; } "
     }
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
alec-georgoff, line 251: Do we want this included?

daniel-heppner-ibigroup: I think we should add this, may as well do it here.

daniel-heppner-ibigroup: yeah... why not. it keeps adding it and it seems like a good thing to have

daniel-heppner-ibigroup:
I went ahead and approved the percy snapshots even though the fare info is missing because:
- the percy tests don't have any fare config, so it shouldn't show
- they have outdated fare data (like pre-faresv2 api)
- we're overhauling fares soon anyway

alec-georgoff (approved):

amy-corson-ibigroup (approved):

PR #1435, opened 2025-07-02
Fix Trip Viewer

binh-dam-ibigroup:
## Description

Trip viewer was still using OTP1 stuff and breaks on OTP instances where the REST API is removed.
This PR updates the trip viewer to use GraphQL and removes corresponding OTP1 API code and mocks. There are no visual changes.

**PR Checklist:**


lib/actions/form.js
```diff
@@ -92,7 +92,14 @@ export function setQueryParam(payload, searchId) {
 export function parseUrlQueryString(params = getUrlParams(), source = null) {
   return function (dispatch, getState) {
     const state = getState()
-    if (state.otp.ui.mainPanelContent !== null) return state
+    // Do process the URL and fetch itineraries when refreshing a page showing the trip viewer,
+    // in addition to pages showing itinerary results.
+    if (
+      state.otp.ui.mainPanelContent !== null &&
+      state.otp.ui.mainPanelContent !== MainPanelContent.TRIP_VIEWER
```
miles-grant-ibigroup, line 100: Can we add a TODO that this isn't a perfect solution? I don't like the trip viewer being such a special case like this

miles-grant-ibigroup (approved):
Incredible fix thank you. I like how much we were able to tear out. My only concern is with how we are handling the routing here. I think this is a good solution for now but we should think about what we want to do here long term

lib/actions/form.js
binh-dam-ibigroup: Thank you, added in https://github.com/opentripplanner/otp-react-redux/pull/1435/commits/a6f7cab44bbc427c331f110c50da8c81a1cc90cd.

lib/reducers/create-otp-reducer.js
```diff
@@ -805,16 +805,6 @@ function createOtpReducer(config) {
             trips: { [action.payload.id]: { $set: action.payload } }
           }
         })
-      case 'FIND_STOPS_FOR_TRIP_RESPONSE':
```
alec-georgoff, line 808: Why are we able to remove these? I assumed they needed to be updated to conform to the new response structure.

binh-dam-ibigroup: With the old REST API, the trip query did not include stops and stop times, so you had to make separate requests to retrieve those (i.e. hit different API endpoints). With GraphQL you can do all of that with a single request.

alec-georgoff (approved):

josh-willis-arcadis (approved):
LGTM!

PR #1436, opened 2025-07-02
Schedule View Cleanup

miles-grant-ibigroup:
**Description:**
Makes a few cleanup changes to the schedule view to accommodate some other changes. Fixes icons, paddings, and icon placements
Screenshot 2025-07-02 at 11 57 19 AM


miles-grant-ibigroup:
Unsure about percy tests... I think they look good but I'll wait to get a second opinion

lib/components/viewers/nearby/stop-card-header.tsx
```diff
@@ -55,10 +55,28 @@ const StopCardHeader = ({
...
+              <Icon
+                Icon={MagnifyingGlassPlus}
+                style={{
+                  bottom: 3,
+                  fontSize: 16,
+                  marginLeft: '0.3em',
+                  position: 'relative'
+                }}
+              />
+              <InvisibleA11yLabel>{zoomButtonText}</InvisibleA11yLabel>
```
amy-corson-ibigroup: You don't need this, you can just add `aria-label` to the button

lib/components/viewers/stop-schedule-viewer.tsx
```diff
@@ -214,8 +214,7 @@ class StopScheduleViewer extends Component<Props, State> {
         <HeaderCard>
           {stopData?.name ? (
             <StopCardHeader
-              // FIXME: What icon should we use?
-              actionIcon={MagnifyingGlass}
+              actionIcon={SatelliteDish}
```
amy-corson-ibigroup: Hmm can we use the pin that we use in the mobile menu for consistency? 
image

lib/components/viewers/nearby/stop-card-header.tsx
```diff
@@ -55,10 +55,28 @@ const StopCardHeader = ({
...
+            </button>
+          ) : null}
         </CardTitle>
         <DistanceDisplay distance={stopData.distance} />
       </CardHeader>
-      <CardBody>
+      <CardBody style={{ marginTop: !stopData.code ? 5 : 'inherit' }}>
```
amy-corson-ibigroup: Can we get a little more margin between the header and body? It's a little cramped 


| Before | After |
|--------|-------|
|image|image
|

amy-corson-ibigroup: I also gave the zoom button `margin-left: 10px`

amy-corson-ibigroup: Also in cases where there's no stop code and the from/to picker moves up, it makes the margin smaller between the schedule picker and the timezone alert, can we make that consistent?

lib/components/viewers/nearby/stop-card-header.tsx
```diff
@@ -69,16 +87,6 @@ const StopCardHeader = ({
...
-              className="link-button"
-              onClick={onZoomClick}
-              title={zoomButtonText}
-            >
-              <Icon Icon={Search} style={{ marginLeft: '0.2em' }} />
-              <InvisibleA11yLabel>{zoomButtonText}</InvisibleA11yLabel>
-            </button>
-          ) : null}
           {actionPath && actionText ? (
             <ActionLink
```
amy-corson-ibigroup: Also, when the f/t picker floats up, it makes the difference in font size between it and the "view nearby" button and everything around it noticeable. Can we up the text size on that button? 
image

amy-corson-ibigroup (changes requested):
Just a few tweaks!

lib/components/viewers/nearby/stop-card-header.tsx
```diff
@@ -1,6 +1,6 @@
 import { connect } from 'react-redux'
 import { FormattedMessage, useIntl } from 'react-intl'
-import { Search } from '@styled-icons/fa-solid/Search'
+import { MagnifyingGlassPlus } from '@styled-icons/fa-solid/MagnifyingGlassPlus'
```
amy-corson-ibigroup: Ugh also i think i prefer the normal looking glass icon sorry, this one just feels like there's a lot going on. I can be convinced to keep it

lib/components/viewers/stop-schedule-viewer.tsx
miles-grant-ibigroup: But it's not for nearby it's for realtime departure times

miles-grant-ibigroup: Screenshot 2025-07-18 at 12 03 15 PM
Also in some cases that icon is already next to the title!

amy-corson-ibigroup: wait im tallking about the "view nearby" button? that's certainly for nearby, no?

i think there should be some consistency in the "nearby view" icon but its fine with me if we choose another icon and change the menu to match. I just don't like this icon because I'm confused about what a satellite would have to do with this. What about either of these? 
image

lib/components/viewers/nearby/stop-card-header.tsx
amy-corson-ibigroup: This is still inconsistent
image
image

lib/components/viewers/stop-schedule-viewer.tsx
```diff
@@ -275,7 +274,7 @@ class StopScheduleViewer extends Component<Props, State> {
     }
 
     return (
-      <div role="group" style={{ marginBottom: '10px' }}>
+      <div role="group" style={{ marginBottom: -15 }}>
```
amy-corson-ibigroup: This is making everything feel pretty cramped

amy-corson-ibigroup: I actually think we could change this to a `margin-top: 5px` to this container just to break up some of the density

amy-corson-ibigroup (changes requested):
looking better! Just clarifying some of my previous notes. The biggest thing is just that inconsistency in spacing between stations and stops

lib/components/viewers/stop-schedule-viewer.tsx
miles-grant-ibigroup: location is great!

lib/components/viewers/stop-schedule-viewer.tsx
miles-grant-ibigroup: Screenshot 2025-07-23 at 9 41 09 AM
Like this? This is no good this makes it look like we don't know how to configure line-height

amy-corson-ibigroup: I get that, I think just in general I wouldn't mind seeing a little more space between these lines, but nbd

amy-corson-ibigroup:
Schedule view is looking good but something about the nearby layout broke :( 
image

amy-corson-ibigroup (approved):
Looks good thanks for the chnages!

alec-georgoff:
My view looks a little different on Safari:

image

compared to your example:

image

Are these differences enough to worry about?

lib/components/viewers/nearby/stop-card-header.tsx
```diff
@@ -55,40 +54,54 @@ const StopCardHeader = ({
...
-            </ActionLink>
-          ) : null}
-        </div>
+        </CardTitle>
+        <DistanceDisplay distance={stopData.distance} />
+      </CardHeader>
+      <CardBody>
+        {(stopData.code || actionPath) && (
+          // TODO: Clean up these conditionals -- perhaps grid one level higher?
+          <div style={{ display: 'grid', height: 20 }}>
```
alec-georgoff, line 81: I notice a fair amount of inline styling; is there a reason we don't move these to CSS?

alec-georgoff:
Also, noticed that some ST routes are missing route names...I'm happy to look into this separately if you'd like!!

image

miles-grant-ibigroup:
> My view looks a little different on Safari:
> image
> 
> compared to your example:
> image
> 
> Are these differences enough to worry about?

These are the result of @amy-corson-ibigroup's comments!

lib/components/viewers/nearby/stop-card-header.tsx
miles-grant-ibigroup: We should probably have a broader discussion about setting an OTP-RR standard for where styling rules should live... It's a horrible mish-mash inside otp-rr. I prefer seeing the styles we're using directly instead of having a million classNames everywhere. Maybe styled-components should be the standard? But I'm not sure those are set up correctly in our project. Part of the problem is we use all 3 approaches

alec-georgoff: I think that each approach has upsides and downsides 🙃 Happy to discuss more offline!

alec-georgoff (approved):

PR #1437, opened 2025-07-03
Migrate findStopsWithinBBox to graphQL

alec-georgoff:
**Description:**

The goal of this PR is to migrate the `findStopsWithinBBox` function from the outdated REST API to the new GraphQL API.

**PR Checklist:**


lib/actions/apiV2.js
```diff
@@ -838,6 +838,43 @@ export const findRoute = (params) =>
     )
   }
 
+const receivedStopsWithinBBoxResponse = createAction(
+  'STOPS_WITHIN_BBOX_RESPONSE'
+)
+const receivedStopsWithinBBoxError = createAction('STOPS_WITHIN_BBOX_ERROR')
+
+// do we need to restrict to a certain boundary box?
```
alec-georgoff: Looking for a little more context on this function to understand how the bounding box is actually defined. These redux functions are very new to me and are obfuscating how this code is used....

alec-georgoff:
I was able to confirm via console logs that the stops are being correctly returned.

[Console.txt](https://github.com/user-attachments/files/21046219/Console.txt)

lib/actions/apiV2.js
miles-grant-ibigroup: This is less important now as our map can handle rendering more than 50 items at a time without crashing

miles-grant-ibigroup: but I'd still keep the comment. Maybe add a TODO

miles-grant-ibigroup (approved):
Great clean solution! We don't need bbox support for now, I think the comment is enough

binh-dam-ibigroup (approved):
Gonna approve, but this code is related to the StopsOverlay component... do we still need it?

miles-grant-ibigroup:
> Gonna approve, but this code is related to the StopsOverlay component... do we still need it?

Yes there are still quite a few uses of it, as the tiles can be a hassle to configure

PR #1438, opened 2025-07-09
i18n Fixes and Updates

amy-corson-ibigroup:
**Description:**
- Fixes a bug where neither Chinese language was being loaded correctly due to a difference in the language code and the file name.
- Creates a config item called `translateExternalSites` that opens an external menu item link with google translate depending on the active language
- Creates a util function called `convertChineseLanguageCode` that supplies a more commonly supported language code for both Chinese (simplified and traditional)

**PR Checklist:**

### DIFF
--- a/lib/components/app/app-menu.tsx
+++ b/lib/components/app/app-menu.tsx
@@ -19,7 +19,7 @@ import * as uiActions from '../../actions/ui'
 import { AppMenuItemConfig, LanguageConfig } from '../../util/config-types'
 import { AppReduxState } from '../../util/state-types'
 import { ComponentContext } from '../../util/contexts'
-import { getLanguageOptions } from '../../util/i18n'
+import { convertChineseLanguageCode, getLanguageOptions } from '../../util/i18n'
 import { isModuleEnabled, Modules } from '../../util/config'
 
 import AppMenuItem from './app-menu-item'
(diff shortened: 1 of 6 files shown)


lib/components/app/app-menu.tsx
```diff
@@ -284,7 +293,7 @@ class AppMenu extends Component<
                 })}
               />
             )}
-            {this._addExtraMenuItems(extraMenuItems)}
+            {this._addExtraMenuItems(extraMenuItems, translateExternalLinks)}
```
binh-dam-ibigroup, line 296: You probably don't need to pass `translateExternalLinks` around. Instead, access it from `_addExtraMenuItems` using `this.props`.

binh-dam-ibigroup (approved):
Just one refactor suggestion. Works well and well written!

lib/components/app/app-menu.tsx
amy-corson-ibigroup: When I used `this.props` I had to write a slightly more complex check inside the function to avoid adding the suffix to the locale header & sub-options on mobile. This seemed a little cleaner but either way works, if you'd prefer to use `this.props` i can!

josh-willis-arcadis (approved):

PR #1440, opened 2025-07-10
Vehicle rental overlay graphql migration

alec-georgoff:
**Description:**

This PR updates the `bikeRentalQuery` to use graphQL instead of the deprecated REST endpoint.

~Note that the updates to OTP-UI found in [this PR](https://github.com/opentripplanner/otp-ui/pull/863) must be merged to correctly use the response that comes back from graphQL.~

**PR Checklist:**


lib/components/map/default-map.tsx
```diff
@@ -333,17 +357,21 @@ class DefaultMap extends Component {
...
-      (station) => station.isFloatingBike === false && station.isFloatingVehicle
+
+    const scooters = rentalVehicles.filter(
+      (vehicle) => vehicle.vehicleType?.formFactor === FormFactor.SCOOTER
+    )
+
+    const micromobility = rentalVehicles.filter(
+      (vehicle) =>
+        vehicle.vehicleType &&
+        vehicle.vehicleType?.formFactor !== FormFactor.CAR
```
daniel-heppner-ibigroup: This will filter out any vehicles where vehicleType is undefined. Is that intentional? Do scooters/bikes never have undefined? Under what circumstance can it be undefined?

You should be able to remove the ? from the second part of the condition since the null check has already happened.

lib/components/map/default-map.tsx
```diff
@@ -145,10 +150,29 @@ function getLayerName(overlay, config, intl) {
...
+  itinerary: any
+  mapConfig: any
+  nearbyViewActive: any
+  pending: any
+  rentalVehicleQuery: any
+  rentalVehicles: RentalVehicle[]
+  setLocation: any
+  setViewedStop: any
+  viewedRouteStops: any
+}
```
daniel-heppner-ibigroup, line 176: I think you could define a bunch of these. There are some low hanging fruit like intl -> IntlShape.

daniel-heppner-ibigroup (changes requested):
Can you add some more types to the DefaultMap type before I approve?

lib/components/map/default-map.tsx
alec-georgoff: I think I was able to get all of them ([commit](https://github.com/opentripplanner/otp-react-redux/pull/1440/commits/98da8e52ff38bc45af122a0f6d2e3e31e7615e75)). Let me know if any stick out to you!

lib/components/map/default-map.tsx
alec-georgoff: Removed the redundant null check.

I think it's better to be more specific here and not include vehicles with an undefined `vehicleType`. It's better for a user to clearly not see their incorrectly-typed vehicles (and then fix them) than for users to incorrectly see vehicles that don't fit the expected type. Open to discussion though.

lib/components/map/default-map.tsx
daniel-heppner-ibigroup: Thank you!! Perfect

lib/components/map/default-map.tsx
daniel-heppner-ibigroup: agree

daniel-heppner-ibigroup (approved):

lib/actions/apiV2.js
```diff
@@ -687,6 +633,127 @@ const getVehiclePositionsForRoute = (routeId) =>
     )
   }
 
+const vehicleRentalStationsQuery = `
+        query VehicleRentalStations {
```
binh-dam-ibigroup: Remove excess indentation for the query block.

lib/actions/apiV2.js
```diff
@@ -687,6 +633,127 @@ const getVehiclePositionsForRoute = (routeId) =>
...
+          vehicleRentalStations {
+            id
+            name
+            lat
+            lon
+            allowDropoff
+            allowPickup
+            rentalNetwork {
+			        networkId
```
binh-dam-ibigroup: Remove excess indentation here too.

lib/actions/apiV2.js
```diff
@@ -687,6 +633,127 @@ const getVehiclePositionsForRoute = (routeId) =>
...
+        }
+      )
+    )
+  }
+}
+
+export const carRentalResponse = createAction('CAR_RENTAL_RESPONSE')
+export const carRentalError = createAction('CAR_RENTAL_ERROR')
+
+export function findCarRentalStations() {
```
binh-dam-ibigroup, line 704: Maybe for another PR: There is probably a way to make that query once and update the bike and rental states at the same time instead of filtering out the results. We could throttle the callbacks from the respective overlays too.

lib/actions/apiV2.js
```diff
@@ -687,6 +633,127 @@ const getVehiclePositionsForRoute = (routeId) =>
...
+    dispatch(
+      createGraphQLQueryAction(
+        vehicleRentalStationsQuery,
+        {},
+        bikeRentalResponse,
+        bikeRentalError,
+        {
+          rewritePayload: (payload) =>
+            payload.data.vehicleRentalStations.filter(
+              vehicleRentalStationFilter('BICYCLE')
```
binh-dam-ibigroup, line 693: Use constant `FormFactor.BICYCLE` that you used elsewhere.

lib/actions/apiV2.js
```diff
@@ -687,6 +633,127 @@ const getVehiclePositionsForRoute = (routeId) =>
...
+    dispatch(
+      createGraphQLQueryAction(
+        vehicleRentalStationsQuery,
+        {},
+        carRentalResponse,
+        carRentalError,
+        {
+          rewritePayload: (payload) =>
+            payload.data.vehicleRentalStations.filter(
+              vehicleRentalStationFilter('CAR')
```
binh-dam-ibigroup, line 715: Use constant `FormFactor.CAR`.

lib/actions/apiV2.js
```diff
@@ -687,6 +633,127 @@ const getVehiclePositionsForRoute = (routeId) =>
...
+            )
+        }
+      )
+    )
+  }
+}
+
+const rentalVehiclesQuery = `
+        query RentalVehicles {
```
binh-dam-ibigroup: Reduce excess indentation for this block.

binh-dam-ibigroup (approved):
Thanks for the changes. Seems to work and looks clean. Consider the suggested style tweaks and constants use before proceeding.

lib/actions/apiV2.js
alec-georgoff: Interestingly, I think my other usages of `FormFactor.CAR` are incorrect. Within the types package, `FormFactor` is a union type:

image

In the default map component, this means that we get type safety when using `'BICYCLE'` to compare against `station.vehicleType.formFactor`, since `formFactor` carries the `FormFactor` type:

image

Unfortunately, this `apiV2` file is not TypeScript, so we don't get that additional safety here.

PR #1441, opened 2025-07-17
use OTP2 standard for badge display

miles-grant-ibigroup:
OTP1 used `1` or `0` to declare true or false. OTP2 has a different approach.


binh-dam-ibigroup (approved):
Thanks for the update!

alec-georgoff (approved):

PR #1442, opened 2025-07-21
feat: Add filtering to nearby view

amy-corson-ibigroup:
**Description:**
Adds configurable filtering inside the nearby view that corresponds to the card types in nearby view. Also filters map layers and removes the layers list in nearby in cases where the filtering config is active.

Curious if folks feel like we need some key value pairs for these card values? Using "Stop" "RentalVehicle" etc for all these values feels a little messy.

**PR Checklist:**

image

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -104,7 +104,7 @@ const TermsOfStorage = () => (
 )
 
 // eslint-disable-next-line complexity
-const SvgIcon = ({ className, iconName, style }) => {
+const SvgIcon = ({ className, FallbackIcon, iconName, style }) => {
   const CustomIcon = getCustomIcon && getCustomIcon(iconName)
   if (CustomIcon) return <CustomIcon className={className} style={style} />
   // Some often used defaults
@@ -135,15 +135,20 @@ const SvgIcon = ({ className, iconName, style }) => {
     case 'clock-o':
       return <Clock className={className} style={style} />
     default:
-      console.warn(
-        `Custom icon provider not found for icon ${iconName}. Using fallback icon ExternalLinkAlt.`
-      )
-      return <ExternalLinkAlt className={className} style={style} />
+      if (FallbackIcon) {
+        return <FallbackIcon />
+      } else {
+        console.warn(
+          `Custom icon provider not found for icon ${iconName}. Using fallback icon ExternalLinkAlt.`
+        )
+        return <ExternalLinkAlt className={className} style={style} />
+      }
   }
 }
 
 SvgIcon.propTypes = {
   className: string,
+  FallbackIcon: object | undefined,
   iconName: string,
   style: object
 }
--- a/lib/app.js
+++ b/lib/app.js
@@ -19,7 +19,7 @@ import React from 'react'
 // import OTP-RR components
 
 // Loads a yaml config file from /tmp folder that contains customizations.
-import { Parking } from '@styled-icons/fa-solid'
+import { Parking } from '@styled-icons/fa-solid/Parking'
 
 import otpConfig from '../tmp/config.yml'
 
@@ -136,7 +136,7 @@ const SvgIcon = ({ className, Fallback, iconName, style }) => {
       return <Clock className={className} style={style} />
     default:
       if (Fallback) {
-        return <Fallback />
+        return <Fallback className={className} style={style} />
       } else {
         console.warn(
           `Custom icon provider not found for icon ${iconName}. Using fallback icon ExternalLinkAlt.`
@@ -148,7 +148,7 @@ const SvgIcon = ({ className, Fallback, iconName, style }) => {
 
 SvgIcon.propTypes = {
   className: string,
-  Fallback: object | undefined,
+  Fallback: object,
   iconName: string,
   style: object
 }
--- a/lib/app.js
+++ b/lib/app.js
@@ -104,7 +104,7 @@ const TermsOfStorage = () => (
 )
 
 // eslint-disable-next-line complexity
-const SvgIcon = ({ className, FallbackIcon, iconName, style }) => {
+const SvgIcon = ({ className, Fallback, iconName, style }) => {
   const CustomIcon = getCustomIcon && getCustomIcon(iconName)
   if (CustomIcon) return <CustomIcon className={className} style={style} />
   // Some often used defaults
@@ -135,8 +135,8 @@ const SvgIcon = ({ className, FallbackIcon, iconName, style }) => {
     case 'clock-o':
       return <Clock className={className} style={style} />
     default:
-      if (FallbackIcon) {
-        return <FallbackIcon />
+      if (Fallback) {
+        return <Fallback />
       } else {
         console.warn(
           `Custom icon provider not found for icon ${iconName}. Using fallback icon ExternalLinkAlt.`
@@ -148,7 +148,7 @@ const SvgIcon = ({ className, FallbackIcon, iconName, style }) => {
 
 SvgIcon.propTypes = {
   className: string,
-  FallbackIcon: object | undefined,
+  Fallback: object | undefined,
   iconName: string,
   style: object
 }
--- a/lib/app.js
+++ b/lib/app.js
@@ -19,6 +19,8 @@ import React from 'react'
 // import OTP-RR components
 
 // Loads a yaml config file from /tmp folder that contains customizations.
+import { Parking } from '@styled-icons/fa-solid'
+
 import otpConfig from '../tmp/config.yml'
 
 // Loads a configuration JavaScript file from the /tmp folder that contains customizations.
@@ -128,6 +130,8 @@ const SvgIcon = ({ className, iconName, style }) => {
       return <Utensils className={className} style={style} />
     case 'plus':
       return <Plus className={className} style={style} />
+    case 'parking':
+      return <Parking className={className} style={style} />
     case 'clock-o':
       return <Clock className={className} style={style} />
     default:
--- a/lib/components/map/default-map.tsx
+++ b/lib/components/map/default-map.tsx
@@ -63,7 +63,7 @@ const MapContainer = styled.div<{ hideLayerFilters: boolean }>`
 
   // If we're using filtering in the nearby view, hide the toggleable layers so there's no confusion.
   ul.layers-list {
-    display: ${(props) => (props.hideLayerFilters ? 'none' : 'block')};
+    visibility: ${(props) => (props.hideLayerFilters ? 'hidden' : 'visible')};
   }
 `
 /**
@@ -165,18 +165,18 @@ class DefaultMap extends Component {
   }
 
   _filterNearbyMapLayers = () => {
-    const { mapConfig, nearbyFilters } = this.props
+    const { activeNearbyFilters, mapConfig } = this.props
     const { overlays } = mapConfig
     const newOverlays = overlays
       .filter((overlay) => {
-        return overlay.cardType ? nearbyFilters[overlay.cardType] : true
+        return overlay.cardType ? activeNearbyFilters[overlay.cardType] : true
       })
       .map((overlay) => {
         if (overlay.layers) {
           return {
             ...overlay,
             layers: overlay?.layers?.filter(
-              (layer) => nearbyFilters[layer.cardType]
+              (layer) => activeNearbyFilters[layer.cardType]
             )
           }
         }
@@ -285,7 +285,7 @@ class DefaultMap extends Component {
   }
 
   componentDidMount() {
-    this._filterNearbyMapLayers()
+    this.props.nearbyFilters && this._filterNearbyMapLayers()
     // HACK: Set state lat and lon to null to prevent re-rendering of the
     // underlying OTP-UI map.
     this.setState({
@@ -295,11 +295,14 @@ class DefaultMap extends Component {
   }
 
   componentDidUpdate(prevProps) {
-    const { mapConfig, nearbyFilters, nearbyViewActive } = this.props
+    const { activeNearbyFilters, mapConfig, nearbyViewActive } = this.props
     // Check if any overlays should be toggled due to mode change
     this._handleQueryChange(prevProps.query, this.props.query)
     // If we're in the nearby view, filter the visible layers according to the nearby view filters.
-    if (prevProps.nearbyFilters !== nearbyFilters && nearbyViewActive)
+    if (
+      prevProps.activeNearbyFilters !== activeNearbyFilters &&
+      nearbyViewActive
+    )
       this._filterNearbyMapLayers()
     // If we leave the nearby view, reset the map overlays back to defaults.
     if (prevProps.nearbyViewActive !== nearbyViewActive && !nearbyViewActive) {
@@ -320,7 +323,7 @@ class DefaultMap extends Component {
       intl,
       itinerary,
       mapConfig,
-      nearbyFiltersConfigured,
+      nearbyFilters,
       nearbyViewActive,
       pending,
       setLocation,
@@ -364,7 +367,7 @@ class DefaultMap extends Component {
     return (
       <MapContainer
         className="percy-hide"
-        hideLayerFilters={nearbyViewActive && nearbyFiltersConfigured}
+        hideLayerFilters={nearbyViewActive && nearbyFilters}
       >
         <BaseMap
           baseLayer={
@@ -504,8 +507,8 @@ class DefaultMap extends Component {
 const mapStateToProps = (state) => {
   const activeSearch = getActiveSearch(state)
   const viewedRoute = state.otp?.ui?.viewedRoute?.routeId
-  const nearbyFilters = state.otp?.ui?.nearbyView?.filters
-  const nearbyFiltersConfigured = state.otp.config.nearbyView.filterConfig
+  const activeNearbyFilters = state.otp?.ui?.nearbyView?.filters
+  const nearbyFilters = state.otp.config?.nearbyView?.filters
   const nearbyViewerActive =
     state.otp.ui.mainPanelContent === MainPanelContent.NEARBY_VIEW
 
@@ -527,13 +530,13 @@ const mapStateToProps = (state) => {
       : null
 
   return {
+    activeNearbyFilters,
     bikeRentalStations: state.otp.overlay.bikeRental.stations,
     carRentalStations: state.otp.overlay.carRental.stations,
     config: state.otp.config,
     itinerary: getActiveItinerary(state),
     mapConfig: state.otp.config.map,
     nearbyFilters,
-    nearbyFiltersConfigured,
     nearbyViewActive:
       state.otp.ui.mainPanelContent === MainPanelContent.NEARBY_VIEW,
     pending: activeSearch ? Boolean(activeSearch.pending) : false,
--- a/lib/components/map/default-map.tsx
+++ b/lib/components/map/default-map.tsx
@@ -167,9 +167,9 @@ class DefaultMap extends Component {
     const { activeNearbyFilters, mapConfig } = this.props
     const { overlays } = mapConfig
     const nearbyViewFilteredOverlays = overlays
-      ?.filter((overlay) => {
-        return overlay.cardType && !!activeNearbyFilters[overlay.cardType]
-      })
+      ?.filter((overlay) =>
+        overlay.cardType ? activeNearbyFilters[overlay.cardType] : true
+      )
       .map((overlay) => {
         if (overlay.layers) {
           return {
@@ -293,23 +293,8 @@ class DefaultMap extends Component {
   }
 
   componentDidUpdate(prevProps) {
-    const { activeNearbyFilters, mapConfig, nearbyViewActive } = this.props
     // Check if any overlays should be toggled due to mode change
     this._handleQueryChange(prevProps.query, this.props.query)
-    // If we're in the nearby view, filter the visible layers according to the nearby view filters.
-    if (
-      prevProps.activeNearbyFilters !== activeNearbyFilters &&
-      nearbyViewActive
-    )
-      if (
-        prevProps.nearbyViewActive !== nearbyViewActive &&
-        !nearbyViewActive
-      ) {
-        // If we leave the nearby view, reset the map overlays back to defaults.
-        this.setState({
-          filteredOverlays: mapConfig?.overlays
-        })
-      }
   }
 
   render() {
--- a/lib/components/map/default-map.tsx
+++ b/lib/components/map/default-map.tsx
@@ -42,7 +42,7 @@ import TripViewerOverlay from './connected-trip-viewer-overlay'
 import VehicleRentalOverlay from './connected-vehicle-rental-overlay'
 import withMap from './with-map'
 
-const MapContainer = styled.div`
+const MapContainer = styled.div<{ hideLayerFilters: boolean }>`
   height: 100%;
   width: 100%;
 
@@ -59,6 +59,10 @@ const MapContainer = styled.div`
   .mapboxgl-popup-content {
     border-radius: 10px;
     box-shadow: 0 3px 14px 4px rgb(0 0 0 / 20%);
+  }
+
+  ul.layers-list {
+    display: ${(props) => (props.hideLayerFilters ? 'none' : 'block')};
   }
 `
 /**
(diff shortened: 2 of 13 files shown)


lib/components/map/default-map.tsx
```diff
@@ -60,6 +60,11 @@ const MapContainer = styled.div`
     border-radius: 10px;
     box-shadow: 0 3px 14px 4px rgb(0 0 0 / 20%);
   }
+
+  // If we're using filtering in the nearby view, hide the toggleable layers so there's no confusion.
+  ul.layers-list {
+    display: ${(props) => (props.hideLayerFilters ? 'none' : 'block')};
```
miles-grant-ibigroup: Should we use `visibility` instead of `display` so there's less potential for rendering issues?

lib/components/map/default-map.tsx
```diff
@@ -152,12 +157,34 @@ class DefaultMap extends Component {
...
       zoom
     }
   }
 
+  _filterNearbyMapLayers = () => {
+    const { mapConfig, nearbyFilters } = this.props
+    const { overlays } = mapConfig
+    const newOverlays = overlays
+      .filter((overlay) => {
+        return overlay.cardType ? nearbyFilters[overlay.cardType] : true
```
miles-grant-ibigroup: I'd prefer for `cardType` to not be a mandatory property. Can we handle it being unset?

lib/util/config-types.ts
```diff
@@ -68,9 +68,15 @@ interface ApiKeyConfig {
 
 export type BugsnagConfig = ApiKeyConfig
 export type MapillaryConfig = ApiKeyConfig
+export type NearbyFilterConfig = {
+  cardType: 'Stop' | 'RentalVehicle' | 'VehicleParking' | 'BikeRentalStation'
+  defaultOn: boolean
```
miles-grant-ibigroup: Can this be optional?

lib/components/map/default-map.tsx
```diff
@@ -462,6 +504,8 @@ class DefaultMap extends Component {
 const mapStateToProps = (state) => {
   const activeSearch = getActiveSearch(state)
   const viewedRoute = state.otp?.ui?.viewedRoute?.routeId
+  const nearbyFilters = state.otp?.ui?.nearbyView?.filters
+  const nearbyFiltersConfigured = state.otp.config.nearbyView.filterConfig
```
miles-grant-ibigroup: These are named quite confusingly. What do you think about this?
```suggestion
  const activeNearbyFilters = state.otp?.ui?.nearbyView?.filters
  const nearbyFilters = state.otp.config.nearbyView.filters
```

lib/util/config-types.ts
```diff
@@ -68,9 +68,15 @@ interface ApiKeyConfig {
...
+  iconName: string
+}
 
 export type NearbyViewConfig = {
   alwaysShowLongName?: boolean
+  filterConfig?: Array<NearbyFilterConfig>
```
miles-grant-ibigroup:
```suggestion
  filters?: Array<NearbyFilterConfig>
```
Putting the word config inside a config

lib/components/map/default-map.tsx
miles-grant-ibigroup: Also, can we add optionals here? This is crashing if the nearby view isn't configured

lib/components/viewers/nearby/nearby-filter.tsx
```diff
@@ -0,0 +1,78 @@
...
+  value: boolean
+}): JSX.Element => {
+  // @ts-expect-error component context
+  const { ModeIcon } = useContext(ComponentContext)
+  const FilterIcon = () =>
+    filter.cardType === 'VehicleParking' ? (
+      <Parking />
+    ) : (
+      <ModeIcon className="mode-svg" mode={filter.iconName} />
+    )
```
miles-grant-ibigroup: This feels like quite the hack... Can we use `SvgIcon` from context instead? With a fallback to ModeIcon.

lib/components/viewers/nearby/nearby-filter.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  return (
+    <>
```
miles-grant-ibigroup: This wrapper is not needed

lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -128,26 +142,41 @@ function NearbyView({
...
+    // Set the filters according to the configuration settings. If there's no filter config, set the filters to default (which is just all filters set to true)
+    filterConfig
+      ? Object.fromEntries(
+          filterConfig.map((filter: Filter) => [
+            filter.cardType,
+            filter.defaultOn
+          ])
+        )
+      : nearbyFilters
+  )
```
miles-grant-ibigroup: Why do we maintain this here? Shouldn't this live in redux?

lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -158,13 +187,20 @@ function NearbyView({
...
   )
-
   const reverseCoords = (coords: LonLatInput) => {
     getGeocoder(geocoderConfig)
       .reverse({ point: coords })
       .then((result: Location) => setReversedPoint(result?.name || ''))
   }
 
+  const scrollToTop = () => {
+    console.log(nearbyContainerRef)
```
miles-grant-ibigroup: whoops!

lib/util/config-types.ts
miles-grant-ibigroup: Can we rename this to `default`?

lib/components/viewers/nearby/nearby-filter.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+import { ComponentContext } from '../../../util/contexts'
+import { getBaseColor } from '../../util/colors'
+import { invisibleCss } from '@opentripplanner/trip-form/lib/MetroModeSelector'
+import { NearbyFilterConfig } from '../../../util/config-types'
+
+import React, { useContext } from 'react'
+import styled from 'styled-components'
+
+const StyledFilterCheckbox = styled.label<{ checked: boolean }>`
```
miles-grant-ibigroup, line 12: `cursor: pointer` on hover?

miles-grant-ibigroup: Could we add some hover effects too? Some kind of opacity adjustment

miles-grant-ibigroup (changes requested):
Behaviorally and visually this is perfect! I am a huge fan and I think it looks great. Code-wise there are some tweaks I would make to clean things up. 

I also think the filtering could be more efficient but let's do all these initial fixes first

lib/components/viewers/nearby/nearby-filter.tsx
miles-grant-ibigroup: Also I'd add `Parking` to `SvgIcon`

lib/components/map/default-map.tsx
amy-corson-ibigroup: Wait can you explain? It's not a mandatory property, that's why there's a check here

lib/components/map/default-map.tsx
amy-corson-ibigroup: Hmm I don't think it matters here, since it's `absolute` but added

lib/components/viewers/nearby/nearby-filter.tsx
amy-corson-ibigroup: How would you add `ModeIcon` as a fallback in this case?

miles-grant-ibigroup: Good point... Probably need a new param for a `default` case override

lib/reducers/create-otp-reducer.js
```diff
@@ -165,6 +165,8 @@ export function getInitialState(userDefinedConfig) {
 
   const sessionId = new URLSearchParams(getUrlParams()).get('sessionId')
 
+  const nearbyFilters = config.nearbyView.filters
```
miles-grant-ibigroup:
```suggestion
  const nearbyFilters = config?.nearbyView?.filters
```
Shouldn't this stay optional!

miles-grant-ibigroup (changes requested):
Getting there, but we should use `SvgIcon` to avoid a lot of the icon code and it's still crashing on configs that don't configure the nearby view!

lib/reducers/create-otp-reducer.js
amy-corson-ibigroup: Fix it crashing and made it crash again in the same commit... my mind 🚀

lib/components/map/default-map.tsx
```diff
@@ -152,12 +157,34 @@ class DefaultMap extends Component {
       initZoom: zoom = 13
     } = props.mapConfig || {}
     this.state = {
+      filteredOverlays: this.props.mapConfig.overlays,
```
miles-grant-ibigroup: We should add some optionals because I don't think overlays is mandatory!

lib/components/map/default-map.tsx
```diff
@@ -152,12 +157,34 @@ class DefaultMap extends Component {
...
       lon,
       zoom
     }
   }
 
+  _filterNearbyMapLayers = () => {
+    const { activeNearbyFilters, mapConfig } = this.props
+    const { overlays } = mapConfig
+    const newOverlays = overlays
+      .filter((overlay) => {
```
miles-grant-ibigroup: Optional here too
```suggestion
      ?.filter((overlay) => {
```

lib/components/map/default-map.tsx
```diff
@@ -267,8 +295,21 @@ class DefaultMap extends Component {
...
+    // If we're in the nearby view, filter the visible layers according to the nearby view filters.
+    if (
+      prevProps.activeNearbyFilters !== activeNearbyFilters &&
+      nearbyViewActive
+    )
+      this._filterNearbyMapLayers()
+    // If we leave the nearby view, reset the map overlays back to defaults.
+    if (prevProps.nearbyViewActive !== nearbyViewActive && !nearbyViewActive) {
+      this.setState({
+        filteredOverlays: mapConfig.overlays
```
miles-grant-ibigroup:
```suggestion
        filteredOverlays: mapConfig?.overlays
```

lib/components/viewers/nearby/nearby-filter.tsx
```diff
@@ -0,0 +1,80 @@
+import { Check2 } from '@styled-icons/bootstrap'
+import { Parking } from '@styled-icons/fa-solid'
```
miles-grant-ibigroup:
```suggestion
import { Check2 } from '@styled-icons/bootstrap/Check2'
import { Parking } from '@styled-icons/fa-solid/Parking'
```
Don't want to break tree shaking!

lib/app.js
```diff
@@ -102,7 +104,7 @@ const TermsOfStorage = () => (
 )
 
 // eslint-disable-next-line complexity
-const SvgIcon = ({ className, iconName, style }) => {
+const SvgIcon = ({ className, FallbackIcon, iconName, style }) => {
```
miles-grant-ibigroup: Can we call this `Fallback` instead of `FallbackIcon`? Since you can send any jsx you like
```suggestion
const SvgIcon = ({ className, Fallback, iconName, style }) => {
```

lib/components/viewers/nearby/nearby-filter.tsx
```diff
@@ -0,0 +1,80 @@
...
+  onChange: (arg: any) => void
+  value: boolean
+}): JSX.Element => {
+  // @ts-expect-error component context
+  const { ModeIcon, SvgIcon } = useContext(ComponentContext)
+  // The icon is either in SvgIcon or ModeIcon so provide both
+  const ModeFilterIcon = () => (
+    <ModeIcon className="mode-svg" mode={filter.iconName} />
+  )
+  const FilterIcon = () => (
```
miles-grant-ibigroup: We can probably inline this

lib/reducers/create-otp-reducer.js
```diff
@@ -232,6 +234,23 @@ export function getInitialState(userDefinedConfig) {
...
+                nearbyFilters.map((filter) => [
+                  filter.cardType,
+                  filter.defaultOn
+                ])
+              )
+            : {
+                BikeRentalStation: true,
+                RentalVehicle: true,
+                Stop: true,
+                VehicleParking: true
```
miles-grant-ibigroup, line 251: @daniel-heppner-ibigroup is there a way to get this list from the type?

miles-grant-ibigroup (changes requested):
Almost there! I can see a vague outline of some larger-scale cleanup but I think overall things look quite good. Just a few more small nitpicks and then we should be good to go!

lib/reducers/create-otp-reducer.js
```diff
@@ -232,6 +234,23 @@ export function getInitialState(userDefinedConfig) {
...
       mainPanelContent: null,
       mobileScreen: MobileScreens.WELCOME_SCREEN,
+      nearbyView: {
+        filters:
+          // Set the filters according to the configuration settings. If there's no filter config, set the filters to default (which is just all filters set to true)
+          nearbyFilters
+            ? Object.fromEntries(
+                nearbyFilters.map((filter) => [
+                  filter.cardType,
+                  filter.defaultOn
```
miles-grant-ibigroup: Sorry for one last nitpick -- can we change this to be `initiallyShown` to match similar config items elsewhere?

miles-grant-ibigroup (approved):
This is really great! Really fast too. Just one last tiny change and then we are good to go!!

lib/reducers/create-otp-reducer.js
daniel-heppner-ibigroup: No there aren't really any ways to convert a type to code afaik, other than enums. But I think the consensus is strongly against enums at this point mostly because they convert types to code which breaks the idea that typescript is just types and doesn't affect the logic of the code.

miles-grant-ibigroup: I mean it's not really logic is it? It's more of a data source. As if we put the list of types in a JSON file somewhere! Still, I'm happy to concede to consensus

daniel-heppner-ibigroup: I mean either way I don't think there is any way to get this list from the types. I could be wrong. Maybe there's a way to get the type from the code?

lib/app.js
```diff
@@ -128,18 +130,25 @@ const SvgIcon = ({ className, iconName, style }) => {
...
+          `Custom icon provider not found for icon ${iconName}. Using fallback icon ExternalLinkAlt.`
+        )
+        return <ExternalLinkAlt className={className} style={style} />
+      }
   }
 }
 
 SvgIcon.propTypes = {
   className: string,
+  Fallback: object | undefined,
```
daniel-heppner-ibigroup: you can't do union type in prop types because it's not actually typescript. what would it take to typescript this file? That would be the best solution. Otherwise you can just remove the `| undefined`.

lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -43,6 +52,7 @@ type CurrentPosition = { coords?: { latitude: number; longitude: number } }
 type ServiceWeek = { end: string; start: string }
 
 type Props = {
+  activeNearbyFilters: any
```
daniel-heppner-ibigroup: I know this is no fun but can you please add valid types to your new props? Since these are isolated new features it shouldn't be too involved to make sure they have valid types. (1x other instance)

lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -61,13 +71,15 @@ type Props = {
...
   nearby: any
+  nearbyFilters?: Array<NearbyFilterConfig>
   nearbyViewCoords?: LatLonObj
   radius?: number
   routeSortComparator: (a: PatternStopTime, b: PatternStopTime) => number
   sessionSearches: any
   setHighlightedLocation: (location: Location | null) => void
   setLocation: SetLocationHandler
   setMainPanelContent: (content: number) => void
+  setNearbyViewFilter: (arg: any) => void
```
daniel-heppner-ibigroup: Remove this any type too

lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -134,20 +147,24 @@ function NearbyView({
...
+  setNearbyViewFilter,
   setViewedNearbyCoords,
   zoomToPlace
 }: Props): JSX.Element {
   const map = useMap().default
   const intl = useIntl()
   const [loading, setLoading] = useState(true)
   const [reversedPoint, setReversedPoint] = useState('')
-  const firstItemRef = useRef<HTMLDivElement>(null)
+  const [filters, setFilters] = useState(activeNearbyFilters)
```
daniel-heppner-ibigroup: Instead of keeping the active nearby filter state in the component locally and trying to keep it synchronized with a useeffect, I think it would be better to remove this entirely and just rely on the parent to manage state.

lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -208,12 +231,19 @@ function NearbyView({
     }
   }, [map, setViewedNearbyCoords, setHighlightedLocation])
 
+  const onFilterChange = (event: FormEvent) => {
+    const { target } = event
+    const { id } = target as HTMLInputElement
+    setFilters({ ...filters, [id]: !filters[id] })
```
daniel-heppner-ibigroup: Put the scrollToTop() call here and then remove the useEffect below.
Instead of calling setFilters() just use the setNearbyViewFilter() action directly

lib/components/viewers/nearby/nearby-view.tsx
daniel-heppner-ibigroup: You can add these types in the app state file

lib/app.js
```diff
@@ -19,6 +19,8 @@ import React from 'react'
 // import OTP-RR components
 
 // Loads a yaml config file from /tmp folder that contains customizations.
+import { Parking } from '@styled-icons/fa-solid'
```
daniel-heppner-ibigroup: Import the tree-shakable one `@styled-icons/fa-solid/Parking`

lib/components/map/default-map.tsx
```diff
@@ -152,12 +157,34 @@ class DefaultMap extends Component {
...
       zoom
     }
   }
 
+  _filterNearbyMapLayers = () => {
+    const { activeNearbyFilters, mapConfig } = this.props
+    const { overlays } = mapConfig
+    const newOverlays = overlays
+      ?.filter((overlay) => {
+        return overlay.cardType ? activeNearbyFilters[overlay.cardType] : true
```
daniel-heppner-ibigroup:
```suggestion
        return overlay.cardType && !!activeNearbyFilters[overlay.cardType]
```

lib/app.js
```diff
@@ -128,18 +130,25 @@ const SvgIcon = ({ className, iconName, style }) => {
...
+      return <Parking className={className} style={style} />
     case 'clock-o':
       return <Clock className={className} style={style} />
     default:
-      console.warn(
-        `Custom icon provider not found for icon ${iconName}. Using fallback icon ExternalLinkAlt.`
-      )
-      return <ExternalLinkAlt className={className} style={style} />
+      if (Fallback) {
+        return <Fallback />
```
daniel-heppner-ibigroup: Fallback should receive the same props as the others right?

lib/components/map/default-map.tsx
```diff
@@ -152,12 +157,34 @@ class DefaultMap extends Component {
...
+          return {
+            ...overlay,
+            layers: overlay?.layers?.filter(
+              (layer) => activeNearbyFilters[layer.cardType]
+            )
+          }
+        }
+        return overlay
+      })
+    return this.setState({ filteredOverlays: newOverlays })
```
daniel-heppner-ibigroup: Was this intentional? this.setState() returns void so you're not actually returning anything here

lib/components/map/default-map.tsx
daniel-heppner-ibigroup, line 1: I think we can do something better for the logic/data flow of the filter variables in this file. I don't think any of the layer stuff actually needs to be stateful at all in this component; all of the data you need to determine which overlays to include is available from the props in each render. It's best to not introduce statefulness into places where its not needed, instead just calculate values that you need from the single source of truth.

Basically it would look something like this (in render()):
`const activeFilters = this.props.nearbyViewActive ? getFilteredOverlays(this.props.activeNearbyFilters) : this.props.mapConfig?.overlays`

The getFilteredOverlays function would have the logic to filter `this.props.mapConfig?.overlays` based on the activeNearbyFilters. You already have this logic in _filterNearbyMapLayers. 

Then I think you can fully remove the new `filteredOverlays` state.

lib/components/viewers/nearby/nearby-filter.tsx
```diff
@@ -0,0 +1,77 @@
...
+  }
+`
+
+export const FilterCheckboxes = ({
+  filter,
+  onChange,
+  value
+}: {
+  filter: NearbyFilterConfig
+  onChange: (arg: any) => void
```
daniel-heppner-ibigroup: Give a type for this `any`

lib/components/viewers/nearby/nearby-filter.tsx
```diff
@@ -0,0 +1,77 @@
+import { Check2 } from '@styled-icons/bootstrap/Check2'
+import { Parking } from '@styled-icons/fa-solid/Parking'
+
+import { ComponentContext } from '../../../util/contexts'
+import { getBaseColor } from '../../util/colors'
+import { invisibleCss } from '@opentripplanner/trip-form/lib/MetroModeSelector'
```
daniel-heppner-ibigroup: Don't import from /lib/ directly. If this isn't exported then it needs to be exported in an otp ui update

lib/reducers/create-otp-reducer.js
```diff
@@ -232,6 +234,23 @@ export function getInitialState(userDefinedConfig) {
       localizedMessages: null,
       mainPanelContent: null,
       mobileScreen: MobileScreens.WELCOME_SCREEN,
+      nearbyView: {
+        filters:
+          // Set the filters according to the configuration settings. If there's no filter config, set the filters to default (which is just all filters set to true)
+          nearbyFilters
+            ? Object.fromEntries(
```
daniel-heppner-ibigroup, line 241: damn this is clever, nice job.

lib/reducers/create-otp-reducer.js
daniel-heppner-ibigroup: One solution here might be to just make the default value for this be `{ }` and then deal with it in the nearby view like `layers[__typename] ?? true`

Then if the property doesn't exist in `layers` you get true, otherwise you get its value. It is important to use `??` instead of `||` for this though because `||` will give you the right hand side value for `false`, whereas `??` only gives the RHS value for undefined or null.

daniel-heppner-ibigroup (changes requested):
Let me know if you want to pair program the fixes to this! I am happy to help you get it done quicker.

lib/components/viewers/nearby/nearby-filter.tsx
amy-corson-ibigroup: On further reflection it's weird we're relying on the trip form library for the CSS util, so i made a different one instead!

daniel-heppner-ibigroup: Good solution! I agree

lib/util/state-types.ts
```diff
@@ -23,8 +32,29 @@ export interface OtpState {
     start: number
   }
   transitIndex: any
-  // TODO: Add other OTP states
-  ui: any // TODO
+  // TODO: Finish ui typing
```
daniel-heppner-ibigroup, line 35: This is great! thanks for the progress on this. I wonder if we could ask an agent to do the rest of the work for us.

lib/util/state-types.ts
```diff
@@ -5,9 +5,18 @@ import {
...
+import { AppConfig, PopupTargetConfig } from './config-types'
+
+export type NearbyFilterKey =
+  | 'Stop'
+  | 'RentalVehicle'
+  | 'VehicleParking'
+  | 'BikeRentalStation'
+export type NearbyFilters = Record<NearbyFilterKey, boolean>
+
+export type LatLonObj = { lat: number; lon: number }
```
daniel-heppner-ibigroup, line 19: We have something like this imported from `@conveyal/lonlat` but I think just writing this here is fine.

daniel-heppner-ibigroup (approved):
Really great work here Amy! I love how it came together.

PR #1444, opened 2025-07-25
better handle blank itineraries

miles-grant-ibigroup:
An undefined itinerary previously crashed the UI. Now we let our error handler render

### DIFF
--- a/lib/components/map/itinerary-summary-overlay.tsx
+++ b/lib/components/map/itinerary-summary-overlay.tsx
@@ -83,6 +83,15 @@ const Card = styled.div`
 `
 
 function addItinLineString(itin: Itinerary): ItinWithGeometry {
+  if (!itin?.legs)
+    return {
+      ...itin,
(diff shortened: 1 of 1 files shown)


alec-georgoff (approved):

daniel-heppner-ibigroup (commented):
I'd like to know when this is happening before I approve. Are you getting nulls back in the array or something? Can you show me where this problem is showing up?

miles-grant-ibigroup:
> I'd like to know when this is happening before I approve. Are you getting nulls back in the array or something? Can you show me where this problem is showing up?

It's happening in cases where OTP is crashing. It returns nulls in the array

daniel-heppner-ibigroup (approved):
Glad that this is being resolved in OTP. Can you add a comment explaining what we talked about, though? I could see myself removing this at some point wondering why it exists since it technically shouldn't be necessary.

PR #1449, opened 2025-08-06
Select scooter mode when starting trip from scooter in nearby view

alec-georgoff:
**Description:**

This PR adds functionality to automatically select the rental vehicle mode in the trip planner when starting a trip from a selected rental vehicle in Nearby View. The logic respects the existing trip modes that are selected.

~Currently blocked until opentripplanner/otp-ui#871 is merged and version bumped.~

**PR Checklist:**


lib/components/map/default-map.tsx
```diff
@@ -319,6 +325,23 @@ class DefaultMap extends Component {
     const baseLayerUrls = baseLayersWithNames?.map((bl) => bl.url)
     const baseLayerNames = baseLayersWithNames?.map((bl) => bl.name)
 
+    const overlayTypes = overlays
```
alec-georgoff: One weird quirk...I realized after writing this that we don't actually specify that this logic should only fire when the `from` is clicked on a rental vehicle popup. And yet, when I click the `from` on other popups (like a stop), it doesn't enable the rental vehicle mode. I'm not really sure why...

lib/components/map/default-map.tsx
```diff
@@ -319,6 +325,23 @@ class DefaultMap extends Component {
...
+      ?.filter((overlay) => overlay?.type === 'otp2')?.[0]
```
miles-grant-ibigroup: Could we not use `find` instead?

lib/components/map/default-map.tsx
```diff
@@ -319,6 +325,23 @@ class DefaultMap extends Component {
...
+      ?.layers?.map((layer) => layer?.type)
+    const handleSetLocation = (location: MapLocationActionArg) => {
```
miles-grant-ibigroup: This is one approach, but could we not fix this by adjusting query params? What if we check for map state in the places where query params are set? Might that be a better separation of concerns than setting modes within the map component?

lib/components/map/default-map.tsx
```diff
@@ -486,6 +509,7 @@ const mapStateToProps = (state) => {
     bikeRentalStations: state.otp.overlay.bikeRental.stations,
     carRentalStations: state.otp.overlay.carRental.stations,
     config: state.otp.config,
+    currentQuery: state.otp.currentQuery,
```
miles-grant-ibigroup: Not a huge fan of importing query info into the map component...

miles-grant-ibigroup (changes requested):
A few architectural questions

lib/components/map/default-map.tsx
alec-georgoff: I like it

lib/components/map/default-map.tsx
```diff
@@ -319,6 +331,35 @@ class DefaultMap extends Component {
...
     const baseLayerNames = baseLayersWithNames?.map((bl) => bl.name)
 
+    const overlayTypes = overlays
+      ?.find((overlay) => overlay?.type === 'otp2')
+      ?.layers?.map((layer) => layer?.type)
+    const handleSetLocation = (location: MapLocationActionArg) => {
+      if (overlayTypes && overlayTypes.includes('rentalVehicles')) {
+        const requiredOptions:
+          | { modeButton: string; modeSetting?: string }
+          | undefined = findRequiredOptionsForTransportMode(
```
alec-georgoff: Requires opentripplanner/otp-ui#871 to be merged and version bumped

alec-georgoff: Done!

miles-grant-ibigroup (changes requested):
My previous comment still stands -- is it possible to avoid doing this work inside the map component? It doesn't feel appropriate

alec-georgoff:
> My previous comment still stands -- is it possible to avoid doing this work inside the map component? It doesn't feel appropriate

Following offline discussion, I've [moved the function outside the component](https://github.com/opentripplanner/otp-react-redux/pull/1449/commits/672843a92a87db491ae41afd14a440e6331c3b55) and [added a clarifying comment](https://github.com/opentripplanner/otp-react-redux/pull/1449/commits/2796903603fb1bfe53650ef1fe902504f60305ba)

lib/components/map/default-map.tsx
```diff
@@ -145,6 +158,48 @@ function getLayerName(overlay, config, intl) {
...
+  enabledModeButtons,
+  setQueryParam,
+  setLocation,
+  overlays
+): ((location: MapLocationActionArg) => void) => {
+  return (location: MapLocationActionArg) => {
+    const overlayTypes = overlays
+      ?.find((overlay) => overlay?.type === 'otp2')
+      ?.layers?.map((layer) => layer?.type)
+    if (overlayTypes && overlayTypes.includes('rentalVehicles')) {
```
miles-grant-ibigroup, line 178: What do we do if the layer name is not named `rentalVehicles`? This approach probably also doesn't work for selecting a scooter from the sidebar

miles-grant-ibigroup (changes requested):
In addition to the other comment, do we have support for bikes?

miles-grant-ibigroup (commented):
Post discussion:

If we are writing this big a diff, can we fix the underlying arch issues that are causing these problems? Shouldn't the URL be the source of truth?

Finally, can we simply by:
- ignoring current button state, and always appending to URL
- don't use complex methods, let the trip-form handle the URL changer
- don't use layer name to detect scooter, instead try to grab it from card properties. This means moving the fromtopicker override stuff into the card itself

miles-grant-ibigroup (changes requested):
What if we tried a different approach (in a differnet PR)

When SET_LOCATION redux action is fired, we also submit a card type. Then, inside of the action, we can set the mode accordingly. Should be much less code

PR #1450, opened 2025-08-06
Attach an agencyName to each monitored trip alert

binh-dam-ibigroup:
## Description

Requires https://github.com/ibi-group/otp-middleware/pull/334.

This PR tweaks the alerts display for a monitored trip, so that the applicable agency name is shown in the alerts link, similar to what is currently happening with the alerts in the itinerary results.

## PR Checklist

| Before | After |
|--------|-------|


lib/components/user/monitored-trip/trip-status.js
```diff
@@ -57,7 +57,14 @@ function MonitoredTripAlerts({ alerts }) {
           values={{ alerts: alerts.length }}
         />
       </h4>
-      <AlertsBody alerts={alerts} />
+      {/* For each alert. render an AlertsBody with the specific agency name */}
+      {alerts.map((alert) => (
+        <AlertsBody
+          agencyName={alert.agencyName}
+          alerts={[alert]}
```
alec-georgoff, line 64: Are there any side effects caused by going from one `AlertsBody` to multiple for each alert?

binh-dam-ibigroup: Maybe slight padding/offset differences from sequencing multiple `AlertsBody` instead of one. `AlertsBody` itself does not perform any side effects.

alec-georgoff (approved):

amy-corson-ibigroup (approved):

PR #1453, opened 2025-08-15
Fix pattern filtering

amy-corson-ibigroup:
**Description:**
- Fixes an issue where patterns were being renamed unnecessarily after all other duplicates were filtered out
- Route viewer will now filter patterns before selecting a pattern to pass to the pattern viewer. This will prevent pattern viewer opening with an unnamed pattern and "select a direction..." in the dropdown.

Here is the updated order of operations when filtering duplicate patterns. Changes are in italics.
- Duplicate patterns are defined as "patterns with the same name"
- We rename patterns that have last stops that are a) different than the headsign and b) _different than the last stop of the previously encountered duplicate headsign._
- The renamed headsign is named "headsign (last stop)"
- If duplicate headsigns have the same last stop, we filter out all but the one with the longest geometry. 

**PR Checklist:**

### DIFF
--- a/lib/components/viewers/route-row.tsx
+++ b/lib/components/viewers/route-row.tsx
@@ -5,6 +5,7 @@ import styled from 'styled-components'
 
 import { blue } from '../util/colors'
 import { ComponentContext } from '../../util/contexts'
+import { extractMainHeadsigns } from '../../util/pattern-viewer'
 import { getFormattedMode } from '../../util/i18n'
 import { getModeFromRoute } from '../../util/viewer'
 import { Icon } from '../util/styledIcon'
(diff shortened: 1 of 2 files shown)


lib/components/viewers/route-row.tsx
```diff
@@ -167,7 +168,8 @@ export class RouteRow extends PureComponent<Props> {
     const { id, longName, mode, patterns, shortName } = route
     const modeFromRoute = getModeFromRoute(route)
     const routePath = `/route/${id}`
-    const firstPattern = patterns && Object.values(patterns)?.[0]?.id
+    const firstPattern =
+      patterns && extractMainHeadsigns(patterns, shortName, () => '')?.[0]?.id
```
binh-dam-ibigroup, line 172: For a subsequent PR, I wonder if we should do that computation elsewhere?

binh-dam-ibigroup (approved):

alec-georgoff (approved):

PR #1455, opened 2025-08-18
Changes to toast promise

amy-corson-ibigroup:
**Description:**
- Fixes a bug wherein if a user is not logged in and attempts to save a place as home or work, a loading toast will pop up that cannot be dismissed.
- Adds a notification to announce when a saved trip has been deleted.
- Also removes the `userSuccess` number result from `rememberPlace` as it wasn't being used anywhere.

**PR Checklist:**


josh-willis-arcadis (approved):
great changes!

lib/components/util/toasts.tsx
```diff
@@ -62,6 +62,7 @@ export const toastPromise = async (
   id?: string,
   silentOnSuccess = false
 ): Promise<any> => {
+  if (!promise) return
```
daniel-heppner-ibigroup: Is `promise` optional? Can you mark it as optional in the argument list?

lib/components/map/connected-endpoints-overlay.tsx
```diff
@@ -14,7 +14,7 @@ import { toastMessageOnPlaceChanged, toastPromise } from '../util/toasts'
 
 type Props = ComponentProps<typeof EndpointsOverlay> & {
   forgetPlace: (place: string, intl: IntlShape) => void
-  rememberPlace: (arg: UserLocationAndType, intl: IntlShape) => number
+  rememberPlace: (arg: UserLocationAndType, intl: IntlShape) => Promise<void>
```
daniel-heppner-ibigroup: I think `rememberPlace` is just returning void,  not Promise<void>.

lib/components/util/toasts.tsx
```diff
@@ -62,6 +62,7 @@ export const toastPromise = async (
   id?: string,
   silentOnSuccess = false
 ): Promise<any> => {
```
daniel-heppner-ibigroup: Ok this is a bit complicated but it would be really cool if this could use a generic type instead of any so that the type that is passed in gets computed and used for the return type automatically.

daniel-heppner-ibigroup: an LLM could probably do this very easily

daniel-heppner-ibigroup (changes requested):
Take a look at my piggyback PR. If you merge that then this is good to go :) #1460

lib/components/util/toasts.tsx
amy-corson-ibigroup: Hmm I dont think this should be optional. This line is here because in the case of `rememberPlace`, if a user isn't logged in, the `promise` gets passed into this function as `undefined`

daniel-heppner-ibigroup: If it can be undefined then it should be optional! There should be a type error somewhere else if there's an undefined that gets passed into this function, since `undefined` is not compatible with `Promise`. If there's no type error is probably because there's an any type somewhere else that maybe should be fixed as well.

amy-corson-ibigroup: You're right! Reworked this in [1fbe9bd](https://github.com/opentripplanner/otp-react-redux/pull/1455/commits/1fbe9bd4a7bf8a8929c240e54939bcde4af94f88), i really don't think the promise should ever be undefined

lib/components/util/toasts.tsx
amy-corson-ibigroup: Gave it a shot in [d17a797](https://github.com/opentripplanner/otp-react-redux/pull/1455/commits/d17a797393f50418f2223944d09efc9fba2a68a5) lmk what you think!

amy-corson-ibigroup: Wait I'm dumb as rocks, you already did this in https://github.com/opentripplanner/otp-react-redux/pull/1460 whoops! Okay well good for me to practice i guess and yours looks better anyway

lib/components/util/toasts.tsx
```diff
@@ -55,13 +55,13 @@ export const toastSettings = (
   return settings
 }
 
-export const toastPromise = async (
-  promise: Promise<any>,
+export const toastPromise = async <T,>(
```
daniel-heppner-ibigroup: cool!

daniel-heppner-ibigroup (approved):
Looks great to me! Thanks for the improvements!

PR #1457, opened 2025-08-20
Dont filter overlays if there are no filters

amy-corson-ibigroup:
**Description:**
Fixes a bug where a lack of `cardType` on the layers (in instances where there's no nearby filters) was causing layers to be filtered out. Better to just avoid filtering at all in cases where there are no configured filters.

**PR Checklist:**


josh-willis-arcadis (approved):
beautiful!

alec-georgoff (approved):

PR #1460, opened 2025-08-26
remove unneded functions and add generic type

daniel-heppner-ibigroup:
piggyback PR for #1455 because I realized I could just do the changes faster than describing them in a PR :)


amy-corson-ibigroup (approved):
Oh way better! Why were these ever getting exported to begin with....

PR #1461, opened 2025-08-27
Fares V2 Support

miles-grant-ibigroup:
**Description:**
Uses new core utils functions to support new Fares V2 Table and display

**PR Checklist:**


miles-grant-ibigroup:
This is now ready for review @daniel-heppner-ibigroup

lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -339,6 +339,18 @@ class MetroItinerary extends NarrativeItinerary {
         />
       )
 
+    if (
+      // Display special message if there are multiple fare products, but no primary product configured
+      // @ts-expect-error TS seems to not understand the optional?
```
daniel-heppner-ibigroup: TS understands the optional, but you you can't use a `>` operator on undefined because it needs two numbers. So you need to provide a default value for `leg.fareProducts?.length` by using `?? 0` or something. Or do a null check some other way.

daniel-heppner-ibigroup (changes requested):
This looks good! I only have the one comment, of course on the ts-expect-error. Please fix!

lib/components/narrative/metro/metro-itinerary.tsx
miles-grant-ibigroup: Beautiful fix thanks! It's so annoying that TS won't respect javascript types...

lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -339,6 +339,17 @@ class MetroItinerary extends NarrativeItinerary {
         />
       )
 
+    if (
+      // Display special message if there are multiple fare products, but no primary product configured
+      itinerary.legs.some((leg: Leg) => leg.fareProducts?.length ?? 0 > 1) &&
+      !defaultFareType
+    ) {
+      console.log("Missing DefaultFareType! Can't display default fare")
```
daniel-heppner-ibigroup, line 347: I feel like this console log isn't necessary

lib/components/narrative/metro/metro-itinerary.tsx
```diff
@@ -339,6 +339,17 @@ class MetroItinerary extends NarrativeItinerary {
...
+      fareInfo = (
+        <FormattedMessage id="common.itineraryDescriptions.noDefaultFareTypeConfigured" />
+      )
+    }
+
```
daniel-heppner-ibigroup, line 352: would it maybe be better to put this in a function that returns the formatttedmessage we want?

daniel-heppner-ibigroup (approved):
works great!

lib/components/narrative/metro/metro-itinerary.tsx
miles-grant-ibigroup: If someone misconfigured their config, how should they be notified?

lib/components/narrative/metro/metro-itinerary.tsx
miles-grant-ibigroup: I agree. This is a holdover from the old architecture of this file. Given that we want to build a new itinerary view anyway, I'll leave it as a todo for now

i18n/fr.yml
```diff
@@ -133,6 +133,7 @@ common:
     "yes": Oui
   itineraryDescriptions:
     fareUnknown: Tarif inconnu
+    noDefaultFareTypeConfigured: Multiple fares apply
```
alec-georgoff, line 136: Just wanna double check that this English is okay for now 😅

miles-grant-ibigroup: Yes! This is what we want. The key is named like this so developers know what's actually going on

miles-grant-ibigroup: Oh shoot I just realized this is in the fr file. Yes I think it is ok we can fix it in weblate!

alec-georgoff (approved):

miles-grant-ibigroup:
Can be merged once https://github.com/opentripplanner/otp-ui/pull/882 is merged and packages here are updated

PR #1468, opened 2025-09-05
Include route.shortname in the transitive route label

amy-corson-ibigroup:
**Description:**
Make sure all available short names are passed to transitive overlay. 

A follow up OTP-UI PR is needed because the comment on this line (that a blank string will render no label) is not the case with the new transitive overlay bubbles.

**PR Checklist:**

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -167,7 +167,7 @@ const components = {
    * @returns A string with the custom label to display for the given leg, or null to render no label.
    */
   getTransitiveRouteLabel: (itineraryLeg) => {
-    return itineraryLeg.routeShortName // null or undefined or empty string will tell transitive-js to not render a route label.
+    return itineraryLeg.routeShortName || itineraryLeg.route.shortName // null or undefined or empty string will tell transitive-js to not render a route label.
   },
 
   ItineraryBody: DefaultItinerary,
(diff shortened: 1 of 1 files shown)


lib/app.js
```diff
@@ -167,7 +167,7 @@ const components = {
    * @returns A string with the custom label to display for the given leg, or null to render no label.
    */
   getTransitiveRouteLabel: (itineraryLeg) => {
-    return itineraryLeg.routeShortName // null or undefined or empty string will tell transitive-js to not render a route label.
+    return itineraryLeg.routeShortName || itineraryLeg.route.shortName // null or undefined or empty string will tell transitive-js to not render a route label.
```
miles-grant-ibigroup: Do we know `route` will exist? Will it always have a shortname?

miles-grant-ibigroup (changes requested):

lib/app.js
amy-corson-ibigroup: I added a conditional to `route`. If there's no short name, it'll return `undefined` which in theory transitive should handle gracefully ( it doesn't but that's out of the scope of this PR)

miles-grant-ibigroup: What if the route object is missing?

amy-corson-ibigroup: Then it'll return undefined.

miles-grant-ibigroup (approved):
I'm nervous but I trust you!

alec-georgoff (approved):
LGTM!!

PR #1471, opened 2025-09-11
Set User's Preferred Language in Auth0 Metadata

amy-corson-ibigroup:
**Description:**
Auth0 can translate account emails into different languages, but only if it has the user's preferred language set in the account metadata. This new method fetches a token from the management API, and uses that token to set the user's account metadata to include the preferred language. This is fired whenever a user creates a new account or if the language changes.

THINGS TO CONSIDER:
- Should this new fetch method be combined with `secureFetch`?
- The account token fetched from the management API expires after 24 hours. Should this be cached in the user info in the middleware? It seems unlikely a user will be changing their language multiple times a day.

**PR Checklist:**

### DIFF
--- a/example/example-config.yml
+++ b/example/example-config.yml
@@ -83,6 +83,10 @@ persistence:
 #  auth0:
 #    domain: your-auth0-domain
 #    clientId: your-auth0-client-id
+###  Management API credentials are required to update user metadata
+#    managementApi:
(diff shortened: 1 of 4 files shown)


amy-corson-ibigroup:
Closing in favor of backend approach https://github.com/ibi-group/otp-middleware/pull/346

PR #1473, opened 2025-09-26
add optional trip-in-past error

miles-grant-ibigroup:
This PR adds a new setting to allow a trip-in-past error. This error alerts the user that one or more itineraries is in the past!
Screenshot 2025-09-26 at 2 43 21 PM

### DIFF
--- a/lib/actions/apiV2.js
+++ b/lib/actions/apiV2.js
@@ -1297,7 +1297,7 @@ export function routingQuery(searchId = null, updateSearchInReducer) {
                     })
                   }
                 }
-                if (displayTripInPastError) {
+                if (displayTripInPastError !== false) {
                   if (
                     response.data.plan.itineraries.some(
                       (itin) => itin.endTime < Date.now()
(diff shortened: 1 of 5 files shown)


josh-willis-arcadis (approved):
great addition!

lib/actions/apiV2.js
```diff
@@ -1294,6 +1297,17 @@ export function routingQuery(searchId = null, updateSearchInReducer) {
                     })
                   }
                 }
+                if (displayTripInPastError !== false) {
```
binh-dam-ibigroup, line 1300: Remove false comparison (I think we can just evaluate `displayTripInPastError`  for truthy like the other settings above)

binh-dam-ibigroup (approved):

lib/actions/apiV2.js
miles-grant-ibigroup: This allows it to be on by default!

PR #1479, opened 2025-10-06
Fix open parenthesis on user saved trips

amy-corson-ibigroup:
**Description:**
Fixes a bug where user saved trips, which are formatted `nickname (placename)` were not showing the close parenthesis in the previous trips pane. 

Open to another approach or laying out the user places in a different way. Are there cases where there could be a "(" character in a name that does not require a close parenthesis?

**PR Checklist:**

| Before | After |
|--------|-------|
|image|image|


binh-dam-ibigroup (approved):
Nice fix

lib/components/form/user-settings-i18n.js
```diff
@@ -16,7 +16,12 @@ function stripAllButNameOfAddress(string) {
 
   // If string contains any letters, split the string!
   if (/[a-zA-Z]/g.test(string)) {
-    return string.split(',')[0]
+    let newString = string.split(',')[0]
+    // If there is an open parenthesis without a close (user saved place), add one.
+    if (newString.includes('(') && !newString.includes(')')) {
```
alec-georgoff: What about this string? `Coffee (Starbucks (Market & 20th)`? I think you may need to actually count the open and close parentheses  🙃 I knew [this leetcode problem](https://leetcode.com/problems/valid-parentheses/) would come in handy one day!!

amy-corson-ibigroup:
Revised the approach! Much easier than I expected it to be, although slightly different than what we discussed on our call. Reassigning @binh-dam-ibigroup to re-review

lib/components/form/user-settings-i18n.js
```diff
@@ -36,6 +36,12 @@ function findLocationType(
     ? getFormattedPlaces(match.type, intl)
     : null
 }
+function findCustomLocationShortName(location, locations = []) {
+  const match = locations.find((l) => matchLatLon(l, location))
+  return match && (match.type === 'custom' || match.type === 'dining')
```
binh-dam-ibigroup: Nit: Try excluding "Home" and "Work" instead?

binh-dam-ibigroup (approved):

alec-georgoff (approved):

PR #1481, opened 2025-10-09
Fix user update before consents to terms

amy-corson-ibigroup:
**Description:**
Fixes a bug where if a user is on the verify email pane and they try to change the locale of the page, the page would reload and reset to the previous language. The issue was coming from a failed user update because the `isNewUser` check was returning true (resulting in a PUT request instead of PATCH) even though the user already exists in the database at this point.

**PR Checklist:**


lib/util/user.js
```diff
@@ -22,6 +22,10 @@ export function isNewUser(loggedInUser) {
   return !loggedInUser.hasConsentedToTerms
 }
 
+export function userExistsInDatabase(loggedInUser) {
```
binh-dam-ibigroup, line 25: Did you mean `userDoesNotExistInDatabase` or `userNotInDatabase`?

lib/actions/user.js
```diff
@@ -391,7 +391,7 @@ export function createOrUpdateUser(userData, intl) {
     )
 
     // Determine URL and method to use.
-    const isCreatingUser = isNewUser(loggedInUser)
+    const isCreatingUser = userExistsInDatabase(loggedInUser)
```
binh-dam-ibigroup: I thought the user entry in Mongo was meant to be created when they click "Next" after accepting the terms of use. I think you uncovered a bug that results in the user being created when they set the language and before they agree to the terms. Congrats!

To fix the bug:
In lib/actions/ui.js near line 471: replace
```
      if (loggedInUser) {
```
with
```
      if (!userNotInDatabase(loggedInUser)) {
```
(with correct import and corrected name/meaning.

lib/util/user.js
```diff
@@ -22,6 +22,10 @@ export function isNewUser(loggedInUser) {
...
+  return !loggedInUser.id
```
binh-dam-ibigroup: `loggedInUser` might be null (after reloading the 'verify' page), so use the `?.` optional chaining syntax.

binh-dam-ibigroup (changes requested):
Rename the method, it was confusing to read at first, and also please fix the related bug that you uncovered too.

lib/util/user.js
amy-corson-ibigroup: Good catch, thank you. I left the name the same and changed the body of the function

binh-dam-ibigroup (approved):

josh-willis-arcadis (approved):
great catch! LGTM!

PR #1482, opened 2025-10-09
Add stop code to place name in nearby view from/to picker

daniel-heppner-ibigroup:
**Description:**
This PR adds the stop code to the nearby view's from/to picker's place name. Clicking on from/to on a stop in the nearby view will now include this information in the location field.
image

### DIFF
--- a/lib/components/viewers/nearby/nearby-view.tsx
+++ b/lib/components/viewers/nearby/nearby-view.tsx
@@ -64,6 +64,7 @@ type Props = {
   defaultLatLon: LatLonObj | null
   displayedCoords?: LatLonObj
   entityId?: string
+  feeds: any[]
   fetchNearby: (
     latLon: LatLonObj,
     radius?: number,
@@ -90,8 +91,15 @@ type Props = {
   zoomToPlace: ZoomToPlaceHandler
 }
 
-const getNearbyItem = (place: any) => {
-  const fromTo = <FromToPicker place={place} />
+const getNearbyItem = (place: any, feeds?: any[]) => {
+  const placeForFromTo = { ...place }
+  if (place.__typename === 'Stop' && feeds) {
+    const feedId = place.gtfsId.split(':')[0]
+    const feed = feeds.find((f) => f.feedId === feedId)
+    const feedName = feed?.publisher?.name
+    placeForFromTo.name = `${place.name} (${feedName} ${place.code})`
+  }
+  const fromTo = <FromToPicker place={placeForFromTo} />
 
   switch (place.__typename) {
     case 'RentalVehicle':
@@ -147,6 +155,7 @@ function NearbyView({
   defaultLatLon,
   displayedCoords,
   entityId,
+  feeds,
   fetchNearby,
   geocoderConfig,
   getCurrentPosition,
@@ -345,7 +354,10 @@ function NearbyView({
           /* eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex */
           tabIndex={0}
         >
-          {getNearbyItem({ ...n.place, distance: n.distance, nearbyRoutes })}
+          {getNearbyItem(
+            { ...n.place, distance: n.distance, nearbyRoutes },
+            feeds
+          )}
         </div>
       </li>
     ))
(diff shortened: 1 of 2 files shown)


lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -64,13 +64,15 @@ type Props = {
...
+  feeds: any[]
   fetchNearby: (
     latLon: LatLonObj,
     radius?: number,
     currentServiceWeek?: ServiceWeek
   ) => void
   geocoderConfig: GeocoderConfig
-  getCurrentPosition: any // TODO
+  getCurrentPosition: any
+  // TODO
```
miles-grant-ibigroup: Why this newline?

package.json
```diff
@@ -5,6 +5,7 @@
   "scripts": {
     "a11y-test": "jest a11y/ --runInBand --force-exit",
     "build": "vite build",
+    "bundle-stats": "yarn build && node -e \"require('open')('stats.html')\"",
```
miles-grant-ibigroup, line 8: Wrong PR?

lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -91,8 +93,15 @@ type Props = {
...
 
-const getNearbyItem = (place: any) => {
-  const fromTo = <FromToPicker place={place} />
+const getNearbyItem = (place: any, feeds?: any[]) => {
+  const placeForFromTo = { ...place }
+  if (place.__typename === 'Stop' && feeds) {
+    const feedId = place.gtfsId.split(':')[0]
+    const feed = feeds.find((f) => f.feedId === feedId)
+    const feedName = feed?.publisher?.name
+    placeForFromTo.name = `${place.name} (${feedName} ${place.code})`
```
miles-grant-ibigroup: What if there is no `place.code`

miles-grant-ibigroup (changes requested):
Looks good, just some small issues. Also, what happens if no stop code is defined? Isn't there already a lot of code for handling what to do with stops in all these cases?

daniel-heppner-ibigroup:
I accidentally broke off from binh's branch but I figured it was getting merged so it wasn't a big deal. I rebased now and it looks good.

daniel-heppner-ibigroup:
So, we do have this PR which sort of has the same logic. However I'm not really sure I see a great way to reuse it here unless we extract a very simple function which I think would be overkill... I can do it if you think that would be better overall.

miles-grant-ibigroup (approved):
Works well! and pretty clean too, although some comments on that ternary might help

lib/components/viewers/nearby/nearby-view.tsx
```diff
@@ -90,8 +91,18 @@ type Props = {
...
-  const fromTo = <FromToPicker place={place} />
+const getNearbyItem = (place: any, feeds?: any[]) => {
+  const placeForFromTo = { ...place }
+  if (place.__typename === 'Stop' && feeds) {
+    const feedId = place.gtfsId.split(':')[0]
+    const feed = feeds.find((f) => f.feedId === feedId)
+    const feedName = feed?.publisher?.name
+    placeForFromTo.name =
+      feedName || place.code
+        ? `${place.name} (${[feedName, place.code].filter(Boolean).join(' ')})`
```
alec-georgoff: Do we want to still display the code if we don't have a `feedName`? I feel like

> 4th & Pine (700)

might be a little odd. Thoughts?

miles-grant-ibigroup: Good catch! I agree. Maybe we handle this inside react-intl and then we can set a default

daniel-heppner-ibigroup:
Here's the PR I meant to mention earlier: https://github.com/opentripplanner/otp-ui/pull/857/files#diff-0779912399fb6ae5700bfb96f7ef55a1ec98581cad1b6281422ba299ed59c62c 

There is some duplicated logic here. Pelias stitch is also doing something similar I guess, but with very different input variables so it wouldn't be possible to reuse the logic there.

I'll go ahead and change this to just require both code and feed name.

alec-georgoff (approved):

PR #1483, opened 2025-10-13
Add Aria-Current to the route viewer

amy-corson-ibigroup:
**Description:**
The Invisible "currently selected" text is not working for all AT, so use `aria-current` [(docs)](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-current) to make sure AT users know the link represents the current page. This might actually be a more appropriate solution to this issue, but both approaches are probably okay. Open to removing the "currently selected" text if people think it would be better.

**PR Checklist:**


lib/components/viewers/route-row.tsx
```diff
@@ -224,6 +224,7 @@ export class RouteRow extends PureComponent<Props> {
     return (
       <StyledRouteRow ref={this.activeRef}>
         <RouteRowLink
+          aria-current={isActive ? 'page' : false}
```
binh-dam-ibigroup: Did you mean to say `true` instead of `page`? (`page` would be used for navigation breadcrumbs from what is in MDN docs)

amy-corson-ibigroup: I think you're right, I thought `page` was more accurate but `true` is probably safer if the intended use for `page` is the breadcrumb nav. Changed in [577e569](https://github.com/opentripplanner/otp-react-redux/pull/1483/commits/577e5691d79f9e5c48a87ddd016225416f6a312c)

binh-dam-ibigroup (approved):

miles-grant-ibigroup (approved):
I love it! I always prefer aria tags to adding text when possible

PR #1485, opened 2025-10-17
Fix order of operations when deleting place

amy-corson-ibigroup:
**Description:**
Fixes a bug where if a user presses "cancel" after being asked to confirm they'd like to delete their saved place, the page goes back and a loading toast gets stuck on the screen.

Also fixes an issue where the user might cancel and still gets navigated back to the saved place list.

**PR Checklist:**


josh-willis-arcadis (approved):
great catch!

alec-georgoff (approved):

PR #1488, opened 2025-10-21
Create new attribution below print map

amy-corson-ibigroup:
**Description:**
The map attribution text was often interfering with the trip overlay in the printed map. This takes the attribution and inserts it below the map, so that it wont interfere with print.

I didn't add any sort of additional AT context, but maybe that would be a good idea? Open to feedback

**PR Checklist:**

| Before | After |
|--------|-------|
|image|image
|image|image|


lib/components/app/trip-preview-layout-base.tsx
```diff
@@ -112,6 +127,11 @@ class TripPreviewLayoutBase extends Component<Props, State> {
 
         {/* The map, if visible */}
         {this.state.mapVisible && mapElement}
+        <CustomAttribution>
+          {attributionHTML && (
+            <div dangerouslySetInnerHTML={{ __html: attributionHTML }} />
```
miles-grant-ibigroup: this is such a nitpick but is it not cleaner to wrap the entire `<CustomAttribution>` in the optional so we don't render an empty tag if there is no attribution!

miles-grant-ibigroup (approved):
Beautiful change and beautifully executed!

lib/components/app/trip-preview-layout-base.tsx
amy-corson-ibigroup: I thought about this, but since I had to move the `margin-bottom` to this element, i figured we'd keep in in both cases so the margin is consistent!

miles-grant-ibigroup: Eep... I'm not sure how I feel about having an empty element that just has some padding! Can we reconsider this margin placement? I'll leave it up to the second reviewer to either tell me my concerns are right or if they can be ignored

alec-georgoff: I think I'm gonna agree with Miles here; coming from some other recent work where margins are all over the place, I think this is a good chance to avoid that kind of thing.

My recommendation would be to just stick the `30px` margin as a `margin-top` on the `PrintableItinerary` component, but open to other suggestions.

alec-georgoff (commented):
Soooo nice!

lib/components/app/trip-preview-layout-base.tsx
amy-corson-ibigroup: My bad, i think i wrote this at 6am and I forgot `margin-top` existed!! Thanks everyone

alec-georgoff (approved):

PR #1489, opened 2025-10-22
always reverse geocode in FromToPicker

josh-willis-arcadis:
**Description:**
Always reverse geocode when using the FromToPicker. This provides the location field with a more user friendly string when planning a trip to or from a rental vehicle in the nearby view. 

**PR Checklist:**

Before:
image
After:
image


miles-grant-ibigroup (approved):
I feel like we've made this change before... But that might just be because there are so many different ways we instantiate the fromto picker

amy-corson-ibigroup (approved):
Looks great! Thanks for the changes

PR #1490, opened 2025-10-23
Force geolocate control to update UI

amy-corson-ibigroup:
**Description:**
Fixes a bug where even if a user's location was clearly blocked on mobile, the geolocate button would not reflect that. Now if the app has an existing location error, we will subtly trigger the geolocate button so it updates its UI. 

**PR Checklist:**


alec-georgoff (approved):

lib/components/map/default-map.tsx
```diff
@@ -506,10 +521,11 @@ const mapStateToProps = (state) => {
   const viewedRoute = state.otp?.ui?.viewedRoute?.routeId
   const activeNearbyFilters = state.otp?.ui?.nearbyView?.filters
   const nearbyFilters = state.otp.config?.nearbyView?.filters
-  const stops = state.otp.transitIndex.stops
```
miles-grant-ibigroup, line 509: No more stops?

lib/components/map/default-map.tsx
```diff
@@ -371,7 +380,12 @@ class DefaultMap extends Component {
           }
           baseLayerNames={baseLayerNames}
           center={[lat, lon]}
-          mapLibreProps={{ reuseMaps: true }}
+          mapLibreProps={{
+            onLoad: () => {
+              return this.setState({ mapLoad: true })
```
miles-grant-ibigroup, line 386: Would it be possible to add a comment here that this is about the hack above?

miles-grant-ibigroup (changes requested):
Clever solution! Just a small quick quesiton

lib/components/map/default-map.tsx
amy-corson-ibigroup: Added in [d06f1bc](https://github.com/opentripplanner/otp-react-redux/pull/1490/commits/d06f1bcae7838100063cc3b506e8fbc5181f83c8)

lib/components/map/default-map.tsx
amy-corson-ibigroup: They weren't doing anything if you can believe it

miles-grant-ibigroup (approved):
Wonderful!

PR #1496, opened 2025-11-05
Keep Russian from breaking mobile layout

amy-corson-ibigroup:
**Description:**
Tweak some styling in the phone form so that Russian does not break the layout

**PR Checklist:**

| Before | After |
|--------|-------|
|image|image|

### DIFF
--- a/lib/components/user/styled.ts
+++ b/lib/components/user/styled.ts
@@ -115,12 +115,16 @@ export const ControlStrip = styled.span`
   & > * {
     margin-right: 4px;
   }
+
+  button {
+    text-wrap: wrap;
+    margin-top: 5px;
+  }
 `
 /** Styles for phone editing fields */
 export const phoneFieldStyle = css`
   display: inline-block;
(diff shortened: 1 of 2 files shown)


lib/components/user/styled.ts
```diff
@@ -115,12 +115,16 @@ export const ControlStrip = styled.span`
...
+  button {
+    text-wrap: wrap;
+    margin-top: 5px;
+  }
 `
 /** Styles for phone editing fields */
 export const phoneFieldStyle = css`
   display: inline-block;
   vertical-align: middle;
-  width: 14em;
```
binh-dam-ibigroup, line 123: For another PR: This input field contains just digits and symbols, so I am wondering whether to replace the placeholder text with a template phone number. That way, we can keep the original width of this field to suggest shorter inputs.

binh-dam-ibigroup (approved):

josh-willis-arcadis (approved):
LGTM

PR #1498, opened 2025-11-13
Clear viewed route in nearby view

amy-corson-ibigroup:
**Description:**
Cleans up the viewed route in the nearby view so that 
- The route overlay is no longer on the map when a user is looking at nearby stops and entities
- Accessing the nearby view via pattern viewer and then clicking the "back" button in the browser will properly re-set the viewed route and zoom the map correctly (previously it would stay in the nearby view map state)

**PR Checklist:**


daniel-heppner-ibigroup (commented):
Let's see if we can find an imperative way to clear the viewed route instead of using a useEffect. Let me know if you want to pair program it.

daniel-heppner-ibigroup (approved):
Thanks! I like this a lot more than the useeffect

josh-willis-arcadis (approved):
LGTM!

PR #1499, opened 2025-11-14
Handle no pattern headsign

binh-dam-ibigroup:
**Description**

This PR fixes the display of pattern headsigns for routes with a single pattern and no headsign. The headsign is the name of the last stop in the pattern, and the dropdown selector is not used. The PR also improves display of patterns with leading "To " text.

**PR Checklist:**

| Before | After |
|--------|-------|
| before screenshot | after screenshot  |


lib/util/viewer.js
```diff
@@ -68,47 +66,43 @@ export function routeIsValid(route, routeId) {
...
+  const { desc, headsign, stops } = pattern
+  const descNotRouteShortName = routeShortName && desc !== `${routeShortName}`
 
-  // If that regex didn't work, try the old regex
-  if (isBlank(headsign)) {
-    const matches = pattern?.desc?.match(/ to ([^(from)]+) \(.+\)/)
-    if (matches) {
-      headsign = matches[1]
+  if (!isBlank(headsign)) {
+    if (headsign.startsWith('To ') || headsign.startsWith('to ')) {
```
amy-corson-ibigroup: Would it be more or less efficient to do `headsign.toLowerCase.startsWith`?

lib/util/viewer.js
```diff
@@ -68,47 +66,43 @@ export function routeIsValid(route, routeId) {
...
-  if (isBlank(headsign) && routeShortName) {
-    const matches = pattern?.desc?.match(/^[^(]*$/)
-    if (matches) {
-      headsign = matches[0].replace(`${routeShortName} `, '')
+    return headsign
+  } else {
+    // In case the headsign is blank, extract headsign from the pattern 'desc' attribute
+    // by attempting the logic below in order:
+    const attempts = [
+      // (format: '49 to <Destination> (<destid>)[ from <Origin> (<originid)]').
```
amy-corson-ibigroup: This is a nitpick, but can we remove the parenthesis around this comment? There's a lot of brackets and parenthesis happening in this line already and removing one set might make it easier to parse

lib/util/viewer.js
```diff
@@ -68,47 +66,43 @@ export function routeIsValid(route, routeId) {
...
+      (d) => d?.match(/(?: to )(.*?)(?: \()/)?.[1],
+      // If that regex didn't work, try the old regex
+      (d) => d?.match(/ to ([^(from)]+) \(.+\)/)?.[1],
```
amy-corson-ibigroup, line 91: Might be reading this regex wrong, but are there actual cases where this one would work and the above one would not? All the cases I'm trying, the regex on 83 does a better job

amy-corson-ibigroup (approved):
Really nice improvement, thank you for all the different considerations on this! Just a few nitpicks/questions but overall looks good.

lib/util/viewer.js
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/1499/commits/52b7a44f9b444310749d8d819b2dc7a6322c6921

lib/util/viewer.js
binh-dam-ibigroup: Yes, clarified in https://github.com/opentripplanner/otp-react-redux/pull/1499/commits/52b7a44f9b444310749d8d819b2dc7a6322c6921

lib/util/viewer.js
binh-dam-ibigroup: I only copied over what was there before. I'm not sure which cases this one was supposed to cover.

alec-georgoff (approved):
Would it be worthwhile to also check for `"toward [destination]"` or `"towards [destination]"`? Not sure if any of those actually exist but it popped into my head.

binh-dam-ibigroup:
> Would it be worthwhile to also check for `"toward [destination]"` or `"towards [destination]"`? Not sure if any of those actually exist but it popped into my head.

@alec-georgoff I added the m in https://github.com/opentripplanner/otp-react-redux/pull/1499/commits/0deb98a47707c1b0d6a5a7b2ed2e9f59ddb1d6bf, let me know how that feels!

alec-georgoff:
> > Would it be worthwhile to also check for `"toward [destination]"` or `"towards [destination]"`? Not sure if any of those actually exist but it popped into my head.
> 
> @alec-georgoff I added the m in [0deb98a](https://github.com/opentripplanner/otp-react-redux/commit/0deb98a47707c1b0d6a5a7b2ed2e9f59ddb1d6bf), let me know how that feels!

Awesome, thanks!

PR #1502, opened 2025-12-03
Move map expansion button to itinerary header

alec-georgoff:
**Description:**

This update moves the "Expand map"/"Show results" button from the map to the itinerary header component, preventing overlap issues with the map attribution on smaller screen sizes in mobile view.

**PR Checklist:**


lib/components/narrative/narrative-itineraries-header.tsx
```diff
@@ -155,52 +159,75 @@ export default function NarrativeItinerariesHeader({
             // because it falls under the "Plan your trip" <h1> header.
             <InvisibleA11yLabel as="h2">{itinerariesFound}</InvisibleA11yLabel>
           )}
-          <ItinerariesHeaderContainer showHeaderText={showHeaderText}>
+          <ItinerariesHeaderContainer
+            showHeaderText={showHeaderText}
+            style={{
+              alignItems: 'flex-end',
```
alec-georgoff: I was going to abstract these into CSS or Styled Components, but it looks like we already use inline styling elsewhere in this file so I stuck with that.

amy-corson-ibigroup:
Looking really good! I haven't spent a ton of time looking at the code yet, thought it might make sense to iron out the below for this first round:
1. Something we discussed was replacing the button text with a chevron, which most mobile users will recognize as a way to expand/collapse the panel. Here's some examples of how this could look 
image
image


2. The container on the button has a scroll and so you can easily accidentally get the page into a weird state: 
image

3. Could be this is outside the scope of this PR, but I'm wondering if we may want to more gracefully animate the map expanding and collapsing? Right now the panel size is animated but the map is not, and there's a huge white gap mid-animation when the map resizes. Either we could animate the map height or if that doesn't work we could delay the map resizing. 

Thank you so so much for the work on this!

alec-georgoff:
@amy-corson-ibigroup I prevented scrolling in the collapsed header area and updated to use the chevron icon! Currently looking into the animation to see if I can clean it up a bit but wanted to get the other fixes over to you for review.

lib/components/mobile/batch-results-screen.js
```diff
@@ -39,19 +34,12 @@ const StyledMobileContainer = styled(MobileContainer)`
...
-`
-
-const NARRATIVE_SPLIT_TOP_PERCENT = 45
+const NARRATIVE_SPLIT_TOP_PERCENT = 41
 
 // Styles for the results map also include prop-independent styles copied from mobile.css.
 const ResultsMap = styled.div`
   bottom: ${(props) =>
-    props.expanded ? '0' : `${100 - NARRATIVE_SPLIT_TOP_PERCENT}%`};
+    props.expanded ? '7%' : `${100 - NARRATIVE_SPLIT_TOP_PERCENT}%`};
```
amy-corson-ibigroup: I think the 7% here should be represented the same way as the `NARRATIVE_SPLIT_TOP_PERCENT`. Can we set a new `TOGGLE_MAP_BUTTON_HEIGHT` variable and use that here?

lib/components/mobile/batch-results-screen.js
```diff
@@ -65,7 +53,7 @@ const narrativeCss = css`
       ? props.expanded
         ? '100px'
         : `${NARRATIVE_SPLIT_TOP_PERCENT}%`
-      : '100%'};
+      : '93%'};
```
amy-corson-ibigroup: Same thing here, this should be `100 - TOGGLE_MAP_BUTTON_HEIGHT`

lib/components/narrative/narrative-itineraries-header.tsx
```diff
@@ -1,5 +1,7 @@
 /* eslint-disable complexity */
 import { ArrowLeft } from '@styled-icons/fa-solid/ArrowLeft'
+import { ChevronDown } from '@styled-icons/fa-solid/ChevronDown'
```
amy-corson-ibigroup: Probably easier to just import one icon and `rotate`

lib/components/narrative/narrative-itineraries-header.tsx
amy-corson-ibigroup: You can add these to `ItinerariesHeaderContainer`, since it's a styled-component! Better to keep all the styles in the same place.

lib/components/narrative/narrative-itineraries-header.tsx
amy-corson-ibigroup: Well, actually I guess we import both in `app-menu-items` so I'll leave that to the second reviewer

lib/components/narrative/narrative-itineraries-header.tsx
```diff
@@ -155,52 +161,82 @@ export default function NarrativeItinerariesHeader({
...
-              className="clear-button-formatting"
-              onClick={onSortDirChange}
-              title={intl.formatMessage({
-                id: 'components.NarrativeItinerariesHeader.changeSortDir'
-              })}
-            >
-              <StyledIconWrapper
-                className={`${customBatchUiBackground && 'base-color-bg'}`}
+            {onClickExpansionButton && (
+              <button
```
amy-corson-ibigroup: This button needs to have a `title` and an `aria-label` for non-sighted users. You can repurpose the i18n strings for "expand map" and "show results"

amy-corson-ibigroup (changes requested):
1. Can we move the zoom buttons to the left hand side? We're still getting some button overlap on smaller screen sizes: !image
2. Something about the button isn't quite right to me, especially when it's collapsed... wondering how we feel about a) making the chevron icon a little wider, and/or b) when the itinerary results are collapsed, making the button the base branding color, so it matches the mobile header? idk, just a couple of ideas, I'm not married to either of them but curious what you and a second reviewer might think!

amy-corson-ibigroup:
> 1. Can we move the zoom buttons to the left hand side? We're still getting some button overlap on smaller screen sizes

Oh, I'm seeing now we have it changed for one environment in the configuration, but I think this needs to be the new default!

lib/components/narrative/narrative-itineraries-header.tsx
alec-georgoff: Yeah, I found these via their usage in another component and figured I'd keep the same pattern here. Happy to go with a rotate strategy if you or others feel strongly, though!

lib/components/mobile/batch-results-screen.js
alec-georgoff: Good suggestion, done!

lib/components/narrative/narrative-itineraries-header.tsx
alec-georgoff: Ah thanks for looking out for that; I assume the same value is okay for `title` and `aria-label`? That's what I implemented. If not, let me know what the difference is and I can fix

alec-georgoff:
> 1. Can we move the zoom buttons to the left hand side? We're still getting some button overlap on smaller screen sizes: !image
> 2. Something about the button isn't quite right to me, especially when it's collapsed... wondering how we feel about a) making the chevron icon a little wider, and/or b) when the itinerary results are collapsed, making the button the base branding color, so it matches the mobile header? idk, just a couple of ideas, I'm not married to either of them but curious what you and a second reviewer might think!

How about this? Made it wider and used the branding color as suggested:

image

alec-georgoff:
Latest update uses the unicode chevron and tweaks the styling a bit to make it less intrusive:

Screenshot 2025-12-16 at 2 31 55 PM

Screenshot 2025-12-16 at 2 32 21 PM

lib/components/narrative/narrative-itineraries-header.tsx
amy-corson-ibigroup: Yes, the same text is good for both! But I would create two variables `showResultsText` and `expandMapText` so you're not duplicating code.

lib/components/narrative/narrative-itineraries-header.tsx
```diff
@@ -161,46 +173,86 @@ export default function NarrativeItinerariesHeader({
...
+                        id: 'components.BatchResultsScreen.expandMap'
+                      })
+                }
+              >
+                <StyledIconWrapper>
+                  {mapExpanded ? (
+                    <UnicodeChevron direction="up">&#10095;</UnicodeChevron>
+                  ) : (
+                    <UnicodeChevron direction="down">&#10095;</UnicodeChevron>
+                  )}
```
amy-corson-ibigroup: Change the `direction` prop to a boolean like `expanded` and then just pass the `mapExpanded` prop, no conditional needed here.

```
<UnicodeChevron expanded={mapExpanded}>&#10095;</UnicodeChevron>
```

lib/components/narrative/narrative-itineraries-header.tsx
```diff
@@ -31,11 +34,18 @@ const SortResultsDropdown = styled(Dropdown)`
   }
 `
 
+const UnicodeChevron = styled.div<{ direction: 'up' | 'down' }>`
+  transform: rotate(${(props) => (props.direction === 'up' ? '270' : '90')}deg)
+    scaleY(2) scaleX(1.5);
+`
```
amy-corson-ibigroup: You're creating a boolean with strings here, I would just change the prop to something true/false!

amy-corson-ibigroup (changes requested):
I like the unicode chevron! thank you for the changes and for your patience! Sorry for taking a few days to review:
1. if you look at this on taller devices, the duration button is peeking out. I think we'll need a solution where the duration button is conditionally rendered? 
image

2. The sort results button is blue now? 
image
 If we can fix it while we're touching this code that would be awesome but feel free to ignore it and we can fix later, I don't want to keep changing the scope of this PR.

lib/components/narrative/narrative-itineraries-header.tsx
alec-georgoff: Done, good idea!

alec-georgoff:
> I like the unicode chevron! thank you for the changes and for your patience! Sorry for taking a few days to review:
> 
> 1. if you look at this on taller devices, the duration button is peeking out. I think we'll need a solution where the duration button is conditionally rendered?
> 
> image
> 2. The sort results button is blue now?
> 
> image If we can fix it while we're touching this code that would be awesome but feel free to ignore it and we can fix later, I don't want to keep changing the scope of this PR.

Good catch on the vertical spacing! I dug into it and realized that we need to be a bit more adaptive in our sizing of the button. I ended up using the [viewport height (`vh`)](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Values_and_units/Numeric_data_types#viewport_units) as the height of the chevron's button. Since our styling was already using a percentage of vertical screen to move the results pane down, using `vh` allows us to use that same percentage value for the button height.

Regarding the blue sort button, I believe that problem is fixed with the updated configuration styling! Let me know if it persists even with the updated config though.

lib/components/map/default-map.tsx
```diff
@@ -540,7 +540,7 @@ class DefaultMap extends Component<DefaultMapProps> {
           {typeof getCustomMapOverlays === 'function' &&
             getCustomMapOverlays(!itinerary && !pending)}
           <NavigationControl
-            position={navigationControlPosition || 'bottom-right'}
```
amy-corson-ibigroup, line 543: I'll leave this to second reviewer to decide but the positioning on desktop here is looking weird to me, it's just so different from every other desktop map application. Wondering if we should follow what others apps do, take these buttons out on mobile and leave them alone on desktop.

lib/components/narrative/narrative-itineraries-header.tsx
amy-corson-ibigroup: Making sure this didn't slip through the cracks (I don't think it was included as part of the review comments, my bad)

lib/components/narrative/narrative-itineraries-header.tsx
```diff
@@ -161,46 +174,93 @@ export default function NarrativeItinerariesHeader({
...
+                <StyledIconWrapper>
+                  {mapExpanded ? (
+                    <UnicodeChevron expanded={mapExpanded}>
+                      &#10095;
+                    </UnicodeChevron>
+                  ) : (
+                    <UnicodeChevron expanded={mapExpanded}>
+                      &#10095;
+                    </UnicodeChevron>
+                  )}
```
amy-corson-ibigroup: You don't need this twice

amy-corson-ibigroup (changes requested):
Unfortunately I can still see the itinerary sort button. It's just a few pixels but it's there: 
image

lib/components/narrative/narrative-itineraries-header.tsx
alec-georgoff: Ah I indeed did forget about this, thanks for the reminder! Taken care of.

lib/components/narrative/narrative-itineraries-header.tsx
alec-georgoff: Fixed!

lib/components/map/default-map.tsx
alec-georgoff: How about we only move them on mobile? Check out [this commit](https://github.com/opentripplanner/otp-react-redux/pull/1502/commits/d0bd9cf2c2aa71db837b8eeca676ba151ecbbe58) and let me know what you think

alec-georgoff:
> Unfortunately I can still see the itinerary sort button. It's just a few pixels but it's there: image

Wow very good find! Tweaked the sizing to take care of that

lib/components/map/default-map.tsx
```diff
@@ -540,7 +546,11 @@ class DefaultMap extends Component<DefaultMapProps> {
           {typeof getCustomMapOverlays === 'function' &&
             getCustomMapOverlays(!itinerary && !pending)}
           <NavigationControl
-            position={navigationControlPosition || 'top-left'}
+            position={
+              overrideNavigationControlPosition ||
+              navigationControlPosition ||
+              'bottom-right'
+            }
```
amy-corson-ibigroup, line 553: Wait would this be easier? Rather than the additional prop:
```
position = { navigationControlPosition || (mobile ? 'top-left' : 'bottom-right') }
```
Let me know, maybe you already tried this

lib/components/narrative/narrative-itineraries-header.tsx
```diff
@@ -89,6 +102,16 @@ export default function NarrativeItinerariesHeader({
...
+  const mapExpansionText = useMemo(() => {
+    return mapExpanded
+      ? intl.formatMessage({
+          id: 'components.BatchResultsScreen.showResults'
+        })
+      : intl.formatMessage({
+          id: 'components.BatchResultsScreen.expandMap'
+        })
+  }, [intl, mapExpanded])
+
```
amy-corson-ibigroup, line 124: Interesting, I think generally we would just create two strings and conditionally render them based on `mapExpanded` below, but this might be more efficient? I'll let the second reviewer decide

lib/components/narrative/narrative-itineraries-header.tsx
```diff
@@ -161,46 +184,71 @@ export default function NarrativeItinerariesHeader({
...
+              <button
+                aria-label={mapExpansionText}
+                className="clear-button-formatting"
+                onClick={onClickExpansionButton}
+                style={{
+                  alignItems: 'center',
+                  display: 'flex',
+                  height: `${TOGGLE_MAP_BUTTON_HEIGHT}vh`,
+                  justifyContent: 'center',
+                  margin: '-10px 0',
```
amy-corson-ibigroup: This is a lot of inline CSS, could be good at this point to create a styled component

amy-corson-ibigroup (commented):
Bad news I am still seeing this: 
image

lib/components/map/default-map.tsx
alec-georgoff: I prefer my solution for two reasons:

1) I couldn't find an easy way to find, within the `default-map` component, whether we are in mobile view or not. From my understanding, this default map is just rendered normally by the `batch-results-screen` component and then auto-fits into the mobile window that is part of `batch-results-screen`
2) The more generic `overrideNavigationControlPosition` prop allows other consumers of the `default-map` to prescribe their own positions, making the `default-map` component more flexible for all consumers

Open to debate on these, though!

lib/components/narrative/narrative-itineraries-header.tsx
alec-georgoff: I went this route just for the best practice of memoization to avoid unnecessary re-renders, but I think the performance benefits are basically zero in this case. If someone else has strong opinions I'm happy to change the approach

lib/components/narrative/narrative-itineraries-header.tsx
alec-georgoff: Good call; styled component created!

alec-georgoff:
> Bad news I am still seeing this: image

Oof...I think different screen resolutions were causing this to still appear. I went ahead and added a very small top-margin to this component as a fix. Really hoping there aren't any more edge cases 😅

lib/components/map/default-map.tsx
amy-corson-ibigroup: 1. To see whether we are in mobile view directly from `default-map` you can import `isMobile` from core-utils which returns a boolean! Apologies, probably should have included the hook in my original comment
2. I agree that the position should be configurable, but as currently written that's not what `overrideNavigationControlPosition` is doing. That's covered by `navigationControlPosition` which can be changed via the config. Right now `overrideNavigationControlPosition` is just a drilled prop with a static value of `top-left`

amy-corson-ibigroup (approved):
Thank you for all the changes! This was a tricky UI change and I think it's well-executed here, nice work! 

LGTM with the caveat that I still think `overrideNavigationControlPosition` is unnecessary (I accidentally commented about that outside the scope of this review, whoops!)

lib/components/map/default-map.tsx
```diff
@@ -167,6 +171,7 @@ interface DefaultMapProps {
   itinerary: Itinerary
   mapConfig: MapConfig
   nearbyViewActive: boolean
+  overrideNavigationControlPosition: ControlPosition | undefined
```
daniel-heppner-ibigroup: can you use a `?` instead of `| undefined`?

lib/components/narrative/narrative-itineraries-header.tsx
daniel-heppner-ibigroup: I like your approach! Thinking of perf!

daniel-heppner-ibigroup (approved):
Pretty hefty change but the code looks good! Thanks!

PR #1504, opened 2025-12-23
Updating Saved Trip Requirements

amy-corson-ibigroup:
**Description:**
- Adds an asterisk to required fields in the trip form 
- Improves yup validation so users can no longer name trips with blank string

**PR Checklist:**


lib/components/user/monitored-trip/trip-basics-pane.tsx
```diff
@@ -377,6 +382,7 @@ class TripBasicsPane extends Component<TripBasicsProps, State> {
           <FormGroup validationState={errorStates.tripName}>
             <ControlLabel htmlFor="tripName">
               <FormattedMessage id="components.TripBasicsPane.tripNamePrompt" />
+              {!isReadOnly && <RequiredIndicator> *</RequiredIndicator>}
```
alec-georgoff: nit: I spent a couple minutes trying to research what the `*</` syntax was for JSX before realizing that it's just an asterisk inside the component 😅

Could we do something like this to make it more clear that the asterisk is just meant to be rendered?

```jsx
<RequiredIndicator>{'*'}</RequiredIndicator>
```

alec-georgoff (approved):
Just one optional nit but otherwise looks good!!

lib/components/user/monitored-trip/trip-basics-pane.tsx
```diff
@@ -377,6 +382,7 @@ class TripBasicsPane extends Component<TripBasicsProps, State> {
           <FormGroup validationState={errorStates.tripName}>
             <ControlLabel htmlFor="tripName">
               <FormattedMessage id="components.TripBasicsPane.tripNamePrompt" />
+              {!isReadOnly && <RequiredIndicator>{' *'}</RequiredIndicator>}
             </ControlLabel>
             {/* onBlur, onChange, and value are passed automatically. */}
             <Field
```
binh-dam-ibigroup, line 389: Should we add a `required` attribute to this field? If so, you could add the asterisk in CSS-only based on the `required` and `disabled` attributes.

lib/components/user/monitored-trip/trip-basics-pane.tsx
```diff
@@ -377,6 +382,7 @@ class TripBasicsPane extends Component<TripBasicsProps, State> {
           <FormGroup validationState={errorStates.tripName}>
             <ControlLabel htmlFor="tripName">
               <FormattedMessage id="components.TripBasicsPane.tripNamePrompt" />
+              {!isReadOnly && <RequiredIndicator>{' *'}</RequiredIndicator>}
```
binh-dam-ibigroup: Instead of adding a space character just for padding, use a CSS margin of equivalent width for the corresponding `span` element. (2 instances)

binh-dam-ibigroup (changes requested):

lib/components/user/monitored-trip/trip-basics-pane.tsx
amy-corson-ibigroup: Wow I'm a little surprised to realize we didn't have the `required` attribute on here already! Thank you for catching this, I agree it's a much better approach

amy-corson-ibigroup: Well once again I have overestimated HTML. We certainly should have `required` as part of the name field, but there is no way to add `required` to a fieldset of checkboxes. The advice I'm seeing is to [make the asterisks available to screenreaders](https://github.com/w3c/wcag/issues/1963#issuecomment-881852022) and if we have to do this for the checkbox fieldset, we should probably keep it for the name field as well, to keep the screenreader readout consistent through the page.

I added the `required` in [f40b6b3](https://github.com/opentripplanner/otp-react-redux/pull/1504/commits/f40b6b3922af68bed08c0885423fd2d0e6b6e9d8)

lib/components/user/monitored-trip/trip-basics-pane.tsx
amy-corson-ibigroup: Added in [f40b6b3](https://github.com/opentripplanner/otp-react-redux/pull/1504/commits/f40b6b3922af68bed08c0885423fd2d0e6b6e9d8)!

binh-dam-ibigroup (approved):
Nice update

PR #1505, opened 2025-12-26
Support Serving Built otp-rr From subdir

miles-grant-ibigroup:
I'm hoping that this small config changes doesn't have any impacts or regressions or side effects. But it seems like a harmless enough change! Dist output is only minimally affected: `/` --> `./`


daniel-heppner-ibigroup (approved):
it's worth a try!

josh-willis-arcadis (approved):
lets see what happens

PR #1506, opened 2025-12-27
Move applying defaults from setUser into form

amy-corson-ibigroup:
**Description:**
Previously our approach to applying user saved trip defaults was causing some issues, namely 1. logging in during trip planning was overriding the current active search, and 2. changing anything about user settings would redirect the user to the home page.

This new approach only applies the trip settings inside the form component and only if there's no active search.

**PR Checklist:**


lib/components/form/batch-settings.tsx
```diff
@@ -16,12 +17,13 @@ import * as narrativeActions from '../../actions/narrative'
 import { ComponentContext } from '../../util/contexts'
 import { getActiveSearch, hasValidLocation } from '../../util/state'
 import { getBaseColor, getDarkenedBaseColor } from '../util/colors'
+import { getDefaultModeButtons } from '../../util/api'
 import { StyledIconWrapper } from '../util/styledIcon'
-import AnimateHeight from 'react-animate-height'
+import { User } from '../user/types'
 
+// eslint-disable-next-line sort-imports-es6-autofix/sort-imports-es6
```
amy-corson-ibigroup: idk why this is here so im taking it out

lib/actions/user.js
```diff
@@ -24,6 +24,7 @@ import { isBlank } from '../util/ui'
 import { secureFetch } from '../util/middleware'
 import { TRIPS_PATH } from '../util/constants'
 
+import { MobileScreens } from './ui-constants'
```
daniel-heppner-ibigroup: what's this import for?

daniel-heppner-ibigroup (approved):
very happy with how this turned out!

lib/util/api.ts
```diff
@@ -79,3 +80,33 @@ export function getDefaultQuery(config: AppConfig) {
...
+    return initialState ?? {}
+  }
+}
+
+export function getDefaultModeButtons(state: AppReduxState): string[] {
+  const userSavedTripDefaults = state.user?.loggedInUser?.userSavedTripDefaults
+  const initialState = state.otp.config.modes.initialState?.enabledModeButtons
+  if (userSavedTripDefaults) {
+    return JSON.parse(userSavedTripDefaults)
+      .mode?.split(',')
```
alec-georgoff, line 110: If `mode` ends up being `undefined` here, then the whole function will return` undefined`, which is not expected given the return type of `string[]`. Also, I'm a little nervous about immediately trying to access properties on the JSON parsed object without knowing whether the parse was successful. Could we add some more safety in here to make sure that we always end up returning a `string[]` type?

amy-corson-ibigroup: Added some fallbacks! Now in both this and the mode setting function, if the parse returns `undefined` it should fall back on `initialState` or `[]`

amy-corson-ibigroup: Thanks for the catch!

alec-georgoff (approved):

PR #1508, opened 2025-12-30
Fix Route Renderer Data

binh-dam-ibigroup:
**Description:**

This PR removes legacy OTP1 fields from the data passed to `RouteRenderer` to address compatibility with custom route renderers that use OTP2-only objects. This is possible because we are already querying the `agency` object with id and name when we query the entire list of routes for the route viewer.

Tested with ST and ATL configs.

**PR Checklist:**


daniel-heppner-ibigroup (approved):
Big improvement! thanks!

josh-willis-arcadis (approved):
LGTM!

PR #1511, opened 2026-01-02
Remove route short names in headsigns

binh-dam-ibigroup:
**Description:**

This PR fixes headsigns where the route short name is prepended multiple times.

**PR Checklist:**

| Before | After |
|--------|-------|

### DIFF
--- a/__tests__/components/viewers/__snapshots__/stop-schedule-viewer.ts.snap
+++ b/__tests__/components/viewers/__snapshots__/stop-schedule-viewer.ts.snap
@@ -100,8 +100,8 @@ exports[`components > viewers > stop viewer should render with initial stop id a
           }
         >
           <StopScheduleViewer
-            calendarMax="2026-12-31"
-            calendarMin="2025-01-01"
+            calendarMax="2027-12-31"
+            calendarMin="2026-01-01"
             findStopTimesForStop={[Function]}
             homeTimezone="America/Los_Angeles"
             intl={
--- a/lib/util/viewer.js
+++ b/lib/util/viewer.js
@@ -60,6 +60,38 @@ export function routeIsValid(route, routeId) {
 }
 
 /**
+ * Try to remove the route short name from the headsign.
+ * Note that the headsign might be added multiple times by OTP, hence a recursive check.
+ */
+function removeRouteShortName(headsign, routeShortName) {
+  let finalHeadsign = headsign
+  if (headsign && routeShortName) {
+    while (finalHeadsign.startsWith(`${routeShortName} `)) {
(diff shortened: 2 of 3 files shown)


__tests__/components/viewers/__snapshots__/stop-schedule-viewer.ts.snap
```diff
@@ -100,8 +100,8 @@ exports[`components > viewers > stop viewer should render with initial stop id a
           }
         >
           <StopScheduleViewer
-            calendarMax="2026-12-31"
-            calendarMin="2025-01-01"
+            calendarMax="2027-12-31"
+            calendarMin="2026-01-01"
```
amy-corson-ibigroup, line 104: happy new year :)

lib/util/viewer.js
```diff
@@ -60,26 +60,49 @@ export function routeIsValid(route, routeId) {
...
-  const descNotRouteShortName = routeShortName && desc !== `${routeShortName}`
+function removeRouteShortName(headsign, routeShortName) {
+  let finalHeadsign = headsign
+  if (headsign && routeShortName) {
+    while (finalHeadsign.startsWith(`${routeShortName} `)) {
+      finalHeadsign = finalHeadsign.substring(routeShortName.length + 1)
+    }
+  }
+  return finalHeadsign
+}
```
amy-corson-ibigroup, line 74: Just to make sure I understand, we're not using something like `replaceAll` here because we wouldn't want to remove characters that aren't actually the short name, right? For example if the shortname was something like "A" we wouldn't want to accidentally take out all the "a"s

amy-corson-ibigroup (approved):
Looks good to me! Just a clarifying question to make sure I'm understanding correctly.

lib/util/viewer.js
binh-dam-ibigroup: Correct.

daniel-heppner-ibigroup (approved):
Code looks good, do we know why the short name was getting prepended to the pattern name multiple times? Is there some other issue we should look into that's causing that?

binh-dam-ibigroup:
> Code looks good, do we know why the short name was getting prepended to the pattern name multiple times? Is there some other issue we should look into that's causing that?

@daniel-heppner-ibigroup This is probably where it is happening: https://github.com/opentripplanner/OpenTripPlanner/blob/24ade9781126307afd08cbe7eda427935e6cfddc/application/src/main/java/org/opentripplanner/graph_builder/module/TripPatternNamer.java#L124, where the `StringBuilder` is initialized with the route short name, and the trip headsign is appended, where the trip headsign also includes the route short name.

PR #1512, opened 2026-01-06
Jan '26 OTP UI Update

miles-grant-ibigroup:
Important crash-fix for map-popup!

Also contains some other new features that are unused.

Package updates break percy tests, these are fixed


binh-dam-ibigroup (approved):

amy-corson-ibigroup (approved):
Sorry this took me so long!

PR #1514, opened 2026-01-15
Prevent trip save if monitored days loading

amy-corson-ibigroup:
**Description:**
Stop the user from saving a trip if there's a chance the monitored days might update. This is to prevent a long server error in the UI unless its absolutely necessary

**PR Checklist:**


josh-willis-arcadis (approved):
LGTM!

lib/components/user/monitored-trip/trip-basics-pane.tsx
```diff
@@ -306,14 +307,31 @@ class TripBasicsPane extends Component<TripBasicsProps, State> {
...
 
   componentDidUpdate(prevProps: TripBasicsProps) {
     this._updateNewTripItineraryExistence(prevProps)
+    const {
+      itineraryExistence,
+      setIsLoading,
+      values: monitoredTrip
+    } = this.props
+    if (
+      (monitoredTrip.itineraryExistence || itineraryExistence) &&
```
alec-georgoff: Are we sure that `monitoredTrip` will always exist? I tried to find the `values` property in `TripBasicsProps` but didn't see it. It's possible I'm misunderstanding which props we're destructuring, though.

If we can't guarantee its existence, we should use `monitoredTrip?.itineraryExistence` to be safe here.

amy-corson-ibigroup: I'm pretty sure the absence of `!monitoredTrip.itineraryExistence` on this page gets handled when the component mounts on line 316, but I don't mind adding it here just to be safe!

alec-georgoff (approved):

PR #1516, opened 2026-01-27
Fix Narrative Header Layout

binh-dam-ibigroup:
**Description:**

This PR fixes the layout for Feedback button in the narrative header when that feature is enabled, whether desktop or mobile view (see screenshots below).

**PR Checklist:**

| Before | After |
|--------|-------|


alec-georgoff (approved):
Nice fix! I believe this was introduced by my #1502 PR, so I appreciate you catching this side effect

josh-willis-arcadis (approved):

josh-willis-arcadis:
not sure why the percy tests are failing??

binh-dam-ibigroup:
> not sure why the percy tests are failing??

Percy test is failing because one of the tests took one second more than "normal", and the printed time is different as a result. But let me update the branch and see if that's still the case.

binh-dam-ibigroup:
> > not sure why the percy tests are failing??
> 
> Percy test is failing because one of the tests took one second more than "normal", and the printed time is different as a result. But let me update the branch and see if that's still the case.

Percy tests pass again after updating the branch.

PR #1522, opened 2026-02-03
Show Vehicles of Subpatterns in Viewed Pattern

binh-dam-ibigroup:
**Description:**

This PR changes the pattern viewer so that vehicles in patterns equal or contained in the viewed pattern are displayed.
Previously, these vehicles would not be displayed except when viewing the entire route.

Behind the scene, each pattern in a route is mapped to the largest containing pattern in that route.
To resolve internal pattern duplication, different stops in the same parent station are considered the same.
Note that patterns previously hidden might emerge from this fix (e.g. "Beacon Hill" in Sound Transit in the screenshot below, which is an actual partial terminus of the 1-Line).

**PR Checklist:**

| Before | After |
|--------|-------|

### DIFF
--- a/__tests__/util/pattern-viewer.ts
+++ b/__tests__/util/pattern-viewer.ts
@@ -93,7 +93,7 @@ describe('util > pattern-viewer', () => {
 
   describe('sortAndRemoveSubpatterns', () => {
     it('should sort and remove subpatterns', () => {
-      // Consider the following patterns P1...P6 of the same route:
+      // Consider the following patterns of the same route:
       // Stops S1 S2 S3 S4 S5 S6 S7 --> direction of travel
       // P1:   o--o--o--o--o
       // P2:         o--o-----o--o
@@ -101,9 +101,10 @@ describe('util > pattern-viewer', () => {
       // P4:   o-----o--o--o
       // P5:   o--o--o--o--o
       // P6: <undefined stops>
+      // P7:         o--o--o
       //
       // One of P1 or P5 should be removed because both have the exact same stops.
-      // P3 should be removed because it is a subset of P1, P4, and P5.
+      // P3 and P7 should be removed because they are a subset of P1, P4, and P5.
       // P1, P2, and P4 should be kept.
       const patterns: Pattern[] = [
         {
@@ -171,6 +172,17 @@ describe('util > pattern-viewer', () => {
           },
           route,
           stops: undefined
+        },
+        {
+          desc: 'P7 Pattern name',
+          headsign,
+          id: 'P7',
+          patternGeometry: {
+            length: 987,
+            points: 'p7-points'
+          },
+          route,
+          stops: createStops(['S3', 'S4', 'S5'])
         }
       ]
       const { containingPatterns, filteredPatterns } =
--- a/__tests__/util/pattern-viewer.ts
+++ b/__tests__/util/pattern-viewer.ts
@@ -98,7 +98,7 @@ describe('util > pattern-viewer', () => {
       // P1:   o--o--o--o--o
       // P2:         o--o-----o--o
       // P3:         o--o--o
-      // P3:   o-----o--o--o
+      // P4:   o-----o--o--o
       //
       // P3 should be removed because it is a subset of P1 and P3.
       // P1, P2, and P4 should be kept.
@@ -148,11 +148,15 @@ describe('util > pattern-viewer', () => {
           stops: createStops(['S1', 'S3', 'S4', 'S5'])
         }
       ]
-      const filteredPatterns = sortAndRemoveSubpatterns(patterns)
+      const { filteredPatterns, subPatterns } =
+        sortAndRemoveSubpatterns(patterns)
       expect(filteredPatterns.length).toBe(3)
       expect(filteredPatterns[0]).toBe(patterns[0])
       expect(filteredPatterns[1]).toBe(patterns[3])
       expect(filteredPatterns[2]).toBe(patterns[1])
+      expect(subPatterns.P1).toContain('P3')
+      expect(subPatterns.P3).toBe(undefined)
+      expect(subPatterns.P4).toBe(undefined)
     })
   })
 })
--- a/__tests__/util/pattern-viewer.ts
+++ b/__tests__/util/pattern-viewer.ts
@@ -173,18 +173,23 @@ describe('util > pattern-viewer', () => {
           stops: undefined
         }
       ]
-      const { filteredPatterns, subPatterns } =
+      const { containingPatterns, filteredPatterns, subPatterns } =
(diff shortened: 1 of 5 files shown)


lib/util/pattern-viewer.ts
```diff
@@ -65,3 +77,69 @@ export function extractMainHeadsigns(
...
+  const patternsSortedByLength = [...patterns]
+    .filter((pattern) => pattern.stops?.length)
+    .sort((a, b) => (a.stops?.length || 0) - (b.stops?.length || 0))
+
+  const containingPatterns: Record<string, string> = {}
+
+  // Keep patterns that are not subsets of larger patterns
+  const filteredPatterns = patternsSortedByLength
+    // Start with the largest for performance
+    .reverse()
```
amy-corson-ibigroup: Could we do `b - a` in  `patternsSortedByLength` so we don't have to reverse here or would that not work?

lib/util/pattern-viewer.ts
```diff
@@ -65,3 +77,69 @@ export function extractMainHeadsigns(
...
+      // Compare to all other patterns TODO: make this beat O(n^2)
+      let includePattern = true
+      const patternStops = pattern.stops?.map(getParentStopOrStopId) || []
+
+      patternsSortedByLength.forEach((p, index) => {
+        // Don't compare against ourself
+        if (p.id === pattern.id) return
+
+        // If the pattern has no stops, exclude it.
+        if (!p.stops || p.stops.length === 0) return
```
amy-corson-ibigroup: I believe you already checked for this on line 100?

lib/components/map/connected-transit-vehicle-overlay.tsx
```diff
@@ -12,23 +10,37 @@
...
+  stopStatus?: string
+  textColor?: string
+}
 
-function VehicleTooltip(props) {
-  const { intl, vehicle } = props
+interface VehicleTooltipProps {
+  vehicle: TransitVehicleExt
+}
+
```
amy-corson-ibigroup: This feels easy enough to inline, do we anticipate adding to  the props for this component in the future?

lib/util/pattern-viewer.ts
```diff
@@ -65,3 +77,69 @@ export function extractMainHeadsigns(
...
+
+/**
+ * Obtains the parent stop id, if available, or the stop id.
+ */
+export function getParentStopOrStopId(stop: StopWithParent): string {
+  return stop.parentStation?.id || stop.id
+}
+
+/**
+ * Returns a list of patterns, in descending length order, in which none is a subpattern of any other.
```
amy-corson-ibigroup: This actually returns two lists of patterns, both `filteredPatterns` and `containingPatterns` would you mind updating to describe both

amy-corson-ibigroup (approved):
This looks good! Just a few questions

lib/util/pattern-viewer.ts
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/1522/commits/99214b38ff3e1130950a7ebf826c2e1d0b8fcd31

lib/util/pattern-viewer.ts
binh-dam-ibigroup: Good catch! (https://github.com/opentripplanner/otp-react-redux/pull/1522/commits/32881aa4b58a9fcf03e66d7571a658f293fe6c6c)

lib/util/pattern-viewer.ts
binh-dam-ibigroup: I removed that in exchange for a type declaration (https://github.com/opentripplanner/otp-react-redux/pull/1522/commits/ff6211527bd68b47903b419c756754cbe5304031)

lib/components/map/connected-transit-vehicle-overlay.tsx
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/1522/commits/b90e46294ef3b2f36667e581112fdf9fe807a760

__tests__/util/pattern-viewer.ts
```diff
@@ -1,17 +1,36 @@
...
-import { extractMainHeadsigns } from '../../lib/util/pattern-viewer'
+import {
+  extractMainHeadsigns,
+  PatternSummary,
+  sortAndRemoveSubpatterns
+} from '../../lib/util/pattern-viewer'
+import { Pattern } from '../../lib/components/util/types'
 
-function createStops(ids) {
+function createStops(ids: string[]): Stop[] {
```
alec-georgoff, line 11: Types, yay!!

__tests__/util/pattern-viewer.ts
```diff
@@ -25,44 +44,146 @@ describe('util > pattern-viewer', () => {
...
       expect(headsignData.length).toBe(2)
       expect(headsignData[0].headsign).toBe(headsign)
       expect(headsignData[1].headsign).toBe(`${headsign} (S7)`)
     })
   })
+
+  describe('sortAndRemoveSubpatterns', () => {
+    it('should sort and remove subpatterns', () => {
+      // Consider the following patterns P1...P6 of the same route:
+      // Stops S1 S2 S3 S4 S5 S6 S7 --> direction of travel
```
alec-georgoff, line 97: Excellent explanation here, thanks for including this

lib/util/pattern-viewer.ts
```diff
@@ -65,3 +81,72 @@ export function extractMainHeadsigns(
...
+ *   Pattern 4: Stops       C, D, E
+ *   Pattern 5: Stops A,    C, D, E
+ * @returns An object with a filteredPatterns field with the filtered (largest) patterns,
+ *   and a containingPatterns field with a map of the containing pattern for each pattern.
+ */
+export function sortAndRemoveSubpatterns(patterns: Pattern[]): SubPatternInfo {
+  const containingPatterns: Record<string, string> = {}
+
+  // Filter out patterns with no stops.
+  const patternsWithStops = [...patterns].filter(
```
alec-georgoff: nit: we don't need to do `[...patterns]`, since [filter creates a shallow copy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter)

alec-georgoff (approved):

lib/util/pattern-viewer.ts
binh-dam-ibigroup: Good catch! A remnant of when that call would be followed by `.reverse()` which operates in place... (https://github.com/opentripplanner/otp-react-redux/pull/1522/commits/46fa25e0ca6ec58feea677da95bab40877893128)

PR #1525, opened 2026-02-09
fix map attribution in trip preview

amy-corson-ibigroup:
**Description:**
- Makes sure to only show the attribution if the map is visible
- Fixes a bug where the map un-mounting would cause the attribution text to be set to `undefined`

**PR Checklist:**


alec-georgoff (approved):

josh-willis-arcadis (approved):
LGTM

PR #1526, opened 2026-02-10
Calltaker: Make trinetReDirect optional

binh-dam-ibigroup:
**Description:**

This PR makes the `trinetReDirect` config param optional.

As a result, for the call button and call history and Field Trip list to appear, a preconfigured `sessionId` must be passed in the web URL. (An alert is shown otherwise.)

**PR Checklist:**
- [ ] Are all languages supported (Internationalization/Localization)? => Not really needed for Calltaker.

### DIFF
--- a/lib/actions/call-taker.js
+++ b/lib/actions/call-taker.js
@@ -107,7 +107,6 @@ export function initializeModules(intl) {
     const { datastoreUrl, trinetReDirect } = getState().otp.config
     // Initialize session if datastore enabled.
     if (datastoreUrl) {
-      // TODO: Generalize for non-TriNet instances.
       const sessionId = getUrlParams().sessionId
       if (sessionId) {
         // Initialize the session if found in URL query params.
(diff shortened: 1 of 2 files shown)


lib/actions/call-taker.js
```diff
@@ -105,15 +106,17 @@ export function initializeModules(intl) {
   return function (dispatch, getState) {
     const { datastoreUrl, trinetReDirect } = getState().otp.config
     // Initialize session if datastore enabled.
-    if (datastoreUrl && trinetReDirect) {
+    if (datastoreUrl) {
       // TODO: Generalize for non-TriNet instances.
```
miles-grant-ibigroup: We can probably remove this TODO now?

miles-grant-ibigroup (approved):
Should we have this also disable field-trip? Since that is at this point a feature which requires trinet

lib/actions/call-taker.js
binh-dam-ibigroup: Yes (https://github.com/opentripplanner/otp-react-redux/pull/1526/commits/7134e4c9d4a05bdce2db5b33663f73a1a7fecf63)

binh-dam-ibigroup:
> Should we have this also disable field-trip? Since that is at this point a feature which requires trinet

Field Trip is already/also covered under the `initializeModules` method. An attempt to open the Field Trip view from the sidebar when no valid session is provided (and after the alert is shown) results in a crash, and that's probably ok?

alec-georgoff (approved):

PR #1527, opened 2026-02-11
Saved trip form

amy-corson-ibigroup:
**Description:**
Beginning the design overhaul for saved trip form and saved trip cards:
- Updating the alert language to be a little more consistent, removing exclamation points.
- Adding the trip summary component inside the save trip form
- Using `Alert` from building-blocks and using it to display realtime alerts and trip alerts

image

Coming up (not included in this PR): 
- Using the Trip name as page header
- Re-figuring Trip notification pane
- Custom checkboxes for monitored days
- Move trip status details into saved trip cards

**PR Checklist:**


lib/components/user/monitored-trip/trip-summary-pane.tsx
```diff
@@ -165,15 +175,20 @@ const TripSummaryPane = ({
 
     const isActiveAndSnoozed = monitoredTrip.isActive && monitoredTrip.snoozed
 
+    const itineraryLegs = monitoredTrip?.itinerary?.legs
+    const fromPlace = from?.name || itineraryLegs[0]?.from?.name
+    const toPlace =
+      to?.name || itineraryLegs[itineraryLegs.length - 1]?.to?.name
```
daniel-heppner-ibigroup, line 181: || will detect an empty string as falsey, is that intended? If not, you can use the `??` operator instead

lib/components/viewers/stop-time-cell.tsx
```diff
@@ -24,7 +24,7 @@ import DepartureTime from './departure-time'
 const { getUserTimezone } = coreUtils.time
 const ONE_HOUR_IN_SECONDS = 3600
 
-const PulsingRss = styled(Rss)`
+export const PulsingRss = styled(Rss)`
```
daniel-heppner-ibigroup: I don't think this is a good place to keep this if it's going to be used in multiple places. can you find a better spot for it? Is there a styled.ts file we can put it in?

daniel-heppner-ibigroup (approved):
Looks good! A few comments on code quality

lib/components/user/monitored-trip/trip-summary-pane.tsx
amy-corson-ibigroup: Yeah that's intended! No sense in displaying an empty string if there's a more useful name in `itineraryLegs`, right?

lib/components/viewers/stop-time-cell.tsx
amy-corson-ibigroup: Good point, I added it to `styledIcon` in utils!

i18n/en-US.yml
```diff
@@ -685,16 +685,16 @@ components:
   TripPreviewLayout:
     previewTrip: Preview Trip
   TripStatus:
-    alerts: "{alerts, plural, one {# alert!} other {# alerts!}}"
+    alerts: "There {alerts, plural, one {is # alert} other {are # alerts}} on your trip"
     deleteTrip: Delete Trip
     planNewTrip: Plan New Trip
   TripStatusRenderers:
     active:
-      delayedHeading: Trip is in progress and is delayed {formattedDuration}!
```
binh-dam-ibigroup, line 694: Evan loved those exclamation marks!

lib/components/user/monitored-trip/trip-status-rendering-strategies/upcoming-trip-renderer.js
```diff
@@ -114,7 +120,7 @@ export default function upcomingTripRenderer({
             }}
           />
         )
-        data.panelBsStyle = 'warning'
+        data.panelBsStyle = COLOR_VARIABLES.warning
       }
     } else {
       data.headingText = (
```
binh-dam-ibigroup, line 126: As written, this will be the same color as the `info` state. I am ok with that, but note that there were two  original colors, one of which is blue.

lib/components/user/monitored-trip/trip-status.js
```diff
@@ -13,14 +14,55 @@ import {
...
-  margin-right: 10px;
+const StyledFooterButton = styled(ToggleNotificationButton)``
+
+const StyledButtonContainer = styled.div`
+  display: flex;
+  gap: 10px;
+  button:not(:last-of-type) {
+    &::after {
+      margin-left: 10px;
+      content: '|';
```
binh-dam-ibigroup, line 29: Nit - sort CSS (2 instances)

lib/components/viewers/trip-viewer.js
```diff
@@ -193,19 +200,23 @@ class TripViewer extends Component {
...
-                    values={{
-                      boardAtStop: (
-                        <strong>{stopTimes?.[fromIndex]?.stop.name}</strong>
-                      ),
-                      disembarkAtStop: (
-                        <strong>{stopTimes?.[toIndex]?.stop.name}</strong>
-                      )
-                    }}
+                <AlertContainer>
+                  <Alert
```
binh-dam-ibigroup, line 204: Nit - Can we remove the icon for the trip viewer? It is taking a lot of space for no reason.

lib/components/user/mobility-profile/trip-readonly-pane.tsx
```diff
@@ -38,12 +39,15 @@ const TripReadOnlyPane = ({
...
 
   return (
-    <Alert bsStyle="warning" style={{ margin: '-30px 0' }}>
-      <FormattedMessage
-        id="components.SavedTripEditor.readOnlyBanner"
-        values={{ creator }}
-      />
-    </Alert>
+    <Alert
+      alertHeader={
```
binh-dam-ibigroup, line 43: We might not need an icon here.

binh-dam-ibigroup (approved):
Looks nice, code is clean. See if you can remove the icon in the Trip Viewer and in the read-only alert in Edit Trip.

i18n/en-US.yml
amy-corson-ibigroup: Oh no! Sorry Evan :(

lib/components/user/monitored-trip/trip-status-rendering-strategies/upcoming-trip-renderer.js
amy-corson-ibigroup: Good catch, thank you! These should have been imported from `colors.ts` which does have the info state set as `blue[50]`

lib/components/viewers/trip-viewer.js
amy-corson-ibigroup: Right now, the `Alert` component doesn't have the flexibility for no icon... With the way the `grid` layout in this component is set up, the easiest thing to do will probably be open a new PR in OTP-UI to make the icon optional. I will make a ticket to follow up on this!

PR #1529, opened 2026-02-12
User saved places fixes

amy-corson-ibigroup:
**Description:**
- Fix bug where toast was not displaying when user saved place is deleted
- Added additional validation to user saved place name to mirror trip name (no blank space as name)

**PR Checklist:**

### DIFF
--- a/lib/actions/user.js
+++ b/lib/actions/user.js
@@ -823,10 +823,11 @@ export function deleteLoggedInUserPlace(placeIndex, intl, placeToSave) {
   return function (dispatch, getState) {
     const { loggedInUser } = getState().user
     loggedInUser.savedLocations.splice(placeIndex, 1)
+    const place = placeToSave || loggedInUser.savedLocations[placeIndex]
 
     return toastPromise(
       dispatch(createOrUpdateUser(loggedInUser, intl)),
-      toastMessageOnPlaceChanged(placeToSave, intl, 'Deleted'),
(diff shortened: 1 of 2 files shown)


lib/actions/user.js
```diff
@@ -823,10 +823,11 @@ export function deleteLoggedInUserPlace(placeIndex, intl, placeToSave) {
   return function (dispatch, getState) {
     const { loggedInUser } = getState().user
     loggedInUser.savedLocations.splice(placeIndex, 1)
+    const place = placeToSave || loggedInUser.savedLocations[placeIndex]
```
binh-dam-ibigroup: We might not need the third parameter `placeToSave` anymore, but do you remember why it was added to begin with in https://github.com/opentripplanner/otp-react-redux/pull/1485?

binh-dam-ibigroup (approved):
Removing the extra param as indicated would be nice to have, but I'm going ahead and approving this.

lib/actions/user.js
amy-corson-ibigroup: It was added to be able to pass the full `place` object to the `toast` but you're right we can just find the object by index

amy-corson-ibigroup: Fixed in [5123bf6](https://github.com/opentripplanner/otp-react-redux/pull/1529/commits/5123bf6338e6156804278177fe9938b02e67c355)!

alec-georgoff (approved):

PR #1530, opened 2026-02-16
Field trip fixes feb 2026

miles-grant-ibigroup:
- [x] Migrate api calls to OTP2
- [x] Avoid crashes on bad play data
- [x] Don't needlessly show location correction popup

### DIFF
--- a/lib/actions/field-trip.js
+++ b/lib/actions/field-trip.js
@@ -491,7 +491,7 @@ export function saveRequestTripItineraries(request, outbound, intl) {
  * Sets the appropriate request parameters for either the outbound or inbound
  * journey of a field trip.
  */
-function prepareQueryParams(request, outbound) {
+function prepareQueryParams(request, outbound, showRePlanPopup) {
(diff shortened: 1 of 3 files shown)


alec-georgoff (approved):

lib/actions/field-trip.js
```diff
@@ -604,16 +608,29 @@ function getMissingTripHashesForActiveItineraries() {
...
+      dispatch(
+        createGraphQLQueryAction(
+          `
+          query {
+            trip (id: "${tripId}") {
+              semanticHash
+             }
+          }
+          `,
+          {},
```
daniel-heppner-ibigroup, line 621: can you use the graphql variables here?

daniel-heppner-ibigroup (approved):
Looks good, thanks for fixes

lib/actions/field-trip.js
miles-grant-ibigroup: I think it would add more complexity than it would reduce

PR #1531, opened 2026-02-16
Save Trip: Reactivate Save button after user dismisses error

binh-dam-ibigroup:
**Description:**

This PR changes the Save Trip form, so that if saving a new or existing trip fails for any reason, the Save button gets reactivated.

Behind the scenes, Formik's `isSubmitting` state is used to correctly determine the Save button state.
You might see the Save button briefly flashing as reenabled just before redirection to the next page occurs.

**PR Checklist:**


lib/components/user/monitored-trip/saved-trip-screen.js
```diff
@@ -101,20 +108,24 @@ class SavedTripScreen extends Component {
...
-    createOrUpdateUserMonitoredTrip(
+    await createOrUpdateUserMonitoredTrip(
       tripToSave,
       isCreating,
       undefined,
       undefined,
       intl
     )
+
+    // After saving, let this trip to be the new undirty reference,
```
amy-corson-ibigroup:
```suggestion
    // After saving, let this trip be the new undirty reference,
```
Can you also elaborate what "undirty" means in this context?

amy-corson-ibigroup (approved):
Great fix! Just a quick question on one of the comments but the code itself looks great

lib/components/user/monitored-trip/saved-trip-screen.js
binh-dam-ibigroup: Formik sets the `isDirty` flag based on an initial value. When users click Save Trip from the search results, a new `MonitoredTrip` instance is created in memory with a blank name, etc. and that is the initial value (clean, "undirty") for Formik. When users click "Save" from the Save Trip form, what I did is set the initial value to be the saved trip, so that Formik treats the state as clean, so we don't warn for lost changes.

josh-willis-arcadis (approved):
LGTM

PR #1532, opened 2026-02-20
[narriative itinerary] Show Bike Time when Bike time exceeds walk time

miles-grant-ibigroup:
I'd like to have a cleaner solution that encompasses all modes. However, as the narrative itinerary is not used outside calltaker, this quick patch which only enables bike time felt cleaner. If we find the narrative itinerary seeing heavier use again, we can revisit this to dynamically determine the longest non-transit mode, and display the time for that mode.

### DIFF
--- a/lib/util/itinerary.tsx
+++ b/lib/util/itinerary.tsx
@@ -121,6 +121,14 @@ export function getLastLegEndTime(legs: Leg[]): number {
   return +legs[legs.length - 1].endTime
 }
 
+// TODO: is there a type for OTP Modes?
+export function getTotalTimeForMode(legs: Leg[], mode: string): number {
+  return legs
+    .filter((l) => l.mode === mode)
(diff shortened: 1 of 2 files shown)


lib/util/itinerary.tsx
```diff
@@ -121,6 +121,14 @@ export function getLastLegEndTime(legs: Leg[]): number {
   return +legs[legs.length - 1].endTime
 }
 
+// TODO: is there a type for OTP Modes?
```
binh-dam-ibigroup, line 124: OTP uses an enumeration for leg mode, see https://docs.opentripplanner.org/api/dev-2.x/graphql-gtfs/types/Mode, so for OTP-RR we would have to introduce something equivalent.

binh-dam-ibigroup (approved):

alec-georgoff (approved):

PR #1533, opened 2026-02-20
Correct Nearby View Stop Card Headsign Merging Behavior

miles-grant-ibigroup:
Previously, this behavior relied on a complicated `if else` tree. This tree turned out not to work the way I expected. Instead of working our way down the tree, we instead skipped out at the first branch that passed. I removed the `else`s, and replicated that behavior by adding an additional check for past checks to also have passed.

I'm sure there's a cleaner way to represent these series of checks, and would appreciate an alternative.

### DIFF
--- a/lib/components/viewers/nearby/stop.tsx
+++ b/lib/components/viewers/nearby/stop.tsx
@@ -63,6 +63,8 @@ export const patternArrayforStops = (
   return stopData?.stoptimesForPatterns
     ?.reduce<PatternStopTime[]>((acc, cur) => {
       const currentHeadsign = extractHeadsignFromPattern(cur.pattern)
+      // TODO: Extract routes out
+      // https://github.com/opentripplanner/otp-react-redux/pull/1533#discussion_r2844679701
       const dupe = acc.findIndex((p) => {
         // TODO: use OTP_generated ids
         let sameRoute = false
--- a/lib/components/viewers/nearby/stop.tsx
+++ b/lib/components/viewers/nearby/stop.tsx
@@ -69,13 +69,15 @@ export const patternArrayforStops = (
         if (p.pattern.route?.shortName && cur.pattern.route?.shortName) {
           sameRoute =
             p.pattern.route?.shortName === cur.pattern.route?.shortName
-        } else if (p.pattern.route?.longName && cur.pattern.route?.longName) {
-          sameRoute = p.pattern.route?.longName === cur.pattern.route?.longName
-        } else if (
-          p?.stoptimes?.[0]?.headsign &&
-          cur?.stoptimes?.[0]?.headsign
-        ) {
+        }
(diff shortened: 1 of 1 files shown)


lib/components/viewers/nearby/stop.tsx
```diff
@@ -69,13 +69,15 @@ export const patternArrayforStops = (
...
-        } else if (p.pattern.route?.longName && cur.pattern.route?.longName) {
-          sameRoute = p.pattern.route?.longName === cur.pattern.route?.longName
-        } else if (
-          p?.stoptimes?.[0]?.headsign &&
-          cur?.stoptimes?.[0]?.headsign
-        ) {
+        }
+        if (p.pattern.route?.longName && cur.pattern.route?.longName) {
+          sameRoute =
+            sameRoute &&
```
alec-georgoff, line 77: Don't we want to use `||` here and below instead of `&&`? If the condition on line 69 evaluates to false, then the assignment on line 74 will always end up as false, regardless of the result of the comparison on line 76...

alec-georgoff: ...or do we need all three conditions to be true in order to consider the routes to be equal? Sorry if I'm misunderstanding

miles-grant-ibigroup: Yeah that's the main fix in this PR! All 3 conditions have to be true, but only if the values are all present. Like I said in my PR description, there's probably a cleaner way to do it!

alec-georgoff: I spent some time trying to come up with something demonstrably cleaner but was unsuccessful...I'm fine with what you've come up with here!

alec-georgoff (approved):

lib/components/viewers/nearby/stop.tsx
daniel-heppner-ibigroup, line 68: extract variables for pRoute and curRoute:
`const pRoute = p.pattern.route` 
etc

lib/components/viewers/nearby/stop.tsx
```diff
@@ -69,13 +69,15 @@ export const patternArrayforStops = (
...
+            p.pattern.route?.longName === cur.pattern.route?.longName
+        }
+        if (p?.stoptimes?.[0]?.headsign && cur?.stoptimes?.[0]?.headsign) {
           sameRoute =
+            sameRoute &&
```
daniel-heppner-ibigroup, line 82: I think it might be nice to extract the isSameRoute logic to a little function above this one to eliminate the `let sameRoute`, but up to you if you're interested in doing that.

daniel-heppner-ibigroup (approved):

lib/components/viewers/nearby/stop.tsx
miles-grant-ibigroup: I just don't see us ever re-using this...

PR #1539, opened 2026-02-26
Calltaker Config for Copy Itinerary URL, Remove Router ID Code

binh-dam-ibigroup:
**Description:**

This PR adds a new calltaker config param that sets the target host of the "Copy Link" button in the itinerary narrative.
That way, the copied URL can be set to a non-calltaker OTP-RR instance for sharing with the general public.

trip-tools.js is converted to TypeScript in the process, and OTP1 router ID code is removed.

**PR Checklist:**

### DIFF
--- a/lib/components/narrative/trip-tools.tsx
+++ b/lib/components/narrative/trip-tools.tsx
@@ -165,7 +165,9 @@ interface LinkButtonProps {
 
 class LinkButton extends Component<LinkButtonProps> {
   _onClick = () => {
-    this.props.url && (window.location.href = this.props.url)
+    if (this.props.url) {
+      window.location.href = this.props.url
+    }
   }
 
   render() {
@@ -225,7 +227,7 @@ const TripTools = ({
     return IconComponent
   }, [SvgIcon, popupTarget])
 
-  const buttonComponents = [] as ReactNode[]
+  const buttonComponents: ReactNode[] = []
   buttonTypes.forEach((type) => {
     switch (type) {
       case 'COPY_URL':
--- a/lib/components/narrative/trip-tools.tsx
+++ b/lib/components/narrative/trip-tools.tsx
@@ -1,32 +1,46 @@
-/* eslint-disable no-case-declarations */
 import { Button } from 'react-bootstrap'
 import { Check } from '@styled-icons/fa-solid/Check'
 import { Clipboard } from '@styled-icons/fa-solid/Clipboard'
 import { connect } from 'react-redux'
 import { Flag } from '@styled-icons/fa-solid/Flag'
-import { FormattedMessage, injectIntl } from 'react-intl'
+import { FormattedMessage, injectIntl, WrappedComponentProps } from 'react-intl'
 import { Print } from '@styled-icons/fa-solid/Print'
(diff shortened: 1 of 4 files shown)


lib/components/narrative/trip-tools.tsx
```diff
@@ -134,29 +143,29 @@ class ReportIssueButtonBase extends Component {
...
+  Icon: ComponentType
+  onClick: () => void
+  text: ReactNode
+  url?: string
+}
+
+class LinkButton extends Component<LinkButtonProps> {
   _onClick = () => {
-    window.location.href = this.props.url
+    this.props.url && (window.location.href = this.props.url)
```
alec-georgoff: nit: I feel like it would be more clear to just use `if (this.props.url) { ... }`

lib/components/narrative/trip-tools.tsx
```diff
@@ -175,37 +184,54 @@ class LinkButton extends Component {
...
-      )
-    }
+    const IconComponent = (componentProps: SVGProps<SVGSVGElement>) => (
+      <SvgIcon iconName={popupTarget} {...componentProps} />
+    )
     return IconComponent
   }, [SvgIcon, popupTarget])
 
-  const buttonComponents = []
+  const buttonComponents = [] as ReactNode[]
```
alec-georgoff: nit: instead of casting, I'd prefer we do `const buttonComponents: ReactNode[] = []`. To me, casting is used when I want to tell the compiler to "just trust me", whereas declaring the type before the `=` feels less like I'm making a one-time exception. Not a huge deal though, just my preference.

alec-georgoff (approved):
Just a couple nits. Yay for new TypeScript file!!

lib/components/narrative/trip-tools.tsx
binh-dam-ibigroup: Sure (https://github.com/opentripplanner/otp-react-redux/pull/1539/commits/48e2d5e508c92b69d7fd2d3d9e3840bbe2368c22)

lib/components/narrative/trip-tools.tsx
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/1539/commits/48e2d5e508c92b69d7fd2d3d9e3840bbe2368c22

lib/components/narrative/trip-tools.tsx
```diff
@@ -134,29 +143,31 @@ class ReportIssueButtonBase extends Component {
     return (
       <Button className="tool-button" onClick={this._onClick}>
         <IconWithText Icon={Flag}>
-          {/* FIXME: Depending on translation, Spanish and French strings may not fit in button width. */}
```
miles-grant-ibigroup, line 137: Is this really fixed?

miles-grant-ibigroup (approved):
Really clean thanks!

lib/components/narrative/trip-tools.tsx
binh-dam-ibigroup: Yes, you will see a CSS `min-width` attribute from a while ago.

PR #1541, opened 2026-03-04
Support Sort by Emissions

miles-grant-ibigroup:
Allows sorting by emissions. I won't lie it's a bit confusing which way shows the highest and lowest emissions... 

It's disabled by default as the co2 values are 0 if the co2 calculation is disabled.

### DIFF
--- a/i18n/en-US.yml
+++ b/i18n/en-US.yml
@@ -381,7 +381,7 @@ components:
     selectCost: Cost
     selectDepartureTime: Departure time
     selectDuration: Duration
-    selectEmissions: CO2 emissions
+    selectEmissions: CO₂ emissions
     selectFare: Transit fare
     selectWalkTime: Walk time
     sortResults: Sort results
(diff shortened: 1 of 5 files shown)


i18n/en-US.yml
```diff
@@ -381,6 +381,7 @@ components:
     selectCost: Cost
     selectDepartureTime: Departure time
     selectDuration: Duration
+    selectEmissions: CO2 emissions
```
binh-dam-ibigroup:
```suggestion
    selectEmissions: CO₂ emissions
```

lib/components/util/sortOptions.ts
```diff
@@ -14,6 +14,7 @@ export const sortOptions = (
     'COST',
     'DEPARTURETIME',
     'FARE'
+    // Emissions is disabled by default, as it does nothing unless co2 calculation is enabled
```
binh-dam-ibigroup:
```suggestion
    // Emissions is disabled by default, as it does nothing unless CO₂ calculation is enabled
```

binh-dam-ibigroup (approved):
Nice one. I prefer we use the subscript₂.

alec-georgoff (approved):
🍃

PR #1545, opened 2026-03-13
Render Multipolygon Flex Zones

binh-dam-ibigroup:
**Description:**

Requires https://github.com/opentripplanner/otp-ui/pull/933.
This PR enables rendering multipolygon flex zones.

There is an undesired type check error that is addressed by https://github.com/opentripplanner/otp-ui/pull/935.

**PR Checklist:**


lib/components/map/connected-stops-overlay.ts
```diff
@@ -36,7 +36,8 @@ const mapStateToProps = (state: AppReduxState) => {
       // However, without changes to GraphQL, getting this data is very expensive
       Object.values(patterns).forEach((p) => {
         p?.stops
-          ?.filter((s: Stop) => s.geometries?.geoJson?.type === 'Polygon')
+          // @ts-expect-error isJsonFlex logic does support undefined arg.
```
miles-grant-ibigroup: Should we wait for that otp-ui patch instead of doing this?

lib/actions/apiV2.js
```diff
@@ -907,7 +912,17 @@ export function findStopsWithinBBox() {
         {
           appendBounds: true,
           noThrottle: true,
-          rewritePayload: (payload) => payload.data,
+          rewritePayload: (payload) => {
+            const mappedStops = payload.data.stops.map((s) =>
```
miles-grant-ibigroup, line 917: Why is this needed? Could you add a comment?

miles-grant-ibigroup (changes requested):
Quick questions

lib/actions/apiV2.js
binh-dam-ibigroup: ~You're right. I thought that would fix some issue while loading routes during dev, but that wasn't that! https://github.com/opentripplanner/otp-react-redux/pull/1545/commits/5dbd4738a0bc182e7eba0409d230454fe1d217bd~

lib/components/map/connected-stops-overlay.ts
binh-dam-ibigroup: Sure

lib/actions/apiV2.js
binh-dam-ibigroup: > Why is this needed? Could you add a comment?

Taking this back. I added the comment.

miles-grant-ibigroup (approved):
I'm honestly confused how this ever worked before now. Thanks for the fix!

alec-georgoff (approved):

PR #1547, opened 2026-03-18
remove superfluous config item

miles-grant-ibigroup:
The way we were handling this config item was silly... the extra item was confusing! It's likely that anyone setting specific valid mode combinations also wants to ensure that they contain at least one requested mode


amy-corson-ibigroup (changes requested):
Wait can you clarify in the description why you're making this change? Are you thinking that if someone is specifying valid mode combinations in the config, we can assume they probably want to also make sure they're returning itineraries that contain at least one explicitly requested mode? Bc that makes sense to me, but as of now these config items are not redundant, they do different things! 

If we do decide to consolidate these into one variable, we should probably combine `strictModes` and `validModeCombinations` into one variable and the two `if` blocks on lines 1237 and 1246 into one.

miles-grant-ibigroup:
That's a great catch thank you! I will update. Much cleaner this way

amy-corson-ibigroup (approved):
Looks good thank you for the changes!

RIP `strictItineraryFiltering` Miles can call you silly but history will remember you as a fix ahead of your time 🕊️🕯️🕊️🕯️🕊️

binh-dam-ibigroup (approved):

PR #1548, opened 2026-03-20
Include forked patterns and fix subpattern filtering

binh-dam-ibigroup:
**Description:**

This PR addresses an issue where of two pattern forks with the same headsigns traveling to the same destination, only one would be shown. ~The PR also hides subpatterns that were previously visible.~

```
o Origin 1
|
o    o Origin 2
|    |
o----/
|
o Destination
```

**PR Checklist:**

| Before | After |
|--------|-------|
| before - atl | after - atl |
| before - st | after - st |


lib/util/pattern-viewer.ts
```diff
@@ -28,51 +29,64 @@ interface PatternWithStops extends Pattern {
...
+      const twoPatternsWithOneAmended =
+        amended.length === 1 && mapped.length === 2
+      // If the last stop is different than the headsign but the same as the last stop
+      // of the previously existing duplicate, there's no point in renaming.
+      if (
+        cur.lastStop &&
+        cur.headsign !== cur.lastStop &&
+        cur.lastStop !== existing.lastStop
+      ) {
+        editToHeadsign(cur)
```
binh-dam-ibigroup: Any suggestions to refactor L63-69 and L71-78?

lib/util/pattern-viewer.ts
```diff
@@ -104,44 +118,46 @@ export function getParentStopOrStopId(stop: StopWithParent): string {
...
-
-      return includePattern
+      }
     })
+  })
+
+  // Keep patterns that are not subsets of larger patterns.
+  const filteredPatterns = sortedPatterns.filter(
+    (pattern) => !containingPatterns[pattern.id]
+  )
```
amy-corson-ibigroup: I think any pattern with an original headsign should not be considered a subset of a larger pattern - only in the case of duplicate headsigns.

lib/util/pattern-viewer.ts
amy-corson-ibigroup: Maybe this is making L63-69 and L71-78 more complicated and worse but maybe we could pull the contents out?

```
const editHeadsignName = (from: boolean) =>  from ? editFromHeadsign(pattern) : editToHeadsign(pattern)

function editHeadsign(from: boolean, twoPatternsWithOneAmended: boolean, amended) {
       // Append 'from' or 'to' + the stop name if the patterns are nearly identical but contain a fork.
       editHeadsignName(from)
       // If there are only two total patterns, then we should rename both of them
       if (twoPatternsWithOneAmended) {
          editHeadsignName(from)
          amended.push(cur)
          return amended
        }
```

amy-corson-ibigroup (commented):
Thank you for this PR and apologies for how long it took me to review! 

Editing duplicate headsigns to include the "from" stop in case of forks is a really smart change. However, I think filtering out subpatterns should _only_ happen in the case of duplicate headsigns. Using the ST 1-Line example: Northgate and Beacon Hill are two valid patterns with valid headsigns. Filtering them out may only cause end-user confusion.

lib/util/pattern-viewer.ts
binh-dam-ibigroup: Addressed in https://github.com/opentripplanner/otp-react-redux/pull/1548/commits/1a7b6ae8ccbd815ebab9b0a7e24babffea1918a1.

lib/util/pattern-viewer.ts
binh-dam-ibigroup: Found a slightly better way: https://github.com/opentripplanner/otp-react-redux/pull/1548/commits/f70c11a904c8790cc73ec0ff2bee2ab7c0bee07e

__tests__/util/pattern-viewer.ts
```diff
@@ -175,15 +210,39 @@ describe('util > pattern-viewer', () => {
...
       expect(filteredPatterns).toContain(patterns[3])
-      expect(['P1', 'P5']).toContain(containingPatterns.P3)
-      expect(containingPatterns.P4).toBe(undefined)
-      expect(
-        containingPatterns.P1 === 'P5' || containingPatterns.P5 === 'P1'
-      ).toBeTruthy()
+      expect(filteredPatterns).toContain(patterns[5])
+      expect(filteredPatterns).toContain(patterns[6])
+      expect(filteredPatterns).toContain(patterns[7])
+      expect(filteredPatterns).toContain(patterns[8])
```
amy-corson-ibigroup: Can we condense these into one test by using `arrayContaining`?

__tests__/util/pattern-viewer.ts
```diff
@@ -175,15 +210,39 @@ describe('util > pattern-viewer', () => {
...
+      // P3:   o--o--o--o--o
+      //
+      const patterns: Pattern[] = [
+        createPattern('P1', 1404, ['S1', 'S2', 'S3', 'S4', 'S5']),
+        createPattern('P2', 1404, ['S1', 'S2', 'S3', 'S4', 'S5']),
+        createPattern('P3', 1404, ['S1', 'S2', 'S3', 'S4', 'S5'])
+      ]
+      const { containingPatterns, filteredPatterns } =
+        sortAndRemoveSubpatterns(patterns)
+      console.log(filteredPatterns, containingPatterns)
```
amy-corson-ibigroup: Is this console.log here for a reason?

lib/actions/apiV2.js
```diff
@@ -836,6 +836,7 @@ export const findRoute = (params) =>
             // Sort and remove all patterns that are subsets of larger patterns.
             const { containingPatterns, filteredPatterns } =
               sortAndRemoveSubpatterns(newRoute.patterns)
+
```
amy-corson-ibigroup: New line can probably be removed here

amy-corson-ibigroup (approved):
Looks good to me, just a few suggestions to clean up code.

Thank you!

__tests__/util/pattern-viewer.ts
binh-dam-ibigroup: nope, good catch!

__tests__/util/pattern-viewer.ts
binh-dam-ibigroup: I couldn't have `arrayContaining` work for some reason, however `expect.toEquals` seems to work (https://github.com/opentripplanner/otp-react-redux/pull/1548/commits/8c91c4dca08b3db3cfc54c9944df64586d523f35)

lib/components/viewers/route-details.tsx
```diff
@@ -87,9 +87,22 @@ class RouteDetails extends Component<Props> {
...
-  _editHeadsign = (pattern: PatternSummary) => {
+  _editToHeadsign = (pattern: PatternSummary) => {
     pattern.headsign = this.props.intl.formatMessage(
-      { id: 'components.RouteDetails.headsignTo' },
+      {
+        defaultMessage: '{headsign} ({lastStop})',
+        id: 'components.RouteDetails.headsignTo'
+      },
+      { ...pattern }
+    ) as string
```
daniel-heppner-ibigroup, line 101: just curious, what happens if you use `satisfies string` here?

lib/components/util/types.ts
```diff
@@ -26,7 +26,7 @@ export interface StopTime {
 
 export interface Pattern {
   desc: string
-  headsign: string
+  headsign: string | null
```
daniel-heppner-ibigroup: If you use `undefined` as the nullish type here then you can avoid using `|| undefined` in pattern-row

lib/util/pattern-viewer.ts
```diff
@@ -25,54 +26,70 @@ interface PatternWithStops extends Pattern {
...
+        cur.lastStop !== existing.lastStop
+      ) {
+        updateHeadsign = editToHeadsign
+      } else if (cur.firstStop !== existing.firstStop) {
+        // Append 'from' + the first stop name if the patterns have the exact same arrival stops but different origins.
+        updateHeadsign = editFromHeadsign
+      }
+      if (updateHeadsign) {
+        // Update headsign if conditions apply.
+        updateHeadsign(cur)
```
daniel-heppner-ibigroup: it might be good to refactor this function to be "pure" and return back the accumulator instead of having it operate on the pattern

lib/util/pattern-viewer.ts
```diff
@@ -104,49 +121,68 @@ export function getParentStopOrStopId(stop: StopWithParent): string {
...
-    })
+  // Keep patterns that are not subsets of larger patterns or, if they are,
+  // have a different headsign from the containing pattern, or if the pattern and immediate coontaining pattern headsigns are null.
+  const filteredPatterns = sortedPatterns.filter((pattern) => {
+    const containingPatternId = immediateContainingPatterns[pattern.id]
+    const containingPattern = patterns.find((p) => p.id === containingPatternId)
+    return (
+      !containingPattern ||
+      containingPattern.headsign !== pattern.headsign ||
+      (!containingPattern.headsign && !pattern.headsign)
```
daniel-heppner-ibigroup: I think you can simplify the last line of this to just `!containingPattern.headsign` because if only one is undefined then the previous line will be truthy.

daniel-heppner-ibigroup (approved):
the undefined/null war strikes again
I haven't looked into it in the code base so maybe replacing that | null with undefined isn't feasible. 

Only thing here I'm pretty sure can definitely be improved is my comment about the binary expression

daniel-heppner-ibigroup (changes requested):
wrong button

lib/components/viewers/route-details.tsx
binh-dam-ibigroup: lol the TS version we're using doesn't support that syntax!

lib/components/util/types.ts
binh-dam-ibigroup: Nice catch! (https://github.com/opentripplanner/otp-react-redux/pull/1548/commits/7660efee70ff4cde67cdb834a7d16b53f2738000)

lib/util/pattern-viewer.ts
binh-dam-ibigroup: Good one, updated in https://github.com/opentripplanner/otp-react-redux/pull/1548/commits/92880a9a146c8408730230072fb5e0a16004f034.

lib/util/pattern-viewer.ts
binh-dam-ibigroup: What do you have in mind for this? `cur.headsign = updateHeadsign(cur)` or something more involved?

daniel-heppner-ibigroup: basically that yeah. I'll approve this now and leave it up to you if you want to do that. Then you can control the types better on the input and output side.

daniel-heppner-ibigroup (approved):

lib/util/pattern-viewer.ts
binh-dam-ibigroup: Refactored in https://github.com/opentripplanner/otp-react-redux/pull/1548/commits/32281dced452cf7555f1a5de24d74164211b7d7c.

lib/components/util/types.ts
binh-dam-ibigroup: I take this one back and reverted the change you suggested because MARTA sends explicit `null` headsigns.

__tests__/util/pattern-viewer.ts
```diff
@@ -16,11 +16,13 @@ function createStops(ids: string[]): Stop[] {
   }))
 }
 
-function editHeadsign(pattern: PatternSummary) {
-  pattern.headsign = `${pattern.headsign} (${pattern.lastStop})`
-}
+const editToHeadsign: HeadsignGenerator = (pattern) =>
+  `${pattern.headsign} (${pattern.lastStop})`
+
+const editFromHeadsign: HeadsignGenerator = (pattern) =>
```
daniel-heppner-ibigroup, line 22: can you include a type for `pattern`?

__tests__/util/pattern-viewer.ts
```diff
@@ -16,11 +16,13 @@ function createStops(ids: string[]): Stop[] {
...
+  `${pattern.headsign} (from ${pattern.firstStop})`
 
-const headsign = 'Everett via Lynnwood'
+const defaultHeadsign = 'Everett via Lynnwood'
```
daniel-heppner-ibigroup: can you write this as DEFAULT_HEADSIGN to indicate that it is a defined constant?

__tests__/util/pattern-viewer.ts
```diff
@@ -31,140 +33,170 @@ const route: Route = {
   sortOrderSet: false
 }
 
+function createPattern(
+  id: string,
+  length: number,
+  stops: string[],
+  headsign: string | null = defaultHeadsign
```
daniel-heppner-ibigroup: use `headsign?: string | null` (add `?`)

daniel-heppner-ibigroup (approved):
only comments are on tests, feel free to fix or not

__tests__/util/pattern-viewer.ts
binh-dam-ibigroup: Unlike type definitions, `?` is not accepted for function parameters.

__tests__/util/pattern-viewer.ts
binh-dam-ibigroup: The type of the parameters is already implied by specifying the `HeadsignGenerator` type to the functions.

PR #1554, opened 2026-04-02
Combine trip and pattern viewer

amy-corson-ibigroup:
**Description:**

- Typescript Trip Viewer
- Create `StopList` component that can be applied to both trip and pattern viewers. 
- Re-use `routeRendererer` in the trip viewer header to emulate the site design in route viewer

**PR Checklist:**

Trip Viewer
| Before | After | Mobile |
|--------|-------|-------|
|image|image|image|

Route Viewer
| Before | After |
|--------|-------|
|image|image|


lib/components/viewers/stop-list.tsx
```diff
@@ -0,0 +1,244 @@
...
+  const tripViewerHighLighter =
+    fromIndex !== undefined &&
+    fromIndex > -1 &&
+    toIndex !== undefined &&
+    toIndex > -1
+
+  // If we're in the trip viewer and there's a highlighted section of the trip, scroll that section into view.
+  const firstStopRef = useRef<HTMLLIElement>(null)
+  useEffect(() => {
+    firstStopRef?.current?.scrollIntoView({
```
daniel-heppner-ibigroup, line 156: this is actually an effect! perfect love it

lib/components/viewers/stop-list.tsx
```diff
@@ -0,0 +1,244 @@
...
+  routeColor = DEFAULT_ROUTE_COLOR,
+  routePattern,
+  setHoveredStop,
+  stopLinkClicked,
+  textColor,
+  toIndex,
+  useRouteColorAsBackground
+}: StopListProps): JSX.Element => {
+  // The stops in the pattern viewer vs the trip viewer are organized slightly differently, so account for that:
+  const stopsArray =
```
daniel-heppner-ibigroup: I think we have a type for the Stop and maybe even the routePattern don't we? Can you see if you can find a type that works here. that would allow you to get rid of a whole bunch of the `any`s here

lib/components/viewers/stop-list.tsx
```diff
@@ -0,0 +1,244 @@
...
+      textColor={textColor}
+      timeColumn={stopsArray[0].scheduledDeparture}
+    >
+      {stopsArray?.map((stop: any, index: number) => {
+        const highlighted = tripViewerHighLighter
+          ? index >= fromIndex && index <= toIndex
+          : true
+
+        // Helpful invisible labels for screenreaders
+        let stopLabel = null
```
daniel-heppner-ibigroup: can you type this? I think string | null. Or just use string | undefined and then replace this with `let stopLabel: string | undefined;`

lib/components/viewers/styled.ts
```diff
@@ -2,7 +2,7 @@ import styled, { css } from 'styled-components'
 
 import { getBaseColor, grey } from '../util/colors'
 
-interface RenderProps {
+export interface RenderProps {
```
daniel-heppner-ibigroup: if RenderProps isn't used in this file anymore maybe move it out

lib/components/viewers/trip-viewer.tsx
```diff
@@ -248,96 +288,25 @@ class TripViewer extends Component {
...
+          )}
         </div>
       </div>
     )
   }
 }
 
-const mapStateToProps = (state) => {
-  const activeSearch = getActiveSearch(state)
+const mapStateToProps = (state: any) => {
```
daniel-heppner-ibigroup: you can use AppState type!

lib/components/viewers/trip-viewer.tsx
```diff
@@ -248,96 +288,25 @@ class TripViewer extends Component {
...
+  const activeSearch: any = getActiveSearch(state)
```
daniel-heppner-ibigroup: and remove this any once you do the above

daniel-heppner-ibigroup (changes requested):
initial typescript pass

lib/components/viewers/stop-list.tsx
```diff
@@ -0,0 +1,244 @@
...
+      opacity: ${lowOpacity};
+    }
+  }
+`
+
+export const StopLink = styled.button<RenderProps>`
+  background-color: transparent;
+  border: none;
+  // Should this just be black to better contrast with the #666 in the faded class?
+  color: ${(props) => props?.textColor + 'da'};
```
daniel-heppner-ibigroup: the da can just go outside the template string {} thing

lib/components/viewers/stop-list.tsx
```diff
@@ -0,0 +1,244 @@
...
+`
+
+interface StopListProps {
+  backgroundColor?: string
+  fromIndex?: number
+  homeTimezone?: any
+  intl: IntlShape
+  routeColor?: string
+  routePattern: any
+  setHoveredStop?: (arg: any) => void
```
daniel-heppner-ibigroup: see setHoveredStop type from route-details.tsx

daniel-heppner-ibigroup (changes requested):
more TS comments

lib/components/viewers/trip-viewer.tsx
amy-corson-ibigroup: Removing this threw some errors on the `pending` that came below this, so I took that out? Seems like we don't need it anymore but I may be wrong

lib/components/viewers/pattern-viewer.tsx
```diff
@@ -81,23 +77,11 @@ const PatternViewer = ({
   if (patternId && route) {
     // Find operator based on agency_id (extracted from OTP route ID).
     const operator = getRouteOperator(route, transitOperators)
-    const { backgroundColor, textColor } = getPatternViewerColors(
-      useRouteColorAsBackground,
-      operator,
-      route
-    )
-    const fill = vehicleIconHighlight === false ? undefined : textColor
+    const fill = vehicleIconHighlight === false ? undefined : DARK_TEXT_GREY
```
daniel-heppner-ibigroup: instead of `=== false` can you do `vehicleIconHighlight ? DARK_TEXT_GREY : undefined`

daniel-heppner-ibigroup: I am seeing now that this line is actually old but the diff is marking it anyway...

lib/components/viewers/trip-viewer.tsx
```diff
@@ -140,31 +141,62 @@ class TripViewer extends Component {
     ]
   }
 
+  _getRouteColor = (operator: TransitOperator) => {
+    const { tripData } = this.props
+    const routeColor =
+      tripData?.route &&
+      ((tripData?.route?.color && `#${tripData?.route?.color}`) ||
+        getRouteColorBasedOnSettings(tripData.route, operator))
```
daniel-heppner-ibigroup: this is really hard to parse, is there any way you could break it up a little bit more so there aren't so many layers in one statement?

lib/components/viewers/trip-viewer.tsx
```diff
@@ -140,31 +141,62 @@ class TripViewer extends Component {
...
+    return routeColor
+  }
+
+  _getOperator = () => {
+    const { transitOperators, tripData } = this.props
+    const operator =
+      (tripData?.route &&
```
daniel-heppner-ibigroup: same here. one thing you could do is add an `if (!tripData?.route) return {}` then strip off that layer of the statement

daniel-heppner-ibigroup (changes requested):
a few more...

lib/components/viewers/trip-viewer.tsx
```diff
@@ -140,31 +141,62 @@ class TripViewer extends Component {
...
+        coreUtils?.route?.getTransitOperatorFromOtpRoute(
+          tripData?.route,
+          transitOperators as TransitOperator[]
```
daniel-heppner-ibigroup: uh oh, I don't think casting this is a good call because that is overriding what typescript thinks this type should be. Let's have a call about this

daniel-heppner-ibigroup (changes requested):
i just saw the casting 😭😭😭

lib/components/viewers/route-details.tsx
```diff
@@ -145,7 +138,7 @@ class RouteDetails extends Component<Props> {
       patternSelectLabel
 
     return (
-      <Container backgroundColor={backgroundColor} full={pattern != null}>
+      <Container full={pattern != null}>
```
daniel-heppner-ibigroup: Can you replace the != null with a `!== undefined`?

lib/components/viewers/stop-list.tsx
```diff
@@ -0,0 +1,231 @@
...
+                style={{ opacity: highlighted ? '100%' : '0%' }}
+              >
+                <DepartureTime originDate={startOfDay} stopTime={stop} />
+              </div>
+            )}
+            <div className="stop-decoration" />
+            <StopLink
+              name={stop.name}
+              onFocus={
+                () => setHoveredStop && setHoveredStop(stop.stopId || null)
```
daniel-heppner-ibigroup: you might want to consider `??` instead of `||` because `||` has some weird behavior around "0" and 0, which are valid stop IDs.

daniel-heppner-ibigroup (approved):
This looks really good!! Thank you for working with me on this. I feel like this is really improving the health of the codebase.

lib/components/viewers/route-details.tsx
```diff
@@ -118,19 +116,14 @@ class RouteDetails extends Component<Props> {
...
-      sortPatternsByVehicleCount,
-      useRouteColorAsBackground
+      sortPatternsByVehicleCount
     } = this.props
     const { agency, patterns = {}, shortName, url } = route
     const pattern = patterns[patternId]
 
     const moreDetailsURL = url || route?.agency?.url
 
-    const { backgroundColor, routeColor, textColor } = getPatternViewerColors(
```
miles-grant-ibigroup, line 129: Can we restore using `getPatternViewerColors`? Even if we remove colored background support, we should still keep the color softening support? Also isn't this how the mode color overrides are set up? Or do we do that elsewhere now. Would prefer to do it elsewhere/remove the softened support

lib/components/viewers/stop-list.tsx
```diff
@@ -0,0 +1,231 @@
...
+import { StopListEntry } from '../util/types'
+import InvisibleA11yLabel from '../util/invisible-a11y-label'
+
+import DepartureTime from './departure-time'
+
+const { time } = coreUtils
+const { getCurrentDate } = time
+
+const timeCellWidth = '45px'
+const lowOpacity = '40%'
```
miles-grant-ibigroup: If we have consts like this can we make them all caps?

lib/components/viewers/stop-list.tsx
```diff
@@ -0,0 +1,231 @@
...
+  are shown when browsers don't calculate 100% sensibly. */
+  height: calc(100% - 140px);
+  overflow-y: ${(props) => (props.timeColumn ? 'visible' : 'scroll')};
+  padding: 5px 10px;
+
+  .highlighted {
+    opacity: 100%;
+
+    div.stop-decoration {
+      box-shadow: 2px 2px 5px 1px rgba(0, 0, 0, 0.1);
```
miles-grant-ibigroup: Can we remove this? Isn't it a bit out of place to have an elevated stop blob but flat text?

lib/components/viewers/trip-viewer.tsx
```diff
@@ -1,79 +1,85 @@
...
+  geometry: { length: number; points: string }
+  id: string
+  route: Route
+  serviceId: string
+  shapeId: string
+  stopTimes: TripStopTime[]
+  tripHeadsign: string
+  wheelchairAccessible: string
+}
+
```
miles-grant-ibigroup, line 64: Doesn't this live in types somewhere?

lib/components/viewers/trip-viewer.tsx
```diff
@@ -140,31 +141,68 @@ class TripViewer extends Component {
     ]
   }
 
+  _getRouteColor = (operator?: TransitOperator) => {
+    const { tripData } = this.props
+    console.log(tripData)
```
miles-grant-ibigroup: Whoops!

lib/components/viewers/trip-viewer.tsx
```diff
@@ -140,31 +141,68 @@ class TripViewer extends Component {
...
+      tripData?.route?.color && `#${tripData?.route?.color}`
+    const routeColorBasedOnSettings = getRouteColorBasedOnSettings(
+      tripData?.route,
+      operator
+    )
+
+    const routeColor = routeColorFromTripData || routeColorBasedOnSettings
+
+    return routeColor
+  }
```
miles-grant-ibigroup, line 151: Could we remove this method and instead run it as one line? 

`const routeColor = `tripData?.route?.color ? `#${tripData?.route?.color}` : getRouteColorBasedOnSettings(tripData?.route, operator)`

lib/components/viewers/trip-viewer.tsx
```diff
@@ -1,79 +1,85 @@
...
-import Strong from '../util/strong-text'
 
-import DepartureTime from './departure-time'
-import ViewStopButton from './view-stop-button'
-
-const { getCurrentDate } = coreUtils.time
+import { RouteRowDetails } from './route-row'
+import StopList from './stop-list'
 
 const AlertContainer = styled.div`
```
miles-grant-ibigroup, line 35: Is there some way we can hide the bell?

miles-grant-ibigroup (changes requested):
I really like these changes! This should really help with maintainability. But I think we can clean this up to result in a net code line reduction.

lib/components/viewers/trip-viewer.tsx
```diff
@@ -182,21 +220,24 @@ class TripViewer extends Component {
...
+                  isMobile()
+                    ? { marginTop: '-0.5em', padding: '3px' }
+                    : { marginTop: '1em' }
+                }
+              >
+                <RouteRowDetails
+                  intl={intl}
+                  ModeIcon={ModeIcon}
+                  operator={operator}
+                  route={tripData?.route}
```
miles-grant-ibigroup, line 228: Is this a full route object? This is crashing in come cases where the route renderer relies on things like the longName.

amy-corson-ibigroup: Yeah it should be! Do you want to send me an example of the crash and I can take a look?

lib/components/viewers/trip-viewer.tsx
amy-corson-ibigroup: nice change!

lib/components/viewers/stop-list.tsx
amy-corson-ibigroup: I don't think so! Plus it serves to further distinguish the trip stops from the non-trip stops in the trip viewer: 
image

lib/components/viewers/trip-viewer.tsx
amy-corson-ibigroup: Just spent some time looking and didn't see anything but lmk if I'm missing something!

lib/components/viewers/trip-viewer.tsx
amy-corson-ibigroup: Oof okay not at this moment but it has been requested before. If you're okay with the bell as-is for now, I can adjust the `Alert` in bb and apply the new no-bell to all requested alerts as a follow up to this PR.

lib/components/viewers/styled.ts
```diff
@@ -23,23 +13,18 @@ export const RouteNameContainer = styled.div`
...
   gap: 10px;
   padding: 15px 10px;
   margin-top: -10px;
 
   a {
-    color: ${(props) => props.textColor};
+    color: ${DARK_TEXT_GREY};
     svg {
-      color: ${(props) =>
-        props.useRouteBgColor ? props.textColor : getBaseColor()};
```
miles-grant-ibigroup, line 42: Why aren't we using `getBaseColor()` anymore?

lib/util/viewer.js
```diff
@@ -317,41 +317,19 @@ function getCleanRouteLongName({ longNameSplitter, route }) {
...
   }`
-  // NOTE: text color is not a part of short response route object, so there
-  // is no way to determine from OTP what the text color should be if the
-  // background color is, say, black. Instead, determine the appropriate
-  // contrast color and use that if no text color is available.
-  const contrastColor = getMostReadableTextColor(
-    backgroundColor,
-    modeColors?.textColor || route?.textColor
-  )
-  const color = `#${defaultRouteTextColor || contrastColor}`
```
miles-grant-ibigroup, line 336: Just to be clear, this is now duplicated in 2 places, right? I think that's OK because I don't see it changing anytime soon, but just want to make sure this wasn't an oversight

lib/components/viewers/stop-list.tsx
```diff
@@ -0,0 +1,231 @@
...
```
miles-grant-ibigroup: Can we get rid of this shadow? It doesn't match any of the other styling in the app, and I believe it's causing this artifacting

Image

miles-grant-ibigroup (changes requested):
Always love to see a diff almost equal `-` and `+`!

Once the css is cleaned up I think this will be good to go!

lib/components/viewers/stop-list.tsx
miles-grant-ibigroup: We have no shadows anywhere else in the app! If you think the color distinction is not enough we could maybe go for a large pill that covers all active stops (similar to how the old trip viewer had it) except with a nicer color. But I think the blur introduces a new design language that is jarring

lib/components/viewers/trip-viewer.tsx
miles-grant-ibigroup: I don't think this can get merged as-is... The padding between each element is so wildly different. Without the bell it still won't be perfect, but I think it'll be acceptable

lib/components/viewers/trip-viewer.tsx
miles-grant-ibigroup: Don't know what happened but in your refactors this has been fixed! It's now a full route object

lib/components/viewers/stop-list.tsx
miles-grant-ibigroup: Sorry I'm seeing now I already asked for this last time...

amy-corson-ibigroup: Oh sorry, this box shadow is actually unrelated so i got confused!! The issue was caused by two low-opacity colors overlapping. I solved it by using `color-mix` to create a "low-opacity` looking color that is opaque! 
image

lib/components/viewers/styled.ts
amy-corson-ibigroup: Wait you're totally right good catch

lib/util/viewer.js
amy-corson-ibigroup: Can you clarify what you mean here?

lib/components/viewers/stop-list.tsx
miles-grant-ibigroup: Wow I didn't know about color-mix! That's really cool. With this when the shadows look good they look good (I like how it really highlights the active section in the trip viewer), but still out of place! We need to find a way of highlighting this without using these types of shadows. Personally I think just the fade is enough to differentiate (and is cleaner) but I'd be willing to compromise and find a second way to differentiate! Maybe some kind of large blob like the current trip viewer? Maybe a highlight under the text?

Here is a good example of why the shadow just doesn't work:
Screenshot 2026-07-13 at 11 53 18 AM

lib/util/viewer.js
miles-grant-ibigroup: This used to be a method, now it's been inlined (twice). So the code is duplicated! Was this intentional

miles-grant-ibigroup (changes requested):
The shadows and the bell are the two blockers for me. There's some regressions around the label padding but we can fix that in a future PR

lib/util/viewer.js
amy-corson-ibigroup: I feel like you're commenting on code that's been deleted so i dont know what being duplicated! Can you point to the lines you're referring to?

lib/util/viewer.js
```diff
@@ -317,41 +317,19 @@ function getCleanRouteLongName({ longNameSplitter, route }) {
...
-  const { defaultRouteColor, defaultRouteTextColor, longNameSplitter } =
-    operator || {}
+  const { defaultRouteColor, longNameSplitter } = operator || {}
 
   const modeColors = operator?.modeColors?.[getModeFromRoute(route)]
 
-  const backgroundColor = `#${
+  const routeColor = `#${
     modeColors?.color || defaultRouteColor || route?.color || '333333'
   }`
```
miles-grant-ibigroup, line 326: This part

lib/util/viewer.js
```diff
@@ -360,40 +338,15 @@ export function getColorAndNameFromRoute(operator, route) {
...
-  operator,
-  route
-) => {
-  const routeColor = getRouteColorBasedOnSettings(operator, route)
-  const backgroundColor = useRouteColorAsBackground ? routeColor : '#fff'
-  const textColor = useRouteColorAsBackground
-    ? getMostReadableTextColor(backgroundColor, route?.textColor)
-    : DARK_TEXT_GREY
-  return { backgroundColor, routeColor, textColor }
+  return routeColor
```
miles-grant-ibigroup, line 349: And this part!

miles-grant-ibigroup (approved):
Here's the 2 segments I was talking about, I don't know what I was thinking with the earlier code segment.

I think it's alright for now though. I worry a bit about maintainability with the multiple ways of reading routeColor, but if we slap a TODO on there I think this is good to go!

lib/util/viewer.js
amy-corson-ibigroup: I think these are okay actually! The `routeColor` returned on line 324 is the default, and the one returned on 349 is the function we pass the default `routeColor` through if the `colorMode` in the config is `disabled`. Might be worth asking if anyone uses the `disabled` color mode tho

PR #1557, opened 2026-04-06
Recheck trip existence

binh-dam-ibigroup:
**Description:**

This PR adds a button to the trip editor, so that users can manually recheck trip existence if what is shown is incorrect and add days to monitor if those days were not previously monitorable.

Note: With this PR alone, the updated itinerary existence will not be updated server-side. OTP-middleware currently ignores updates to `MonitoredTrip.itineraryExistence`, so when the user refreshes the trip, the unavailable days remain.

**PR Checklist:**

| Before | After |
|--------|-------|
| before screenshot | After screen recording |

### DIFF
--- a/lib/components/user/monitored-trip/trip-basics-pane.tsx
+++ b/lib/components/user/monitored-trip/trip-basics-pane.tsx
@@ -1,12 +1,11 @@
+import { connect } from 'react-redux'
 import {
-  Button,
   ControlLabel,
   FormControl,
   FormGroup,
   HelpBlock,
   Radio
 } from 'react-bootstrap'
-import { connect } from 'react-redux'
 import { Field, FormikProps } from 'formik'
 import { FormattedMessage, injectIntl } from 'react-intl'
 import { Prompt } from 'react-router'
--- a/lib/components/user/monitored-trip/trip-basics-pane.tsx
+++ b/lib/components/user/monitored-trip/trip-basics-pane.tsx
@@ -401,6 +401,25 @@ class TripBasicsPane extends Component<TripBasicsProps, State> {
       /* Hack: because the selected days checkboxes are not grouped, we need to assign this error to one of the 
       checkboxes so that the FormikErrorFocus works. */
       const selectOneDayError = errorStates.monday
+      const dayButtons = (
+        <>
+          <RenderAvailableDays
+            errorCheckingTrip={errorCheckingTrip}
+            errorSelectingDays={
+              disableSingleItineraryDays ? selectOneDayError : undefined
+            }
+            finalItineraryExistence={finalItineraryExistence}
+            isReadOnly={isReadOnly}
+            monitoredTrip={monitoredTrip}
+          />
+          {!isCreating && (
+            <Button onClick={this._handleRecheckItineraryExistence}>
+              Check again
+            </Button>
+          )}
+        </>
+      )
+
       return (
         <div>
           {/* TODO: This component does not block navigation on reload or using the back button.
@@ -448,18 +467,7 @@ class TripBasicsPane extends Component<TripBasicsProps, State> {
                 <FormattedMessage id="components.TripBasicsPane.tripDaysPrompt" />
                 {!isReadOnly && <RequiredIndicator>*</RequiredIndicator>}
               </ControlLabel>
-              <RenderAvailableDays
-                errorCheckingTrip={errorCheckingTrip}
-                errorSelectingDays={selectOneDayError}
-                finalItineraryExistence={finalItineraryExistence}
-                isReadOnly={isReadOnly}
-                monitoredTrip={monitoredTrip}
-              />
-              {!isCreating && (
-                <Button onClick={this._handleRecheckItineraryExistence}>
-                  Check again
-                </Button>
-              )}
+              {dayButtons}
             </FormGroup>
           ) : (
             <FormGroup>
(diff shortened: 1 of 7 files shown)


alec-georgoff (approved):
Nice!

lib/components/user/monitored-trip/trip-basics-pane.tsx
```diff
@@ -391,6 +204,32 @@ class TripBasicsPane extends Component<TripBasicsProps, State> {
...
+            isReadOnly={isReadOnly}
+            monitoredTrip={monitoredTrip}
+          />
+          {!isCreating && (
+            <Button
+              disabled={isRecheckingExistence}
+              onClick={this._handleRecheckItineraryExistence}
+            >
+              {isRecheckingExistence ? (
+                <InlineLoading />
```
amy-corson-ibigroup, line 229: Right now there's no feedback for AT users when this button is pressed. Could you use `isRecheckingExistence` to trigger the status region in `HelpBlock` for this check? And could we give this button an `aria-label` and change the label to "Rechecking" for screen readers when the button is pressed? They're not going to be able to see the "loading" icon.

amy-corson-ibigroup (commented):
The code looks good, and I really like the functionality. I'm wondering if you'd be open to exploring other button designs? Right now the button is very prominent and feels like an important part of saving a trip, rather than a fallback option a user only needs if something appears incorrect. 

My suggestion would be we restyle it to appear more like a link, and add it either at the end of the line, or just below the text. There are other examples of this styling around the site, such as the "resend code" button in the phone number editor, but I'm not married to this idea if you have another one you'd prefer! 

Image

Image

lib/components/user/monitored-trip/trip-basics-pane.tsx
binh-dam-ibigroup: AT feedback already exists in the status content of the progress bar.

binh-dam-ibigroup:
> My suggestion would be we restyle it to appear more like a link, and add it either at the end of the line, or just below the text.

I thought about it. How about the same style as "Snooze" in the trip notification box?

lib/components/user/monitored-trip/trip-basics-pane.tsx
amy-corson-ibigroup: That's interesting, you're right that it should be working.  think there may be an issue with how the `ProgressBar` is interacting with the `aria-live` region. The "checking itinerary existence" string isn't being read to AT users when I test on VO + Safari

amy-corson-ibigroup: I am hearing the "Trip is available on days indicated" message though.

binh-dam-ibigroup:
Updated style (https://github.com/opentripplanner/otp-react-redux/pull/1557/commits/c283c484643cea15313b0878f9b0c970b7952110): 
image

lib/components/user/monitored-trip/trip-basics-pane.tsx
binh-dam-ibigroup: You were right @amy-corson-ibigroup, the progress bar was masking the content to Voice Over. With https://github.com/opentripplanner/otp-react-redux/pull/1557/commits/92ff6fb75756a26cea93164e62a956687bfa7436, that text is now read during recheck.

lib/components/user/monitored-trip/trip-monitored-day-selector.tsx
```diff
@@ -0,0 +1,203 @@
...
+    return <MonitoredDays days={dayFieldsToArray(monitoredTrip)} />
+  }
+
+  return (
+    <>
+      {errorCheckingTrip && (
+        <>
+          {/* FIXME: Temporary solution until itinerary existence check is fixed. */}
+          <br />
+          <FormattedMessage id="actions.user.itineraryExistenceCheckFailed" />
```
amy-corson-ibigroup, line 130: This might also need to be in a status region? But if it's a temporary solution maybe not.

lib/components/user/monitored-trip/trip-basics-pane.tsx
```diff
@@ -391,6 +203,32 @@ class TripBasicsPane extends Component<TripBasicsProps, State> {
...
+            errorCheckingTrip={errorCheckingTrip}
+            errorSelectingDays={
+              disableSingleItineraryDays ? selectOneDayError : undefined
+            }
+            finalItineraryExistence={finalItineraryExistence}
+            isReadOnly={isReadOnly}
+            monitoredTrip={monitoredTrip}
+          />
+          {!isCreating && (
+            <ToggleNotificationButton
```
amy-corson-ibigroup: Could the color of this be `#737373` to match the text above?

I also think we could make the margin between this button and the `TripMonitoredDaySelector` a little smaller to match the spacing above.

Image

amy-corson-ibigroup (approved):
Looks really nice, thanks for the changes! Just some nitpicky styling suggestions and an opportunity to improve the AT readout

lib/components/user/monitored-trip/trip-monitored-day-selector.tsx
binh-dam-ibigroup: There is also an `alert` box that shows the same message (visually and to AT) if the check doesn't complete, so I will leave this one here.

lib/components/user/monitored-trip/trip-basics-pane.tsx
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/1557/commits/e4645deae9f9ca7db010eb268a56945a37a455c5

PR #1559, opened 2026-04-07
Field Trip Support New Endpoints

miles-grant-ibigroup:
Make use of https://github.com/opentripplanner/otp-datastore/pull/11

### DIFF
--- a/lib/actions/field-trip.js
+++ b/lib/actions/field-trip.js
@@ -69,10 +69,13 @@ const FieldTripEndPoints = {
   DELETE_NOTE: 'deleteNote',
   DELETE_TRIP: 'deleteTrip',
   EDIT_SUBMITTER_NOTES: 'editSubmitterNotes',
+  SET_DESTINATION_ARRIVAL_TIME: 'setRequestArriveDestinationTime',
+  SET_DESTINATION_DEPART_TIME: 'setRequestLeaveDestinationTime',
   SET_REQUEST_DATE: 'setRequestDate',
   SET_REQUEST_GROUP_SIZE: 'setRequestGroupSize',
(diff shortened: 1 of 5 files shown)


lib/actions/field-trip.js
amy-corson-ibigroup, line 1: These three new methods `setArriveDestinationTime` `setLeaveDestinationTime` and `setArriveSchoolTime` are all completely identical checks, with the exception of the `FieldTripEndPoints` value. Can you create one `setFieldTripTime` and just pass in the appropriate `FieldTripEndPoints`?

lib/components/admin/call-taker.css
```diff
@@ -23,4 +23,11 @@
 
 .otp .search-plan-button-container {
   padding: 1rem;
+}
+
+.field-trip-time-edit form {
+  display: flex;
+  flex-direction: row-reverse;
+  align-items: baseline;
+  justify-content: left;
```
amy-corson-ibigroup, line 32: Can you sort these

amy-corson-ibigroup (changes requested):
Looks good but I think we can create one method instead of three!

lib/actions/field-trip.js
miles-grant-ibigroup: I struggled with this one since field-trip is quite well split into model/view/controller, but I think I found a way that doesn't put too much load into the view, although a purist would probably get quite mad at me. 

This solution means that the view is now coupled to, and will now need to change if the server architecture changes, but honestly given the state of the server I don't see that as a risk.

amy-corson-ibigroup (approved):
Looks good to me! Thanks for the changes

josh-willis-arcadis (approved):
LGTM. I like the solution for `setFieldTripTime`!

PR #1562, opened 2026-04-14
Fix auto refresh during typing in nearby view

alec-georgoff:
**Description:**

This PR addresses negative behavior where the auto-refresh interval would cause a page re-render even if the user was typing in the location field, causing their input to be erased. This behavior was originally noticed on nearby view on mobile web; however, the issue was also present on desktop once the refresh interval was adjusted back to `15000` ms (it was changed to `15000000` for some reason a couple of years ago, which doesn't make much sense).

See video below for the solution in action (the auto refresh interval was shortened to 3 seconds for the video just to show the refresh without waiting a full 30 seconds).

<details>
<summary>Expand for video</summary>

</details>

Note that the solution requires a bump to the location-field package (opentripplanner/otp-ui#939)

**PR Checklist:**


amy-corson-ibigroup (approved):
Looks good and works well! 

There is something kind of unintuitive for users about all updates stopping if you're focused on the location field, but after looking at the way we handle these `currentLocation` and nearby view fetches on mobile, I do agree this is the best approach currently. Just flagging, because it could be worth looking into new ways to watch users current location on mobile in the future!

josh-willis-arcadis (approved):
good catch! LGTM!

PR #1564, opened 2026-04-15
Clean Up URL when clearing an active search

amy-corson-ibigroup:
**Description:**
Leaving outdated state variables in the URL was causing the `popstate` event listener to fire unnecessary, making it impossible for users to navigate to different pages. Instead, clean up the `urlParams` whenever the search UI is cleared. 

To reproduce the bug:
- Plan a trip
- Select an itinerary (`activeItinerary` should populate in the URL)
- Click either the "switch locations" or "clear location" button (The search results in the UI should clear)
- Try to navigate to the route viewer via the view switcher on the nav ribbon

I also took out the `console.log` in `itineraries.tsx` because I think that was mistakenly left in at some point.

**PR Checklist:**


binh-dam-ibigroup (approved):
Nice fix!

miles-grant-ibigroup (approved):
Don't mind the hack, but could you add a comment to the method saying that we should try to move these changes into the redux store/find a more redux-ey way of doing this?

amy-corson-ibigroup:
> Don't mind the hack, but could you add a comment to the method saying that we should try to move these changes into the redux store/find a more redux-ey way of doing this?

@miles-grant-ibigroup wait is there a better way to do this? This wasn't intended to be hacky so I'm open to other approaches if you have ideas!

miles-grant-ibigroup:
> > Don't mind the hack, but could you add a comment to the method saying that we should try to move these changes into the redux store/find a more redux-ey way of doing this?
> 
> @miles-grant-ibigroup wait is there a better way to do this? This wasn't intended to be hacky so I'm open to other approaches if you have ideas!

The "redux-ey" way to do it would be to take that CLEAR_ACTIVE_SEARCH action and have it clear out everything else that it's supposed to clear out. But we'd probably have to add some other listeners as well and rework on what triggers some of those URL params are updated and that all seems like much more hassle than it's worth!

binh-dam-ibigroup:
> The "redux-ey" way to do it would be to take that CLEAR_ACTIVE_SEARCH action and have it clear out everything else...

I thought we had an internal backlog item to make the URL the source of truth instead?

miles-grant-ibigroup:
> > The "redux-ey" way to do it would be to take that CLEAR_ACTIVE_SEARCH action and have it clear out everything else...
> 
> I thought we had an internal backlog item to make the URL the source of truth instead?

This is exactly why I approved this PR the way it is!

PR #1566, opened 2026-04-16
Modify Leg ID for FIFA-Altered Routes

alec-georgoff:
**Description:**
This PR updates the FIFA stop adjustment logic, adding logic to update the leg ID of the altered transit leg so that it accurately reflects the new transit leg's stops. Because the only change to the transit leg is the new origin/destination stop, it's possible to just replace the old stop code with the new one in the decoded leg ID, then re-encode it.

**PR Checklist:**

### DIFF
--- a/lib/util/stop-adjustments.ts
+++ b/lib/util/stop-adjustments.ts
@@ -716,6 +716,29 @@ const itineraryIsInZoneTimeRange = (
   )
 }
 
+const updateLegIdWithNewStop = (
+  leg: Leg,
+  adjustment: StopAdjustment
+): string | undefined => {
+  const oldLegId = leg.id
+  const oldStopCode = leg.to.stop?.gtfsId
+  const newStopCode = adjustment.newStop.gtfsId
+  if (oldLegId && oldStopCode) {
+    try {
+      const bytes = Uint8Array.fromBase64(oldLegId, {
+        alphabet: 'base64url'
+      })
+      const decoded = new TextDecoder().decode(bytes)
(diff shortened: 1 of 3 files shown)


lib/util/stop-adjustments.ts
```diff
@@ -716,6 +716,29 @@ const itineraryIsInZoneTimeRange = (
...
+  const oldLegId = leg.id
+  const oldStopCode = leg.to.stop?.gtfsId
+  const newStopCode = adjustment.newStop.gtfsId
+  if (oldLegId && oldStopCode) {
+    try {
+      const bytes = Uint8Array.fromBase64(oldLegId, {
+        alphabet: 'base64url'
+      })
+      const decoded = new TextDecoder().decode(bytes)
+      const updatedDecoded = decoded.replace(oldStopCode, newStopCode)
```
binh-dam-ibigroup, line 732: If by luck the stop code appears multiple times in the leg ID components, what is preventing this line to replace the wrong occurrence?

binh-dam-ibigroup (approved):
Seems to work, but you will have to add spelling exceptions to CodeQL.

lib/util/stop-adjustments.ts
alec-georgoff: Seems like an unlikely edge case, but I added a quick duplicate check. I think there could be some more advanced logic to actually determine which of the duplicates should be used, but in this case I decided to just have it abort the leg ID manipulation if it finds duplicates, since this is all for a temporary feature anyway.

PR #1569, opened 2026-04-21
Fix Nearby View Magnifying Glass Size

miles-grant-ibigroup:
I'm a little confused why this worked... I guess something changed inside the location field. It's not perfectly centered. Perhaps there's a solution I'm not seeing where we can get it more properly centered?

### DIFF
--- a/__tests__/components/viewers/__snapshots__/nearby-view.js.snap
+++ b/__tests__/components/viewers/__snapshots__/nearby-view.js.snap
@@ -161,8 +161,7 @@ exports[`components > viewers > nearby view renders nothing on a blank page 1`] 
                             <Search
                               style={
                                 {
-                                  "marginRight": 5,
-                                  "padding": 5,
+                                  "padding": 3,
                                 }
                               }
                             >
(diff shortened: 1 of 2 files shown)


amy-corson-ibigroup:
Kinda feel like a better approach might be just using the `size` prop, no? 

I think the icon is technically centered but I see what you mean, it looks off next to the placeholder text... could add a small `marginTop` of, say, like `-3px` to scoot it up?

Also, do you think while we're in here, we can make the search and clear icons the same color? And can that color be `#666` (`grey[700]`)? That'll make sure it matches the placeholder text and has appropriate contrast.

miles-grant-ibigroup:
Size didn't quite work... But the margin looks great, thanks for that! Color is excellent too

amy-corson-ibigroup (changes requested):
Padding isn't a good approach, unfortunately. Not only is it inherently confusing in the code, the icon is changing sizes at different screen widths.

I'm seeing `size={10}` work fine, can you elaborate on why it doesn't work?

miles-grant-ibigroup:
> Padding isn't a good approach, unfortunately. Not only is it inherently confusing in the code, the icon is changing sizes at different screen widths.
> 
> I'm seeing `size={10}` work fine, can you elaborate on why it doesn't work?

Maybe I was putting it in the wrong place. If you got it working would you mind making a commit here/a PR that merges into here?

amy-corson-ibigroup (approved):
Bye tiny magnifying glass i will miss you forever

alec-georgoff (approved):

PR #1571, opened 2026-04-22
Encode and decode polylines instead of directly replacing encoded substrings

alec-georgoff:
**Description:**
Previously, polylines were adjusted by looking for specific substrings in the encoded polyline and removing and/or replacing them with other snippets of encoded polyline. This was not a great approach, and is directly discouraged in the polyline documentation. After multiple bugs were seen with the resulting polylines, the approach was changed.

Now, we always decode polylines before attempting to add to or remove from them. We are able to use the change in the `legGeometry.length` property of the itineraries to determine how many coordinate pairs to remove from a decoded polyline. When we need to add points to the polyline, we decode the additional points and then directly concatenate them with the existing points.

**PR Checklist:**

### DIFF
--- a/lib/util/STOP_ADJUSTMENTS.md
+++ b/lib/util/STOP_ADJUSTMENTS.md
@@ -132,15 +132,11 @@ This property contains the following fields:
 
 #####  `length`
 
-The *change* in length of the leg geometry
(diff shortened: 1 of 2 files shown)


josh-willis-arcadis (approved):
much cleaner! Great use of the `polyline` library.

PR #1580, opened 2026-05-26
hide itin summary overlay if more than 6 routes are present

miles-grant-ibigroup:
This overlay needs a lot more work, but this is a quick bandaid that should help quite a bit.


daniel-heppner-ibigroup (approved):
looks good to me

alec-georgoff (commented):
(potentially consequential) nit: wouldn't this hide the overlay if exactly 6 routes are present? The PR title says "if more than 6". Just want to check to make sure that's intended.

miles-grant-ibigroup:
> (potentially consequential) nit: wouldn't this hide the overlay if exactly 6 routes are present? The PR title says "if more than 6". Just want to check to make sure that's intended.

It doesn't really matter it's just a vague heuristic. Really we should take into account how wide each badge will be but for now this is fine. Thanks for double checking though!

alec-georgoff (approved):

PR #1581, opened 2026-05-27
create `allowFlexWithoutTransit` setting, clean up flex duplication support

miles-grant-ibigroup:
There should be no changes to requests generated unless `allowFlexWithoutTransit` is enabled.

Removing the restriction on merging flex trips together should help with cleanliness

### DIFF
--- a/lib/actions/apiV2.js
+++ b/lib/actions/apiV2.js
@@ -1177,7 +1177,13 @@ export function routingQuery(searchId = null, updateSearchInReducer) {
     // this removes superfluous flex requests.
     combinations = combinations.filter((c) => {
       const flexCount = countFlexModes(c.modes)
-      // We need either all 3 flex modes or none! Anything in-between is invalid
+      // Slight hack to remove combinations that feature 3 flex modes and no transit
+      if (
+        config.modes?.allowFlexWithoutTransit === false &&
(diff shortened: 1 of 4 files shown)


lib/actions/apiV2.js
```diff
@@ -1177,7 +1177,13 @@ export function routingQuery(searchId = null, updateSearchInReducer) {
     // this removes superfluous flex requests.
     combinations = combinations.filter((c) => {
       const flexCount = countFlexModes(c.modes)
-      // We need either all 3 flex modes or none! Anything in-between is invalid
+      // Slight hack to remove combinations that feature 3 flex modes and no transit
+      if (
+        config.modes?.allowFlexWithoutTransit === false &&
```
amy-corson-ibigroup, line 1182: Should this just be `!config.modes?.allowFlexWithoutTransit`

amy-corson-ibigroup (approved):
Looks good just one cleanup

lib/actions/apiV2.js
miles-grant-ibigroup: No! We want this on by default, so undefined has to equal `true`

josh-willis-arcadis (approved):
LGTM

PR #1586, opened 2026-06-15
Support Extra Views

miles-grant-ibigroup:
Allows for mounting any component as a full-width component via `config.js`. 

There is a field that allows the component to appear in the view switcher


lib/util/config-types.ts
```diff
@@ -406,6 +406,14 @@ export interface DateTimeConfig {
   timeFormat: string
 }
 
+export type ExtraView = {
+  content: JSX.Element
+  icon?: JSX.Element
+  name: string
+  path: string
+  viewStitcher?: boolean
```
miles-grant-ibigroup: Can we rename this?
```suggestion
  showInHeaderBar?: boolean
```

lib/components/app/app-frame.tsx
```diff
@@ -21,13 +22,17 @@ const AppFrame = ({ children, SubNav }: Props): JSX.Element => (
...
     <main tabIndex={-1}>
-      <div className="container">
-        <Row>
-          <Col md={8} mdOffset={2} sm={10} smOffset={1} xs={12}>
-            {children}
-          </Col>
-        </Row>
-      </div>
+      {fullWidth ? (
+        <div>{children}</div>
```
miles-grant-ibigroup, line 26: Why'd you remove the margin here?

josh-willis-arcadis: It was pushing the component off the screen. There is css in the component to handle any padding/margins.

josh-willis-arcadis (approved):

alec-georgoff (approved):

PR #1593, opened 2026-06-30
Vehicle Popup: don’t show duration ago if lastUpdated is null

miles-grant-ibigroup:
This code really should be moved back to OTP-UI... The handling there is correct...

### DIFF
--- a/lib/components/map/connected-transit-vehicle-overlay.tsx
+++ b/lib/components/map/connected-transit-vehicle-overlay.tsx
@@ -35,6 +35,7 @@ interface TransitVehicleExt extends TransitVehicle {
   textColor?: string
 }
 
+// eslint-disable-next-line complexity
 function VehicleTooltip({
   vehicle
 }: {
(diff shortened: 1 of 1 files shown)


amy-corson-ibigroup (approved):

lib/components/map/connected-transit-vehicle-overlay.tsx
```diff
@@ -70,20 +71,22 @@ function VehicleTooltip({
...
-              duration: formatDuration(
-                Math.floor(Date.now() / 1000 - (vehicle?.seconds || 0)),
-                intl,
-                true
-              )
-            }
-          )
-        )}
-      </div>
+      {!!vehicle?.seconds && (
```
alec-georgoff: If the update is ultra-fresh (`seconds === 0`), this would evaluate to false and we wouldn't see the duration. Is that desired?

miles-grant-ibigroup: Good catch, thanks!

alec-georgoff (approved):

PR #1594, opened 2026-06-30
Display map with open itinerary on mobile

amy-corson-ibigroup:
**Description:**
Previously there was no way to see the itinerary you choose on the map in mobile view. This PR maintains the map movement of the results screen through to the itinerary view.

Because we're moving the button out of a parent container, the positioning is a little different which is triggering a diff in the percy tests. These changes look fine to me but happy to discuss if anyone disagrees!

**PR Checklist:**

| Before | After |
|--------|-------|
|image|image|

### DIFF
--- a/lib/components/narrative/map-expansion-button.tsx
+++ b/lib/components/narrative/map-expansion-button.tsx
@@ -4,7 +4,7 @@ import styled from 'styled-components'
 
 import { StyledIconWrapper } from '../util/styledIcon'
 
-export const TOGGLE_MAP_BUTTON_HEIGHT = 7
+export const TOGGLE_MAP_BUTTON_HEIGHT = 5
 
 const StyledMapExpansionButton = styled.button`
   align-items: center;
@@ -48,7 +48,7 @@ const MapExpansionButton = ({
       onClick={onClickExpansionButton}
       title={mapExpansionText}
     >
-      <StyledIconWrapper>
+      <StyledIconWrapper style={{ marginTop: '3px' }}>
         <UnicodeChevron expanded={mapExpanded}>&#10095;</UnicodeChevron>
       </StyledIconWrapper>
     </StyledMapExpansionButton>
--- a/lib/components/narrative/map-expansion-button.tsx
+++ b/lib/components/narrative/map-expansion-button.tsx
@@ -44,7 +44,7 @@ const MapExpansionButton = ({
       aria-label={mapExpansionText}
       // CSS classes 'base-color-bg' and 'itinerary' let us set the chevron color to the
       // same color as the color used the H3 headings over the narrative background.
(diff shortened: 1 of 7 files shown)


lib/components/narrative/map-expansion-button.tsx
```diff
@@ -0,0 +1,58 @@
...
+    <StyledMapExpansionButton
+      aria-label={mapExpansionText}
+      // CSS classes 'base-color-bg' and 'itinerary' let us set the chevron color to the
+      // same color as the color used the H3 headings over the narrative background.
+      className="clear-button-formatting base-color-bg itinerary"
+      onClick={onClickExpansionButton}
+      title={mapExpansionText}
+    >
+      <StyledIconWrapper style={{ marginTop: '3px' }}>
+        <UnicodeChevron expanded={mapExpanded}>&#10095;</UnicodeChevron>
```
miles-grant-ibigroup, line 52: Image

Something funky is going on here!

lib/util/ui.ts
```diff
@@ -113,6 +115,7 @@ export function getItineraryView({
   ui_activeItinerary,
   ui_itineraryView
 }: UrlParams): ItineraryView {
+  console.log(ui_itineraryView)
```
miles-grant-ibigroup: Whoops!

miles-grant-ibigroup (approved):
Good implementation! The only issue is the chevron display, but that seems to be an existing problem?

Once the console.log is gone I think this is good to go!

alec-georgoff (approved):
Awesome! One non-blocking question: should this be a configuration option? I could see some consumers wanting the increased viewing space for an itinerary instead of viewing the map

lib/components/narrative/map-expansion-button.tsx
amy-corson-ibigroup: Hmm I'm only seeing this on one configuration and that's septa... wondering if there's some custom css that's causing it? But I don't think it's in this repo

amy-corson-ibigroup:
> Awesome! One non-blocking question: should this be a configuration option? I could see some consumers wanting the increased viewing space for an itinerary instead of viewing the map

@alec-georgoff  It's a good question! I feel personally like seeing the map alongside the itinerary feels important (crazy we didn't have it before). I don't know if I would want to add complexity to accommodate a view where you can't see the map. But I see what you mean and I could see the map movement going full screen itinerary to half screen itinerary, rather than halfscreen itinerary to full screen map? Maybe we can add to the tech call as a follow up?

miles-grant-ibigroup:
I believe the only reason we weren't showing it before is because the web on mobile view was designed during the era of the iphone 5S :) I think what you have here is a sane default!

amy-corson-ibigroup:
Thanks guys!!

PR #1595, opened 2026-07-01
Load Stop Closures Data

alec-georgoff:
**Description:**
This PR adds the fetching of closed stops from the OTP API and implements in them into the timetable. In addition, it overhauls the logic for creating and displaying a React Portal.

## Closed Stops

The `alerts` query in OTP is used to determine closed stops. The results are used to build a map:

```ts
Map<[ROUTE GTFSID], Set<[CLOSED STOP GTFSID]>>
```

This map is stored in the application state for use in other parts of the application. In this PR, the closed stops are passed to the timetable component so that closed stops are struck through in the timetable.

## React Portals

The logic for displaying React Portals was overhauled. Previously, the styles from the main application were not being applied to the content within the Portal. Largely following guidance from [this article](https://dev.to/noriste/the-challenges-of-rendering-an-openlayers-map-in-a-popup-through-react-2elh), the logic was updated to correctly copy over the styles to the Portal content.

**PR Checklist:**


lib/actions/apiV2.js
```diff
@@ -636,6 +636,41 @@ const getVehiclePositionsForRoute = (routeId) =>
     )
   }
 
+const stopClosuresQuery = `
+  query StopClosures {
+    alerts(effect: NO_SERVICE) {
```
alec-georgoff, line 641: We could potentially make the effect type argument configurable here?

lib/components/app/responsive-webapp.js
```diff
@@ -203,6 +205,8 @@ class ResponsiveWebapp extends Component {
       }
     }
 
+    stopClosuresQuery()
```
alec-georgoff, line 208: I think that loading all of the stop closures upon initial load and storing them in state makes sense (so they can be used in multiple areas of the application), but I'm open to disagreement on that

lib/components/app/responsive-webapp.js
```diff
@@ -395,7 +400,7 @@ class RouterWrapperWithAuth0 extends Component {
             style: { minWidth: '250px' }
           }}
         />
-        {timetableEnabled && <TimetableWrapper />}
+        {timetableEnabled && portal && <TimetableWrapper />}
```
alec-georgoff, line 403: Without this addition, the `TimetableWrapper` component was sticking around even after closing the portal window, which would cause the state (such as which direction is shown) to preserve when the timetable is re-opened later.

lib/actions/apiV2.js
josh-willis-arcadis: What other `effects` would be helpful to know about?

lib/reducers/create-otp-reducer.js
```diff
@@ -500,6 +500,34 @@ function createOtpReducer(config) {
...
+          })
+        })
+        return update(state, {
+          ui: {
+            stopClosures: {
+              $set: map
+            }
+          }
+        })
+      case 'STOP_CLOSURES_ERROR': // TODO: add graceful handling of errors
```
josh-willis-arcadis, line 523: could this error be displayed in the timetable portal? Just a way to visually see that there is potential data missing.

lib/components/app/responsive-webapp.js
josh-willis-arcadis: I don't have any strong opinions. An alternative approach could be querying this data when the route viewer is rendered.

lib/components/viewers/timetable-wrapper.tsx
```diff
@@ -27,8 +32,15 @@ const TimeTableWrapper = (props: TimeTableWrapperProps): JSX.Element => {
...
 
+  const closedStopsSet = closedStops?.get(portal || '')
+
   return portal ? (
-    <WindowPortal onClose={() => setPortal(false)}>
+    <PortalWrapper
+      onClose={() => {
+        setPortal(undefined)
+      }}
+      title="timetable"
```
josh-willis-arcadis, line 42: More information here would be nice. Can we change the "about:blank" text?

Image

lib/components/viewers/popout.tsx
```diff
@@ -1,56 +1,94 @@
...
-    clearInterval(intervalId)
+/** Injects the node ref into the newly created window upon load
+ */
+const onLoad = (
+  windowRef: React.MutableRefObject<Window | null>,
+  nodeRef: React.MutableRefObject<HTMLElement>,
+  title: string,
+  setIsLoading: (loading: boolean) => void
+) => {
+  if (windowRef.current) {
```
josh-willis-arcadis, line 29: is there a way to bring the portal into focus if it has already loaded? Use case is the timetable portal is open already for a route. Timetable remains open, but a new route is selected. Data is loaded into react portal correctly, but it is not in focus or in view after clicking the timetable button.

josh-willis-arcadis (approved):
data loads and can be viewed in timetable component. nothing blocking this PR in my eyes. A few items that would be nice to have. Could be addressed in future PRs.

lib/components/app/responsive-webapp.js
amy-corson-ibigroup: I think I disagree on this placement... I could see us moving this query here in the future if we implement closed stops more thoroughly into the UI, but for now this call is only used for twin cities calltaker. It feels a bit unnecessary to include for everyone in all cases.

lib/components/app/responsive-webapp.js
amy-corson-ibigroup: woah nice catch!

lib/components/viewers/timetable-wrapper.tsx
```diff
@@ -5,16 +5,21 @@ import TimeTable from '@opentripplanner/timetable'
...
 
 interface TimeTableWrapperProps {
-  portal: boolean
-  setPortal: (portal: boolean) => void
+  /** A map of closed stops. Keys are route gtfsIds, values are sets of gtfsIds for stops that are closed on that route */
+  closedStops: Map<string, Set<string>>
+  /** The value of the portal in application state. If defined, it refers to the gtfsId of the route whose timetable
+   * is to be displayed in the portal
+   */
+  portal: string | undefined
```
amy-corson-ibigroup, line 16: Could this not just be `portal?: string`

lib/components/viewers/timetable-wrapper.tsx
```diff
@@ -5,16 +5,21 @@ import TimeTable from '@opentripplanner/timetable'
...
+  setPortal: (portal: string | undefined) => void
```
amy-corson-ibigroup, line 17: might be worth it to pull the `portal` type out so you can reference it here too.

lib/components/viewers/timetable-wrapper.tsx
amy-corson-ibigroup: also kinda wondering if this variable name should change and get more specific? Maybe like `routeIdForPortal`? The way it's used below in line 35 for example gets a little confusing at first glance

lib/components/viewers/pattern-viewer.tsx
```diff
@@ -28,7 +28,7 @@ import VehiclePositionRetriever from './vehicle-position-retriever'
 interface Props {
   findRoutesIfNeeded: () => void
   getTimetableData: (params: any) => any
-  setPortal: (arg: boolean) => void
+  setPortal: (arg: string | undefined) => void
```
amy-corson-ibigroup, line 31: again, would `arg?: string` be better here?

amy-corson-ibigroup (approved):
Nothing blocking! Just a few thoughts/questions

lib/components/viewers/timetable-wrapper.tsx
alec-georgoff: Unfortunately I wasn't able to find a way to change that...I did update the title to use the route GTFS ID for now, though (instead of `timetable`)

lib/components/viewers/popout.tsx
alec-georgoff: That makes sense...one of my next tasks is to add support for multiple portals being open at once. I'll add a note to also work on this focus issue there!

lib/reducers/create-otp-reducer.js
alec-georgoff: Yup, good idea. I added logic to display an error in the portal if we get an error during stop closures loading

lib/components/viewers/timetable-wrapper.tsx
alec-georgoff: Updated to `portalId` for now; this will change anyway when we add support for multiple portals being open at once. Also updated the optional props to use `?`

lib/components/viewers/timetable-wrapper.tsx
alec-georgoff: Will work on this when support for multiple portals is added!

lib/components/app/responsive-webapp.js
alec-georgoff: Yeah that makes sense. The plan is to expand it a bit into the itinerary view and the stop viewer, but since it's only used in timetable right now I'll go ahead and move it to the pattern viewer for the moment.

alec-georgoff:
Note: I forgot to push my PR response commits before merging this...those are located in #1597

PR #1598, opened 2026-07-09
Add support for omitCanceled option in config

alec-georgoff:
**Description:**
Adds support for the `omitCanceled` option of the [OTP Plan query](https://docs.opentripplanner.org/api/dev-2.x/graphql-gtfs/queries/plan), via the config file. Requires version bump after merging/deploying opentripplanner/otp-ui#971

**PR Checklist:**

### DIFF
--- a/lib/actions/apiV2.js
+++ b/lib/actions/apiV2.js
@@ -1147,6 +1147,8 @@ export function routingQuery(searchId = null, updateSearchInReducer) {
         config.routeModeOverrides
       )?.join(',')
 
+    const { omitCanceled = true } = config.itinerary || {}
+
     const baseQuery = {
       arriveBy,
       banned: {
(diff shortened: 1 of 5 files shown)


josh-willis-arcadis (approved):
routing requests look good to me.

lib/actions/apiV2.js
```diff
@@ -1157,6 +1157,13 @@ export function routingQuery(searchId = null, updateSearchInReducer) {
       from: currentQuery.from,
       modes: modes || activeModes,
       modeSettings,
+      // Need to be careful here: the default value for OTP queries is true. If a config
+      // doesn't include the omitCanceledTrips flag, we need to make sure we keep that true
+      // default value
+      omitCanceled:
+        config.itinerary?.omitCanceledTrips !== undefined
+          ? config.itinerary.omitCanceledTrips
+          : true,
```
amy-corson-ibigroup: You can pull out this variable via destructuring and set a default to `true`

So on line 1149: 
```
const { omitCanceledTrips = true } = config.itinerary
```

and then just `omitCanceled: omitCanceledTrips`. No need for comment/ternary

amy-corson-ibigroup (commented):
I think there's a cleaner way to do this!

lib/actions/apiV2.js
alec-georgoff: I like it!! I added a fallback just in case `config.itinerary` ends up undefined:

```ts
const { omitCanceled = true } = config.itinerary || {}
```

alec-georgoff:
Note: I version bumped core-utils to include the now-deployed updated query that actually includes `omitCanceled`

amy-corson-ibigroup (approved):

PR #1599, opened 2026-07-13
Update itineraries to show canceled trip warning

alec-georgoff:
**Description:**
Adds a warning to the itinerary summary if an itinerary contains a canceled trip. Note that including canceled trips [requires a config flag](https://github.com/opentripplanner/otp-react-redux/pull/1598), so these updates will not show up for any implementation until they turn on that config flag.

## Example

image

**PR Checklist:**

### DIFF
--- a/lib/components/narrative/default/default-itinerary.js
+++ b/lib/components/narrative/default/default-itinerary.js
@@ -89,6 +89,10 @@ const ItinerarySummaryWrapper = styled.div`
   justify-content: space-between;
 `
 
+const CanceledTripMessage = styled.div`
+  color: rgb(183, 38, 32);
+`
+
 const ITINERARY_ATTRIBUTES = [
   {
     alias: 'best',
@@ -335,6 +339,10 @@ class DefaultItinerary extends NarrativeItinerary {
       )
     }
 
+    const containsCanceledTrip = itinerary.legs?.some(
+      (leg) => leg.realtimeState === 'CANCELED'
+    )
+
     return (
       <div
         className={`option default-itin${visible ? ' active' : ''}${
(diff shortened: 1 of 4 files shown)


lib/components/narrative/default/default-itinerary.js
```diff
@@ -89,6 +89,10 @@ const ItinerarySummaryWrapper = styled.div`
   justify-content: space-between;
 `
 
+const CanceledTripMessage = styled.div`
+  color: rgb(183, 38, 32);
```
josh-willis-arcadis: opportunity to use building blocks colors here.

josh-willis-arcadis (approved):
1 nit, but lgtm

lib/components/narrative/default/default-itinerary.js
alec-georgoff: Done!

lib/components/narrative/default/default-itinerary.js
```diff
@@ -359,6 +368,11 @@ class DefaultItinerary extends NarrativeItinerary {
               </h3>
             )}
             <ItinerarySummary itinerary={itinerary} LegIcon={LegIcon} />
+            {containsCanceledTrip && (
+              <CanceledTripMessage>
+                <FormattedMessage id="common.itineraryDescriptions.containsCanceledTrip" />
```
amy-corson-ibigroup, line 379: can you add an `ExclamationTriangle` icon here to draw more attention to this?

amy-corson-ibigroup (approved):
If we can add iconography to this, I think that would be an improvement! otherwise everything looks good

lib/components/narrative/default/default-itinerary.js
amy-corson-ibigroup: You can use `<IconWithText>`

alec-georgoff: image

Awesome

PR #1603, opened 2026-07-16
Field Trip Restructruing 

miles-grant-ibigroup:
As part of moving otp-rr to use the new `planConnection` endpoint, the field trip module would need extensive work. As there is no active maintainer for the field trip module, this PR removes the field trip code from the dev branch. The field trip code (as is in working order) can be maintained in a new branch, which will still use the old `plan` endpoint.

If field trip becomes actively maintained again and is updated to use `planConnection`, then it can be merged back into the dev branch


lib/components/admin/trip-status.js
binh-dam-ibigroup: Can we get rid of this file? It talks about "schools".

binh-dam-ibigroup (approved):
I never looked at how deep field trip and call taker were coupled together!

lib/components/admin/trip-status.js
miles-grant-ibigroup: Great catch thanks!

lib/components/app/call-taker-panel.js
```diff
@@ -129,48 +128,24 @@ class CallTakerPanel extends Component {
...
-    if (
-      // Only enabled when field trip is enabled, as otherwise
-      // this flashes a lot, possibly to an annoying extent
-      this.props.fieldTripVisible &&
-      (prevProps?.currentQuery?.from?.name !==
-        this.props?.currentQuery?.from?.name ||
-        prevProps?.currentQuery?.to?.name !==
-          this.props?.currentQuery?.to?.name)
-    ) {
-      this.setState({ queryChanged: true }, () => {
```
daniel-heppner-ibigroup, line 148: is the queryChanged state prop still needed?

lib/components/narrative/plan-first-last-buttons.js
```diff
@@ -36,9 +36,7 @@ function PlanFirstLastButtons(props) {
 const mapStateToProps = (state) => {
   const itineraries = getActiveItineraries(state)
   return {
-    enabled:
-      state.otp.config.itinerary?.showPlanFirstLastButtons &&
-      state.callTaker.fieldTrip.visible === false,
+    enabled: state.otp.config.itinerary?.showPlanFirstLastButtonse,
```
daniel-heppner-ibigroup: buttonse

daniel-heppner-ibigroup (approved):
i gotta say, it feels nice to remove all this code!

lib/components/app/call-taker-panel.js
miles-grant-ibigroup: Great catch thanks!

PR #1604, opened 2026-07-17
`PlanConnection`

miles-grant-ibigroup:
##### TODO JUST AFTER MERGING: Create a `PLAN` branch that is protected and have the only commit on it be a revert of this PR. That way, adding commits to non-planconnection otp-rr will be trivial

Blocked by https://github.com/opentripplanner/otp-ui/pull/977

- [x] Percy tests will be annoying to fix


miles-grant-ibigroup:
All failing tests are a result of wrong otp-ui versions

lib/actions/apiV2.js
binh-dam-ibigroup: `generateOtp2Query` does not preserve stuff in `baseQuery`, so custom params need to be injected here:

```suggestion
          {
            ...query.variables,
            mobilityProfile
          },
```

binh-dam-ibigroup, line 1185: Need to inject custom query params here because `generateOtp2Query` from https://github.com/opentripplanner/otp-ui/pull/977 does not preserve them.

Instead of 
```
          query.variables,
```
write, for instance:
```
          {
            ...query.variables,
            mobilityProfile: baseQuery.mobilityProfile
          },
```

binh-dam-ibigroup (commented):
I have not looked at the other changes yet, but migrating custom query params isn't too bad.

lib/actions/apiV2.js
miles-grant-ibigroup: Is `mobilityProfile` supported in the current main `dev` branch?

binh-dam-ibigroup: @miles-grant-ibigroup `mobilityProfile` is currently supported because OTP-UI leaves any custom params in `query.variables` untouched, and they get passed to the graphql endpoint. (This is not the case with https://github.com/opentripplanner/otp-ui/pull/977.)

miles-grant-ibigroup: Is there a solution you can think of that doesn't explicitly include `mobilityProfile`? Where in opentripplanner/otp-ui#977 would you want the variables passed through?

PR #1607, opened 2026-07-23
Move timetable to url scheme

alec-georgoff:
**Description:**
This PR removes the React portals logic and instead opts for a URL-based approach to loading a timetable. This allows the pattern viewer to just trigger a new `window.open` with the necessary URL and not deal with the intricacies of portals.

There are also some general improvements.

**PR Checklist:**


lib/components/viewers/timetable-wrapper.tsx
```diff
@@ -79,23 +116,27 @@ const TimeTableWrapper = (props: TimeTableWrapperProps): JSX.Element => {
...
   ) : (
-    <></>
+    <FormattedMessage id="components.Timetable.errorLoadingTimetable" />
   )
 }
 
 const mapStateToProps = (state: AppReduxState) => {
+  const { pathname } = state.router.location
+  const split = pathname.split('/')
+  const routeId = split[split.length - 1]
```
alec-georgoff: Open to suggestions on better ways to extract this if folks have any!

lib/components/viewers/timetable-wrapper.tsx
```diff
@@ -1,48 +1,81 @@
...
   const [timepointsOnly, setTimepointsOnly] = useState(true)
+  const [loading, setLoading] = useState(true)
+
+  const closedStopsSet = useMemo(
+    () => closedStops?.get(routeId || ''),
+    [closedStops, routeId]
+  )
+
+  useEffect(() => {
+    stopClosuresQuery()
```
alec-georgoff, line 46: We can rip out this stop closures query once #1611 is merged, since that moves the query to the top level of the application.

lib/components/viewers/pattern-viewer.tsx
```diff
@@ -163,8 +143,8 @@ const PatternViewer = ({
             )}
           </h1>
           {timetableEnabled && (
-            <button onClick={() => setPortalId(routeId || undefined)}>
-              Timetable
+            <button onClick={handleTimetableButtonClick}>
```
miles-grant-ibigroup: Why is this a button? Can't it just be an anchor tag

lib/components/viewers/timetable-wrapper.tsx
miles-grant-ibigroup: Doesn't react-router extract this for you?

miles-grant-ibigroup: check the printable-itinerary wrapper for more detail

lib/components/viewers/timetable-wrapper.tsx
miles-grant-ibigroup: Can we block this PR until that's merged? I really don't like all this extra processing here

lib/components/viewers/timetable-wrapper.tsx
```diff
@@ -1,48 +1,81 @@
...
-  portalId: string | undefined
-  setPortalId: (portalId?: string) => void
+  getTimetableData: (params: GetTimetableDataParams) => void
+  routeId: string
   stopClosuresError?: string
-  timetable: any // TODO: add typing
+  stopClosuresQuery: () => void
 }
 
 const TimeTableWrapper = (props: TimeTableWrapperProps): JSX.Element => {
```
miles-grant-ibigroup, line 23: In my mind I'm imagining a future "bare-bones" timetable wrapper that only displays the date picker, timepoints picker, and timetable. How can we refactor this component to make that future wrapper easier to build?

lib/components/viewers/pattern-viewer.tsx
```diff
@@ -163,8 +143,8 @@ const PatternViewer = ({
             )}
           </h1>
           {timetableEnabled && (
```
miles-grant-ibigroup: Should we replace this with a generic calltaker check? That way regular users get the to-be-built "regular" timetable viewer, and calltaker users get the power user timetable viewer

miles-grant-ibigroup (changes requested):
I think this should be blocked until that other PR is merged

lib/components/viewers/timetable-wrapper.tsx
alec-georgoff: I think the pattern of

```js
const requestId = parseInt(state.router.location.query.requestId)
```

only works for parameters that are in the query string (assuming that's what you're referring to). It looks like we still need to split the pathname:

image

lib/components/viewers/timetable-wrapper.tsx
alec-georgoff: Added blocked label!

lib/components/viewers/timetable-wrapper.tsx
alec-georgoff: I'm not sure I fully understand...right now we're only displaying timepoints button, direction button, and the timetable (and pattern names for the direction, which I guess we could move into the timetable UI component), along with error/loading displays.

Are you saying we should move the state handlers and/or data fetching logic up a level? Or something else?

lib/components/viewers/pattern-viewer.tsx
alec-georgoff: I think I'd prefer to leave it as a granular option instead of assuming that all calltaker implementations will want to include a timetable. I could envision an implementer that uses a different timetable program/solution and they wouldn't want this functionality included in calltaker. I'm open to being persuaded though if you feel super strongly.

lib/components/viewers/pattern-viewer.tsx
alec-georgoff: Good call, [updated](https://github.com/opentripplanner/otp-react-redux/pull/1607/commits/48bfeb84ee1246622bce1db03af5967a26cc3752)

lib/components/viewers/timetable-wrapper.tsx
miles-grant-ibigroup: Hm should we just use query params then?

lib/components/viewers/timetable-wrapper.tsx
miles-grant-ibigroup: So the direction button feels like behavior that only makes sense in a url-based popup, rather than the modal timetable I believe we are considering for non-calltaker users.

I think a non-calltaker user should only see the timepoints button, a date picker, and a timetable. All other controls should be visible to calltaker users only

lib/components/viewers/pattern-viewer.tsx
miles-grant-ibigroup: I would prefer to reduce the number of feature flags. We can always hide the timetable link via css for special cases where people don't want to see a timetable

lib/components/viewers/timetable-wrapper.tsx
alec-georgoff: Actually I just had to dig a bit deeper to find the react-router `matchPath` function. [Implemented here](https://github.com/opentripplanner/otp-react-redux/pull/1607/commits/d69a08da291e8b3df395c775fdb7fb8ba4ff0cd0)

lib/components/viewers/timetable-wrapper.tsx
alec-georgoff: Okay so how about a structure like this?

```js
isCalltaker ? (
   <CalltakerTimetableWrapper>
      <BaseTimetableWrapper />
   </CalltakerTimetableWrapper>
) : (
   <BaseTimetableWrapper />
)
```

`BaseTimetableWrapper` just handles state for date and timepoints, and directly implements `Timetable` from otp-ui.

`CalltakerTimetableWrapper` implements all the extra power-user goodies for calltaker.

lib/components/viewers/pattern-viewer.tsx
alec-georgoff: Makes sense. [Implemented here](https://github.com/opentripplanner/otp-react-redux/pull/1607/commits/d00f562467cd1700e776e917af7d63cb8d284d41)

lib/components/viewers/timetable-wrapper.tsx
miles-grant-ibigroup: I can get on board with this!

miles-grant-ibigroup:
I'll review once the timetable package is merged!

lib/components/viewers/timetable-wrapper.tsx
alec-georgoff: @miles-grant-ibigroup Update here: I attempted to remove this query, but it made me realize that we actually need to preserve these two queries within the timetable wrapper component.

Because the timetable loads in its own separate window/instance, the state from the "parent" application is not carried over to the new window. After some reflection and discussion with @daniel-heppner-ibigroup, I think this is actually a good thing. This means that a user can send the URL of a timetable to someone else and that person can load it without needing an instance of the parent application.

I've removed the blocked label.

lib/components/util/types.ts
```diff
@@ -125,3 +125,18 @@ export type ZoomToPlaceHandler = (
 ) => void
 
 export type QueryParamChangeHandler = (event: any) => void
+
+export interface GetTimetableDataParams {
```
amy-corson-ibigroup: "Get" as a prefix for a type feels odd, would this not just be `TimetableDataParams`?

lib/components/viewers/timetable-wrapper.tsx
```diff
@@ -1,48 +1,83 @@
...
+      const names = (map.get(dirId) || []).concat([pattern.name])
+      map.set(dirId, names)
+    })
+
+    return map
+  }, [timetable])
+
+  if (loading) {
+    return <Loading />
+  }
```
amy-corson-ibigroup, line 73: Can we add a TODO here to add a loading aria status region to the body? we can do that when we implement the UI designs but I just don't want to forget

lib/components/viewers/viewers.css
```diff
@@ -50,6 +50,7 @@
 
 .otp .trip-viewer .header-with-back-button,
 .otp .pattern-viewer .header-with-back-button {
+  align-items: center;
```
amy-corson-ibigroup, line 53: Crazy this wasn't on there before

lib/components/viewers/pattern-viewer.tsx
```diff
@@ -140,10 +123,15 @@ const PatternViewer = ({
...
-            </button>
+          {callTakerEnabled && (
+            <a
+              href={timetableHref}
+              onClick={handleTimetableButtonClick}
+              rel="noreferrer"
+              target="_blank"
+            >
+              <FormattedMessage id="components.Timetable.timetable" />
+            </a>
```
amy-corson-ibigroup: Could you use `LinkOpensNewWindow` util to indicate it pops out? You may have to add optional `onClick` prop

amy-corson-ibigroup (approved):
I think this makes sense, thank you for the changes! Just a few nitpicks. The URL is really much cleaner than the portal.

lib/components/viewers/pattern-viewer.tsx
alec-georgoff: I tried, but it ended up using the browser-default "open in new tab" behavior, which isn't what we want. I was able to add the icon showing that it will open in a new window, though:

image

lib/components/viewers/timetable-wrapper.tsx
alec-georgoff: Done!

lib/components/util/types.ts
alec-georgoff: Yeah good call, updated.

lib/components/viewers/timetable-wrapper.tsx
```diff
@@ -1,48 +1,84 @@
...
-    const names = (directionNames.get(dirId) || []).concat([pattern.name])
-    directionNames.set(dirId, names)
-  })
+  const closedStopsSet = useMemo(
+    () => closedStops?.get(routeId),
+    [closedStops, routeId]
+  )
+
+  useEffect(() => {
+    stopClosuresQuery()
```
daniel-heppner-ibigroup, line 46: Actually not sure this needs to be in the useEffect. stopClosuresQuery happens through redux so I don't think it's an effect.

lib/components/viewers/timetable-wrapper.tsx
```diff
@@ -1,48 +1,84 @@
...
+    const today = new Date()
+    const tomorrow = new Date(today)
+    tomorrow.setDate(tomorrow.getDate() + 1)
+
+    getTimetableData({
```
daniel-heppner-ibigroup, line 48: perhaps getTimetableData should accept JS date objects and format them internally?

lib/components/viewers/timetable-wrapper.tsx
```diff
@@ -1,48 +1,84 @@
...
+      gtfsId: routeId,
+      serviceDate: format(today, 'yyyyMMdd'),
+      start: format(today, 'yyyy-MM-dd')
+    })
+  }, [getTimetableData, routeId, stopClosuresQuery])
 
-  const closedStopsSet = closedStops?.get(portalId || '')
+  useEffect(() => {
+    if (timetable?.route) setLoading(false)
+  }, [timetable])
```
daniel-heppner-ibigroup, line 56: can you add a note somewhere that handling the query like this is strictly okay only because this component loads in its own window? Handling data loading in useEffects is bug prone. 

For example: if input props change (like routeId) it can trigger the getTimetableData query again, which wouldn't cause the loading indicator to happen again. And it's also possible that if the routeId changes twice quickly, the first response could take longer than the second, causing the displayed data to be stale.

Also, error handling is a risk. If a request fails what happens? Loading indicator never goes away.

daniel-heppner-ibigroup: I'm okay with this since this component only gets rendered on its own with props that never get mutated, but I think we should document it well in the comments.

daniel-heppner-ibigroup (changes requested):

lib/components/viewers/timetable-wrapper.tsx
alec-georgoff: I think it does have to be here, unfortunately. See [this comment above](https://github.com/opentripplanner/otp-react-redux/pull/1607#discussion_r3786996495). Because our new window is not connected to the main application at all, any redux queries we want to happen need to be called explicitly. If you check the console of the new window while running locally, you'll see that very few redux actions actually take place.

lib/components/viewers/timetable-wrapper.tsx
alec-georgoff: Yeah, good call. Updated the params.

lib/components/viewers/timetable-wrapper.tsx
alec-georgoff: I agree with your concern here. And the props actually will end up getting mutated once we add the functionality of a date picker...I've added a TODO note to tackle this at that time and make it more resilient. Like you said, for now the props are static so shouldn't be an issue.

daniel-heppner-ibigroup (approved):

miles-grant-ibigroup (commented):
Looks like we'll have to open a new PR to merge this into dev. 

Also I'm seeing an issue where nothing is getting displayed? It's maybe due to the fact that the directionId is always `-1`?

Image

In terms of having the requests live inside otp-rr, I think that's fine but then I'd prefer it if we followed the pattern we have for all other requests, which is to say that we have a strongly typed otp-ui component that demands a data structure that cleanly matches the OTP graphql response. 

Then, I think we should put _all_ the timetable graphql stuff inside otp-rr, and get rid of the wrapper method.

miles-grant-ibigroup:
Also, this white screen is not a great failure mode, and we should probably try to add in some better error handling inside otp-ui to deal with bad data like this

miles-grant-ibigroup:
Also also, is there any way we can wrap instead of doing ellipses here? 
Screenshot 2026-09-01 at 2 05 43 PM

PR #1608, opened 2026-07-23
Remove TriMet Vehicle Label Rule

binh-dam-ibigroup:
**Description:**

This PR removes an arcane rule about vehicle position `label` field,
so that we can display correct labels with vehicle position feeds that use them.

An old JS construct about vehicle status is inlined.

**PR Checklist:**


amy-corson-ibigroup (approved):
LGTM!

miles-grant-ibigroup (approved):
Much cleaner thanks!

PR #1610, opened 2026-07-24
Remove 'Walk is better' error if faster transit itinerary

binh-dam-ibigroup:
**Description:**

With this PR  "Walk faster than transit" is removed if a transit itinerary from a flex query is faster than a walk-only itinerary.

**PR Checklist:**


miles-grant-ibigroup (changes requested):
This solution is clean, but I've seen this error falsely occur in other mode combinations too! I think the simpler solution is to add `WALKING_BETTER_THAN_TRANSIT` to `mutedErrors` in the config?

binh-dam-ibigroup:
> This solution is clean, but I've seen this error falsely occur in other mode combinations too! I think the simpler solution is to add `WALKING_BETTER_THAN_TRANSIT` to `mutedErrors` in the config?

@miles-grant-ibigroup How about we remove that indication all together unless no transit itineraries are returned?

miles-grant-ibigroup:
> > This solution is clean, but I've seen this error falsely occur in other mode combinations too! I think the simpler solution is to add `WALKING_BETTER_THAN_TRANSIT` to `mutedErrors` in the config?
> 
> @miles-grant-ibigroup How about we remove that indication all together unless no transit itineraries are returned?

But what about the case where there are transit itineraries returned that are in fact slower than walking? The annoying thing is that really the only way to get it right is to re-do the entire check client-side once all the mode combinations have finished loading. The issue is that that is way more overhead than I think makes sense. Therefore, the mutedError fix

binh-dam-ibigroup:
Closing this PR for now, there doesn't seem to be a generic way to handle that.

PR #1611, opened 2026-07-24
Feed Closed Stops to Stops Overlay

alec-georgoff:
**Description:**
Passes closed stops info to the stop overlay logic to add annotation when a stop is closed. Requires opentripplanner/otp-ui#979

**PR Checklist:**


josh-willis-arcadis (approved):

lib/components/map/default-map.tsx
```diff
@@ -544,7 +546,8 @@ class DefaultMap extends Component<DefaultMapProps> {
                   viewedRouteStops,
                   config.companies,
                   this.getEntityPrefix,
-                  feeds
+                  feeds,
+                  new Set(closedStops.values().flatMap((v) => v.values()))
```
amy-corson-ibigroup: I think this needs to be `closedStops?.values` because my app is crashing without it!

lib/components/app/responsive-webapp.js
```diff
@@ -203,6 +205,8 @@ class ResponsiveWebapp extends Component {
       }
     }
 
+    stopClosuresQuery()
```
amy-corson-ibigroup, line 209: Nitpicky, but can you add a comment here? It probably goes without saying but all the other methods we're calling here have explanations

amy-corson-ibigroup (changes requested):
Overall looking good, but I'm seeing a crash without optional chaining

lib/components/map/default-map.tsx
alec-georgoff: Oh great catch! I also updated the props type to allow it to be undefined for extra clarity

lib/components/app/responsive-webapp.js
alec-georgoff: Yup!

amy-corson-ibigroup (approved):
LGTM thanks for the changes!!

PR #1612, opened 2026-07-27
Import new transitive and pass ref to base-map

amy-corson-ibigroup:
**Description:**
Using custom vector tiles can sometimes break the transitive overlay. This is because our custom `Layer` code requests certain fonts that are not always available in the tiles, resulting in an error.

To fix this, we will instead fetch the available fonts from the tiles on map load, and then pass them to the transitive overlay. 

Requires https://github.com/opentripplanner/otp-ui/pull/981

**PR Checklist:**


miles-grant-ibigroup (approved):
Excellent fix! Very clean

lib/components/map/default-map.tsx
```diff
@@ -239,6 +241,19 @@ class DefaultMap extends Component<DefaultMapProps> {
...
+  _getFontsFromVectorTiles = (mapRef) => {
+    const fonts = []
+    mapRef?.current?.getStyle().layers.forEach((l) => {
+      const font = l.layout && l.layout['text-font']
+      font?.forEach((f) => !fonts.includes(f) && fonts.push(f))
+    })
+    this.setState({ vectorTileFonts: fonts })
+  }
+
   /**
```
miles-grant-ibigroup: Wait I'm just realizing -- why not do this inside of `base-map`? Wouldn't it be nice if every `base-map` user could benefit from the auto-detection?

miles-grant-ibigroup (dismissed):
Wait

lib/components/map/default-map.tsx
amy-corson-ibigroup: i added it to transitive! i guess it could put it in base map if we needed but since transitive is the package that needs the fonts and it needs to get routed through otp-rr, this seemed cleaner? lmk

lib/components/map/default-map.tsx
```diff
@@ -434,9 +436,9 @@ class DefaultMap extends Component<DefaultMapProps> {
             reuseMaps: true
           }}
           maxZoom={maxZoom}
-          onClickLayerSelector={onClickLayerSelector}
           // In Leaflet, this was an onclick handler. Creating a click handler in
           // MapLibreGL would require writing a custom event handler for all mouse events
```
alec-georgoff, line 441: It looks like this comment refers to the `onContextMenu` prop; can we move it back to that one?

alec-georgoff (approved):
Just one nit!

lib/components/map/default-map.tsx
amy-corson-ibigroup: Good catch, thank you!

binh-dam-ibigroup (approved):

PR #1614, opened 2026-07-29
Add flex symbol next to route block

binh-dam-ibigroup:
**Description:**

This PR adds a symbol next to route blocks in the itinerary summary (see screenshot) for routes that require reservation for the portion traveled. The reservation symbol can be customized (text or a React component).

image

**PR Checklist:**


miles-grant-ibigroup (changes requested):
Shouldn't we try to put this next to the bus icon instead?

binh-dam-ibigroup:
> Shouldn't we try to put this next to the bus icon instead?

Does it mean adding a separate mode icon if a flex bus follows or precedes a regular bus? Not sure I'd like that. How about adding the symbol to the left of the route number instead?

miles-grant-ibigroup:
> > Shouldn't we try to put this next to the bus icon instead?
> 
> Does it mean adding a separate mode icon if a flex bus follows or precedes a regular bus? Not sure I'd like that. How about adding the symbol to the left of the route number instead?

That's what I meant! As long as it's clearly "bundled" with the route icon

binh-dam-ibigroup:
> > > Shouldn't we try to put this next to the bus icon instead?
> > 
> > 
> > Does it mean adding a separate mode icon if a flex bus follows or precedes a regular bus? Not sure I'd like that. How about adding the symbol to the left of the route number instead?
> 
> That's what I meant! As long as it's clearly "bundled" with the route icon

Getting there with https://github.com/opentripplanner/otp-react-redux/pull/1614/commits/dca8fd511437bc04834d14f15372868c145f8d32 (a bit more work to do)

image

lib/components/narrative/metro/metro-itinerary-routes.tsx
```diff
@@ -74,16 +74,23 @@ const MetroItineraryRoutes = ({
...
 }: Props): JSX.Element => {
   const intl = useIntl()
+  // @ts-expect-error No type on ComponentContext
+  const { FlexReservationIcon, LegIcon } = useContext(ComponentContext)
   const routeLegs = itinerary.legs
     .filter(showAllWalkLegs ? () => true : removeInsignificantWalkLegs)
     .filter((leg) => !leg.interlineWithPreviousLeg)
   const transitRoutes = getItineraryRoutes(itinerary, intl)
 
+  const reservationWrapper = (
```
miles-grant-ibigroup:
```suggestion
  const ReservationWrapper = (
```

This is sort of a component isn't it

lib/components/narrative/metro/metro-itinerary-routes.tsx
```diff
@@ -107,24 +114,28 @@ const MetroItineraryRoutes = ({
       </InvisibleHeader>
       <Routes aria-hidden className="routes-container" enableDot={enableDot}>
         {routeLegs.map((leg: Leg, index: number, filteredLegs: Leg[]) => {
-          const previousLegMode =
-            (index > 0 && filteredLegs[index - 1].mode) || undefined
+          const prevLeg = filteredLegs[index - 1]
```
miles-grant-ibigroup: Should we guard on this somehow? I can imagine WSOD (white screen of death) if this leg is missing somehow

lib/components/narrative/metro/metro-itinerary-routes.tsx
```diff
@@ -107,24 +114,28 @@ const MetroItineraryRoutes = ({
...
               hideLongName
               key={index}
               leg={leg}
               LegIcon={LegIcon}
-              previousLegMode={previousLegMode}
+              modeIconDecoration={needReservation && reservationWrapper}
+              previousLegMode={
+                needReservation === previousLegReservation
+                  ? previousLegMode
+                  : undefined
```
miles-grant-ibigroup, line 139: Why is this now conditional? There are some non-flex `RouteBlock`s that rely on this info

lib/components/narrative/metro/metro-itinerary-routes.tsx
```diff
@@ -107,24 +114,28 @@ const MetroItineraryRoutes = ({
...
+            index > 0 && coreUtils.itinerary.isReservationRequired(prevLeg)
+          const needReservation = coreUtils.itinerary.isReservationRequired(leg)
           return (
             <RouteBlock
-              aria-hidden
               footer={
                 showLegDurations && (
-                  <FormattedDuration
-                    duration={leg.duration}
-                    includeSeconds={false}
```
miles-grant-ibigroup, line 119: Why was this removed?

lib/components/narrative/metro/metro-itinerary-routes.tsx
```diff
@@ -74,16 +74,23 @@ const MetroItineraryRoutes = ({
...
   const intl = useIntl()
+  // @ts-expect-error No type on ComponentContext
+  const { FlexReservationIcon, LegIcon } = useContext(ComponentContext)
   const routeLegs = itinerary.legs
     .filter(showAllWalkLegs ? () => true : removeInsignificantWalkLegs)
     .filter((leg) => !leg.interlineWithPreviousLeg)
   const transitRoutes = getItineraryRoutes(itinerary, intl)
 
+  const reservationWrapper = (
+    <span style={{ marginLeft: '-1ch' }}>
```
miles-grant-ibigroup: Is there no way to center vertically?

miles-grant-ibigroup: Or I guess if you're going for an asterisk is there no way to position it a bit above the icon?

miles-grant-ibigroup (changes requested):
I love the concept and the execution is clean, but I have a few small questions

lib/components/narrative/metro/metro-itinerary-routes.tsx
binh-dam-ibigroup: No, that's just rendered content.

lib/components/narrative/metro/metro-itinerary-routes.tsx
binh-dam-ibigroup: `includeSeconds` is already false by default in `FormattedDuration`'s definition.

lib/components/narrative/metro/metro-itinerary-routes.tsx
binh-dam-ibigroup: Good point, improved in https://github.com/opentripplanner/otp-react-redux/pull/1614/commits/55d2446555a0923c87761623d47532184deee0a4.

lib/components/narrative/metro/metro-itinerary-routes.tsx
binh-dam-ibigroup: I added this condition so that, if the reservation requirement changes between legs of the same mode (flex or not), the mode icon and its notice icon are drawn again. Otherwise the mode icon will not be redrawn.

lib/components/narrative/metro/metro-itinerary-routes.tsx
binh-dam-ibigroup: I ended up extracting a component, but it was for reuse to show in itinerary body.

lib/components/narrative/metro/metro-itinerary-routes.tsx
binh-dam-ibigroup: The vertical alignment is tricky and is influenced by the route number and mode icon rendering that introduce vertical offsets. For asterisks and text labels, I like how the decoration sits as implemented (see screenshot). 

image

To guarantee vertical content centering, configure `FlexNoticeIcon` to be a block with 100% height (see screenshot). People can also control horiz. margins through that block's CSS attributes as well.

image

lib/components/narrative/metro/metro-itinerary-routes.tsx
```diff
@@ -107,24 +109,34 @@ const MetroItineraryRoutes = ({
...
         {routeLegs.map((leg: Leg, index: number, filteredLegs: Leg[]) => {
-          const previousLegMode =
-            (index > 0 && filteredLegs[index - 1].mode) || undefined
+          let previousLegMode
+          let previousLegReservation
+
+          if (index > 0) {
+            const prevLeg = filteredLegs[index - 1]
+            previousLegMode = prevLeg.mode
+            previousLegReservation =
```
alec-georgoff: nit: Update name to `previousLegNeedsReservation`; this current naming makes me thing that this is an object describing the previous leg's reservation, not a boolean

alec-georgoff (approved):
One nit but otherwise looks good!

lib/components/narrative/metro/metro-itinerary-routes.tsx
miles-grant-ibigroup: I see. Can we add a pixel or 2 of padding to the left of the asterisk then?

miles-grant-ibigroup:
Should be good to go once the padding is there!

lib/components/narrative/metro/metro-itinerary-routes.tsx
binh-dam-ibigroup: Updated in https://github.com/opentripplanner/otp-react-redux/pull/1614/commits/df334ab9478f6659fc29b7cc6515ef85f6106e1d

miles-grant-ibigroup (approved):
Exciting! Thanks for the changes

PR #1617, opened 2026-08-27
Account module

binh-dam-ibigroup:
**Description:**

This PR is a first attempt to modularize OTP-RR, starting with the user account and mobility profile screens.
With these modules, related JS code and internationalization (i18n) messages are consolidated and loaded on demand instead of at startup (see bundle diffs below). The effect is mostly seen on the i18n messages that are now separate from the main messages.

Note: There is an existing concept of module at configuration level which is not conflicting with the `AppModule` terminology, but I'm happy to entertain a separate name for that.

**PR Checklist:**

| Before | After |
|--------|-------|
| before bundle stats  |  after bundle stats  |


lib/actions/ui.js
```diff
@@ -484,6 +491,26 @@ export function setLocale(locale, isUpdatingLocale = false) {
...
+  return (
+    !loadingMessages &&
+    appModules.every((appModule) => loadedModules.has(appModule))
+  )
+}
+
+export function loadAppModules(appModules) {
+  return async function (dispatch, getState) {
+    const state = getState()
+    if (!state.otp.ui.loadingMessages && !areModulesLoaded(state, appModules)) {
```
daniel-heppner-ibigroup: I believe loadingMessages is already checked inside `areModulesLoaded`

lib/actions/ui.js
```diff
@@ -484,6 +491,26 @@ export function setLocale(locale, isUpdatingLocale = false) {
...
+      dispatch(loadingMessages(true))
+      dispatch(loadModules(appModules))
+      await dispatch(setLocale(state.otp.ui.locale, false, true))
+      dispatch(loadingMessages(false))
```
daniel-heppner-ibigroup: maybe put loadingMessages(false) in a finally block

i18n/account/en-US.yml
```diff
@@ -0,0 +1,267 @@
...
+    confirmDeleteMonitoredTrip: Would you like to remove this trip?
+    confirmDeletePlace: Would you like to remove this place?
+    emailVerificationResent: The email verification message has been resent.
+    genericError: "An error was encountered: {err}"
+    itineraryExistenceCheckFailed: Error checking whether your selected trip is possible.
+    mustAcceptTermsToSavePlace: Please accept the Terms of Use (under My Account) to save locations.
+    mustBeLoggedInToSavePlace: Please log in to save locations.
+    placeDeleted: Your place has been deleted.
+    placeRemembered: The settings for this place have been saved.
+    preferencesSaved: Your preferences have been saved.
```
daniel-heppner-ibigroup: this message is rendered in a toast from the advanced-settings-panel. Although maybe it's fine because that code path never runs unless the accounts module is loaded.

lib/components/user/places/place-shortcut.js
```diff
@@ -44,7 +44,7 @@ class PlaceShortcut extends Component {
       } else if (persistenceMode.isLocalStorage) {
         window.alert(
           intl.formatMessage(
-            { id: 'components.Place.enterAlert' },
+            { id: 'components.PlaceShortcut.enterAlert' },
             { type: getFormattedPlaces(place.type, intl) }
```
daniel-heppner-ibigroup: should this be `placeType` rather than `type`?  (similar to in user.js)

daniel-heppner-ibigroup (approved):
Approving with minor comments

PR #1619, opened 2026-08-31
Include annotation on closed stops in nearby view

alec-georgoff:
**Description:**
Show a warning in the nearby view card when a stop is closed.

Screenshot 2026-08-31 at 3 16 59 PM

**PR Checklist:**

### DIFF
--- a/lib/components/util/colors.ts
+++ b/lib/components/util/colors.ts
@@ -33,6 +33,10 @@ const GREY_ON_WHITE = grey[700]
  */
 const BLUE_ON_WHITE = blue[800]
 
+const CANCELED_TRIP_RED = red[700]
+
+const CLOSED_STOP_RED = red[700]
+
 const DARK_TEXT_GREY = '#333333'
 
 const DEFAULT_ROUTE_COLOR = grey[800]
@@ -49,6 +53,8 @@ export {
   grey,
   red,
   blue,
+  CANCELED_TRIP_RED,
+  CLOSED_STOP_RED,
   DARK_TEXT_GREY,
   DEFAULT_ROUTE_COLOR,
   ELEVATION_BLUE,
--- a/lib/components/util/colors.ts
+++ b/lib/components/util/colors.ts
@@ -33,9 +33,7 @@ const GREY_ON_WHITE = grey[700]
  */
 const BLUE_ON_WHITE = blue[800]
 
-const CANCELED_TRIP_RED = red[700]
-
-const CLOSED_STOP_RED = red[700]
+const DISABLED_RED = red[700]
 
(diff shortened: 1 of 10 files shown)


lib/components/util/colors.ts
```diff
@@ -33,6 +33,10 @@ const GREY_ON_WHITE = grey[700]
  */
 const BLUE_ON_WHITE = blue[800]
 
+const CANCELED_TRIP_RED = red[700]
+
+const CLOSED_STOP_RED = red[700]
```
josh-willis-arcadis: do we need two different vars if they are the same thing? 
```suggestion
const DISABLED_RED = red[700]
```

josh-willis-arcadis (approved):
lgtm. one nit about colors.

lib/components/util/colors.ts
alec-georgoff: Makes sense!

lib/components/util/colors.ts
```diff
@@ -33,6 +33,8 @@ const GREY_ON_WHITE = grey[700]
  */
 const BLUE_ON_WHITE = blue[800]
 
+const DISABLED_RED = red[700]
```
amy-corson-ibigroup: To me, `DISABLED` brings to mind "disabled" in the context of HTML attributes (i.e. like a button being disabled). Could we do like `CANCELLED_RED` instead?

lib/util/itinerary.tsx
```diff
@@ -461,3 +461,14 @@ export function copyAndRemoveRouteModeOverrides(
...
 }
+
+/** Take a Map<string, Set<string>> of routes and their closed stops
+ * and flatten to a Set<string> of just stop IDs
+ */
+export function flattenStopClosures(
+  map: Map<string, Set<string>>
+): Set<string> {
+  return new Set(
+    Array.from(map.values()).flatMap((v) => Array.from(v.values()))
```
amy-corson-ibigroup, line 472: Maybe a TODO here to pull this eventually from OTP-UI?

amy-corson-ibigroup (approved):
Looks good! Just a couple nitpicks

lib/components/util/colors.ts
alec-georgoff: Since it's used for closures and cancelations, I ended up going with `CLOSED_OR_CANCELED_RED`

lib/util/itinerary.tsx
alec-georgoff: I have a ticket to simplify the stop closures data structure which will make this whole function go away! Will tackle that once the branches are in a more cohesive state.

PR #1621, opened 2026-09-02
Timetable Integration & Closed Stops

alec-georgoff:
**Description:**
This PR integrates the usage of the new timetable OTP-UI package into the pattern viewer (when enabled in the configuration). It also pulls in stop closure information and feeds it to the timetable for correct annotation.

See #1607 for previous discussion of timetable integration

**PR Checklist:**


alec-georgoff:
Copying over a few comments I need to address from #1607:

- [x] Handle `directionId` of `-1` or no `directionId` at all more gracefully ([comment](https://github.com/opentripplanner/otp-react-redux/pull/1607#pullrequestreview-5077763420))
- [ ] Strongly type the graphQL response for the timetable data (same comment as above)
- [ ] Update error handling to avoid mostly-blank white screen with no information ([comment](https://github.com/opentripplanner/otp-react-redux/pull/1607#issuecomment-5493653396))
- [x] Handle long route names re: placement of timetable button in pattern viewer ([comment](https://github.com/opentripplanner/otp-react-redux/pull/1607#issuecomment-5493670878))

alec-georgoff:
@miles-grant-ibigroup I'm looking for a bit of clarification on this comment:

> In terms of having the requests live inside otp-rr, I think that's fine but then I'd prefer it if we followed the pattern we have for all other requests, which is to say that we have a strongly typed otp-ui component that demands a data structure that cleanly matches the OTP graphql response.

Currently, the OTP-UI timetable package [is strongly typed for the graphQL response](https://github.com/ibi-group/otp-ui/blob/master/packages/timetable/src/index.tsx#L54). Is that what you're aiming for here, or are you saying something additional should be done on the OTP-RR side? All the reducers are in untyped JS files...although I suppose we could add typing for the timetable to the `OtpState` type? Is that what you're referring to?

alec-georgoff:
Handled placement of timetable link cutting off long route names by [moving it to its own row below](https://github.com/opentripplanner/otp-react-redux/pull/1621/commits/10db508ebd08b4ba3e2f067ba6e1700bb9855613)

Screenshot 2026-09-03 at 10 13 42 AM

alec-georgoff:
Added logic to strictly enforce proper `directionId` field for all patterns before rendering: https://github.com/opentripplanner/otp-react-redux/pull/1621/commits/4d90a2c11694ce48f4bb5a2bab49f7d1a8d258d4

alec-georgoff:
@miles-grant-ibigroup regarding the better error handling overall, I agree but think that should be addressed in the timetable package rather than here in OTP-RR. If you agree, I can create a ticket to do a general error-handling overhaul within the package, separate from this PR.

miles-grant-ibigroup:
> @miles-grant-ibigroup I'm looking for a bit of clarification on this comment:
> 
> > In terms of having the requests live inside otp-rr, I think that's fine but then I'd prefer it if we followed the pattern we have for all other requests, which is to say that we have a strongly typed otp-ui component that demands a data structure that cleanly matches the OTP graphql response.
> 
> Currently, the OTP-UI timetable package [is strongly typed for the graphQL response](https://github.com/ibi-group/otp-ui/blob/master/packages/timetable/src/index.tsx#L54). Is that what you're aiming for here, or are you saying something additional should be done on the OTP-RR side? All the reducers are in untyped JS files...although I suppose we could add typing for the timetable to the `OtpState` type? Is that what you're referring to?

Not quite... I do agree that it is annoying to move from typed code to untyped code but I think unless we are really deliberate about moving more graphql stuff to core-utils I would much rather have it all in one place. I was referring to the fact that we should keep the data structure similar to OTP's. But I really think we should have all the query generation happen in one place.

miles-grant-ibigroup:
> Handled placement of timetable link cutting off long route names by [moving it to its own row below](https://github.com/opentripplanner/otp-react-redux/pull/1621/commits/10db508ebd08b4ba3e2f067ba6e1700bb9855613)
> Screenshot 2026-09-03 at 10 13 42 AM

I'm not sure I love this... should we discuss further offline? Maybe there is a happy medium. Perhaps an icon where the text shoots out on hover? Although that wouldn't really match anything we've done in the past...

miles-grant-ibigroup:
> @miles-grant-ibigroup regarding the better error handling overall, I agree but think that should be addressed in the timetable package rather than here in OTP-RR. If you agree, I can create a ticket to do a general error-handling overhaul within the package, separate from this PR.

That's fine! But then I think that should maybe be blocking for this PR. I'm not sure we can merge code in that's this prone to displaying a completely broken interface

PR #1624, opened 2026-09-07
Timetable Integration & Closed Stops

alec-georgoff:
**Description:**
This PR integrates the usage of the new timetable OTP-UI package into the pattern viewer (when enabled in the configuration). It also pulls in stop closure information and feeds it to the timetable for correct annotation.

See #1607 and #1621 for previous discussion of timetable integration

**PR Checklist:**


lib/components/viewers/pattern-viewer.tsx
```diff
@@ -48,6 +62,12 @@ const PatternViewer = ({
   const patternId = viewedRoute?.patternId
   const routeId = viewedRoute?.routeId || null
 
+  const timetableHref = useMemo(() => `/#/timetable/${routeId}`, [routeId])
+
+  const handleTimetableButtonClick = useCallback(() => {
```
miles-grant-ibigroup: This is somehow causing the underlying window to reset?

lib/components/viewers/pattern-viewer.tsx
```diff
@@ -115,6 +135,17 @@ const PatternViewer = ({
             )}
           </h1>
         </div>
+        {callTakerEnabled && (
+          <TimetableLink
```
miles-grant-ibigroup, line 144: Can we un-center this?

lib/components/viewers/timetable-wrapper.tsx
```diff
@@ -0,0 +1,159 @@
...
+        {(stopClosuresError || !closedStops) && (
+          <FormattedMessage id="components.Timetable.errorLoadingStopClosures" />
+        )}
+        <button
+          onClick={() => {
+            setTimepointsOnly(!timepointsOnly)
+          }}
+        >
+          {timepointsOnly ? (
+            <FormattedMessage id="components.Timetable.showAllStops" />
```
miles-grant-ibigroup, line 113: I'm assuming this is temporary, but if it's not could we have these buttons in a row instead?

lib/components/viewers/timetable-wrapper.tsx
```diff
@@ -0,0 +1,159 @@
...
+    <div>
+      <div
+        style={{
+          alignItems: 'center',
+          display: 'flex',
+          flexDirection: 'column'
+        }}
+      >
+        {(stopClosuresError || !closedStops) && (
+          <FormattedMessage id="components.Timetable.errorLoadingStopClosures" />
```
miles-grant-ibigroup, line 105: Can we pass in the error somehow?

lib/components/app/responsive-webapp.js
```diff
@@ -202,6 +204,9 @@ class ResponsiveWebapp extends Component {
       }
     }
 
+    // Load closed stops into state for usage throughout UI (map popup, timetable, itinerary, etc.)
+    stopClosuresQuery()
```
miles-grant-ibigroup: We already have FIND_FEEDS_RESPONSE and FIND_ROUTES_RESPONSE. Could we try to merge this in with them?

lib/components/viewers/timetable-wrapper.tsx
```diff
@@ -0,0 +1,159 @@
...
+  ) : (
+    <FormattedMessage id="components.Timetable.errorLoadingTimetable" />
+  )
+}
+
+const mapStateToProps = (state: AppReduxState) => {
+  const { pathname } = state.router.location
+  const match = matchPath<{ routeId: string }>(pathname, {
+    exact: true,
+    path: TIMETABLE_PATH,
```
miles-grant-ibigroup: Can we re-use this up above where we've hardcoded it?

lib/reducers/create-otp-reducer.js
```diff
@@ -498,6 +503,52 @@ function createOtpReducer(config) {
             }
           }
         })
+      case 'STOP_CLOSURES_RESPONSE':
+        const alerts = action.payload.data.alerts
+        const map = new Map()
+        alerts.forEach((alert) => {
+          const entities = alert.entities
+          entities.forEach((entity) => {
+            if (entity.__typename === 'StopOnRoute') {
```
miles-grant-ibigroup, line 512: This makes me nervous about the case where `entities` is somehow blank. Could we either add fail-safe checking or use postprocessing in createGraphQLQuery?

miles-grant-ibigroup (changes requested):
A few nitpicks but overall this is quite well integrated!

The biggest worries I have are around error display and the button placement

lib/components/viewers/timetable-wrapper.tsx
alec-georgoff: Yup! Had to update the constant to be a function that takes in a routeId, since the path that the router needs uses the `:routeId` string. [524f74d](https://github.com/opentripplanner/otp-react-redux/pull/1624/commits/524f74dbc0bd1343130e4b42544717df3657809b)

lib/reducers/create-otp-reducer.js
alec-georgoff: I added an optional chaining operator in case there are no `entities` on the object. I was also curious about ending up with an undefined `alerts` object, but in my testing OTP will always return an `alerts` array, even if it's empty. So we should be good there [ef2f234](https://github.com/opentripplanner/otp-react-redux/pull/1624/commits/ef2f2344262dcdccd978c87dede4d3e37900655a)

lib/components/viewers/pattern-viewer.tsx
alec-georgoff: Uncentered and looking much better [b065f86](https://github.com/opentripplanner/otp-react-redux/pull/1624/commits/b065f8627804417fdcafdc8958d2eb635828254f)

lib/components/viewers/pattern-viewer.tsx
alec-georgoff: Ah yup, needed to throw a `preventDefault` on the click handler event: [4e5d050](https://github.com/opentripplanner/otp-react-redux/pull/1624/commits/4e5d050f4d68102968d6f1af28eba0661417fecf)

lib/components/viewers/timetable-wrapper.tsx
alec-georgoff: This is indeed very temporary! We have updated UI designs that will be implemented once core functionality is in a stable place (which seems very soon, fingers crossed).

lib/components/viewers/timetable-wrapper.tsx
alec-georgoff: Good question...I'm inclined to say we shouldn't show the raw error message because it will most likely be some kind of graphQL error? I'm not sure that's useful for the end user to see. What are your thoughts on passing the error into a `console.warn()` so it's a bit easier to debug?

lib/components/app/responsive-webapp.js
alec-georgoff: I'd prefer we keep it as its own query. Since the timetable is in its own window, it needs to fire off any of the graphQL queries required for the data being presented (it does not have access to the underlying state from the "parent" application). If we combine the stop closures query with one of the other queries, it introduces a ton more unnecessary data in the timetable window.

Combining the queries would indeed remove the need for this specific query call in `responsive-webapp`, but conceptually I feel better about having a single "source of truth query" that is used throughout the application.

What are your thoughts?

miles-grant-ibigroup: This makes sense. I'm ok keeping it as a separate query, but I'm not a huge fan of adding this manual function call in responsive-webapp. Any way we can bundle the call in to where the other 2 queries are fired off?

lib/components/viewers/timetable-wrapper.tsx
miles-grant-ibigroup: That sounds great!

lib/components/viewers/pattern-viewer.tsx
miles-grant-ibigroup: That's alright for now but could we shift to using a raw anchor tag? If it's too complicated then I'd be ok with a TODO to move to a raw anchor tag instead

miles-grant-ibigroup:
Let me know if you'd like a hand with moving away from `window.open`... I _think_ it should be possible

lib/components/viewers/pattern-viewer.tsx
alec-georgoff: We do use an anchor tag! `TimetableLink` is just a styled anchor tag. Interestingly, the anchor tag is part of what causes the problem: the browser both executes the `onClick` function _and_ tries to navigate to the `href`.

In theory, we could remove the `href` attribute, but [this has accessibility impacts](https://www.sitelint.com/docs/accessibility/missing-attribute-href-on-link#additional_informations) and removes the timetable button from the tab index, so I don't think it's worth it.

lib/components/app/responsive-webapp.js
alec-georgoff: Yup, moved it to where we fire the find feeds call inside default map: [dff74e6](https://github.com/opentripplanner/otp-react-redux/pull/1624/commits/dff74e694c69fb7bd4d0395183bcbe9bf022ceb0)

lib/components/viewers/timetable-wrapper.tsx
alec-georgoff: Done! [401ed9e](https://github.com/opentripplanner/otp-react-redux/pull/1624/commits/401ed9e9ea5904a3b2b58bcdac69e37c53d89564)

alec-georgoff:
> Let me know if you'd like a hand with moving away from `window.open`... I _think_ it should be possible

@miles-grant-ibigroup is there a reason you want to move away from `window.open`? It gives us a lot of flexibility in window placement and size.

lib/components/viewers/pattern-viewer.tsx
miles-grant-ibigroup: Didn't know about the a11y stuff with anchor tags... That's annoying

miles-grant-ibigroup (approved):
I know this was quite a bit of back and forth but I really like where we've ended up! Thanks for all your effort

lib/actions/api.js
```diff
@@ -402,6 +407,12 @@ export const findingRoutes = createAction('FINDING_ROUTES')
 export const findRoutesResponse = createAction('FIND_ROUTES_RESPONSE')
 export const findRoutesError = createAction('FIND_ROUTES_ERROR')
 
+// Timetable lookup query
+
```
binh-dam-ibigroup: Remove extra blank line.

lib/actions/api.js
```diff
@@ -345,6 +345,11 @@ export function parkAndRideQuery(
   return createQueryAction(endpoint, responseAction, errorAction, options)
 }
 
+// Load closed stops
+export function stopClosuresQuery() {
```
binh-dam-ibigroup: Nit - rename to `getClosedStops` or `fetchClosedStops` or similar.

lib/actions/apiV2.js
```diff
@@ -1056,6 +1092,65 @@ export function findFeeds() {
   }
 }
 
+const timetableResponse = createAction('TIMETABLE_ROUTE_RESPONSE')
+const timetableError = createAction('TIMETABLE_ROUTE_ERROR')
+
+// serviceDate: YYYYMMDD
+// start, end: YYYY-MM-DD
+// end is exclusive
+const timetableQuery = (end, gtfsId, serviceDate, start) => `
```
binh-dam-ibigroup: Rename `gtfsId` to `routeGtfsId` and make it the first arg. `end` goes to the end. I would also be in favor of passing a single date argument and moving the date format logic in here.

lib/components/app/responsive-webapp.js
```diff
@@ -18,6 +18,7 @@ import PropTypes from 'prop-types'
 import qs from 'qs'
 import React, { Component } from 'react'
 
+import * as apiActions from '../../actions/api'
```
binh-dam-ibigroup: Revert this change?

lib/components/util/types.ts
```diff
@@ -125,3 +125,10 @@ export type ZoomToPlaceHandler = (
 ) => void
 
 export type QueryParamChangeHandler = (event: any) => void
+
+export interface TimetableDataParams {
+  /** The desired service date for the timetable, as a Date object */
+  date: Date
+  /** The GTFS ID of the route for which a timetable is being fetched */
+  gtfsId: string
```
binh-dam-ibigroup: Nit- Rename to `routeGtfsId`?

lib/components/viewers/timetable-wrapper.tsx
```diff
@@ -0,0 +1,164 @@
...
+      gtfsId: routeId
+    })
+  }, [getTimetableData, routeId, stopClosuresQuery])
+
+  useEffect(() => {
+    // TODO: improve handling of data fetching to avoid issues with useEffect and stale data.
+    // This will be important when the capability to fetch timetables for different dates via a calendar
+    // is added.
+    if (timetable?.route) setLoading(false)
+  }, [timetable])
```
binh-dam-ibigroup: You might be able to address/remove the TODO above by extracting a const for `timetable?.route` and put it here as the dependency for this `useEffect`.

lib/components/viewers/timetable-wrapper.tsx
```diff
@@ -0,0 +1,164 @@
...
+      (pattern: any) => ![0, 1].includes(pattern?.directionId)
+    )
+
+    if (invalid)
+      console.warn(
+        'Direction IDs of timetable data are not valid (must be 0 or 1 for every pattern)'
+      )
+
+    return invalid
+  }, [timetable])
```
binh-dam-ibigroup: Yeah, I think extracting a const for `timetable.route` for this entire method is the way to go for all of these `useEffect`s.

binh-dam-ibigroup (changes requested):
Nice changes overall. Please fix the one method that is messed up, then I'll approve. (And see the other nits)

lib/components/app/responsive-webapp.js
alec-georgoff: Done: [06ca182](https://github.com/opentripplanner/otp-react-redux/pull/1624/commits/06ca182d0fdecf95cf61b512fa82dc32197a475c)

lib/actions/apiV2.js
alec-georgoff: Done: [76b459e](https://github.com/opentripplanner/otp-react-redux/pull/1624/commits/76b459ede5ee1ec36a28832ccaeadac1e0b4b7f6)

lib/actions/api.js
alec-georgoff: [8ec6564](https://github.com/opentripplanner/otp-react-redux/pull/1624/commits/8ec656497a119761b3f438a089435edaeddafdec)

lib/actions/api.js
alec-georgoff: [b1687d6](https://github.com/opentripplanner/otp-react-redux/pull/1624/commits/b1687d6812501a08498e5f0149781ad135112395)

lib/components/util/types.ts
alec-georgoff: Done!

lib/components/viewers/timetable-wrapper.tsx
alec-georgoff: Implemented here: [da3979c](https://github.com/opentripplanner/otp-react-redux/pull/1624/commits/da3979cb6623270346dda130bb1418d0a59bd005)

lib/components/viewers/pattern-viewer.tsx
```diff
@@ -48,6 +63,16 @@ const PatternViewer = ({
   const patternId = viewedRoute?.patternId
   const routeId = viewedRoute?.routeId || null
 
+  const timetableHref = useMemo(() => `/#${TIMETABLE_PATH(routeId)}`, [routeId])
+
+  const handleTimetableButtonClick = useCallback(
+    (e) => {
+      e.preventDefault()
+      window.open(timetableHref, undefined, 'width=1000,height=800')
```
binh-dam-ibigroup, line 71: Not blocking, but on first time opening a timetable window, the main window reloads for me. Do you see that?

binh-dam-ibigroup (approved):
Thanks for the changes, much cleaner!

lib/components/viewers/pattern-viewer.tsx
alec-georgoff: Weird...Miles saw that issue and [it should have been fixed with the preventDefault call](https://github.com/opentripplanner/otp-react-redux/pull/1624#discussion_r3958153565). I just tried in Safari and Chrome with Incognito windows and wasn't able to observe it happening. I'll keep an eye out to see if it pops up again.

binh-dam-ibigroup: I'm using Firefox, not sure if that made a difference

miles-grant-ibigroup: I was using firefox as well.

PR #1625, opened 2026-09-07
Fix Route Long Name Display Logic

alec-georgoff:
**Description:**
Currently, route long name information can be carried in one of two ways in an itinerary leg:
- Directly on the leg object (`leg.routeLongName`), or
- As part of the `route` object on the leg (`leg.route.longName`)

This can be seen [here in OTP-UI](https://github.com/opentripplanner/otp-ui/blob/master/packages/core-utils/src/itinerary.ts#L888)

When the `RouteLongName` component is conditionally rendered in `RouteBlock`, one of the conditions is that `leg.routeLongName` is truthy. This is not accommodating to the `route` object approach; the change here fixes that.

**PR Checklist:**

### DIFF
--- a/lib/components/narrative/metro/route-block.tsx
+++ b/lib/components/narrative/metro/route-block.tsx
@@ -146,6 +146,9 @@ const RouteBlock = ({
   }
   const showModeIcon = leg.mode !== previousLegMode
 
+  const longNameExists =
+    leg.routeLongName || (typeof leg.route === 'object' && leg.route.longName)
+
   return (
     <>
       {showDivider && previousLegMode && (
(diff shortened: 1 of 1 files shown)


miles-grant-ibigroup (commented):
Happy to approve for now (if the types can be fixed) but I am a little worried that we might be patching this in the wrong location. Shouldn't this all be on OTP2 standard by now?

alec-georgoff:
> Happy to approve for now (if the types can be fixed) but I am a little worried that we might be patching this in the wrong location. Shouldn't this all be on OTP2 standard by now?

Fixed the type issues. I agree with the question about OTP2...it seems like, at least for now, this component is built to accept both OTP1 and 2. The conditional above this one references `leg.routeShortName || leg.route || leg.routeLongName`.

I'd argue this patch is worthwhile for now regardless, but am open to a discussion of removing OTP1 references.

miles-grant-ibigroup (approved):
Wonderful

binh-dam-ibigroup (approved):

PR #1629, opened 2026-09-14
Subtract transfer leg durations from total itin walk times

amy-corson-ibigroup:
**Description:**

`ExtraTransferLegOnSameStop` feature means that wait times on transfer legs are being added to the total walk time. Account for this by manually adding together the walk time excluding transfer legs

**PR Checklist:**


lib/actions/apiV2.js
```diff
@@ -1282,9 +1282,25 @@ export function routingQuery(searchId = null, updateSearchInReducer) {
...
+                  return { ...itin, walkTime: adjustedWalkTime }
+                }
+              )
+
               // Filter itineraries to collapse short names and hide unnecessary errors.
-              const withCollapsedShortNames = filteredItineraries?.map(
-                (itin) => ({
+              const withCollapsedShortNames =
+                itinerariesWithAdjustedWalkTimes?.map((itin) => ({
                   ...itin,
```
miles-grant-ibigroup, line 1304:
```suggestion
                  ...itin,
                  walkTime: legs.filter(duration > 0).map(leg=>walktime)
```

Wouldn't this be simpler? Just one line

miles-grant-ibigroup (changes requested):
I think we can make this a lot cleaner!

lib/actions/apiV2.js
amy-corson-ibigroup: This would be awesome but `leg` doesn't have a `walkTime` that's only on `itinerary`

miles-grant-ibigroup: there's a duration though right? so if we add another filter that only gives us walk legs, then this should work

`legs.filter(leg=>iswalkingleg).filter(duration > 0).map(leg=>walktime)`

amy-corson-ibigroup: I tried to apply this, but needed to filter for distance too and add a reducer

miles-grant-ibigroup (approved):
Love it!

josh-willis-arcadis (approved):

PR #1632, opened 2026-09-16
Create text-only itinerary button and import textonly itin function

amy-corson-ibigroup:
**Description:**
Import new text-only itinerary function and create a new trip tool button to call it. Requires https://github.com/opentripplanner/otp-ui/pull/987

**PR Checklist:**


lib/components/narrative/copy-text-itin-button.tsx
```diff
@@ -0,0 +1,72 @@
+import { Button } from 'react-bootstrap'
+import { Check, FileText } from '@styled-icons/fa-solid'
```
miles-grant-ibigroup: These need to be broken out into separate imports! To avoid a huge bundle

miles-grant-ibigroup: I actually spotted 2 other commits from you in `back-button.tsx` and `styledIcon.tsx` that have the same problem. Could you please fix those too? This will really improve the tree shaking

miles-grant-ibigroup (changes requested):
Love the text change! Just need to repair the tree shake and this is good to go

lib/components/narrative/copy-text-itin-button.tsx
amy-corson-ibigroup: Good catch!

miles-grant-ibigroup (approved):
Excited for this!

alec-georgoff (approved):

PR #1635, opened 2026-09-21
Remove incorrect longNameExists blocker

alec-georgoff:
**Description:**
#1625 was recently merged in, which fixed a bug related to route long name logic. However, that was actually just applying a fix to a larger issue.

The `RouteBlock` component uses the `RouteLongName` component from the itinerary-body OTP-UI package. As can be seen [here](https://github.com/opentripplanner/otp-ui/blob/master/packages/itinerary-body/src/defaults/route-long-name.tsx#L34), that component already has logic for handling the existence of route long names. When a long name doesn't exist, the `RouteLongName` component falls back to the headsign.

By having a long name check in OTP-RR, we're preventing the `RouteLongName` from actually performing the required logic that would cause it to fallback to the headsign. This means that, before this, any route without a route long name would have nothing displayed here, instead of falling back to the headsign.

**PR Checklist:**


binh-dam-ibigroup (approved):
Good finding!

miles-grant-ibigroup (approved):
Clean fix! We must've forgotten about this when we added headsign parsing

